PR #2, opened 2021-01-29
RFC0002: MVP

kiwicopple:
We would like to create a basic CLI for Supabase. At this stage we can assume that the user won't need to create orgs/projects from the CLI, but they will need to:

- develop locally
- push database migrations to their Postgres database

### DIFF
--- a/rfcs/0002-mvp.md
+++ b/rfcs/0002-mvp.md
@@ -1,0 +1,159 @@
+---
+feature: mvp
+start-date: 2020-01-28
+author: kiwicopple
+co-authors: steve-chavez
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+We would like to create a basic CLI for Supabase. At this stage we can assume that the user won't need to create orgs/projects from the CLI, but they will need to:
+
+- develop locally
+- push database migrations to their Postgres database
+
+# Motivation
+[motivation]: #motivation
+
+- Make it easy to run Supabase locally
+- Make it easy to manage database migrations
+- Make it easy to do "database branching", where each git branch can have it's own data to prevent conflicts. This is especially important for working with platforms like Vercel/Netlify
+
+
+# Detailed design
+[design]: #detailed-design
+
+
+### Initializing
+
+```sh
+supa init     # creates a .supabase folder with relevant config, connection strings etc. Also scaffolds out a database folder.
+supa start    # Starts Supabase (by running docker-compose), and pulls up the database
+supa eject    # See the .supabase/docker-compose.yml details below
+```
+This will likely need: 
+
+- `supabase.js` - a config file with the Supabase Project details
+- `.supabase/config.txt` - CLI config and details
+- `.supabase/docker-compose.yml` - for running Supabase locally. We may not even need this, as we can just run it directly from the `npm_modules` folder, which would prevent conflicts when we add upgrades or make changes. If we do generate a file, this should be stored in the `.supabase` folder so that it doesn't conflict with the user's docker-compose.yml file. When the user runs `supa eject`, it can move the file out to their root folder (similar to a `cra eject` model), and then the user can manage this themselves.
+
+
+For the database scaffold, [this](https://gitlab.com/delibrium/delibrium-postgrest/-/tree/master/database) format is very nice, although I think we can make it a bit clearer. Here is a suggested format:
+
+```sh
+│   # Data: database fixed/seed data.
+├── data
+│   ├── some_data.{sql|csv|json}        
+│   # Models: Tables and Views
+├── models
+│   ├── public.table                    # Personal preference to explicitly add the schema, and we should prefer dot notation over nesting.
+│   │   ├── data.{sql|csv|json}         # For convenience, table specific data can be stored here too.   
+│   │   ├── model.sql                   # This is the core. We should also figure out a nice, readable format for SQL files. Should include constraints and indexes and table/column comments.
+│   │   ├── permissions.sql             # Grants and Policies.
+│   │   ├── test.sql                    # PGTAP tests?
+│   │   └── triggers.sql                # table triggers 
+│   # Functions/RPC: functions. At the moment the function are written in SQL. If we want to support something like plv8, we would have to include a builder (like webpack) to bundle the function.
+├── functions                                
+│   ├── public.func1.sql                
+│   ├── public.func2.sql│   
+│   # Postgres roles and grants.
+├── roles                                
+│   ├── role_name.sql    
+│   # Types: custom database types
+└── types
+    ├── _init.sql
+    ├── public.some_type.sql
+    └── extensions.some_type.sql
+```
+
+
+### Branching
+
+This is going to be a key part of Supabase development. Since so many of our users are on Vercel/Netlify, we should have the ability to support [branch/preview-development](https://vercel.com/docs/platform/deployments).
+
+We can do this by connecting to a different database, an idea that I got from [pgsh](https://github.com/sastraxi/pgsh). For example:
+
+`postgresql://postgres@localhost:5432/master`
+`postgresql://postgres@localhost:5432/develop`
+`postgresql://postgres@localhost:5432/chore_add_types`
+
+This is the same from CI (like github actions), except that instead of running against a local database, it is running against the supabase database. We will need a way inside our config to match branch names to database names too. For example 
+
+
+### Dump
+
+```sh
+supa db dump                # dumps the currently active (local) database to disk
+supa db restore             # restores the (local) database from disk
+supa migrate                # generate the migrations file
+supa up                     # runs the "up" migrations on Prod
+supa down                   # runs the "down" migrations on Prod
+```
+
+The `supa migrate` task is the trickiest part here. See [#migration-questions](#migration-questions) below.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Supabase is about "time to value", and making things as simple/easy as possible. A CLI can be complicated, so we should make sure that the experience is as amazing as our hosted platform
+
+# Alternatives
+[alternatives]: #alternatives
+
+The hosted platform provides much of this functionality. We could continue to build this functionality into the platform however it will be a great dev experience if they can develop locally and push their changes later.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+
+### Migration questions
+[migration-questions]: #migration-questions
+
+The `supa migrate` task is the trickiest part. How do we determine what migrations should run on our database, since the database is created declaratively? We would need to compare the databases somehow, and figure out the `up/down` migrations. Alternatively we could do "stateless" migrations, where we just run `restore` on a brand new database, then re-route traffic, but then how do we get the user's prod data into the new database. Bear in mind that enterprise users might have terrabytes of data, and so copying it across will be slow. 
+
+
+### Other questions
+- what is the prefix? should it be `supabase` (clearer) or `supa` (might have a few people calling it "supper/suppa")
+- Which CLI tool/framework/libraries should we use?
+- `supa start` requires docker-compose. Is it reasonable to expect that everyone will have docker set up on their computer? Can we make this easier? 
+- How much functionality should we push to postgres-meta? 
+
+
+
+# Future work
+[future]: #future-work
+
+### Supabase platform 
+
+```sh
+supa deploy   # Deploy a local app to prod. On the first time this is run, it should generate a `./supabase.js` with relevant Supabase config.
+supa orgs     # Manage orgs
+supa project  # Manage projects
+supa logs     # Fetch logs
+supa scale    # Scale the current database
+...etc
+```
+
+
+### Generators (bonus)
+
+Since I have been using typescript in one of our repos, I have discovered a lot of limitations around our [Typescript Generators](https://supabase.io/docs/client/generating-types). This is because Open API spec doesn't have all the tables (if they aren't exposed to `postgres` role), and sometimes it expects some inputs where the database already has defaults (for example our OpenAPI for `created_at` expects a string, when it is actually auto-generated).
+
+I propose here a few ideas for generators. While they are exposed through the CLI, I think we should actually build these into [pg-api](https://github.com/supabase/pg-api), so that anyone can use them without the supabase CLI. (I also think this should be the case with the "dump" functionality).
+
+```sh
+supa gen json api # Dumps the JSON Schema for the API
+supa gen json db # generate AJV schemas for input validation
+
+supa gen typescript api  # generate typescript interfaces for api
+supa gen typescript db # generate typescript interfaces for database. Note that these types should have an interface for each table with "insert types" and "select types". These are actually different because the database can generate default values, so some fields aren't "required" 
+
+
+# Models: 
+# This is my suggestion for a "local first" database which has all of the rules/relationships we need to build a a realtime user store with automatic synchronization via our Realtime engine.
+# A bit like https://github.com/mobxjs/mst-gql (but hopefully much simpler)
+supa gen typescript store 
+```


rfcs/0002-mvp.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Migration questions
+[migration-questions]: #migration-questions
+
+The `supa migrate` task is the trickiest part. How do we determine what migrations should run on our database, since the database is created declaratively? We would need to compare the databases somehow, and figure out the `up/down` migrations. Alternatively we could do "stateless" migrations, where we just run `restore` on a brand new database, then re-route traffic, but then how do we get the user's prod data into the new database. Bear in mind that enterprise users might have terrabytes of data, and so copying it across will be slow. 
+
+
+### Other questions
+- what is the prefix? should it be `supabase` (clearer) or `supa` (might have a few people calling it "supper/suppa")
+- Which CLI tool/framework/libraries should we use?
```
darora, line 120: If we use golang (static binaries, easy to distribute), https://github.com/spf13/cobra is used by most (all?) of the nicer modern CLIs (e.g. kubectl, I think also flyctl under the wraps)

kiwicopple: Looks great, although I suspect we will need to use `postgres-meta`/`pg-api` for a lot of this functionality. 

Unfortunately it creates a dependency on NodeJS

rfcs/0002-mvp.md
```diff
@@ -0,0 +1,159 @@
...
+supa deploy   # Deploy a local app to prod. On the first time this is run, it should generate a `./supabase.js` with relevant Supabase config.
+supa orgs     # Manage orgs
+supa project  # Manage projects
+supa logs     # Fetch logs
+supa scale    # Scale the current database
+...etc
+```
+
+
+### Generators (bonus)
```
BeeeQueue, line 141: This is a fantastic idea that I think will help a lot of people!

For some inspiration you can take a look at the fantastic [GraphQL Code Generator](https://graphql-code-generator.com/) project!

rfcs/0002-mvp.md
```diff
@@ -0,0 +1,159 @@
...
+supa gen json db # generate AJV schemas for input validation
+
+supa gen typescript api  # generate typescript interfaces for api
+supa gen typescript db # generate typescript interfaces for database. Note that these types should have an interface for each table with "insert types" and "select types". These are actually different because the database can generate default values, so some fields aren't "required" 
+
+
+# Models: 
+# This is my suggestion for a "local first" database which has all of the rules/relationships we need to build a a realtime user store with automatic synchronization via our Realtime engine.
+# A bit like https://github.com/mobxjs/mst-gql (but hopefully much simpler)
+supa gen typescript store
```
kiwicopple, line 158: @joshnuss 's ActiveRecord library could be a good starting place for JS: 
https://github.com/joshnuss/supabase-active-record

rfcs/0002-mvp.md
```diff
@@ -0,0 +1,159 @@
...
+
+```sh
+│   # Data: database fixed/seed data.
+├── data
+│   ├── some_data.{sql|csv|json}        
+│   # Models: Tables and Views
+├── models
+│   ├── public.table                    # Personal preference to explicitly add the schema, and we should prefer dot notation over nesting.
+│   │   ├── data.{sql|csv|json}         # For convenience, table specific data can be stored here too.   
+│   │   ├── model.sql                   # This is the core. We should also figure out a nice, readable format for SQL files. Should include constraints and indexes and table/column comments.
```
kiwicopple, line 53: I did a small mock of this to figure out how to structure the file. Some initial thoughts here:
https://github.com/supabase/cli/pull/3/files#r575771489

rfcs/0002-mvp.md
```diff
@@ -0,0 +1,159 @@
...
+- `postgresql://postgres@localhost:5432/develop`
+- `postgresql://postgres@localhost:5432/chore_add_types`
+
+This is the same from CI (like github actions), except that instead of running against a local database, it is running against the supabase database. We will need a way inside our config to match branch names to database names too. For example 
+
+
+### Dump
+
+```sh
+supa db dump                # dumps the currently active (local) database to disk
```
kiwicopple, line 88: I used `postgres-meta` to create a small mock here: https://github.com/supabase/cli/pull/3

I'm not entirely sure if it's the right approach. On the one hand, we have full control to manage the layout/structure of the dump. On the other hand, we will need to cover all the nuances of Postgres. 

Perhaps it's better rely on `pgdump` from the outset. doing some sort of system where we either:

Option 1:

1. Run pg_dump on the entire database
2. Run some sort of AST parser on the full dump
3. Clean it up and process it into separate files

or Option 2:

1. Run `pg_dump` on individual tables/functions/types etc (eg: `pg_dump -Fp -t $table -f $table.dump;`). We can use `postgres-meta` to fetch the "structure" of the database
2. Run some some sort of "prettier" formatter to make it look readable


I'm leaning towards option 2, as it also give the develop better control. For example they might want to dump just their "functions".

rfcs/0002-mvp.md
```diff
@@ -0,0 +1,159 @@
...
+supa db restore             # restores the (local) database from disk
+supa migrate                # generate the migrations file
```
kiwicopple, line 90: I'm still not sure how we will handle going from a declarative structure, into a set of migration scripts.

There are tools like [migra](https://github.com/djrobstep/migra) which work on a schema comparison - this might be the way to do it as long as the limitations aren't too restrictive. 

eg: "custom types/domains Basic support (drop-and-create only, no alter)". [ref](https://databaseci.com/docs/migra)

rfcs/0002-mvp.md
soedirgo: https://github.com/vadimdemedes/ink looks good, apparently Prisma used to use it but I couldn't figure out why they migrated away from it.

rfcs/0002-mvp.md
```diff
@@ -0,0 +1,159 @@
...
+│   ├── some_data.{sql|csv|json}        
+│   # Models: Tables and Views
+├── models
+│   ├── public.table                    # Personal preference to explicitly add the schema, and we should prefer dot notation over nesting.
+│   │   ├── data.{sql|csv|json}         # For convenience, table specific data can be stored here too.   
+│   │   ├── model.sql                   # This is the core. We should also figure out a nice, readable format for SQL files. Should include constraints and indexes and table/column comments.
+│   │   ├── permissions.sql             # Grants and Policies.
+│   │   ├── test.sql                    # PGTAP tests?
+│   │   └── triggers.sql                # table triggers 
+│   # Functions/RPC: functions. At the moment the function are written in SQL. If we want to support something like plv8, we would have to include a builder (like webpack) to bundle the function.
```
steve-chavez, line 57: I would say yes for pgtap tests. Should we generate some simple pgtap tests by default? Like table [has_column](https://pgtap.org/documentation.html#has_column)? That would give the CLI  a more framework-like experience.

Also, maybe we should stick to SQL for this RFC, plv8 can come later(more complicated).

rfcs/0002-mvp.md
```diff
@@ -0,0 +1,159 @@
...
+- `.supabase/config.txt` - CLI config and details
+- `.supabase/docker-compose.yml` - for running Supabase locally. We may not even need this, as we can just run it directly from the `npm_modules` folder, which would prevent conflicts when we add upgrades or make changes. If we do generate a file, this should be stored in the `.supabase` folder so that it doesn't conflict with the user's docker-compose.yml file. When the user runs `supa eject`, it can move the file out to their root folder (similar to a `cra eject` model), and then the user can manage this themselves.
+
+
+For the database scaffold, [this](https://gitlab.com/delibrium/delibrium-postgrest/-/tree/master/database) format is very nice, although I think we can make it a bit clearer. Here is a suggested format:
+
+```sh
+│   # Data: database fixed/seed data.
+├── data
+│   ├── some_data.{sql|csv|json}
```
steve-chavez, line 48: Regarding branching, this would allow us to just rely on different hardcoded master data for test/staging/preview environments right?

It does seem easier compared to something like snapshotting the db with ZFS. Much more lighter on resources as well.

rfcs/0002-mvp.md
```diff
@@ -0,0 +1,159 @@
...
+│   ├── public.func1.sql                
+│   ├── public.func2.sql│   
+│   # Postgres roles and grants.
+├── roles                                
+│   ├── role_name.sql    
+│   # Types: custom database types
+└── types
+    ├── _init.sql
+    ├── public.some_type.sql
+    └── extensions.some_type.sql
```
steve-chavez, line 68: These seems to be composed of really short files: https://github.com/Delibrium/delibrium-postgrest/tree/master/database/types.

Perhaps we could have a single `types.sql` by default instead. Is there an advantage for migrations that comes with using single files?

I'd be fine either way, but it's more jumping around files to find something when having many tiny ones.

rfcs/0002-mvp.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Branching
+
+This is going to be a key part of Supabase development. Since so many of our users are on Vercel/Netlify, we should have the ability to support [branch/preview-development](https://vercel.com/docs/platform/deployments).
+
+We can do this by connecting to a different database, an idea that I got from [pgsh](https://github.com/sastraxi/pgsh). For example:
+
+- `postgresql://postgres@localhost:5432/master`
+- `postgresql://postgres@localhost:5432/develop`
+- `postgresql://postgres@localhost:5432/chore_add_types`
```
steve-chavez, line 80: On these different envs, would the users also get different realtime/postgrest instances?

For preview workflows, the users would like to interact with the full stack right?

rfcs/0002-mvp.md
steve-chavez: > I'm still not sure how we will handle going from a declarative structure, into a set of migration scripts.

Ideally the developer should not look at migrations scripts and instead it should focus on declarative* SQL. So perhaps we could keep the migrations off git?(perhaps on a table)

* Of course there's a limit of how declarative SQL can be. Column renames for example always have to use ALTER.

steve-chavez: [metagration](https://github.com/michelp/metagration) also has an interesting approach. It allows generating up/down migration scripts with stored procedures. It also has its own table for tracking applied migrations.

Maybe we can use metagration plus migra for providing a template migration that can be edited.

steve-chavez: metagration also creates a restore point for down migrations, so it's possible to recover dropped columns for example(no need for a restore from a backup).

rfcs/0002-mvp.md
kiwicopple: Agreed

rfcs/0002-mvp.md
kiwicopple: Yeah exactly - this would give "reproducible builds". It's not as powerful as ZFS cloning, but it is less "magic"

ZFS cloning might solve a different use-case in the future for managing terrabyte-scale databases

rfcs/0002-mvp.md
kiwicopple: This completely slipped my mind. Yes they are going to need it. Hmm, how will we keep the costs down? 

Perhaps it will have to be "local only" for now? It should be manage-able with the docker files

rfcs/0002-mvp.md
kiwicopple: OK cool - metagration looks interersting. 

I think it's the declarative structure -> up/down scripts that's the tricky part here though. The tools I found:

- migra (python)
- https://github.com/gimenete/dbdiff (JS, looks limited)
- https://github.com/joncrlsn/pgdiff (Go)
- https://github.com/fordfrog/apgdiff (Java)
- https://github.com/eulerto/pgquarrel (C)

Or we can of course code this directly into `postgres-meta`, perhaps using `migra` as an example since it seems the most feature complete. It will be a lot of work though.

rfcs/0002-mvp.md
steve-chavez: > Hmm, how will we keep the costs down?

A temporary postgrest with `db-pool = 1`(one connection max) should be lightweight, it would fit on the same instance. Realtime would need to limit its pool as well, but we'd definitely need to set its systemd `MemoryLimit` to be really low. But I'm not sure how low it can get.

> Perhaps it will have to be "local only" for now? It should be manage-able with the docker files

Yes, local-only it's a good start. Other envs can be done in a later RFC - they need more thought.

rfcs/0002-mvp.md
Jolg42: @soedirgo It had issues on Windows for us (Prisma) and we simplified things a lot so it was not needed anymore.

rfcs/0002-mvp.md
steve-chavez: There's also pgadmin schema diff:

- doc: https://www.pgadmin.org/docs/pgadmin4/5.0/schema_diff.htm
- source: https://github.com/postgres/pgadmin4/tree/master/web/pgadmin/tools/schema_diff (python)

steve-chavez: > https://github.com/gimenete/dbdiff

Last commit on 2017.

> https://github.com/fordfrog/apgdiff

From experience(I added [RLS support](https://github.com/fordfrog/apgdiff/pull/213)), this one is not actively maintained. It has many open issues/PRs. 

> https://github.com/eulerto/pgquarrel

Has unanswered issues(also unmerged PRs), it also has partial support for many pg features: https://github.com/eulerto/pgquarrel#features. 

> https://github.com/joncrlsn/pgdiff

Actively maintained(latest commit 2 days ago). Also being Go it makes it easy to distribute.

> https://github.com/djrobstep/migra

Also has some unanswered issues, but seems the more mature of all.

---

IMO, the choice should be between pgdiff and migra.

(Maybe the pgadmin differ will become the best later)

steve-chavez: Seems pgadmin is already more complete than migra. Discussion here: https://github.com/djrobstep/migra/issues/148

steve-chavez: Some examples of how they generate their diff SQL:

- policy update: [code](https://github.com/postgres/pgadmin4/blob/master/web/pgadmin/browser/server_groups/servers/databases/schemas/tables/templates/row_security_policies/sql/default/update.sql)
- column update: [code](https://github.com/postgres/pgadmin4/blob/master/web/pgadmin/browser/server_groups/servers/databases/schemas/tables/templates/columns/sql/default/update.sql)
- trigger update: [code](https://github.com/postgres/pgadmin4/blob/master/web/pgadmin/browser/server_groups/servers/databases/schemas/tables/templates/triggers/sql/pg/default/update.sql)

Lots of work in the templates there.

kiwicopple: wow, there is some amazing stuff in there. Any way we can use/support this in our CLI though? I guess we would need to copy the SQL to `postgres-meta`?

steve-chavez: pgadmin schema diff looks pretty complete. I've noted it even does unusual cases like `ALTER VIEW v OWNER`([migra doesn't](https://github.com/djrobstep/migra/issues/160)).

2021-03-10-164926_1351x643_scrot

> Any way we can use/support this in our CLI though?

pgadmin4 seems to have a notion of CLI on their codebase, but doesn't seem to be working now. I've requested exposing the Schema Diff tool through CLI on both their mailing list and redmine - still no reply though.

If the above gets done, we could use a: `pgadmin4 --schema-diff db1 db2` in our CLI.

>  I guess we would need to copy the SQL to postgres-meta?

The SQL code I shared was the gist of the schema diff, but there's some other [logic](https://github.com/postgres/pgadmin4/blob/master/web/pgadmin/tools/schema_diff/directory_compare.py).

Translating all the logic to JS could be done, but wouldn't be better to reuse the python code? That way staying in sync with upstream changes(in case we fork it) would be easier.

kiwicopple: > wouldn't be better to reuse the python code?

Definitely, and nicer to support pgadmin too. Let's wait and see what they say, but very cool that the PG community are building this already.

steve-chavez: Unfortunately they're not open to provide a CLI mode: https://redmine.postgresql.org/issues/6304 (you'll need to sign in to see this). They mention that it would take a lot of work.

I still think the pgadmin diff is the best choice. It's already the most complete and has dedicated support. It would provide the best DX.

I've already managed to run it from source. It's a Flask application plus a Backbone.js frontend(seems they're migrating to React).

The functionality for the schema diff was started here: https://github.com/postgres/pgadmin4/commit/45f2e35a99ee691d67c45982211238b8ad40d7de. That gives me an idea of the modules involved for making it work.

How about if I try isolating that python code from the GUI and create a package of our own?

rfcs/0002-mvp.md
steve-chavez: > For example they might want to dump just their "functions".

AFAICT, `pg_dump` cannot dump only functions([SO ref](https://stackoverflow.com/questions/13758003/how-to-take-backup-of-functions-only-in-postgres)).

So if we go the above way, we'd need a combination of 1 and 2

---

Maybe there's another way.

If you look at the Schema Diff output image in https://github.com/supabase/cli/pull/2#discussion_r591909137, you'll notice that it also gives the source DDL. And all the database objects are neatly separated.

The Flask endpoint of the pgadmin schema diff process gives this [big json](https://gist.github.com/steve-chavez/e0f857cb223cfc8a539cd1feceb0f102) as its response. The diff related to the above image is located [here](https://gist.github.com/steve-chavez/e0f857cb223cfc8a539cd1feceb0f102#file-schema_diff_output-json-L4953-L4967).

Notice that the ddl output also has COMMENTs included and it's pretty printed(has newlines).

Checking the big json, all the different types of db objects(event triggers, tables, functions, etc) are there. If we could (ab)use that output, we'd have most of what we need already organized.

rfcs/0002-mvp.md
steve-chavez: > How about if I try isolating that python code from the GUI

Succeeded doing that. Code here: https://gist.github.com/steve-chavez/8c6825abee788394cb5424380c06487c.

Now I would need to wrap the code in a CLI.

kiwicopple: Nice job Steve - that's an awesome effort. 

> Now I would need to wrap the code in a CLI

Yeah, it would be very cool if we can use it here somehow.

steve-chavez: I have the CLI ready on this branch: https://github.com/steve-chavez/pgadmin4/blob/cli/web/cli.py.

I'm now wrapping it on a docker container so it can be tried without building the whole pgadmin4 package.

rfcs/0002-mvp.md
steve-chavez: We can also take advantage of systemd [socket activation](https://blog.developer.atlassian.com/docker-systemd-socket-activation/), which essentially lets a service sleep until a request comes. That coupled with a systemd timer to stop the service after some idle time([so question](https://unix.stackexchange.com/questions/470880/deactivate-a-systemd-service-after-idle-time)), would make resource usage minimum.

rfcs/0002-mvp.md
steve-chavez: Docker container here: https://hub.docker.com/r/stevechavez/pgadmin-schema-diff/

Usage:
```bash
docker run stevechavez/pgadmin-schema-diff \
  'postgres://postgres@host:5432/diff_source' \
  'postgres://postgres@host:5432/diff_target' \
  > diff_db.sql
```

soedirgo:
Closing this as we're transitioning to the Go CLI.

PR #3, opened 2021-02-14
Rfc/dump

kiwicopple:

Mocks up a dump concept


temp/database/models/public.users/columns.sql
```diff
@@ -0,0 +1,82 @@
...
+/*   
+ * data  
+*/
+alter table "public"."users" add column 
+"data"                
+    jsonb -- type
+    default  -- default
+;
+comment on column "public"."users"."data" is 
+"";
```
kiwicopple, line 13: The trickiest part of making the models "clean" seems to be figuring out how to group the comments with the column.

This looks clean:

```sql
create table users (
  id           uuid primary_key,
  first_name   text,
  last_name    text
)
```

but then the comments have to sit below the actual table:


```sql
create table users (
  id           uuid primary_key,
  first_name   text,
  last_name    text
)
comment on column "public"."users"."firstname" is "Some comment";
comment on column "public"."users"."lastname" is "Some comment";
```

As a result it may be better to create all the columns in separate `alter` statements

steve-chavez: The problem with the ALTER COLUMNs is that they're too verbose. I think that our structure should make the point that "SQL is code". Verbosity won't aid in that goal, and instead it discourages looking into the generated SQL.

The comments at the bottom seem good to me. Personally I only use them for documenting APIs, and prefer using regular comments(`-- a comment`) inlined with the columns or constraints.

steve-chavez: I wonder if our `migrator` could somehow generate the `COMMENT ON` statements. So if we have this:

```sql
create table users (
  id           uuid primary_key,
  -- | The user first name
  first_name   text,
  -- | The user last name
  last_name    text
);
```

It would auto-generate:
```sql
create table users (
  id           uuid primary_key,
  first_name   text,
  last_name    text
);

comment on column "public"."users"."first_name" is "The user first name";
comment on column "public"."users"."last_name" is "The user last name";
```

I think that would gets us to a more declarative SQL.

PR #5, opened 2021-03-25
Replace emulator images

soedirgo:

Feature.

The emulator's docker-compose.yml is using opaque images.

## What is the new behavior?

- Use the docker files from [here](https://github.com/supabase/supabase/tree/master/docker),
- remove `eject` command (redundant since we dump the docker-compose.yml on `init`),
- don't `.gitignore` `.supabase` on `init`. It will contain configs etc. in the future which we don't want to ignore (right?), for things like temporary migration scripts (?) we can add individual `.gitignore`s in the corresponding directory.


kiwicopple:
> don't .gitignore .supabase on init. It will contain configs etc.

My line of reasoning was that we use this folder for anything that is epemeral/secret, and then we add a `supabase.config.js` or something similar to the user's root folder for anything that needs to be tracked in git. When the user upgrades the CLI we can also make updates to the files in the `.supabase` folder without it polluting their git history

Vercel's have a similar concept (although they don't use the folder as heavily as we will):

```
# ./.vercel/README.txt

> Why do I have a folder named ".vercel" in my project?
The ".vercel" folder is created when you link a directory to a Vercel project.

> Should I commit the ".vercel" folder?
No, you should not share the ".vercel" folder with anyone.
Upon creation, it will be automatically added to your ".gitignore" file.
```

> remove eject command

I think this is still useful. The reasoning is:

As a frontend dev I might have my own `docker-compose` in the root of my folder, and I don't want it polluted with backend setup (especially one as complicated as Supabase's). However, if I want to "take control" then I can eject. Once I eject, I have to manage everything myself (eg, the `supabase cli` probably won't work as expected)


wdyt?

soedirgo:
Ah, makes sense. I was thinking along the lines of `.github`.

> However, if I want to "take control" then I can eject

I think for some form of control (e.g. changing kong's port) we can do it through the config file. Also, the compose file needs a bunch of other directories now (does the user have `kong`/`postgres` in the project root?), and if I want to take control and use it without the CLI, I can always fetch it from `.supabase`.

PR #15, opened 2021-04-01
Update stop.ts

alteredorange:
Issue docker-compose stop command, instead of down command, to keep containers and data. (So any edits persist between supabase start and supabase stop, instead of wiping containers).

Bug fix, so supabase stop doesn't destroy current docker containers.

supabase stop currently issues docker-compose down, which erases all docker containers. Issuing docker-compose stop will keep containers and allow for persisted data to be kept.

## What is the new behavior?

Docker containers are not destroyed.

### DIFF
--- a/src/commands/stop.ts
+++ b/src/commands/stop.ts
@@ -29,7 +29,7 @@ export default {
     const spinner = spin('Stopping local Supabase...')
 
     await run(
-      'docker-compose --file .supabase/docker/docker-compose.yml --project-name supabase down'
+      'docker-compose --file .supabase/docker/docker-compose.yml --project-name supabase stop'
     ).catch(() => {
       spinner.fail('Error running docker-compose.')
       process.exit(1)


kiwicopple:
At the moment we remove the containers because you might have multiple different Supabase projects and we don't want them to collide.

@soedirgo I think for this to work with different projects we would have to generate the `supabase` docker folder with a suffix somehow? Is that sufficient to namespace the docker-compose?

we would need to add some sort of `remove` and `rebuild` workflows. This needs thought because the user might want to remove the docker containers, but be afraid they are removing all the other supabase config too.

Perhaps we add in a `supabase pause` command, which stops the containers without removing them and leave the `supabase stop` command to function like it does now?

soedirgo:
Yeah, I think specifying a project name would be the way to go, e.g. by prompting for it during `init`.

Also +1 on `supabase pause`.

soedirgo:
On second thought, I think it's better to `build` and `create` during `supabase init`, so `supabase start` does `docker-compose start` and `supabase stop` does `docker-compose stop`. WDYT?

kiwicopple:
when we have migrations ready, we will need this to work with branches too. i.e. if a dev switches to a new branch that has an updated schema, the database needs to get re-created from scratch with the schema that they have in their repo. 

Is there a workflow that suits this use-case?

soedirgo:
> if a dev switches to a new branch that has an updated schema, the database needs to get re-created from scratch

That's where you do a `rebuild` right? Or is the use-case you're referring to the init/start/stop I mentioned above?

kiwicopple:
OK I had a think about your suggested workflow - perhaps `rebuild` works. 

Is there a way that we can detect the git branch when they run `supabase start`? 
Then if the dev switches branches we can compare the previous run - if it is different, we can say something like "looks like you changed branches, would you like to rebuild"?

soedirgo:
Hmm, I think we talked about how each branch has a separate database ala pgsh?

But yeah, the branching feature needs more thought and can be discussed off thread.

kiwicopple:
> each branch has a separate database ala pgsh

Ah yeah - you're absolutely right. Agreed - we can do that in a separate thread/RPC. Should we go ahead with `rebuild`? Also do we need a flag on the `stop` command which can remove the containers too? Otherwise devs are going to end up with a lot of zombie stacks

soedirgo:
Yes, let's go with `rebuild`. For removing containers, maybe `supabase remove` would work (again, to be discussed off thread).

Issue #24, opened 2021-04-21
service role key provided by cli is not working as expected


ff6347:
# Bug report

The service role key provided by the cli seems not to be the right one. When I try to run this command in the shell:

```
curl -X POST 'http://localhost:8888/auth/v1/signup' \
-H "apikey: eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJzdXBhYmFzZSIsImlhdCI6MTYwMzk2ODgzNCwiZXhwIjoyNTUwNjUzNjM0LCJyb2xlIjoic2VydmljZV9yb2xlIn0.necIJaiP7X2T2QjGeV-FhpkizcNTX8HjDDBAxpgQTEI" \
-H "Content-Type: application/json" \
-d '{
  "email": "someone@email.com",
  "password": "123456"
}'
```

I get this response

```
{
  "message":"Invalid authentication credentials"
}
```

The JWT_SECRET (`super-secret-jwt-token-with-at-least-32-characters-long`) is indeed the one that was used to sign this key, but something seems to be missing?

## To Reproduce

clone the repo https://github.com/ff6347/supabase-cli-gives-non-functional-service-role-key
run `npx supabase start`
run `node index.mjs` or the command above
## Expected behavior

>A clear and concise description of what you expected to happen.

I expect the key to work.

## System information

- OS: macOS 11.2.3
- Version of supabase-js:

```
  "dependencies": {
    "@supabase/supabase-js": "1.11.7",
    "supabase": "0.3.8"
  }
```

- Version of Node.js: 14.16.0

kiwicopple:
Hey @ff6347 - the anon key actually acts as the apikey, and you can use the SERVICE_KEY as an Auth Bearer token like this:

```
curl -X POST 'http://localhost:8888/auth/v1/signup' \
-H "apikey: ANON_KEY" \
-H "Authorization: Bearer SERVICE_KEY" \
-H "Content-Type: application/json" \
-d '{
  "email": "someone@email.com",
  "password": "123456"
}'
```

This is also how "logged in users" work too:

```
curl -X POST 'http://localhost:8888/auth/v1/signup' \
-H "apikey: ANON_KEY" \
-H "Authorization: Bearer USER_JWT" \
-H "Content-Type: application/json" \
-d '{
  "email": "someone@email.com",
  "password": "123456"
}'
```

I'll close this, and if I have misunderstood the issue or you have any other questions, feel free to reopen 👍

[...]

PR #25, opened 2021-04-23
add serviceRoleApiKey to kong

sroussey:

bugfix

Only client key is available to use

## What is the new behavior?

service key is also available

## Additional context

https://github.com/supabase/supabase/discussions/1253

Not sure if this is the right kind of fix.


sroussey:
Also see https://github.com/supabase/cli/issues/24

PR #37, opened 2021-05-28
Replace docker-compose start with docker-compose up in start command

ianwalter:

Bug fix or enhancement depending on your point of view

Currently the cli assumes the containers have already been created (via `supabase init`) so running `start` works but if you haven't run `supabase init` and have instead commited the `.supabase` directory, it doesn't work. Using `up` allows you to run it in this case. I can't run `supabase init` because I'm trying to run it in GitHub Actions and it has prompts that I don't see a way of bypassing, but that's a separate issue.

## What is the new behavior?

Using `up` instead of `start` is essentially the same thing as the existing functionality since if the containers have already been created it just runs `start`.

### DIFF
--- a/src/commands/start.ts
+++ b/src/commands/start.ts
@@ -29,7 +29,7 @@ export default {
     const spinner = spin('Starting local Supabase...')
 
     await run(
-      'docker-compose --file .supabase/docker/docker-compose.yml --project-name supabase start'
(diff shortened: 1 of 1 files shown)


soedirgo:
Hmm, gotcha. Thanks for the PR!

PR #40, opened 2021-06-30
Refactor code in menu options

backsoul:
- Feature: The way of validating the options' menu has been changed so that in the future it will be easier to add new options in the menu.

For each option a yes was added, if in the future they want to add more options, the file will be filled with unnecessary conditionals.

## What is the new behavior?

Change the conditionals to literal objects, because when you add a new option, you just add the option name as a brace and the function as a value on the object and don't create a if method.

## Additional context
Before:
Captura de Pantalla 2021-06-30 a la(s) 10 01 30 a  m

After:
Captura de Pantalla 2021-06-30 a la(s) 10 01 50 a  m


src/commands/supabase.ts
```diff
@@ -35,24 +35,21 @@ export default {
...
-    // options.length === 0
+    const versionSupa = () => info(version())
+    for (const opt in options) {
+      let options_functions: any = {
+        h: printHelp,
+        help: printHelp,
+        version: versionSupa,
+        v: versionSupa,
+      }
+      options_functions[opt] || keyNotFound(opt)
```
soedirgo, line 51: Hmm, I think the `options_functions[opt]` is missing the call? Also, there should be a `return` after a valid option.

I think a better alternative without introducing closures is to just `switch` on `opt`.

soedirgo (commented):
Thanks for the PR! Though to be level with you we're planning to rewrite the CLI (WIP [here](https://github.com/soedirgo/cli)).

backsoul:
> Thanks for the PR! Though to be level with you we're planning to rewrite the CLI (WIP [here](https://github.com/soedirgo/cli)).
I would be happy to contribute!!

backsoul (commented):
> Thanks for the PR! Though to be level with you we're planning to rewrite the CLI (WIP [here](https://github.com/soedirgo/cli)).
I would be happy to contribute!

soedirgo:
Closing this as we're transitioning to the Go CLI.

PR #48, opened 2021-08-29
Add more error informations for start command when docker-compose fail

PaulRosset:

I was using the CLI to start the local env, but the `start` was failing, I had no idea why cause the error message was not very informative.

The error was because the port was already in use.

So, I added a more descriptive message to know about the error.

A simple error message:

`Error running docker-compose.`

## What is the new behavior?

It now displays the error message entirely.


soedirgo (approved):
Thanks for the PR!

PR #78, opened 2021-11-25
Add inbucket to local setup

### LINKED ISSUE #71: Add back inbucket (supabase-mail) and disable autoconfirm

mr-bjerre:
Fixes #71 

I've never written in Go so forgive me if this is bad. It seems to be working for me so that's a good start at least. There is one issue though; the generated link which is captured by inbucket have a wrong domain.

It is 
```
http://localhost:54324/?redirect_to=http%3A%2F%2Flocalhost%3A3000%2F&token=J_H61qmv_JcfnXHvkn_QNw&type=signup
```

but it should be

```
http://localhost:8000/?redirect_to=http%3A%2F%2Flocalhost%3A3000%2F&token=J_H61qmv_JcfnXHvkn_QNw&type=signup
```


mr-bjerre:
Last commit fixes the above issue I had.

Currently `GOTRUE_SITE_URL` is hardcoded to port 3000 though. It probably should be a setting instead

mr-bjerre:
Great progress! 😎

Maybe you want to allow to set custom GOTRUE_SITE_URL as well

soedirgo:
I went ahead and made inbucket opt-in (i.e. only when you specify `ports.inbucket`). Can you let me know if it works?

> Currently GOTRUE_SITE_URL is hardcoded to port 3000 though. It probably should be a setting instead

+1, but I'd do this after the OSS dashboard for this one, as I'd like to avoid designing the config up front. It does mean you wouldn't be redirected automatically, but at least you can confirm emails, which is at least something I guess...

mr-bjerre:
Great work. I'll test after lunch here 🙏🏼

mr-bjerre:
Looks pretty good @soedirgo . A couple of things though

1. if you don't opt in then you have `GOTRUE_MAILER_AUTOCONFIRM` to false still.
2. if you don't opt in then you still see `Inbucket URL: http://localhost:` in console.
3. In the console there seems to be empty spaces after `Inbucket URL: http://localhost:54324` so you can't click the link in your terminal. See the underline
Screenshot 2021-11-26 at 12 41 26

mr-bjerre (commented):
Looking good from my point of view !

soedirgo:
Thanks! Re: point 2 above, there actually isn't any space after it, so not sure why it underlined it.

bhvngt:
@mr-bjerre @soedirgo I am trying to use [inbucket/inbucket:latest](https://hub.docker.com/layers/inbucket/inbucket/latest/images/sha256-1f10a0efea694592c06799c729aee1d6d71c9a4f72b73031d4a426ef5f26dfc1?context=explore) docker image as part of [supabase docker compose](https://github.com/supabase/supabase/blob/master/docker/docker-compose.yml). While doing so I get following error when I signup for a user by calling `http://<gotrue>/auth/v1/signup`.

```
 level=error msg="500: Error sending confirmation mail" component=api error="Error sending confirmation email: unencrypted connection" method=POST path=/signup
```

When I use docker container using `supabase/cli`, I see that `inbucket` works without any problem. While going through the supabase/gotrue code trail leading to golang's net/smtp [implementation](https://go.dev/src/net/smtp/auth.go?s=1820:1882#L68) I realised that it is expecting `TLS` connection unless host is `localhost`. 

I was wondering how you are able to bypass this requirement. I went through `supabase/cli` code and this PR and see that you have not used [INBUCKET_SMTP_TLSENABLED](https://github.com/inbucket/inbucket/blob/master/doc/config.md) flag and yet your setup works with `gotrue` server. 

I am using exactly the same [docker image](https://github.com/supabase/cli/blob/2806bc88b6aa0e7e237be8e219c8d1ea242191c0/internal/utils/utils.go#L45) and yet I keep facing this issue.

soedirgo:
> I am trying to use inbucket/inbucket:latest docker image

I think that might be the issue. The CLI uses `inbucket/inbucket:stable` which is a couple months older.

Btw, you might want to consider raising the issue on supabase/supabase instead as the setup is different.

Also you may have better luck trying [this](https://github.com/supabase/supabase/blob/master/docker/dev/docker-compose.dev.yml) setup, though we only use it for smoke tests.

bhvngt:
Thanks a ton @soedirgo. You were right. It was indeed the `inbucket/inbucket:latest` image. With `stable`, it's working out fine for me. 

> Btw, you might want to consider raising the issue on supabase/supabase instead as the setup is different.

Sure will do that. I guess this will still be an issue once inbucket is upgraded to `v3` final version which is what is  `latest` is pointing to.

PR #83, opened 2021-11-29
Upgrade realtime image to v0.19.0

byudaniel:

Upgrades realtime image tag to v0.19.0.

It is currently on v0.18.0, which gives a socket connection error.

## What is the new behavior?

The v0.19.0 resolves the socket connection error.


byudaniel:
After running `supabase start` with the default settings I would get a `REFUSED JOIN realtime:public:todos` error when connecting. After upgrading to v0.19.0 with the same configuration it works. I'm assuming [the fix in v0.18.1](https://github.com/supabase/realtime/releases/tag/v0.18.1) resolved the connection error.

soedirgo:
Looking good, thanks for the PR!

PR #98, opened 2021-12-07
fix: Apply extensions.sql before migrations

everzet:

Some of the migrations might depend on certain extensions. Running extensions *after* migrations breaks such migrations.

This PR should fix the issue by applying `extensions.sql` before any migrations. This is also the current behaviour of [`start`](https://github.com/supabase/cli/blob/main/internal/start/start.go#L548-L581) command.

`extensions.sql` is executed after all migrations on `db reset`

## What is the new behavior?

`extensions.sql` is executed before any migrations on `db reset`

### DIFF
--- a/internal/db/reset/reset.go
+++ b/internal/db/reset/reset.go
@@ -182,8 +182,37 @@ EOSQL
 		}
 	}
 
-	// 3. Apply migrations + extensions + seed.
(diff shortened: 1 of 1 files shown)


soedirgo:
Ah yes, this was a bug. Thanks!

PR #107, opened 2021-12-12
Update docker volume mount flags for SELinux compatibility.

### LINKED ISSUE #106: SELinux compatibility
# Feature request

The current configuration is not using the docker mount flags required for SELinux compatibility ([see this note in the Docker documentation](https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label)). It would be great to make this change so that people using operatings systems with SELinux enabled (ie. Fedora) can easily use the command line interface too.

Replacing the existing mount flags with the SELinux compatible ones should be all that's needed to gain this compatibility. 

There might be alternative ways, but this seems to be the most straight-forward solution, and one officially recommended by Docker. This change should also work out of the box with Moby project and Podman, so it seems to give all the benefits without any downsides (that I know of, at least).

Thank you kindly in advance for your consideration!

jibsaramnim:

Bug fix/feature addition: SELinux compatible volume mounts, useful ie. when the host OS is Fedora. Fixes #106.

Mounts the `pgbouncer.ini`, `userlist.txt`, and `kong.yml` files with the read only (`:ro`) flag.

## What is the new behavior?

Mounts the `pgbouncer.ini`, `userlist.txt`, and `kong.yml` files with the SELinux compatible flag. I have chosen the lowercase z variant (indicating that the bind mount content is shared among multiple containers) just in case the user also uses their own docker containers alongside Supabase's (ie. VSCode devcontainer). This does mean the files are no longer mounted in a read-only way, unfortunately you can't use both flags at the same time. 

## Additional context

I had originally submitted the feature request as I am not familiar with Go development and wasn't sure if I could figure out how to locally build and test things if I made the changes myself, but as it turns out, building was super simple and I was able to test it locally to ensure it's working in my SELinux enabled environment. Go seems pretty cool :).

Thank you!

### DIFF
--- a/internal/start/start.go
+++ b/internal/start/start.go
@@ -696,8 +696,8 @@ EOSQL
 			},
 			&container.HostConfig{
 				Binds: []string{
-					cwd + "/supabase/.temp/pgbouncer.ini:/etc/pgbouncer/pgbouncer.ini:ro",
-					cwd + "/supabase/.temp/userlist.txt:/etc/pgbouncer/userlist.txt:ro",
+					cwd + "/supabase/.temp/pgbouncer.ini:/etc/pgbouncer/pgbouncer.ini:z",
+					cwd + "/supabase/.temp/userlist.txt:/etc/pgbouncer/userlist.txt:z",
 				},
 				PortBindings: nat.PortMap{"5432/tcp": []nat.PortBinding{{HostPort: utils.DbPort}}},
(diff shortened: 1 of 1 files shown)


soedirgo:
Interesting note, thanks! These are temporary files anyhow, so no problem making them non-ro.

soedirgo (approved):

Issue #80, opened 2021-11-27
Documentation improvement: how to access auth settings in local development


emckay:
# Improve documentation

## Link 

[Local development](https://supabase.com/docs/guides/local-development)

## Describe the problem

This page does not describe how to change auth settings when I'm running supabase in local dev. I'd like to change settings like requiring the email address to be confirmed before logging in and enable additional auth providers.

## Describe the improvement

Adding instructions on how to change the auth settings (if this is possible).

## Additional context

Issue field per [this request](https://discord.com/channels/839993398554656828/839993398554656831/913942123064209449) on Discord.

soedirgo:
Thanks for opening this @emckay! I was waiting for the OSS dashboard integration to see if I can offload the settings there, but looks like you'd have to use container env vars both for the CLI and self-host.

This is currently not possible, so I'm turning this issue into a feature request instead.

SamHep0803:
hi, any update on this or a temporary work around?

ydennisy:
@soedirgo any update here?

PR #138, opened 2022-01-14
feat: add external OAuth settings from .env

YuheiNakasaka:
Feature.

Third Party Login feature cannot be enabled easily with cli.

This could be a workaround for #80.

## What is the new behavior?
Add an environment variable prefixed with `GOTRUE_EXTERNAL` from `supabase/.env` in local environment when starting a Docker container of GoTrue with cli. In this way, users can enable/disable the external OAuth service at the time of `supabase start` just by writing the following settings in `supabase/.env`.

```env:.env
# Google OAuth config
GOTRUE_EXTERNAL_GOOGLE_ENABLED=true
GOTRUE_EXTERNAL_GOOGLE_CLIENT_ID=000000000-01234567890abcdefghijk.apps.googleusercontent.com
GOTRUE_EXTERNAL_GOOGLE_SECRET=ABCDEFG-ABC-01234567890ABCDEFG
GOTRUE_EXTERNAL_GOOGLE_REDIRECT_URI=http://localhost:54321/auth/v1/callback
GOTRUE_EXTERNAL_URI=https://accounts.google.com
```

See here for more information on environment variables that can be set.
[https://github.com/supabase/gotrue/blob/master/example.env](https://github.com/supabase/gotrue/blob/master/example.env)

### DIFF
--- a/internal/start/start.go
+++ b/internal/start/start.go
@@ -641,6 +641,14 @@ EOSQL
 			)
 		}
 
+		// Append external OAuth settings if exists.
+		envs := os.Environ()
+		for _, e := range envs {
+			if strings.HasPrefix(e, "GOTRUE_EXTERNAL") {
+				env = append(env, e)
+			}
+		}
+
 		if _, err := utils.DockerRun(
 			ctx,
 			utils.GotrueId,


soedirgo:
Hey Yuhei, thanks for the PR!

This works as a temporary solution, but what I had in mind was to expose this through `supabase/config.json`. `supabase/.env` was an implementation detail for `supabase db remote set` - I might change this in the future.

Let me see if I can put out something in 1-2 weeks' time - otherwise we'll merge this for the time being.

ydennisy:
@soedirgo @YuheiNakasaka - this would be a great feature even if temporarily using `.env` with a plan to move to `config.json`

Could make sense to merge this?

soedirgo:
Clearly I'm way overdue for handling this one, sorry about that. I've finally made a PR here that addresses the issue: https://github.com/supabase/cli/pull/172 - it's part of a larger change related to extending the config.

soedirgo:
Superseded in #172.

PR #156, opened 2022-01-30
Remove a check that normalized URL is equal to raw

ktosiek:
This check was problematic when using passwords with special characters.
Those must be encoded, and this check forced users to guess which exact characters Go would escape, otherwise the URL was considered invalid.

Bug fix

Currently URLs must be provided in the exact form they would be normalized to by net/url.

## What is the new behavior?

Any URL that could be parsed by net/url is accepted.

### DIFF
--- a/internal/db/remote/set/set.go
+++ b/internal/db/remote/set/set.go
@@ -17,8 +17,6 @@ func Run(url string) error {
 	{
 		if parsedUrl, err := u.Parse(url); err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


soedirgo:
Thanks for the PR!

I don't think removing the check helps - the reason it's there in the first place was because some URLs was accepted by `net/url` but not by Postgres's psql tool. IIRC something about having `@` in the password.

Either way, this is now superseded by #170, so closing this. Let me know if that works!

PR #161, opened 2022-02-02
feat: Add dry-run flag to db push command

mcintyre94:

Adds a new flag `--dry-run` on the `supabase db push` command. If enabled this will print out the migrations that would be applied **without** executing anything on the database. 

No support for this flag, AFAIK no way to view which migrations will be applied by `supabase db push`

## What is the new behavior?

Flag is added with the behaviour described above.

Updated help:

```
go run . db push --help                          
Push new migrations to the remote database.

Usage:
  supabase db push [flags]

Flags:
  -d, --dry-run   Print the migrations that would be applied, but don't actually apply them.
  -h, --help      help for push
```

Example with dry-run enabled:

```
go run . db push --dry-run
DRY RUN: migrations will *not* be pushed to the database.
Would apply migration 20220202121715_m1.sql:
CREATE TABLE films (
    code        char(5) CONSTRAINT firstkey PRIMARY KEY,
    title       varchar(40) NOT NULL,
    did         integer NOT NULL,
    date_prod   date,
    kind        varchar(10),
    len         interval hour to minute
);

---

Would apply migration 20220202121756_m2.sql:
CREATE TABLE distributors (
     did    integer PRIMARY KEY DEFAULT nextval('serial'),
     name   varchar(40) NOT NULL CHECK (name <> '')
);

---

Finished supabase db push.
```

Without the flag behaviour remains the same:

```
$ go run . db push
Applying unapplied migrations...
Finished supabase db push.
```

## Additional context

This will make my days less stressful :)

### DIFF
--- a/cmd/db.go
+++ b/cmd/db.go
@@ -68,11 +68,13 @@ var (
 		},
 	}
 
+	dryRun bool
+
 	dbPushCmd = &cobra.Command{
 		Use:   "push",
 		Short: "Push new migrations to the remote database.",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return push.Run()
+			return push.Run(dryRun)
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


mcintyre94:
@soedirgo Would you mind taking a look at this please? Happy to close it if you don't want it but would be good to get your thoughts :)

soedirgo:
Hey, sorry for the late response on this. Can we have the `--dry-run` flag without a shorthand? I'd like to reserve them for frequent flags.

Otherwise looks good - let's get this merged :+1:

mcintyre94:
Thanks! Updated to remove the shorthand:

```
$ go run . db push --help                                                                                                    
Push new migrations to the remote database.

Usage:
  supabase db push [flags]

Flags:
      --dry-run   Print the migrations that would be applied, but don't actually apply them.
  -h, --help      help for push
```

soedirgo (approved):

PR #162, opened 2022-02-03
docs: add error message hint when db push fails because of a prepared statement

mcintyre94:

This commit updates the error message when supabase db push fails to select migrations.

- If the error message contains "prepared statement" and "already exists", then the hint
suggests adding statement_cache_mode=describe to the connection string

I was seeing an error from supabase db push like:

```
Error: Error querying remote database: ERROR: prepared statement "lrupsc_1_0" already exists (SQLSTATE 42P05).
Try running supabase db remote set.
```

## What is the new behavior?

This error now looks like:

```
Error: Error querying remote database: ERROR: prepared statement "lrupsc_1_0" already exists (SQLSTATE 42P05).
Try adding statement_cache_mode=describe to your connection string.
```

## Additional context

Hopefully this will save future devs some debugging time!!

The error is thrown by the first conn.Query (which selects applied migrations)

I found that the cause of this is in pgx and it's caused by having pgbouncer in Transaction mode,
which is the recommended mode in supabase docs for serverless:
https://supabase.com/docs/guides/database/connecting-to-postgres#transaction

This can be resolved without any code changes by adding statement_cache_mode=describe to the connection string:
https://github.com/jackc/pgx/issues/602#issuecomment-532236718

### DIFF
--- a/internal/db/push/push.go
+++ b/internal/db/push/push.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"strings"
 
 	pgx "github.com/jackc/pgx/v4"
 	"github.com/supabase/cli/internal/utils"
@@ -28,7 +29,14 @@ func Run() error {
 	// `schema_migrations` must be a "prefix" of `supabase/migrations`.
 
 	rows, err := conn.Query(ctx, "SELECT version FROM supabase_migrations.schema_migrations ORDER BY version")
+
 	if err != nil {
+		error := err.Error()
+		if (strings.Contains(error, "prepared statement") && strings.Contains(error, "already exists"))	{
+			return fmt.Errorf(`Error querying remote database: %w.
+Try adding `+utils.Aqua("statement_cache_mode=describe")+" to your connection string.", err)
+		}
+
 		return fmt.Errorf(`Error querying remote database: %w.
 Try running `+utils.Aqua("supabase db remote set")+".", err)
 	}


soedirgo:
Does `supabase db remote commit` work with this on your end? I'm getting `Error: unexpected end of JSON input`, which means something wrong happened during schema diff. Using a direct connection (w/o `statement_cache_mode=describe`) works fine.

Taking a step back, I'm leaning towards not supporting PgBouncer - I've seen some gotchas when using PgBouncer with [other tools](https://www.prisma.io/docs/guides/performance-and-optimization/connection-management/configure-pg-bouncer#prisma-migrate-and-pgbouncer-workaround), and Supabase projects always have the option of direct connection anyway. I don't think the performance benefits of connection pooling matters here - the use-case is running migrations, which is infrequent, as opposed to the typical serverless load.

mcintyre94:
Ah that's my mis-understanding, I didn't realise that direct connection was available in addition to connection pooling. I thought that was a kind of DB-level setting that affected all connections. I definitely agree that there's no reason to use connection pooling for the CLI. A much better fix for my use case will be to move the CLI URL to the direct connection. Thanks for the pointer!

On your specific question, I'm unable to get a new supabase instance spun up ATM to check, but I haven't used that command before and wouldn't have tested it.

I'll close this, thanks again for taking a look and giving me a nudge in the better direction! :)

soedirgo:
Thanks for the PR regardless, I think the docs could be improved in this case :+1:

Issue #152, opened 2022-01-26
`supabase start` is unusable in CI due to requirement of interactive shell


everzet:
# Bug report

## Describe the bug

It is near impossible to use CLI in automation scenarios due to it requiring TTY session.

## To Reproduce

Run `supabase start > output` and you would get:

```
Error: inappropriate ioctl for device
```

The above example looks innocent until you realise that it means you wouldn't be able to run even a simpler `supabase start` command during Docker container build or any other non-interactive background process.

## Expected behavior

`supabase start` should either recognize that it is not inside interactive shell or provide an explicit option to disable tty use (e.g. `--no-tty`).

## Additional context

The reason above is an issue is that it makes it impossible to create an isolated automated test environment inside a container. Think Docker or Github CI or Earthly.

PR #163, opened 2022-02-03
Use simple text output if a tty is not available

wilhuff:

Fixes supabase/cli#152.

Currently, if you run `supabase start` in a GitHub Action, or locally, piping output e.g. to a file, you get the following error:

```
Error: inappropriate ioctl for device
```

## What is the new behavior?

With this PR, if both stdin and stdout are not a terminal, the CLI bypasses BubbleTea's UI drawing methods and just logs status and progress messages to stdout.

```bash
supabase start | cat
Pulling images...
Starting database...
Restoring branches...
Setting up initial schema...
Applying supabase/extensions.sql...
Applying migration 20220201010915_add_shared_configs.sql...
Applying supabase/seed.sql...
Starting containers...
Started local development setup.

         API URL: http://localhost:54321
          DB URL: postgresql://postgres:postgres@localhost:54322/postgres
      Studio URL: http://localhost:54323
        anon key: ***
service_role key: ***
```

If running interactively, the behavior is unchanged.

### DIFF
--- a/internal/db/changes/changes.go
+++ b/internal/db/changes/changes.go
@@ -60,7 +60,7 @@ var (
 	diff string
 )
 
-func run(p *tea.Program) error {
+func run(p utils.Program) error {
 	p.Send(utils.StatusMsg("Creating shadow database..."))
 
 	// 1. Create shadow db and run migrations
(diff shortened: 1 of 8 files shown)


wilhuff:
Also, FYI, I can confirm that with this fix I'm able to run `supabase start` as part of a GitHub Actions workflow (unfortunately in a private repo, so I can't link to its output).

soedirgo:
This is great, thanks! Didn't think of using a fake for this.

soedirgo (approved):

PR #178, opened 2022-02-23
env variables for external auth services in config.toml

### LINKED ISSUE #177: pull environment variables into `config.toml`
# Feature request

The new `config.toml` file is cool. I'd like to version it to keep track of its changes, but I don't want to commit e.g. my Google Oauth ClientID value.

For consistency I'd expect the CLI to pull things from my `.env.local` file already setup for my supabase thing. 

Don't know if the following is valid TOML, but just `"{{MY_SECRET}}"` would do as well.

```toml
[auth.external.google]
enabled = true
client_id = {GOOGLE_CLIENT_ID}
secret = {GOOGLE_OAUTH_SECRET}
```

theo-m:

Feature.
Fixes #177 

No parsing of env variables

## What is the new behavior?

Config parser will look up env variables in external auth values. Will only look for unique envs, e.g. `[[ENV1]]-[[ENV2]]` is not handled

## Additional context

- Wrote tests - not too happy about having to write configs to disk but hey.
- Went with `[[XXX]]` instead of `{{XXX}}` as the latter messed up golang's template lib.
- No documentation, don't know where it should go, tell me.


soedirgo:
Btw I'm guessing the env substitution approach you suggested is something like a [Prisma schema](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#specify-a-postgresql-data-source-via-an-environment-variable). Any thoughts of doing `"env(MY_ENV)"` instead? I think it's good to make it clear to new users that it's supplied by env. Or even better: `"${MY_ENV}"` which is more familiar to most. `[[MY_ENV]]` is a bit too similar to TOML's `[[a.b.c]]` I feel like.

> Wrote tests

The first test in the repo 🤡 I definitely feel guilty about this.

> No documentation

Meant to be https://supabase.com/docs/reference/cli/config-reference but it's been outdated for a while so... don't worry about it for now 😅

theo-m:
Went with `env(key)` which I feel is cleaner than `"${KEY}"` - it's a very simple change to make, see [bbdf8bd](https://github.com/supabase/cli/pull/178/commits/bbdf8bd560db3c2718ae886e3470bbc891f3ca44), so feel free to adapt it if you do prefer the other way around :)

PR #196, opened 2022-03-25
fix: don't include a period next to the generated fn filename

darora:
Periods aren't usually a word break character in most terminals, so if
you double-click the name to copy it, the period will be selected as
well. The user would then need to erase the period quite
unnecessarily.

Additionally, impose further restrictions on what a valid function name can be.

### DIFF
--- a/internal/functions/new/new.go
+++ b/internal/functions/new/new.go
@@ -19,12 +19,12 @@ func Run(slug string) error {
 
 	// 2. Validate Function slug.
 	{
-		matched, err := regexp.MatchString(`^[A-Za-z0-9_-]+$`, slug)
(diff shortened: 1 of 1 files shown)


internal/functions/new/new.go
```diff
@@ -48,6 +48,6 @@ serve(() => new Response("Hello World"));
 		}
 	}
 
-	fmt.Println("Created new Function at " + utils.Bold("supabase/functions/"+slug+".ts") + ".")
+	fmt.Println("Created new Function at \"" + utils.Bold("supabase/functions/"+slug+".ts") + "\"")
```
soedirgo, line 51: don't need the `"` I think - it's already constrasted with bold

soedirgo (approved):

PR #208, opened 2022-03-30
fix: auth default disable email confirmations

thorwebdev:
I think for local development we should default to `enable_confirmations = false` to make it easier for folks to get started testing with auth. Wdyt? cc @kangmingtay

### DIFF
--- a/internal/utils/config.go
+++ b/internal/utils/config.go
@@ -343,7 +343,7 @@ enable_signup = true
 # addresses. If disabled, only the new email is required to confirm.
 double_confirm_changes = true
 # If enabled, users need to confirm their email address before signing in.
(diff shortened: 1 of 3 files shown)


kangmingtay:
I'm good with this change as long as its not breaking (something @soedirgo would know better):
1. Are there developers who use the cli for their own local developement / self-hosted setup? 
2. If so, enabling confirmations by default might break the self-hosted setup since they might not have set up a custom smtp server

soedirgo:
Should be fine I think - the CLI is only for local dev, and this change only affects the default config generated by `supabase init`.

Self-hosted setup should still enable confirmations by default.

soedirgo (approved):

PR #215, opened 2022-03-31
Chore/move apikey location

awalias:
rationale is that devs are more likely to look for the apikey config inside `~/` than Application Support dir for dev tools

I'm not as concerned with the deno binary but probably makes sense to have all the config together so devs can check its version if they want to, so I moved it over as well (this one is also easier to move again later if we want since the state of it is managed fully automatically by the CLI)

### DIFF
--- a/internal/functions/deploy/deploy.go
+++ b/internal/functions/deploy/deploy.go
@@ -55,11 +55,7 @@ func Run(slug string, projectRefArg string) error {
 	var newFunctionBody string
 	{
 		fmt.Println("Bundling " + utils.Bold("supabase/functions/"+slug))
-
-		denoPath, err := xdg.ConfigFile("supabase/deno")
-		if err != nil {
-			return err
-		}
+		denoPath := xdg.Home + "/.supabase/deno"
 
 		cmd := exec.Command(denoPath, "bundle", "--quiet", "supabase/functions/"+slug+"/index.ts")
 		var outBuf, errBuf bytes.Buffer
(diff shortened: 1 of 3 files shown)


soedirgo:
Hmm I don't think I'm fully sold on this. The idea of XDG basedirs is that it's a way to store user data in a cross-platform way - I don't have to think about whether the CLI is running on macOS, Linux, Windows, or even FreeBSD, OpenBSD, the spec handles that gracefully. I don't know what the implication of storing in `~` is on Windows - is it going to store on `C:\`?

Another point is that this is really an implementation detail. We don't want the user to actually *see* this directory, everything inside it is managed by the CLI. Not putting it in `$HOME` gets it out of the way. Note that this is a different usecase from a global config file, which the user edits manually - in that case it might be sensible to put it in `$HOME/supabase.toml` to aid discovery.

awalias:
main motivation is to stay consistent with most dev command line tools, including tools like pulumi, aws, ssh etc. who all store their credentials there and in this format:

```bash
% cat ~/.pulumi/
credentials.json  plugins/          workspaces/
```

image

compare this to my Application Support dir which is almost exclusively "Big Letter" desktop applications

image

It personally tripped me up when I went looking to ensure that the correct key was being persisted after running `supabase link`, which is why I think it should only be hidden in a place where most devs know to look if they need to

awalias:
also looks like it's also common practice to store it there on Windows:
> By default, the credentials file is stored here: On Windows: C:\Users\ username \. aws\credentials.

soedirgo (approved):

Issue #149, opened 2022-01-22
`supabase status` to get information about an already started supabase ap


osaxma:
# Feature request

## Use Case

In some cases, I clear/close the terminal, and I lose the following information:

```
Started local development setup.

        API URL: http://localhost:54321
        DB URL: postgresql://postgres:postgres@localhost:54322/postgres
        Studio URL: http://localhost:54323
        anon key: <anon_key>
        service_role key: <service_role_key>
```

If I run `supabase start` again, I get the following error (expected, but does not provide the information mentioned above):
```
Error: supabase start is already running. Try running supabase stop first.
```

## Suggested Solution

It'd be nice to retrieve the information using something like `supabase status`.

osaxma:
Also another thought, it'd be nice to have the ability to specify the supabase project directory without `cd`ing into it :

```
supabase start --project path/to/supabase/directory 
supabase db reset --project path/to/supabase/directory 
supabase stop --project path/to/supabase/directory 
...
...
and so on
```

PR #237, opened 2022-04-04
feature(cmd): add status|info|container (list|restart)

fernandolguevara:

feature

## What is the new behavior?

`supabase status`

image

image

## Additional context

add status cmd #149

`supabase start` shows project info if project is started 

image

`supabase info`

image

 `supabase containers list`

image

`supabase containers restart`

image

### DIFF
--- a/cmd/status.go
+++ b/cmd/status.go
@@ -9,23 +9,10 @@ var statusCmd = &cobra.Command{
 	Use:   "status",
 	Short: "Get Supabase local development status.",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		useShortId := true
-		showKeys := false
-
-		if useFullId, err := cmd.Flags().GetBool("full-id"); err == nil {
-			useShortId = !useFullId
-		}
-
-		if _showKeys, err := cmd.Flags().GetBool("show-keys"); err == nil {
-			showKeys = _showKeys
-		}
-
-		return status.Run(useShortId, showKeys)
+		return status.Run()
 	},
 }
 
 func init() {
-	statusCmd.Flags().Bool("full-id", false, "Display full id")
-	statusCmd.Flags().Bool("show-keys", false, "Display supabase keys")
 	rootCmd.AddCommand(statusCmd)
 }
--- a/cmd/status.go
+++ b/cmd/status.go
@@ -1,0 +1,25 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/supabase/cli/internal/status"
+)
+
+var statusCmd = &cobra.Command{
+	Use:   "status",
+	Short: "Get Supabase local development status.",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		useShortId := true
+
+		if showFullId, err := cmd.Flags().GetBool("full-id"); err == nil {
+			useShortId = !showFullId
+		}
+
+		return status.Run(useShortId)
+	},
+}
+
+func init() {
+	statusCmd.Flags().Bool("full-id", false, "Display full id")
+	rootCmd.AddCommand(statusCmd)
+}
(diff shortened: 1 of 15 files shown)


internal/utils/config.go
```diff
@@ -15,6 +15,11 @@ import (
 	"github.com/spf13/viper"
 )
 
+var (
+	DEFAULT_ANON_KEY        string = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZS1kZW1vIiwicm9sZSI6ImFub24ifQ.625_WdcF3KHqz5amU0x2X5WWHP-OEs_4qj0ssLNHzTs"
+	DEFAULT_SERVER_ROLE_KEY string = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZS1kZW1vIiwicm9sZSI6InNlcnZpY2Vfcm9sZSJ9.vI9obAHOGyVVKa3pD--kJlyxp-Z2zV9UUMAhKpNLAcU"
+)
+
```
soedirgo, line 22: Can use `const` instead of `var`. There's no plan to allow overriding these keys, so can drop the `DEFAULT_` for now. Also to follow Go's naming convention this could be `AnonKey` & `ServiceRoleKey` instead.

cmd/status.go
```diff
@@ -0,0 +1,31 @@
...
+			showKeys = _showKeys
+		}
+
+		return status.Run(useShortId, showKeys)
+	},
+}
+
+func init() {
+	statusCmd.Flags().Bool("full-id", false, "Display full id")
+	statusCmd.Flags().Bool("show-keys", false, "Display supabase keys")
```
soedirgo, line 29: Let's just include everything for now and drop the options - I don't imagine these would pollute the output much.

internal/start/start.go
```diff
@@ -543,7 +539,11 @@ EOSQL
 	// Start Kong.
 	{
 		var kongConfigBuf bytes.Buffer
-		if err := kongConfigTemplate.Execute(&kongConfigBuf, struct{ ProjectId string }{ProjectId: utils.Config.ProjectId}); err != nil {
+		if err := kongConfigTemplate.Execute(&kongConfigBuf, struct {
+			ProjectId     string
+			AnonKey       string
+			ServerRoleKey string
```
soedirgo, line 545: Should be `ServiceRoleKey` instead of `ServerRoleKey`

internal/status/status.go
```diff
@@ -0,0 +1,76 @@
...
+func Run(useShortId bool, showKeys bool) error {
+	// Sanity checks.
+	if err := utils.AssertDockerIsRunning(); err != nil {
+		return err
+	}
+	if err := utils.LoadConfig(); err != nil {
+		return err
+	}
+
+	// List containers.
```
soedirgo, line 24: Let's drop the containers list for now - this might be confusing for new users, and makes the output noisy. Can keep the `Supabase stopped` output though :+1:

soedirgo (commented):
Thanks a lot for the PR! Left some suggestions, but otherwise we're good :+1:

fernandolguevara:
@soedirgo when you have time, can you review the PR again?

soedirgo:
Can you keep the old `supabase status` and omit `info` and `containers`? Containers info is not really in the purview of the CLI, it's just implementation detail - the Docker CLI is better served for this.

fernandolguevara:
@soedirgo after running some supabase projects on my local enviroment I had to deal with some docker issues ... I want to have a cmd on the supabase cli to know the state of project containers or restart all of them without having to call the docker cli by myself making the filter stuff

soedirgo:
Sure, but as I mentioned this is out of the scope of the CLI. If we task the CLI with containers management it will soon become unwieldy - before long you'd have the CLI reimplementing half of the Docker CLI.

PR #238, opened 2022-04-04
feat(start): support pg_graphql

### LINKED ISSUE #233: Enable Graphql in Kong
[Graphql routes are missing from Kong Config](https://github.com/supabase/cli/blob/160e898f83aaf58e826bf726fe5742f39561621e/internal/start/templates/kong_config#L46)

# Feature request
Graphql should be enabled in Kong just like it is in [the master supabase repo self-hosted example](https://github.com/supabase/supabase/blob/21f9e79799bdc37278951c1e0cf600422d38cf25/docker/volumes/api/kong.yml#L98)

Yes, currently we cannot query graphql in the version of supabase produced by the CLI, though #232 fixes the database-level concern there.

I *think* all that needs to happen is the addition of the new route to Kong... but we may also have to update the postgrest container version. Not sure yet.

soedirgo:

Support pg_graphql

## Additional context

Fixes #233.

### DIFF
--- a/internal/start/templates/kong_config
+++ b/internal/start/templates/kong_config
@@ -56,6 +56,24 @@ services:
       - name: key-auth
         config:
           hide_credentials: true
+  - name: graphql-v1
+    _comment: "PostgREST: /graphql/v1/* -> http://rest:3000/rpc/graphql"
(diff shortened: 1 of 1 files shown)


Marviel:
@soedirgo Tried this locally and it seemed to work fine.

kiwicopple (approved):

PR #239, opened 2022-04-05
kong container repeated crashed due to incorrect config

alexanderl19:

Bug fix

Kong container repeatedly crashes, resulting in function serving not working.

## What is the new behavior?

Kong container starts properly and the `supabase function serve` command works as expected.

## Additional context

Kong CORS Config: https://github.com/Kong/kong/blob/21cc6f895bcf8af22da233e929e36ac0e4963321/kong/plugins/cors/schema.lua#L40

Thanks @falsidge for helping me hunt this one down.

### DIFF
--- a/internal/start/templates/kong_config
+++ b/internal/start/templates/kong_config
@@ -73,7 +73,7 @@ services:
         config:
           add:
             headers:
-              - Content-Profile: graphql_public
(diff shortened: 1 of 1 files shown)


soedirgo:
Thanks for the fix!

soedirgo (approved):

PR #241, opened 2022-04-05
feat: make GoTrue SMTP rate limit configurable

vbaluch:

This change makes the GoTrue SMTP rate limit configurable.

Currently the CLI tool starts up GoTrue without a SMTP rate limit configuration. Therefore it uses its one minute interval default, which is annoying for E2E testing involving authentication.

## What is the new behavior?

Now it is possible for a CLI user to adjust the rate limit by setting `max_frequency` in the `[auth.email]` block of `config.toml`.


soedirgo:
How about just hardcoding it to 1 second? I don't imagine people are relying on exact rate limit down to the second, but we can cross that bridge when we get to it.

vbaluch:
> How about just hardcoding it to 1 second? I don't imagine people are relying on exact rate limit down to the second, but we can cross that bridge when we get to it.

Would also work for me, created a [new PR](https://github.com/supabase/cli/pull/244), closing this one.

PR #245, opened 2022-04-08
feat (config.db): [logs] pass the min log level arg using commands

ferantivero:

feature

when using `cli` to work with docker locally `postgres` container size growth wildly based on the current desired configuration as detailed below: 
- `log_statement='all'`
- `log_min_messages='warning'`

related: https://github.com/supabase/realtime/issues/231; https://github.com/supabase/supabase/pull/5481; https://github.com/supabase/supabase/issues/5505

## What is the new behavior?

- No new behavior by default. 
- Devs can opt to change min level of messages from configuration. 

### current version (configured with `warning`)
I've no screenshots but I've seen the container size growth up to `11GiBs` after a few days 

### configured with `fatal`

container size
image

new command
image

### DIFF
--- a/internal/start/start.go
+++ b/internal/start/start.go
@@ -273,8 +273,15 @@ func run(p utils.Program) error {
 	// Start Postgres.
 	{
 		cmd := []string{}
+		var logMinMessagesCommand strings.Builder
+		logMinMessagesCommand.WriteString("log_min_messages=")
+		logMinMessagesCommand.WriteString(utils.Config.Db.LogMinMessages)
 		if utils.Config.Db.MajorVersion >= 14 {
-			cmd = []string{"postgres", "-c", "config_file=/etc/postgresql/postgresql.conf"}
+			cmd = []string{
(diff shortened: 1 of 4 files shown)


soedirgo:
Thanks for the PR! As an alternative, do you know if there's a way to limit Postgres logs to a reasonable size? That way we can alleviate the problem without extending the config.

sweatybridge:
Hi @ferantivero , appreciate your contribution. We have enabled log rotation in local Postgres. I believe this solves the main problem of logs growing indefinitely. 

There's another issue with reducing verbosity of the logs. Adding a min log level config could be one solution. However, I think it's more flexible to expose the log explorer interface so that users can filter by the log level they desire. 

If you agree, I will close this PR in favour of the log explorer.

evelant:
@sweatybridge Is there a way to enable the log explorer during local development? The postgresql logs are still unreadable when realtime is enabled due to the constant printing of huge queries that are polled.

PR #261, opened 2022-04-22
add more descriptive error messages for db push

maxkostow:

Feature - more detailed error messages for `db push`

It's difficult to tell which migration is causing a problem because the migration logs are overwritten each time.

## What is the new behavior?

Includes the actual error that happened and which migration timestamp it happened on.

### DIFF
--- a/internal/db/push/push.go
+++ b/internal/db/push/push.go
@@ -59,7 +59,7 @@ Try running `+utils.Aqua("supabase db remote set")+".", err)
 	conflictErr := errors.New("supabase_migrations.schema_migrations table conflicts with the contents of " + utils.Bold("supabase/migrations") + ".")
 
 	if len(versions) > len(migrations) {
-		return fmt.Errorf("%w; Found " + len(versions) + " versions and " + len(migrations) + " migrations.", err)
+		return fmt.Errorf("%w; Found " + len(versions) + " versions and " + len(migrations) + " migrations.", conflictErr)
 	}
 
 	if !dryRun {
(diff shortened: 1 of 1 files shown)


soedirgo:
Nice one, thanks!

internal/db/push/push.go
```diff
@@ -59,7 +59,7 @@ Try running `+utils.Aqua("supabase db remote set")+".", err)
 	conflictErr := errors.New("supabase_migrations.schema_migrations table conflicts with the contents of " + utils.Bold("supabase/migrations") + ".")
 
 	if len(versions) > len(migrations) {
-		return conflictErr
+		return fmt.Errorf("%w; Found " + len(versions) + " versions and " + len(migrations) + " migrations.", err)
```
soedirgo, line 62: should be `conflictErr`

internal/db/push/push.go
```diff
@@ -80,7 +80,7 @@ Try running `+utils.Aqua("supabase db remote set")+".", err)
 		} else if versions[i] == migrationTimestamp {
 			continue
 		} else {
-			return conflictErr
+			return fmt.Errorf("%w; Expected version" + versions[i] + " but found migration " + migrationTimestamp + " at index " + i + ".", err)
```
soedirgo, line 83: should be `conflictErr`

soedirgo (changes requested):
Just need some minor fixes, then we can merge :+1:

soedirgo (approved):

PR #282, opened 2022-06-20
fix: remove unnecessary exec start call

egor-romanov:

Bug fix: removes second call to start exec

Currently exec start is called twice in docker utils: once in `ExecAttach` and another one in `ExecStart`

## What is the new behavior?

`ExecStart` call is removed


soedirgo:
Thanks!

soedirgo (approved):

PR #283, opened 2022-06-21
feat: add tests and simple docker mock

egor-romanov:

adding tests for cli


.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
```
soedirgo, line 1: Let's omit this? Editor-specific configs shouldn't be committed.

egor-romanov: hey, yeah there is such practice 
also this is debugger setup which can be very handy if anyone would like to run/debug tests on their own

even in other editor there are very similar configs

egor-romanov: that's not prefs, that's debug config...

darora: Is it something more exotic than go test?
If so, I'd suggest including in an editor-agnostic fashion. If not, then there's minimal value in including it

darora: Eh, I don't care about it that much 🤷

egor-romanov: do you have any reasons to not include it?
(I understand when it comes to configs - cause anyone can have different configs and then there may be interference)
 
but when it comes to debug configuration - it is something that can be shared between all devs and it would only add some comfort cause you don't need to write it by yourself

I mean best practices exist for a reason but I just don't see it here in debug configurations being omitted  

(yep, for this case it contains envs. And also it is much easier to run tests when you have configured debug action, that's from personal experience)

soedirgo:
What do you think about doing E2E tests to start with, instead of integration tests with mocks?

A lot of the issues we (& users) have faced wouldn't surface without real Docker containers. I'm worried about using test doubles because it introduces maintenance burden, and this is problematic for this repo especially since we have bus factor == 1. Also, I'm expecting some major refactors in the future, so I don't want to risk having to redo the test setup.

I kind of lapsed on setting up proper testing here, but roughly my idea is to start with E2E testing centered on regressions (I can comb through old issues and include these on our internal manual testing board), and once we have a stable architecture in place, start doing more fine-grained testing.

egor-romanov:
> What do you think about doing E2E tests to start with, instead of integration tests with mocks?

Hey, yeah I was thinking about e2e testing, but as we have intensive docker API usage in cli, it can really be the source of very long running tests (since it is required to pull all the images, wait till it running, applying migration, etc). But what is even more dangerous - is that it can be the source of high flakiness.

Instead with the current setup all tests may be run in fractions of a second. And there should be no flakiness at all.

Also these are not unit tests, cause everything is mocked on API level, we are mocking `docker daemon` with real tcp server and providing another docker host url to the CLI and also `supabase API` is mocked as the real http server. Where we can control if we would like to return error or make the request "stuck" for example.

Also all CLI calls are made with constructions like `init, _, err := suite.cmd.Find([]string{"init"})` ` init.RunE(init, []string{})` which almost completely mimics the real CLI calls. And should be no difference between running it like e2e or integration tests.

> A lot of the issues we (& users) have faced wouldn't surface without real Docker containers. I'm worried about using test doubles because it introduces maintenance burden, and this is problematic for this repo especially since we have bus factor == 1. Also, I'm expecting some major refactors in the future, so I don't want to risk having to redo the test setup.

Not really very worried here from the view of bus factor and everything, cause e2e is less explanatory from this point of view. With current testing I am adding [extensive checks](https://github.com/supabase/cli/pull/283/files#diff-2f1494c5f844cb6bb73d2a3a37a56ae7d9ca319ca60390cd30b9e689ad2d3d08R23-R37) on what exact calls were made to the docker API and/or Supabase API. Which btw spoiled some issues as for me, for example we are checking only if db container is running in status checks.

Of course it is required that you check at least once that the new feature is working: to be sure that your `docker exec` commands are doing what it is supposed to do.

> I kind of lapsed on setting up proper testing here, but roughly my idea is to start with E2E testing centered on regressions (I can comb through old issues and include these on our internal manual testing board), and once we have a stable architecture in place, start doing more fine-grained testing.

I am not so sure in that rule btw, if we have a lot of reoccurring bugs that repeat from time to time on releases, then of course these cases have to be automated. But in most cases with correct git flow and everything it shouldn't be the issue in the modern times. Like what can be the reason to add the same bug again and again (especially when there is only one maintainer)?

We should add some e2e tests after all for the most used user scenarios. But as this app is not very complicated and huge, there won't be a lot of them. For example "start the local infra".

egor-romanov:
Yeah, jic I ve already been in the position of testing go projects, cli included, and I was referring to "how google does it" and their kubectl for example in this case

soedirgo:
Hmm ok, I've got to admit I'm new to testing in Go - for now let's enable these tests but not block merges? Then once we have more confidence we can require passing checks on PRs.

.github/workflows/ci.yml
```diff
@@ -29,6 +29,11 @@ jobs:
...
-        with:
-          version: v1.46
+      - run: |
+          if [ ! -z $(gofmt -l .) ]; then echo 'Make sure to run "go fmt ./..." before commit!' && exit 1; fi
+          go test ./... -race -v -count=1 -cover
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_PAT }}
+          DOCKER_HOST: tcp://127.0.0.1:2375
+          DOCKER_API_VERSION: "1.41"
+          SUPABASE_API_URI: http://127.0.0.1:2376
```
sweatybridge, line 39: Any reason to change the order of these GHA jobs? I would prefer minimising changes in this PR.

internal/secrets/list/list.go
```diff
@@ -38,7 +38,11 @@ func Run() error {
 			return err
 		}
 
-		req, err := http.NewRequest("GET", "https://api.supabase.io/v1/projects/"+projectRef+"/secrets", nil)
+		supabaseAPI := os.Getenv("SUPABASE_API_URI")
+		if supabaseAPI == "" {
+			supabaseAPI = "https://api.supabase.io"
+		}
+		req, err := http.NewRequest("GET", supabaseAPI+"/v1/projects/"+projectRef+"/secrets", nil)
```
sweatybridge, line 45: Instead of introducing a new env var, shall we use [gock](https://github.com/h2non/gock) to mock out any supabase api requests for now? Eventually we want to expose `SUPABASE_HOST` properly as a `viper` config.

internal/link/link.go
```diff
@@ -27,7 +27,11 @@ func Run(projectRef string) error {
 			return err
 		}
 
-		req, err := http.NewRequest("GET", "https://api.supabase.io/v1/projects/"+projectRef+"/functions", nil)
+		supabaseAPI := os.Getenv("SUPABASE_API_URI")
+		if supabaseAPI == "" {
+			supabaseAPI = "https://api.supabase.io"
+		}
+		req, err := http.NewRequest("GET", supabaseAPI+"/v1/projects/"+projectRef+"/functions", nil)
```
sweatybridge, line 34: ditto

internal/secrets/unset/unset.go
```diff
@@ -42,7 +42,12 @@ func Run(args []string) error {
 		}
 		reqBody := bytes.NewReader(secretsNamesBytes)
 
-		req, err := http.NewRequest("DELETE", "https://api.supabase.io/v1/projects/"+projectRef+"/secrets", reqBody)
+		supabaseAPI := os.Getenv("SUPABASE_API_URI")
+		if supabaseAPI == "" {
+			supabaseAPI = "https://api.supabase.io"
+		}
+		req, err := http.NewRequest(
+			"DELETE", supabaseAPI+"/v1/projects/"+projectRef+"/secrets", reqBody)
```
sweatybridge, line 50: ditto

internal/secrets/set/set.go
```diff
@@ -79,7 +79,11 @@ func Run(envFilePath string, args []string) error {
 		}
 		reqBody := bytes.NewReader(secretsBytes)
 
-		req, err := http.NewRequest("POST", "https://api.supabase.io/v1/projects/"+projectRef+"/secrets", reqBody)
+		supabaseAPI := os.Getenv("SUPABASE_API_URI")
+		if supabaseAPI == "" {
+			supabaseAPI = "https://api.supabase.io"
+		}
+		req, err := http.NewRequest("POST", supabaseAPI+"/v1/projects/"+projectRef+"/secrets", reqBody)
```
sweatybridge, line 86: ditto

internal/functions/deploy/deploy.go
```diff
@@ -126,7 +126,11 @@ Enter your project ref: `)
 			return err
 		}
 
-		req, err := http.NewRequest("GET", "https://api.supabase.io/v1/projects/"+projectRef+"/functions/"+slug, nil)
+		supabaseAPI := os.Getenv("SUPABASE_API_URI")
+		if supabaseAPI == "" {
+			supabaseAPI = "https://api.supabase.io"
+		}
+		req, err := http.NewRequest("GET", supabaseAPI+"/v1/projects/"+projectRef+"/functions/"+slug, nil)
```
sweatybridge, line 133: ditto

internal/functions/delete/delete.go
```diff
@@ -63,7 +63,11 @@ func Run(slug string, projectRefArg string) error {
 			return err
 		}
 
-		req, err := http.NewRequest("GET", "https://api.supabase.io/v1/projects/"+projectRef+"/functions/"+slug, nil)
+		supabaseAPI := os.Getenv("SUPABASE_API_URI")
+		if supabaseAPI == "" {
+			supabaseAPI = "https://api.supabase.io"
+		}
+		req, err := http.NewRequest("GET", supabaseAPI+"/v1/projects/"+projectRef+"/functions/"+slug, nil)
```
sweatybridge, line 70: ditto

sweatybridge (commented):
Just a few comments to minimise changes to `internal`. The rest of integration tests look great!

.vscode/launch.json
```diff
@@ -0,0 +1,21 @@
...
+      "name": "Test Package",
+      "type": "go",
+      "request": "launch",
+      "mode": "test",
+      "cwd": "${workspaceRoot}",
+      "program": "integration",
+      "env": {
+        "DOCKER_HOST": "tcp://127.0.0.1:2375",
+        "DOCKER_API_VERSION": "1.41",
+        "SUPABASE_API_URI": "http://127.0.0.1:2376"
```
sweatybridge, line 17: Remember to remove this env var once we introduce gock.

.github/workflows/ci.yml
egor-romanov: Yes, lint is common to run before tests, because it is faster and will fail pipeline faster

sweatybridge: Jobs on GitHub actions are run in parallel by default https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow

internal/secrets/list/list.go
egor-romanov: We agreed on using gock in unit tests. 

Since this extremely unlikely to break anything, i think it is ok as temp solution, and can be reworked in future (because not everything should be in one pr right).

sweatybridge: Well, I'm hoping to reduce the scope of this PR by reverting this env var.

Especially since we have no control over users who set this env var once it's released. It's a small problem given the number of cli users right now. But for the peace of mind, I'd rather not introduce new config that's used for testing only. 

When we are ready to expose it for changing the supabase host that cli talks to, we can look into it again. That way we can also plan its naming convention more carefully. What do you think?

.github/workflows/ci.yml
egor-romanov: Oh, damn, you right, I have not noticed that these are jobs and not steps
Will fix it, thanks 👍

internal/secrets/list/list.go
egor-romanov: I don't think this is such a problem, but we may change env var to be sure noone will override it. 
To `SUPABASE_CLI_SUPABASE_API_URI` for example. With this I doubt any of our users accidentally override it. 

This is good for us since we actually will be able to test some new things on staging also.

sweatybridge: > This is good for us since we actually will be able to test some new things on staging also.

Yup, I agree it is good for us.

@soedirgo suggests `SUPABASE_INTERNAL_API_URL` to be more concise. I'm ok with that also.

soedirgo: I don't mind having it as an env, as long as we make it clear that it's for internal use only, e.g. by renaming it to `_SUPABASE_ADMIN_API_HOST` or `SUPABASE_INTERNAL_ADMIN_API_HOST`.

btw we started off using `viper`, but moved off of it - iirc something to do with flexibility and fine-grained error messages.

sweatybridge: > btw we started off using viper, but moved off of it - iirc something to do with flexibility and fine-grained error messages.

Hmm I would be interested to read the discussion thread ^. The way `viper` standardises precedence order sounds too good to pass up.

egor-romanov: `SUPABASE_INTERNAL_API_HOST` 👍

egor-romanov: I will apply all suggested changes later today

soedirgo: Err, `s/we started off/I started off/` 😄. We still use it here: https://github.com/supabase/cli/blob/9c4b8a9801a50a553fb08e5b79180999482913c9/internal/utils/config.go#L289-L307

These are for `supabase/config.{json,toml}` though which is associated w/ the project and not the CLI - these would be different once we have some sort of global config (i.e. the ones in `~/.supabase`, not `./supabase`

soedirgo: also we can prob remove this deprecated config handling code now - will do it in another PR

cmd/root.go
```diff
@@ -22,3 +22,7 @@ func Execute() {
 func init() {
 	rootCmd.SetVersionTemplate("{{.Version}}\n")
 }
+
+func NewRootCmd() *cobra.Command {
```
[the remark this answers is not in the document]
egor-romanov: instantiate will require full rework of cmd package, so I changed func name to `GetRootCmd`

sweatybridge (approved):

.github/workflows/ci.yml
```diff
@@ -17,7 +19,14 @@ jobs:
 
       - run: |
           if [ ! -z $(gofmt -l .) ]; then echo 'Make sure to run "go fmt ./..." before commit!' && exit 1; fi
-          go test ./...
+          mkdir  /home/runner/.supabase && chown -R runner /home/runner/.supabase
+          go test ./... -race -v -count=1 -cover -coverprofile=coverage.out
+          go tool cover -func coverage.out | grep total
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_PAT }}
```
sweatybridge, line 26: probably not needed

sweatybridge (commented):
we should probably move the integration tests to `tests` directory following the standard layout here https://github.com/golang-standards/project-layout/tree/master/test

.vscode/launch.json
sweatybridge: > (yep, for this case it contains envs. And also it is much easier to run tests when you have configured debug action, that's from personal experience)

How about calling `t.Setenv` in all test suites so that `go test ./...` command can succeed without requiring any env? It will also help the tests to be more specific to whether they require `DOCKER_HOST` or `SUPABASE_INTERNAL_API_HOST` only.

sweatybridge (changes requested):
we should probably move the integration tests to `tests` directory following the standard layout here https://github.com/golang-standards/project-layout/tree/master/test

.vscode/launch.json
sweatybridge: @egor-romanov This is how I removed explicit env var from running tests https://github.com/supabase/cli/pull/283/files#diff-7f11b9ec624784dcb36925610adb78aa39e9c1e06bf867ddfba5bf99cb9b1f62R38. I believe `launch.json` is no longer necessary after this change.

.github/workflows/ci.yml
soedirgo: This is needed for the goreleaser Action to push to supabase/homebrew-tap: https://github.com/goreleaser/goreleaser-action#limitation

sweatybridge: This is on ci.yaml, not release.yml.

.github/workflows/ci.yml
```diff
@@ -2,6 +2,8 @@ name: CI
 
 on:
   pull_request:
+    branches:
+      - main
```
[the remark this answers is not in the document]
egor-romanov: I can’t say why exactly, but workflows were not triggered since yesterday and until i ve added branch name. 
also I don’t see any other maintained branches

.github/workflows/ci.yml
```diff
@@ -30,5 +34,3 @@ jobs:
           go-version: 1.18
 
       - uses: golangci/golangci-lint-action@v3
-        with:
-          version: v1.46
```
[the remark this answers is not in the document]
egor-romanov: @sweatybridge

egor-romanov:
@sweatybridge after some of your latest changes total coverage dropped by 10%

it was 27% on my last commit and now 16%

egor-romanov:
https://github.com/supabase/cli/runs/7462491975?check_suite_focus=true#step:4:263

https://github.com/supabase/cli/runs/7455234657?check_suite_focus=true#step:4:267

.github/workflows/ci.yml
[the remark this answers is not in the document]
sweatybridge: ok my bad here. lets pump to version 1.47

sweatybridge:
> @sweatybridge after some of your latest changes total coverage dropped by 10% it was 27% on my last commit and now 14.6%

@egor-romanov  I believe the drop is due to inclusion of more files, such as `integration/mock`. The flag we need to tweak is `-coverpkg ./...` which adds all files to coverage.

Previously 27% was from unit test only, not including integration tests. You can verify it by checking out the current `main` branch and run the following script.

```bash
go test ./... -coverprofile=coverage.out
go tool cover -func coverage.out
```

egor-romanov:
Hey, yep, I’ve checked commits and found this, we can add except for test package

.github/workflows/ci.yml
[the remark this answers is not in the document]
soedirgo: It can be useful for short-lived branches, e.g. if I want to make a PR into this PR

sweatybridge (approved):
lgtm once we agree on `launch.json`

PR #293, opened 2022-07-01
fix: handle SIGINT/SIGTERM for the start command

PaulRosset:

This PR introduces a graceful shutdown for SIGINT/SIGTERM signals for the start command.

Not implemented yet

## What is the new behavior?

Handling SIGINT/SIGTERM signals.


sweatybridge:
Hi @PaulRosset , apologies for getting back to you late on this PR.

The bubbletea terminal already handles signals. What's missing is to propagate cancellation to the command context. I've updated your PR to handle this more consistently https://github.com/supabase/cli/blob/main/internal/db/remote/commit/commit.go#L57.

sweatybridge (approved):

Issue #277, opened 2022-06-13
No `seed.sql` after `supabase init`


geyermichael:
# Bug report

## Describe the bug

There is no `supabase/seed.sql` after using `supabase init` as it is mentioned in the documentation.

## To Reproduce

Just follow the documentation on `https://supabase.com/docs/guides/local-development`.

## Expected behavior

After using `supabase init` there should be a file called `seed.sql`.

## System information

- OS: macOS
- Version of supabase cli: 0.26.1

## Additional context

I created the `seed.sql` file myself and could follow the documentation without any additional error.
So it is just the part of creating the file by default after `supabase init`

vejja:
Another comment here: 
If the <code>seed.sql</code> file contains too many lines, the seeding procedure will fail silently.
It makes it hard to understand why the seeding does not work.

Not sure how many lines is too many, but this happened to me because I was generating the <code>seed.sql</code> file from a <code>pg_dump</code> of an existing test database.
It would be good to have some kind of console output for <code>supabase db reset</code> on failure to help debugging.

snorremd:
@vejja This explains a lot. I've been struggling to get my `seed.sql` file to work. It performs a bunch of SQL insert commands to populate the database with data from my old hobby project so it's easier to check that my queries returns the same values as before. Somehow it has been failing for me as well and I've needed to manually run the queries after running the migration.

thomasmol:
The problem @vejja is describing is still happening to me. I am on version 1.11.7 of the cli.
My seed.sql has roughly 270 insert statements but I get this error:
``` 
Seeding data supabase/seed.sql...
Error: ERROR: relation "challenges" does not exist (SQLSTATE 42P01)
```
```challenges``` is one of my table names.

sweatybridge:
Hi @thomasmol, is your `challenges` table created in `seed.sql` or a separate migration file? DDL statements like create table are not designed to be used in `seed.sql`.

If you are sure that the table is created in its own migration, you might also want to prefix references to `challenges` table with its schema, for eg. `"public"."challenges"`. This helps to avoid ambiguity resolving tables with default search path.

Let me know if either solution works for you.

saltcod:
Quick note to anyone who stumbles on this: 

I ran across this today, not realizing I was running an old version of the cli. Make sure you're version is up to date if you run into this! 

https://supabase.com/docs/guides/cli#updates

CareTiger:
[...]

Issue #274, opened 2022-05-31
Remove "Body exceeded 1mb limit" error


vbylen:
Hi there,

I'm trying to seed my local database with 1000 users and 1000 posts.

However I'm greeted with the following error:

`"Body exceeded 1mb limit"`

So postgres is complaining that the `seed.sql` file is too big.

I've tested inserting a smaller number of rows and it works fine.

Any ideas how I could get around this?

Thanks!

eeston:
Hello,

I've currently toying around with a seed file that's around 3mb which is full of mostly JSON plus a few functions for inserting the data.

I'm running 1.8.2 and I'm getting the following error while trying to seed... `Error: bufio.Scanner: token too long`
I also tried to seed in the dashboard (local) but the limit there is 1mb. 

Would importing from a csv be the prefered method here?

eeston:
Hello @sweatybridge,

I've currently toying around with a seed file that's around 3mb which is full of mostly JSON plus a few functions for inserting.

I'm running 1.8.2 and I'm getting the following error while trying to seed...
Error: bufio.Scanner: token too long.

I also tried to seed in the dashboard (local) but the limit there is still 1mb (Body exceeded 1mb limit).

Is there another way to seed large amounts of data?

sweatybridge:
Hi @eeston , we have made the scanner buffer size configurable. You might need to play around with buffer sizes to fit your seed file. Perhaps 5mb could be a good start

```bash
SUPABASE_SCANNER_BUFFER_SIZE=5mb supabase db reset
```

> I'll need to do the same in my cloud instance of supabase too. I assume the 1mb will still be an issue on that dashboard?

Yes, the hosted dashboard has the same body size limit. We will look into extending the CLI to push your seed data to cloud instance using `supabase db push`. The related ticket is https://github.com/supabase/cli/issues/160

eeston:
Fantastic...I'll test this at the weekend. Thanks for your help @sweatybridge

eeston:
Hi @sweatybridge,

Just had a play around with this and unfortunately it still fails with the same error. Not a big deal locally but for the hosted dashboaed I think I'll end up splitting the data into multiple chunks. I assume the csv upload option is also limited to 1mb?

sweatybridge:
> I assume the csv upload option is also limited to 1mb?

I just checked with our support team. The csv upload option does not impose such size limits. You should be able to upload 3mb of seed data using it.

kiwicopple:
[...]

PR #295, opened 2022-07-03
Support bigger seed.sql files using docker copy

snorremd:

Bug fix

For seed.sql files that exceed the maximum argument length of the Linux distribution used to run the Postgres container the `postgres -c` command will fail. Supabase CLI will then silently fail to apply the seed file.

See related [StackOverflow Max size of string CMD that can be passed to Docker](https://stackoverflow.com/a/70738704) answer about command argument sizes. 

https://github.com/supabase/cli/issues/277 (See related comment)
https://github.com/supabase/cli/issues/274 (I think this might be related, but not sure)

## What is the new behavior?

A new Docker utility function DockerAddFile can be used to write a byte buffer to a container, in this case the seed.sql file. Then migrating the seed file is just a matter of using the file option of the `pgsql` command rather than the command option.

## Additional context

Let me know if the code needs some adjustments and if the selected approach is good or not.

### DIFF
--- a/internal/db/reset/reset.go
+++ b/internal/db/reset/reset.go
@@ -206,9 +206,15 @@ EOSQL
 				// skip
 			} else if err != nil {
 				return err
+			}
+
+			err = utils.DockerAddFile(ctx, utils.DbId, "seed.sql", content)
+
+			if err != nil {
+				return err
 			} else {
 				out, err := utils.DockerExec(ctx, utils.DbId, []string{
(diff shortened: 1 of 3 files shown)


soedirgo:
LGTM, thanks for this!

soedirgo (approved):

snorremd:
> LGTM, thanks for this!

My pleasure. Thank you for reviewing and merging this so fast! This will improve my development flow so much!

Issue #268, opened 2022-05-17
Support for GitPod


Haberkamp:
# Bug report

## Describe the bug

It seems that it's not possible to start supabase in a gitpod workspace

## To Reproduce

Steps to reproduce the behaviour, please provide code snippets or a repository:

1. Create a new gitpod workspace
2. Install supabase via homebrew
3. Run supabase init and after that supabase start

## Expected behaviour

All containers should start successfully without an error.

## Screenshots

![Bildschirmfoto 2022-05-17 um 14 19 26](https://user-images.githubusercontent.com/35109813/168809694-738f028f-3e41-40ca-8742-33f4636460a6.png)

## System information

- OS: Remote GitPod Server (probably a Linux machine)
- Browser (if applies) Not relevant
- Version of supabase-js: 1.34.0
- Version of supabase-cli: 0.26.0
- Version of Node.js: 16.15.0

soedirgo:
I'm not familiar with GitPod, but can you do everything the Docker CLI can do inside it?

Haberkamp:
Yes, I was able to pull a docker image for adminer (a tool for database management with a web ui) and I could reach the application. So you're able to use the Docker CLI in GitPod.

sweatybridge:
> When working inside a GitPod instance you are already inside a docker container.

Running CLI in docker would require docker-in-docker support. I have not tested this myself but it appears to be supported on GitPod https://github.com/gitpod-io/gitpod/issues/52#issuecomment-738629624

zivbk1:
I have a similar problem, but different errors when I run `supabase start`.

```
failed to register layer: ApplyLayer exit status 1 stdout:  stderr: failed to Lchown "/usr/local/lib/node_modules/pm2/node_modules/nssocket/.npmignore" for UID 1389985163, GID 386085923 (try increasing the number of subordinate IDs in /etc/subuid and /etc/subgid): lchown /usr/local/lib/node_modules/pm2/node_modules/nssocket/.npmignore: invalid argument
```

and

```
Error: failed to register layer: ApplyLayer exit status 1 stdout:  stderr: failed to Lchown "/usr/local/lib/node_modules/pm2/node_modules/nssocket/.npmignore" for UID 1389985163, GID 386085923 (try increasing the number of subordinate IDs in /etc/subuid and /etc/subgid): lchown /usr/local/lib/node_modules/pm2/node_modules/nssocket/.npmignore: invalid argument
Try rerunning the command with --debug to troubleshoot the error.
[...]

PR #296, opened 2022-07-04
Gitpod Support

akoenig:

We use Gitpod internally and therefore we were eagerly searching for a solution of https://github.com/supabase/cli/issues/268. Due to the fact that we were able to run `supabase` in a Gitpod environment by using the official `docker-compose` [setup](https://github.com/supabase/supabase/tree/bc9992ff15a833b5bb0d9714f876eed668bf2757/docker), we had our Proof of Concept that running supabase within Gitpod is doable. Next, we thought it might be a good idea to look into the `cli` and see if we can make it work there too 🙂 

So, although Gitpod is a containerized environment, it turned out that the culprit is something else. Instead, it looks like the CLI is behaving differently when the environment is using Docker in [Rootless mode](https://docs.docker.com/engine/security/rootless/).

Using the `supabase` CLI within a Gitpod environment failed as described in https://github.com/supabase/cli/issues/268.

## What is the new behavior?

Compiling a CLI with the changes in this PR leads to a scenario in which all `supabase` commands, like `supabase start`, work as expected.

## Additional context

We also tested the change in a non-containerized environment and it behaves as expected there too.

### DIFF
--- a/internal/db/branch/delete/delete.go
+++ b/internal/db/branch/delete/delete.go
@@ -37,7 +37,7 @@ func Run(branch string) error {
 
 	{
 		out, err := utils.DockerExec(context.Background(), utils.DbId, []string{
-			"dropdb", "--username", "postgres", "--host", "localhost", branch,
+			"dropdb", "--username", "postgres", "--host", "127.0.0.1", branch,
 		})
 		if err != nil {
 			return err
--- a/internal/db/changes/changes.go
+++ b/internal/db/changes/changes.go
@@ -70,7 +70,7 @@ func run(p utils.Program) error {
 		out, err := utils.DockerExec(
 			ctx,
 			utils.DbId,
-			[]string{"createdb", "--username", "postgres", "--host", "localhost", utils.ShadowDbName},
+			[]string{"createdb", "--username", "postgres", "--host", "127.0.0.1", utils.ShadowDbName},
 		)
 		if err != nil {
 			return err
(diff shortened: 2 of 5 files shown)


soedirgo:
Interesting, do you know why it behaves different with `localhost` vs. `127.0.0.1`?

There's a failing test - `go fmt ./...` should fix that - and then we can merge this. Thanks for the PR!

soedirgo (approved):

Haberkamp:
@akoenig thank you for your contribution. I have one question for you. How are you able to access the database via supabase studio. It tells me that it is always connecting. I use the default settings.

Bildschirmfoto 2022-07-18 um 13 27 36

soedirgo:
@Haberkamp can you check the logs of the containers from Docker Desktop etc.? I suspect the `supabase_rest_...` container is having issues.

Haberkamp:
Yes, this is the error I get. I cannot connect to the database. The postgres container seems to be working fine.
Bildschirmfoto 2022-07-19 um 08 03 07

soedirgo:
How about `docker logs supabase_db_coespace-next`? Could be a failing migration.

Haberkamp:
This is the log output.

Bildschirmfoto 2022-07-19 um 10 38 54

soedirgo:
Hmm, I don't have any lead so far - I'll just reopen the issue for now.

egor-romanov:
I have seen a lot of access control issues when in browser console, when I tried to open studio from gitpod

PR #299, opened 2022-07-12
Update Postgres14 image version

alex-ketch:

Updates the image version used for Postgres v14 to use the [latest published tag](https://hub.docker.com/r/supabase/postgres/tags).

## Additional context

During local development I encountered a bug with the `pg_graphql` plugin that has already been [fixed upstream](https://github.com/supabase/pg_graphql/pull/154).
I imagine there are other fixes that would be nice to have during local development too.

Thanks, and let me know if this isn't the proper way to update image versions.

### DIFF
--- a/internal/utils/config.go
+++ b/internal/utils/config.go
@@ -176,7 +176,7 @@ func LoadConfig() error {
 			DbImage = "supabase/postgres:13.3.0"
 			InitialSchemaSql = initialSchemaPg13Sql
 		case 14:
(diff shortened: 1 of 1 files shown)


soedirgo:
This is perfect, thanks!

soedirgo (approved):

soedirgo:
Had to revert this because of https://github.com/supabase/cli/issues/300 - will update the image & resolve the issue in a separate PR

Issue #300, opened 2022-07-12
`supabase start` command fails after upgrading to `0.29.2`


KirioXX:
# Bug report

## Describe the bug

After updating to supabase CLI version 0.29.2, when running `supabase start` I get this error
```sh
Error: Error starting database: WARNING:  you need to manually restart any running background workers after this command
```
I already tried to remove all containers and images with the same result.

## To Reproduce

Steps to reproduce the behaviour, please provide code snippets or a repository:

1. Upgrade to subase cli version 0.29.2
2. Remove all containers and images with for example `docker container prune && docker image prune -a`
3. Run `supabase start`

## Expected behaviour

The service starts without errors.

## System information

- OS: macOS Monterey (12.4)
- Version of supabase-cli: 0.29.2

soedirgo:
Should be fixed now - can you try upgrading to 0.29.3?

djhi:
Confirmed on my side :+1:

KirioXX:
Yes works for me too. Thanks, @soedirgo 🙌

PR #301, opened 2022-07-12
feat: (inbucket) expose smtp port

djhi:

This PR exposes the inbucket SMTP port so that an application does not need to set up another local dev tool for SMTP.

Only the UI port is exposed.

## What is the new behavior?

The SMTP port is exposed and configurable through the TOML file.

## Tasks

- [x] Exposes the SMTP port
- [x] Exposes the POP3 port
- [ ] Test everything locally

## Notes

I couldn't test it locally yet because of #300 which also happens with a freshly built supabase binary :sob:


djhi:
Hi supabase team! Is there anything you'd like me to change or add on this PR?

soedirgo:
Sorry for the late review! Can we make these ports optional, and simply not expose the ports when not specified? Existing users wouldn't have these ports configured, and the CLI would error out on loading the config.

djhi:
Sure, I'll try to do that. Thanks!

djhi:
Done. I have not much experience in Go though so I'd welcome any feedback you might have.

soedirgo:
LGTM, thanks!

PR #303, opened 2022-07-20
Add unit tests for init command

sweatybridge:

Unit tests with in-memory file system.

Config tests writes out to local filesystem.

## What is the new behavior?

- Uses afero as the abstract file system (`io/fs` only supports read operations, [read/write proposal](https://github.com/golang/go/issues/45757) is in progress)
- Simplifies config tests to use afero
- Adds unit test for init command

## Additional context


internal/init/init.go
```diff
@@ -21,33 +22,33 @@ var (
...
 )
 
-func Run() error {
-	if err := run(); errors.Is(err, errAlreadyInitialized) {
+func Run(fsys afero.Fs) error {
+	if err := run(fsys); errors.Is(err, errAlreadyInitialized) {
 		return err
 	} else if err != nil {
-		_ = os.RemoveAll("supabase")
+		fsys.RemoveAll("supabase")
```
sweatybridge, line 29:
```suggestion
		_ = fsys.RemoveAll("supabase")
```

soedirgo (approved):
Nice one! 🔥

PR #306, opened 2022-07-20
test: add unit tests for login command

sweatybridge:

unit tests

login reads from `os.Stdin` and writes to `~/.supabase/access_token`

## What is the new behavior?

- dependency inject `io.Reader` and `afero.Fs` for unit testing
- refactor regex compilation to a global variable
- generate random token in test cases
- 94.7% coverage

## Additional context

### DIFF
--- a/internal/init/init_test.go
+++ b/internal/init/init_test.go
@@ -48,8 +48,6 @@ func TestInitCommand(t *testing.T) {
 		assert.Error(t, Run(fsys))
 	})
 
-	// TODO: throws error on failure to find git root
-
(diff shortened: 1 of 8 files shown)


internal/init/init_test.go
```diff
@@ -48,6 +48,8 @@ func TestInitCommand(t *testing.T) {
 		assert.Error(t, Run(fsys))
 	})
 
+	// TODO: throws error on failure to find git root
+
```
soedirgo, line 52: It should do nothing on failure to find git root, e.g. if the user isn't using git

soedirgo (approved):

PR #323, opened 2022-08-01
fix(functions/deploy): pin deno version for now

soedirgo:
Fixes: https://github.com/supabase/supabase/issues/7878

Latest Deno version's bundling is having some issues - downgrade to 1.20.3 for now, and later we can switch back to `latest` and use ESZip

### DIFF
--- a/internal/functions/deploy/deploy.go
+++ b/internal/functions/deploy/deploy.go
@@ -89,7 +89,7 @@ Enter your project ref: `, utils.GetSupabaseDashboardURL())
 			}
 		}
 
-		cmd := exec.Command(denoPath, "bundle", "--quiet", functionPath+"/index.ts")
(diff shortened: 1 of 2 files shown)


internal/functions/deploy/deploy.go
```diff
@@ -89,7 +89,7 @@ Enter your project ref: `, utils.GetSupabaseDashboardURL())
 			}
 		}
 
-		cmd := exec.Command(denoPath, "bundle", "--quiet", functionPath+"/index.ts")
+		cmd := exec.Command(denoPath, "bundle", "--no-check=remote", "--quiet", functionPath+"/index.ts")
```
soedirgo, line 92: latest version uses `no-check` by default, but 1.20.3 does not

thorwebdev (approved):
LGTM, thanks.

PR #326, opened 2022-08-03
feat: add db test command using pgtap

sweatybridge:

feature: `supabase db test`

no cli native way to run tests

## What is the new behavior?

- installs `pg_prove` on `supabase_db_cli` container once
- copies `supabase/tests` into `supabase_db_cli` container
- enable pgtap extension temporarily
- runs `pg_prove` via docker exec and prints out results

## Additional context

Example `order_test.pg`
```sql
BEGIN;
SELECT plan(1);

SELECT has_column(
    'public',
    'order',
    'status',
    'status should exist'
);

SELECT * FROM finish();
ROLLBACK;
```

Output
```bash
$ supabase db test
supabase/tests/order_test.pg .. ok
supabase/tests/pet_test.pg .... ok
All tests successful.
Files=2, Tests=2,  1 wallclock secs ( 0.01 usr  0.00 sys +  0.04 cusr  0.02 csys =  0.07 CPU)
Result: PASS
```

### DIFF
--- a/internal/utils/templates/initial_schemas/14.sql
+++ b/internal/utils/templates/initial_schemas/14.sql
@@ -104,6 +104,20 @@ COMMENT ON EXTENSION pgjwt IS 'JSON Web Token API for Postgresql';
 
 
 --
+-- Name: pgtap; Type: EXTENSION; Schema: -; Owner: -
+--
+
+CREATE EXTENSION IF NOT EXISTS pgtap WITH SCHEMA extensions;
+
+
+--
+-- Name: EXTENSION pgtap; Type: COMMENT; Schema: -; Owner:
+--
+
+COMMENT ON EXTENSION pgtap IS 'Unit testing for PostgreSQL';
+
+
+--
 -- Name: uuid-ossp; Type: EXTENSION; Schema: -; Owner: -
 --
 
--- a/internal/utils/templates/initial_schemas/14.sql
+++ b/internal/utils/templates/initial_schemas/14.sql
@@ -150,20 +150,6 @@ COMMENT ON EXTENSION pgjwt IS 'JSON Web Token API for Postgresql';
 
 
 --
--- Name: pgtap; Type: EXTENSION; Schema: -; Owner: -
---
-
-CREATE EXTENSION IF NOT EXISTS pgtap WITH SCHEMA extensions;
-
-
---
--- Name: EXTENSION pgtap; Type: COMMENT; Schema: -; Owner:
---
-
-COMMENT ON EXTENSION pgtap IS 'Unit testing for PostgreSQL';
-
-
---
 -- Name: uuid-ossp; Type: EXTENSION; Schema: -; Owner: -
 --
 
--- a/cmd/db.go
+++ b/cmd/db.go
@@ -158,7 +158,8 @@ var (
 		Use:   "test",
 		Short: "Tests local database with pgTAP.",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return test.Run(afero.NewOsFs())
+			ctx, _ := signal.NotifyContext(cmd.Context(), os.Interrupt)
+			return test.Run(ctx, afero.NewOsFs())
 		},
 	}
 )
--- a/cmd/db.go
+++ b/cmd/db.go
@@ -150,7 +150,8 @@ var (
 		Use:   "lint",
 		Short: "Checks local database for typing error",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return lint.Run(cmd.Context(), schema, level.Value, afero.NewOsFs())
+			ctx, _ := signal.NotifyContext(cmd.Context(), os.Interrupt)
+			return lint.Run(ctx, schema, level.Value, afero.NewOsFs())
 		},
 	}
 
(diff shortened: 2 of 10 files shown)


internal/utils/templates/initial_schemas/14.sql
```diff
@@ -103,6 +103,20 @@ CREATE EXTENSION IF NOT EXISTS pgjwt WITH SCHEMA extensions;
...
+CREATE EXTENSION IF NOT EXISTS pgtap WITH SCHEMA extensions;
+
+
+--
+-- Name: EXTENSION pgtap; Type: COMMENT; Schema: -; Owner:
+--
+
+COMMENT ON EXTENSION pgtap IS 'Unit testing for PostgreSQL';
+
+
```
soedirgo, line 119: No need to install pgtap here - pg_prove installs it automatically. pgtap is also not installed by default on the hosted platform

soedirgo, line 119: pgTAP isn't installed by default on the hosted platform - we need to install it there first before updating the initial schemas to reflect it.

sweatybridge: How do you check if an extension is installed on hosted platform? I know it's listed in our docker hub image https://hub.docker.com/r/supabase/postgres and also available though GUI on a test project I created.

soedirgo: More specifically, this needs 2 PRs:
- for new & unpaused/restored projects: creating a new migration [here](https://github.com/supabase/infrastructure/tree/develop/init-scripts/migrations) to `CREATE EXTENSION IF NOT EXISTS pgtap WITH SCHEMA extensions`
- for existing projects: creating a new ansible migration script ([example](https://github.com/supabase/helper-scripts/pull/203)) and pinging the backend team to roll it out

soedirgo: Oh I just created a new project and checked which extensions are active

sweatybridge: ^ Does that enable pgTAP by default on all projects? I'm not sure if we want that because production db doesn't need these functions.

soedirgo: We'll need it eventually for doing e.g. https://github.com/supabase/pgtap-poc, no harm in enabling it. pgjwt isn't needed either and it's still installed.

Otherwise we need another way to enable the extension (e.g. checking if pgTAP is not enabled, if so enable it and disable it at the end)

sweatybridge: My preference is to make this feature opt-in for now. If many customers ask for it, and already have a bunch of tests written for their local db, then we think about how to make their lives easier by auto configuring staging / production db.

Right now `db test` only works on a local db through docker. This is sufficient to demo CI.

soedirgo: In that case we should do the `CREATE EXTENSION` elsewhere, otherwise this might mess up schema diffs

sweatybridge: hmm let me check pgAdmin. `--migra` ignores extension from its diff.

sweatybridge: Seems like pgtap only adds functions to `extensions` schema, which are excluded from schema diff entirely. https://github.com/theory/pgtap/blob/master/sql/pgtap.sql.in#L1

soedirgo: Just saw that `--migra` only diffs the `public` schema - I don't think that's a correct behavior, the user may have objects on other schemas.

Also, the schema diff *should* include `CREATE EXTENSION`s, these may be required e.g. if a table is using a type coming from an extension (postgis, etc.)

sweatybridge: > user may have objects on other schemas

will add a flag for selecting schema

> Also, the schema diff should include CREATE EXTENSIONs

How about we punt on this until a customer requests for it?

soedirgo: OK, let's do that for now

psteinroe:
this is awesome!! Are tests within nested directories also supported?

soedirgo:
I think instead of manually traversing the dirs, we can instead use one file as an entrypoint and use the [`\i`/`\include` directive](https://github.com/theory/pgtap/blob/ff4a30b59d602ab7a3ea48ce91402fa9bc83abdf/test/setup.sql#L1) to include other files.

Or use the [`--recurse` flag](https://pgtap.org/pg_prove.html) from pg_prove.

psteinroe:
> I think instead of manually traversing the dirs, we can instead use one file as an entrypoint and use the [`\i`/`\include` directive](https://github.com/theory/pgtap/blob/ff4a30b59d602ab7a3ea48ce91402fa9bc83abdf/test/setup.sql#L1) to include other files.
> 
> 
> 
> Or use the [`--recurse` flag](https://pgtap.org/pg_prove.html) from pg_prove.

+1 for the recurse flag. The \include directive would require the user to manually include hundreds of files eventually.

sweatybridge:
Thanks for the suggestions! Added `--recurse` flag and undid changes to initial schema.

soedirgo (approved):

psteinroe:
@sweatybridge why do we only search for `.pg` files? All our tests are `.sql`. I can fix my nvim config, but still curious since most editors do not recognise `.pg` files.

PR #328, opened 2022-08-04
feat: allow configuring migra schema diff

sweatybridge:

feature

only `public` schema is diffed by migra

## What is the new behavior?

exposes a `-s` or `--schema` flag
```bash
$ supabase db commit --migra --schema extensions my_ext
```

## Additional context

Generates `20220804072957_my_ext.sql`

```sql
create extension if not exists "pgtap" with schema "extensions" version '1.1.0';

create type "extensions"."_time_trial_type" as ("a_time" numeric);
```

### DIFF
--- a/cmd/db.go
+++ b/cmd/db.go
@@ -63,6 +63,7 @@ var (
 	}
 
 	useMigra bool
+	schema   string
 
 	dbCommitCmd = &cobra.Command{
 		Use:   "commit <migration name>",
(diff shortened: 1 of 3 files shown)


soedirgo:
How would this take care of diffing multiple schemas? e.g. if I have new tables in `public` as well as `private`

sweatybridge:
I assume it's trivial to run the command multiple times.

soedirgo (approved):

PR #333, opened 2022-08-05
feat(db remote commit): pg_dump special case

soedirgo:
Use pg_dump instead of schema diff for `db remote commit` if it's the first migration.

This is to handle the common case of users wanting to use the CLI midway after developing their schemas using the dashboard. At that point the schema diff is likely very large and prone to errors.

### DIFF
--- a/internal/db/remote/commit/commit.go
+++ b/internal/db/remote/commit/commit.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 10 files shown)


internal/utils/templates/initial_schemas/14.sql
```diff
@@ -2314,6 +2325,16 @@ GRANT ALL ON FUNCTION graphql.variable_definitions_sort(variable_definitions jso
...
 
 
+-- --
+-- -- Name: FUNCTION graphql("operationName" text, query text, variables jsonb, extensions jsonb); Type: ACL; Schema: graphql_public; Owner: supabase_admin
+-- --
+-- 
+-- GRANT ALL ON FUNCTION graphql_public.graphql("operationName" text, query text, variables jsonb, extensions jsonb) TO postgres;
+-- GRANT ALL ON FUNCTION graphql_public.graphql("operationName" text, query text, variables jsonb, extensions jsonb) TO anon;
+-- GRANT ALL ON FUNCTION graphql_public.graphql("operationName" text, query text, variables jsonb, extensions jsonb) TO authenticated;
+-- GRANT ALL ON FUNCTION graphql_public.graphql("operationName" text, query text, variables jsonb, extensions jsonb) TO service_role;
```
sweatybridge, line 2335: Shall we remove these? I will look into ways to not let pg_dump emit these statements altogether.

internal/db/remote/commit/commit.go
```diff
@@ -172,7 +173,62 @@ func run(p utils.Program, url string) error {
...
+			// Use pg_dump instead of schema diff
+			out, err := utils.DockerRun(
+				ctx,
+				dbId,
+				&container.Config{
+					Image: utils.DbImage,
+					Env:   []string{"POSTGRES_PASSWORD=postgres"},
+					Entrypoint: []string{
+						"sh", "-c",
+						"pg_dump --schema-only --quote-all-identifier --exclude-schema '" + strings.Join(utils.InternalSchemas, "|") + `' --schema '*' --extension '*' --dbname '` + url + `' | sed 's/CREATE SCHEMA "public"/-- CREATE SCHEMA "public"/'`,
```
sweatybridge, line 197: Probably more maintainable to externalise these scripts so we can lint with shellcheck. Not necessarily in this PR.

sweatybridge (approved):

internal/db/remote/commit/commit.go
soedirgo: :+1: can convert all these in another PR

internal/utils/templates/initial_schemas/14.sql
soedirgo: I think it's more helpful to just show it but comment it out, so people can see what the raw pg_dump output looks like. Tuning pg_dump to omit these would be ideal though

PR #344, opened 2022-08-10
feat: prompt for password when pushing

sweatybridge:

feature

`db push` requires `db remote set`

## What is the new behavior?

- prompt for db password when calling `db push` and `db remote commit`
- requires a project to be linked
- removes `db remote set` command (will delete code separately)

## Additional context


soedirgo (approved):

drewbietron:
Is there an environment variable we can set to by pass the prompt?  I had a deploy workflow set to run `db set [url]` which would set the config and allow me to push the changes in a CI setting.  Now that the database prompt is introduced I am curious how you all might approach a use case such as mine?   Thank you!

sweatybridge:
Hi @drewbietron , yes having an env var would be nice. For now you can use `-p` flag to pass in the password without prompt. Here's an example CI workflow https://github.com/sweatybridge/supabase-dev/blob/develop/.github/workflows/staging.yaml

PR #347, opened 2022-08-10
fix: init exits 0 when config file exists

sweatybridge:

patch

return 1 when config file exists

## What is the new behavior?

return 0 so that ci works more gracefully

## Additional context


soedirgo:
Returning 1 is expected behavior - you're not supposed to reinit a project, but you can relink. See also npm init, git init, vercel init, etc

sweatybridge:
hmm ok

PR #366, opened 2022-08-15
chore: use ECR registry for images

darora:
(will leave it to @sweatybridge to finalize this)


sweatybridge:
Oh I just saw this after merging #364. Same idea, but i added an internal env var, in case this blows up and we want to switch back.

PR #378, opened 2022-08-17
fix: remove registry prefix from docker hub image

sweatybridge:

Bug fix

fails to pull docker hub image

## What is the new behavior?

pulls locally

## Additional context


egor-romanov (approved):
You should get rid of this duplication ~~and preferably use existing mocks and not gock~~

sweatybridge:
> You should get rid of this duplication and preferably use existing mocks and not gock

There's a lot of deduplication needed in this repo that's more pressing than this one :)

Last time we agreed to use gock for unit tests. Did something change?

egor-romanov:
sry missed that it was in internal folder

PR #393, opened 2022-08-20
fix: remove docker network when supabase start is interrupted

### LINKED ISSUE #425: docker network not removed when supabase start is interrupted
# Bug report

## Describe the bug

When 

## To Reproduce

run `supabase start`
press ctrl + c before it finishes

created containers get's removed, docker network is not removed causing a conflict

## Expected behavior

created containers and network to be removed

## Additional context

the conflict does not make the command fail because the error is not handled (fixed in #424)

kandros:

Studying the codebase I've noticed the function `DockerRemoveAll` used as a cleanup function always paired with another function to remove the docker network.

I've refactored the function having a misleading name `DockerRemoveAll` to `DockerRemoveContainers` as `DockerRemoveAll` only removes containers I renamed it to `DockerRemoveContainers` and replaces the usage of the old functions (always used together) to the new version of `DockerRemoveAll` which now calls both.

A spot where the old `DockerRemoveAll` were used lacked the removal of network, is was a bug and is not handler properly.

fixes #425

### DIFF
--- a/internal/db/remote/changes/changes.go
+++ b/internal/db/remote/changes/changes.go
@@ -92,7 +92,7 @@ func run(p utils.Program, ctx context.Context, username, password, database stri
 			},
 		},
 	)
-	defer cleanup()
+	defer utils.DockerRemoveAll(context.Background(), netId)
 
 	p.Send(utils.StatusMsg("Pulling images..."))
 
@@ -277,10 +277,6 @@ EOSQL
 	return nil
 }
 
-func cleanup() {
-	_ = utils.DockerRemoveAll(context.Background(), netId)
-}
-
 type model struct {
 	cancel      context.CancelFunc
 	spinner     spinner.Model
@@ -303,7 +299,7 @@ func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			// Stop future runs
 			m.cancel()
 			// Stop current runs
-			cleanup()
+			utils.DockerRemoveAll(context.Background(), netId)
 			return m, tea.Quit
 		default:
 			return m, nil
--- a/internal/db/remote/changes/changes.go
+++ b/internal/db/remote/changes/changes.go
@@ -278,8 +278,7 @@ EOSQL
 }
 
 func cleanup() {
-	utils.DockerRemoveAll()
-	_ = utils.Docker.NetworkRemove(context.Background(), netId)
+	_ = utils.DockerRemoveAll(context.Background(), netId)
 }
 
 type model struct {
--- a/internal/start/start.go
+++ b/internal/start/start.go
@@ -61,8 +61,7 @@ func Run(ctx context.Context) error {
 		return errors.New("Aborted " + utils.Aqua("supabase start") + ".")
 	}
 	if err := <-errCh; err != nil {
-		utils.DockerRemoveAll()
-		_ = utils.Docker.NetworkRemove(context.Background(), utils.NetId)
+		_ = utils.DockerRemoveAll(context.Background(), utils.NetId)
 		return err
 	}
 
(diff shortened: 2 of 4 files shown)


internal/start/start.go
```diff
@@ -816,7 +815,7 @@ func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			// Stop future runs
 			cancelCtx()
 			// Stop current runs
-			utils.DockerRemoveAll()
```
AaronDewes, line 819: I'm assuming you have tested this, because previously, the network had not been removed there. Are you sure this won't cause issues?

kandros: Ouch! You are right, I overlooked the suppressed error here https://github.com/supabase/cli/blob/main/internal/start/start.go#L85
(now is handled)

It was still working "thanks" to https://github.com/kandros/supabase-cli/blob/e76bfec2a2efd2239774204a7cb48effd85d623d/internal/start/start.go#L64

This is indeed a bug for me.

fun fact: To prevent a mistake like that is exactly what this refactor was made for in the first place 😅 shame on me

internal/start/start.go
```diff
@@ -93,6 +92,9 @@ func run(p utils.Program) error {
 			},
 		},
 	)
+	if err != nil {
+		return err
+	}
```
AaronDewes, line 97: This should not fail if the network already exists. Previous versions did not remove the network in some cases (https://github.com/supabase/cli/pull/393/files#r950759922), so if someone migrates from an older version, this could fail now because the network will exist.

AaronDewes (changes requested):

internal/start/start.go
kandros: Ok, I see the problem.

Not  handling the error at all may hide other bugs in the future.

We could handle the specific error "network already exist" and make it continue without blocking the user and return any other type of error so that they can be catch in development?

What do you think?

kandros:
as @AaronDewes pointed out this PR fixes a bug but handling the network creation error creates a situation that requires additional code that is out of scope of this PR.

I've reverted the error handling and open another issue and PR to deal with ti

kandros:
handled "network already exist" error gracefully in #424

internal/db/remote/changes/changes.go
```diff
@@ -278,8 +278,7 @@ EOSQL
 }
 
 func cleanup() {
-	utils.DockerRemoveAll()
-	_ = utils.Docker.NetworkRemove(context.Background(), netId)
+	_ = utils.DockerRemoveAll(context.Background(), netId)
```
sweatybridge, line 281: Since there's only 1 method, let's unwrap the cleanup function and call `DockerRemoveAll` in place instead. For eg.

```go
defer utils.DockerRemoveAll(context.Background(), netId)
```

internal/utils/docker.go
```diff
@@ -113,6 +113,11 @@ func DockerRemoveAll() {
 	wg.Wait()
 }
 
+func DockerRemoveAll(ctx context.Context, netId string) error {
+	DockerRemoveContainers(ctx)
+	return Docker.NetworkRemove(ctx, netId)
+}
```
sweatybridge, line 119: No need to return the error if we are ignoring it everywhere

```suggestion
func DockerRemoveAll(ctx context.Context, netId string) {
	DockerRemoveContainers(ctx)
	_ = Docker.NetworkRemove(ctx, netId)
}
```

sweatybridge (commented):
Thanks @kandros for helping to check the docker cleanup code, and apologies for the late response.

Just a couple of comments before we merge it.

internal/utils/docker.go
kandros: Makes, sense I wanted to return in to handle them in different PRs but is such a thing that is worth doing all at once (returning error and handling all their usage).

TL;DR will remove the return and open a new PR to handle them

sweatybridge (approved):

PR #394, opened 2022-08-20
chore: clearer error message when project already initialized

kandros:

With current error it may seems like the supabase cli needs to be reinstalled intead of the directory removed

Current:

`Error: Project already initialized. Remove supabase to reinitialize.`

## What is the new behavior?
New:
`Error: Project already initialized. To allow reinitialization remove the directory ./supabase`

### DIFF
--- a/internal/init/init.go
+++ b/internal/init/init.go
@@ -14,7 +14,7 @@ var (
 	//go:embed templates/init_gitignore
 	initGitignore []byte
 
-	errAlreadyInitialized = errors.New("Project already initialized. To allow reinitialization remove the directory " + utils.Bold("./supabase"))
+	errAlreadyInitialized = errors.New("Project already initialized. Remove " + utils.Bold(utils.ConfigPath) + " to reinitialize.")
 )
 
 func Run(fsys afero.Fs) error {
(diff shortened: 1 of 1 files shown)


internal/init/init.go
```diff
@@ -14,7 +14,7 @@ var (
 	//go:embed templates/init_gitignore
 	initGitignore []byte
 
-	errAlreadyInitialized = errors.New("Project already initialized. Remove " + utils.Bold("supabase") + " to reinitialize.")
+	errAlreadyInitialized = errors.New("Project already initialized. To allow reinitialization remove the directory " + utils.Bold("./supabase"))
```
sweatybridge, line 17:
```suggestion
	errAlreadyInitialized = errors.New("Project already initialized. Remove " + utils.Bold(utils.ConfigPath) + " to reinitialize.")
```

sweatybridge (commented):
Thanks for helping to improve the error message. Just a minor comment:

Instead of suggesting to delete the whole directory, we can be clearer that deleting the `supabase/config.toml` is sufficient to reinitialise the project.

sweatybridge (approved):

Issue #298, opened 2022-07-08
new cmd: `supabase db lint`


Marviel:
# Feature request
I'm writing this spec to gauge interest in a PR I'm going to put together that wraps up a script we use internally to type-check the functions in our database.

## TL;DR
- Runtime exceptions should be avoided.
- Type safety at the database level is critical for avoiding runtime exceptions, b/c the database should be a source of truth and clarity
- Functions in Postgres are effectively not - Type Safe at "compile time" / "commit time" by default
- [Tools like plpgsql_check exist](https://github.com/okbob/plpgsql_check) to type-check Postgres functions to a relatively high degree of accuracy, however, while robust, they are not as user-friendly.
- It is impractical to force "on commit" type checking (a-la passive mode) (probably? for speed reasons?)
- Because it cannot be done "on commit", and this data is imho critical to having a consistent database -> it should be made easier for users to type-check their schema, at any time they choose.
- Therefore, there should exist a `supabase db lint` command (exact naming up for grabs) which is inspired by the `eslint` / `tslint` ecosystem, and follows the following rough spec:
    - [x] `supabase db lint` should check the existing database (after migrations have been run) to identify any typing errors, warnings, and notes, at different log levels, as available on the console environment they are running. (Can be achieved via the [`plpgsql_check` Postgres extension](https://github.com/okbob/plpgsql_check), as described above -- or by other, similar projects)
    - [x] Users should be able to specify what level of lint errors they would like to be output, via a flag.
    - [x] Another flag should exist, to allow the user to optionally drop and recreate their local database at the same time -> this would be helpful as a "pre-PR" smoke test, so one could validate that new migrations haven't caused any issues.
- Other Nice-To-Haves / Future Work
    - [ ] Users should be able to easily write their own custom linters for both their migrations, and their built schemas, in whatever language is found most suitable for this purpose. 

## Longform

Runtime exceptions should be avoided whenever possible.

One cause of Runtime Exceptions is calling a function (procedure, etc) "in the wrong way" -- i.e. with the wrong parameter count, or with the wrong parameter types.

Typed languages (~Typescript, Rust, C#, et. al.) afford the **compiler** the ability to check ahead of time whether or not a given function call is valid in a given context, based on the type signature of the function (i.e. the types of its parameters, and the type of its return value.) This reduces the cognitive load of a programmer, improves clarity, self-documents, and also allows more individuals to collaborate on the same schema at scale.

[...]

PR #398, opened 2022-08-22
feat: add db lint command

sweatybridge:

feature #298 

no linting

## What is the new behavior?

- adds `supabase db lint` with `--schema` flag
- uses [plpgsql_check](https://github.com/okbob/plpgsql_check)  extension
- uses prepared statement to support multiple schema

TODO:
- [x] format output (json for now)
- [x] add unit tests
- [x] separate warning vs error
- [ ] create ignore file for different types of error (comment might be hard)
- [x] custom lint check (might be able to use pg_tap)

use case
- run check on anything that touches supabase schema
- useful to have the line of code where error occurs (body of function)
- want to check columns are created, ie. created_by, modified_by, etc

## Additional context

Add `f1()` example as a migration script, then run

```
$ supabase db lint -s public,private
Linting schema: public
Linting schema: private
[
  {
    "function": "public.f1",
    "issues": [
      {
        "level": "error",
        "message": "relation \"t1\" does not exist",
        "statement": {
          "lineNumber": "4",
          "text": "FOR over SELECT rows"
        },
        "query": {
          "position": "15",
          "text": "SELECT * FROM t1"
        },
        "sqlState": "42P01"
      }
    ]
  }
]
```

### DIFF
--- a/internal/db/lint/lint.go
+++ b/internal/db/lint/lint.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	"github.com/jackc/pgx/v4"
 	"github.com/spf13/afero"
@@ -28,14 +29,14 @@ type LintLevel int
 
 func toEnum(level string) LintLevel {
 	for i, curr := range AllowedLevels {
-		if curr == level {
+		if strings.HasPrefix(level, curr) {
 			return LintLevel(i)
 		}
 	}
 	return -1
 }
 
-func Run(ctx context.Context, schema []string, level string, stdout io.Writer, fsys afero.Fs, opts ...func(*pgx.ConnConfig)) error {
+func Run(ctx context.Context, schema []string, level string, fsys afero.Fs, opts ...func(*pgx.ConnConfig)) error {
 	// Sanity checks.
 	if err := utils.LoadConfigFS(fsys); err != nil {
 		return err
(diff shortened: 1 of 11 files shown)


internal/db/lint/lint.go
```diff
@@ -0,0 +1,174 @@
...
+	// Connect to database
+	return pgx.ConnectConfig(ctx, config)
+}
+
+func LintDatabase(ctx context.Context, conn *pgx.Conn, schema []string) ([]Result, error) {
+	// Enable plpgsql_check
+	enable := "CREATE EXTENSION IF NOT EXISTS plpgsql_check"
+	if _, err := conn.Exec(ctx, enable); err != nil {
+		return nil, err
+	}
```
soedirgo, line 108: This command should have no side effects, i.e. we should either:
- fail if `plpgsql_check` is not enabled, or
- enable it if not enabled, but disable it at the end (but don't disable it if it's already enabled)

soedirgo: Or if there's no need to commit, we can simply rollback the txn

sweatybridge: Is there any benefit for disabling this extension on local dev? I'm thinking to add it to default schema after this PR.

soedirgo: Same reason for disabling pgtap: the schema will diverge from the hosted platform's

sweatybridge: Sure, also added a task to include extensions in schema diff.

soedirgo (approved):

PR #424, opened 2022-08-29
chore: handle name conflict in docker network creation

kandros:

Bug fix

Error is not handler and get's lost

## What is the new behavior?

propagate error unless is "network already exists" as it is ok to handle internally and continue without giving an error the the use


internal/start/start.go
```diff
@@ -94,6 +95,13 @@ func run(p utils.Program) error {
 		},
 	)
 
+	// matching against public interface because concrete type is private, error.As only works for types */
+	if _, ok := err.(errdefs.ErrConflict); ok {
```
sweatybridge, line 99: We should add a unit test for this code path to be sure. There's a few examples of [mocking docker api](https://github.com/supabase/cli/blob/main/internal/utils/docker_test.go) in tests.

Let me know if you have to time to do it or I can take a look.

kandros: I'll give it a shot 👍 thanks for the example link

sweatybridge (approved):

PR #426, opened 2022-08-29
show migration new help when called without arguments

kandros:
print help instead of validation error when passing 0 or more than 1 arguments to `migration new`

feature

image

## What is the new behavior?

image

### DIFF
--- a/cmd/migration.go
+++ b/cmd/migration.go
@@ -31,7 +31,13 @@ var (
 	migrationNewCmd = &cobra.Command{
 		Use:   "new <migration name>",
 		Short: "Create an empty migration script",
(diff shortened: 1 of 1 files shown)


kandros:
@soedirgo @sweatybridge I've seen this behaviour in many commands, if the change I made is ok for you I can use it to more commands

what do you think?

soedirgo:
Hey there, sorry for the late response. I think you can just remove this line to show usage for all commands:

https://github.com/supabase/cli/blob/3968f1cb2c75cdd797e266453086851d7d493c45/cmd/root.go#L26

sweatybridge:
Closing this in favour of #562

PR #434, opened 2022-09-02
fix: update initial schema to include pgsodium

### LINKED ISSUE #383: Error while running db migrations
# Bug report

## Describe the bug

I'm running into the following error when I'm running `supabase db remote commit` or `supabase start` after running the first migration by running `supabase db remote commit`.

```
Error: Error starting shadow database: ERROR:  relation "pgsodium.key_key_id_seq" does not exist
ERROR:  current transaction is aborted, commands ignored until end of transaction block
```

## To Reproduce

1. Create a new Supabase project
2. Install CLI, and link to the project
3. Inside Supabase dashboard, go to SQL editor and run _Stripe subscription starter_ template to create the tables.
4. Run `supabase db remote commit` in the terminal
5. Run `supabase start` or run `supabase db remote commit` again.

## Expected behavior

Run the db migrations without errors

## System information

- OS: macOS
- Version of CLI: 1.0.7

sweatybridge:

fixes #383 

missing pgsodium schema

## What is the new behavior?

include pgsodium init scripts in upstream image

## Additional context

depends on 
- [x] https://github.com/supabase/postgres/pull/286


soedirgo:
Getting
```
Error response from daemon: manifest for public.ecr.aws/t3w2s2c9/postgres:14.1.0.66 not found: manifest unknown: Requested image not found
```
There's an extra `v` here: https://github.com/supabase/postgres/blob/615ea0a16b881788a10650991a7cfc086b45c415/.github/workflows/dockerhub-release.yml#L38

internal/db/remote/changes/changes.go
```diff
@@ -170,25 +170,9 @@ EOSQL
...
-			}
-			var errBuf bytes.Buffer
-			if _, err := stdcopy.StdCopy(io.Discard, &errBuf, out); err != nil {
-				return err
-			}
-			if errBuf.Len() > 0 {
-				return errors.New("Error starting shadow database: " + errBuf.String())
-			}
+		p.Send(utils.StatusMsg("Resetting database..."))
+		if err := differ.ResetDatabase(ctx, dbId, utils.ShadowDbName); err != nil {
```
soedirgo, line 174: Should be `postgres` which is the default `POSTGRES_DB`
```suggestion
		if err := differ.ResetDatabase(ctx, dbId, "postgres"); err != nil {
```
We might also need `PGOPTIONS='--client-min-messages=error` in `reset.sh`

internal/db/remote/changes/changes.go
```diff
@@ -221,7 +205,7 @@ EOSQL
 			}
 
 			out, err := utils.DockerExec(ctx, dbId, []string{
-				"sh", "-c", `PGOPTIONS='--client-min-messages=error' psql postgresql://postgres:postgres@localhost/postgres <<'EOSQL'
+				"sh", "-c", "PGOPTIONS='--client-min-messages=error' psql postgresql://postgres:postgres@localhost/" + utils.ShadowDbName + ` <<'EOSQL'
```
soedirgo, line 208: Ditto

internal/db/remote/changes/changes.go
```diff
@@ -253,7 +237,7 @@ EOSQL
 				Entrypoint: []string{
 					"sh", "-c", "/venv/bin/python3 -u cli.py --json-diff" +
 						" '" + conn.Config().ConnString() + "'" +
-						" 'postgresql://postgres:postgres@" + dbId + ":5432/postgres'",
+						" 'postgresql://postgres:postgres@" + dbId + ":5432/" + utils.ShadowDbName + "'",
```
soedirgo, line 240: Ditto

internal/db/remote/commit/commit.go
```diff
@@ -153,8 +153,6 @@ func run(p utils.Program, ctx context.Context, username, password, database stri
 		}
 
 		if len(localMigrations) == 0 {
-			strings.Join(utils.InternalSchemas, "|")
-
```
soedirgo, line 157: Oops... don't know how this got past the lints

internal/db/remote/commit/commit.go
```diff
@@ -246,25 +244,9 @@ EOSQL
...
-			}
-			var errBuf bytes.Buffer
-			if _, err := stdcopy.StdCopy(io.Discard, &errBuf, out); err != nil {
-				return err
-			}
-			if errBuf.Len() > 0 {
-				return errors.New("Error starting shadow database: " + errBuf.String())
-			}
+		p.Send(utils.StatusMsg("Resetting database..."))
+		if err := differ.ResetDatabase(ctx, dbId, utils.ShadowDbName); err != nil {
```
soedirgo, line 248: Ditto

internal/db/remote/commit/commit.go
```diff
@@ -294,7 +276,7 @@ EOSQL
 			}
 
 			out, err := utils.DockerExec(ctx, dbId, []string{
-				"sh", "-c", `PGOPTIONS='--client-min-messages=error' psql postgresql://postgres:postgres@localhost/postgres <<'EOSQL'
+				"sh", "-c", "PGOPTIONS='--client-min-messages=error' psql postgresql://postgres:postgres@localhost/" + utils.ShadowDbName + ` <<'EOSQL'
```
soedirgo, line 279: Ditto

internal/db/remote/commit/commit.go
```diff
@@ -326,7 +308,7 @@ EOSQL
 				Entrypoint: []string{
 					"sh", "-c", "/venv/bin/python3 -u cli.py --json-diff" +
 						" '" + conn.Config().ConnString() + "'" +
-						" 'postgresql://postgres:postgres@" + dbId + ":5432/postgres'",
+						" 'postgresql://postgres:postgres@" + dbId + ":5432/" + utils.ShadowDbName + "'",
```
soedirgo, line 311: Ditto

internal/utils/templates/globals.sql
```diff
@@ -55,7 +53,7 @@ ALTER ROLE authenticated SET statement_timeout TO '8s';
 -- User Config "authenticator"
 --
 
-ALTER ROLE authenticator SET session_preload_libraries TO 'supautils', 'safeupdate';
```
soedirgo, line 58: Any reason to change this?

internal/utils/templates/initial_schemas/14.sql
```diff
@@ -348,6 +394,7 @@ BEGIN
 
     IF func_is_graphql_resolve
     THEN
+        DROP FUNCTION IF EXISTS graphql_public.graphql;
```
soedirgo, line 397: Where's this line coming from? This function hasn't changed since.

internal/db/remote/commit/templates/dump_initial_migration.sh
```diff
@@ -1,13 +1,17 @@
...
+#   --extension '*'   prevents event triggers from being dumped, bash escaped with single quote
 pg_dump \
     --schema-only \
     --quote-all-identifier \
     --exclude-schema "$EXCLUDED_SCHEMAS" \
-    --schema '*' \
     --extension '*' \
+    --no-comments \
     --dbname "$DB_URL" \
-| sed 's/CREATE SCHEMA "public"/-- CREATE SCHEMA "public"/' \
```
soedirgo, line 11: We shouldn't remove this - the `public` schema already exists.

internal/db/remote/commit/templates/dump_initial_migration.sh
```diff
@@ -1,13 +1,17 @@
...
+#   --exclude-schema  omit internal schemas as they are maintained by platform
+#   --no-comments     only object owner can set comment, omit to allow restore by non-superuser
+#   --extension '*'   prevents event triggers from being dumped, bash escaped with single quote
 pg_dump \
     --schema-only \
     --quote-all-identifier \
     --exclude-schema "$EXCLUDED_SCHEMAS" \
-    --schema '*' \
     --extension '*' \
+    --no-comments \
```
soedirgo, line 15: We shouldn't omit *all* comments, only those on extensions. `postgres` can't comment on extensions (since they're created by `supabase_admin`), so we assume extension comments are unchanged.

internal/db/remote/changes/changes.go
sweatybridge: > Should be postgres which is the default POSTGRES_DB

This is a new change introduced in `14.1.0.66`: [docker entrypoint](https://github.com/supabase/postgres/blob/develop/ansible/playbook-docker.yml#L21) now creates additional roles and schemas in default `postgres` db. Our initial schema tries to create the same schema and fails. Hence, we use shadow db. Also, this change makes remote diff consistent with local diff.

internal/utils/templates/globals.sql
sweatybridge: Yes, `supautils` extension is not installed on docker image. So this errors at runtime (more noticeable in self hosted setup).

internal/utils/templates/initial_schemas/14.sql
sweatybridge: It's coming from here https://github.com/supabase/infrastructure/pull/9042. I found this issue by running all infra migration scripts in self hosted docker.

internal/db/remote/commit/templates/dump_initial_migration.sh
sweatybridge: I just checked that `create schema "public"` statement is no longer dumped since the removal of `--schema '*'` flag.

internal/db/remote/commit/templates/dump_initial_migration.sh
sweatybridge: This is explained in comments of `dump_initial_migration.sh`. Only object owner can set comment. So any object owned by `supabase_admin` will need to be commented. Leaving it there is a time bomb as infra creates more migration scripts in the future.

internal/utils/templates/globals.sql
soedirgo: Note that we need `supautils` in docker eventually, when we move to [extension custom scripts](https://github.com/supabase/supautils/pull/26).

sweatybridge: Hopefully we will have centralised migrations by then :)

internal/db/remote/commit/templates/dump_initial_migration.sh
soedirgo: Sure, but that also removes all comments from users' objects. Anyway, aren't internal schemas supposed to be excluded in the first place?

sweatybridge: I think it's a trade off between getting complaints when cli users fail to `db push` vs complaints when user's object comments are not synced. The latter is less severe and would be my preference.

Perhaps we can add it back when we finally switch sql editor to postgres role.

soedirgo (approved):

Issue #451, opened 2022-09-12
Add the ability to export values from `supabase start` to an `.env.local` file 


NatoNathan:
# Feature request

## Is your feature request related to a problem? Please describe.

So I'm working on a project where I'm using codespaces for an ephemeral development environment, but this could equally apply to gitpod as well. 

One of the issues I'm coming up against is that when I make a new codespace/container, I need to manually create a new `.env.local` file with the anon and service_role keys. This is not major, but it would be super nice to be able to start work without this step.

## Describe the solution you'd like

I was thinking there could be a option flag on `supabase start` like `supabase start --value-file` and perhaps some config in `supabase/config.toml` like 

```toml
[value_file]
path = "./.env.local"
file_type = "env"
anon_key = "NEXT_PUBLIC_SUPABASE_ANON_KEY"
service_role_key = "SUPABASE_SERVICE_ROLE_KEY"
api_url = "NEXT_PUBLIC_SUPABASE_URL"
```

Which would create the correct file based on the path, file_type and mappings. In this case a `.env.local` file at the projects root.

## Describe alternatives you've considered

Another solution could be creating environment variables directly however that would not be as flexible and would still need the mappings.

Another solution could be to provide the values to use like
```sh
supabase start --anon-key="sknsdjkbnhn" --service-role-key="cndjkskcnsdjkn"
```

maxkostow:
Another benefit could be to compile your frontend in parallel with the docker image download / seed

NatoNathan:
I have attempted to have go at this myself, but only supporting `env` files here #452

churichard:
I think this would be really useful for CI purposes as well. For example, it would be nice to have the proper environment variables automatically filled in so that Cypress tests can be run in CI.

ryanalta:
Just wound up on this because it looks like the default keys are all hardcoded and not generated. Would be extremely nice to have something better!

sweatybridge:
Exporting env vars is now possible with the `status` command:

```bash
supabase status -o env
```

More examples are available under `-h` flag. 

@ryanalta do you mind creating a separate issue for making the hardcoded values dynamic/configurable?

ryanalta:
@sweatybridge can do!

PR #452, opened 2022-09-13
feat: support output variables with status command

NatoNathan:

Add an `--value-file` flag to `supabase start`. This will add the ability to export values from supabase start to an `.env.local` file as discussed in #451.

## What is the new behavior?

The `config.toml` now has support for the following section
```toml
[value_file]
path = "./.env.local"
type = "env"
anon_key = "NEXT_PUBLIC_SUPABASE_ANON_KEY"
service_role_key = "SUPABASE_SERVICE_ROLE_KEY"
api_url = "NEXT_PUBLIC_SUPABASE_URL"
```

This config is used when you use the `--value-file` to make a new file. in this case it would output a file like the following
```sh
# ./.env.local

NEXT_PUBLIC_SUPABASE_ANON_KEY=some_anon_key
SUPABASE_SERVICE_ROLE_KEY=some_service_role_key
NEXT_PUBLIC_SUPABASE_URL=some_api_url
```

At the moment I have only added support for `env` files. Also this is the first time I have used golang so any suggestions for improvements or corrections would be appreciated.


internal/utils/templates/init_config.toml
```diff
@@ -60,3 +60,10 @@ enable_confirmations = false
 enabled = false
 client_id = ""
 secret = ""
+
+[value_file]
+path = "./.env.local"
+type = "env"
+anon_key = "NEXT_PUBLIC_SUPABASE_ANON_KEY"
+service_role_key = "SUPABASE_SERVICE_ROLE_KEY"
+api_url = "NEXT_PUBLIC_SUPABASE_URL"
```
sweatybridge, line 69: This stanza actually deals with two separate concerns: setting the export path which applies to all services and configuring individual service keys. I'm thinking the following alternative instead:
- set `anon_key` and `service_role_key` in [auth] stanza
- set `api_url` in [api] stanza
- remove the rest of [value_file] stanza

The last point is to avoid introducing additional CLI flag. Instead of pretty printing the connection details in `utils.ShowStatus`, we can write to stdout in env var format. For example,

```
$ supabase start
Started supabase local development setup.

API_URL=http://localhost:54321
DB_URL=postgresql://postgres:postgres@localhost:54322/postgres
STUDIO_URL=http://localhost:54323
INBUCKET_URL=http://localhost:54324
ANON_KEY=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZS1kZW1vIiwicm9sZSI6ImFub24ifQ.625_WdcF3KHqz5amU0x2X5WWHP-OEs_4qj0ssLNHzTs
SERVICE_ROLE_KEY=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZS1kZW1vIiwicm9sZSI6InNlcnZpY2Vfcm9sZSJ9.vI9obAHOGyVVKa3pD--kJlyxp-Z2zV9UUMAhKpNLAcU
```

Users can then redirect stdout to a file for setting up code editors: `supabase start > .env.local`

Other logs will go to stderr so they won't be redirected to env file. @soedirgo

sweatybridge (commented):
Thanks for opening this PR. In general, I agree it is a useful feature and would be happy to merge it either this week or next.

internal/utils/templates/init_config.toml
NatoNathan: Ok, I can make those changes,

Also the names of the env mapping should probably be changed too, like `anon_key_env_var`?

NatoNathan: Ok, I have updated `utils.ShowStatus()` however the docker pull logs are written to `stdout` instead of `stderr` so when you run `supabase start > .env.local` has all the docker pull logs and stuff.

sweatybridge: Yup, that seems fine for now. I will take a look at the docker code tmr to pipe those logs to `stderr`. Appreciate your help in driving this change.

NatoNathan:
I have synced this pr with main, and have fixed the issue with the logs getting printed to `stdout` instead of `stderr`

sweatybridge:
Thank you very much. Would it work for you if we move this `--env-file` flag to `supabase status` command instead?

Something like

```bash
supabase start
supabase status --env-file > .env.local
```

I think it's more flexible and if it works for you, I can update your PR directly.

NatoNathan:
Yeah that could work.

sweatybridge (approved):

PR #461, opened 2022-09-16
chore: add tests for function deploy

sweatybridge:

tests

## What is the new behavior?

- refactored deno path to utils
- pass in cancellation context
- extracted string and regex constants
- switch to codegen api client

### DIFF
--- a/internal/functions/deploy/deploy_test.go
+++ b/internal/functions/deploy/deploy_test.go
@@ -2,7 +2,6 @@ package deploy
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 	"os"
 	"testing"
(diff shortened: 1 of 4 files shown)


internal/functions/deploy/deploy_test.go
```diff
@@ -0,0 +1,87 @@
...
+func init() {
+	// Setup fake deno binary
+	if denoPath, err := os.Executable(); err == nil {
+		utils.DenoPathOverride = denoPath
+	}
+}
+
+func TestMain(m *testing.M) {
+	if len(os.Args) > 1 && (os.Args[1] == "bundle" || os.Args[1] == "upgrade") {
+		fmt.Println("ea09bff03b3894b79e335f77a0f71d75adf263e9")
```
soedirgo, line 28: What's the magic number?

sweatybridge: Can be deleted. Was using it to check stdout.

soedirgo (approved):

PR #464, opened 2022-09-16
chore: add tests for function serve

sweatybridge:

tests

## What is the new behavior?

- use the same service role key and anon key from template
- moves login and project ref prompt to cmd
- tests for specific error string
- increase coverage of function deploy

## Additional context


internal/functions/serve/serve.go
```diff
@@ -101,32 +89,26 @@ func Run(slug string, envFilePath string, verifyJWT bool) error {
...
-			"deno", "cache", "/home/deno/functions/" + slug + "/index.ts",
-		})
-		if err != nil {
-			return err
-		}
-		if _, err := stdcopy.StdCopy(io.Discard, io.Discard, out); err != nil {
-			return err
-		}
+	localFuncDir := filepath.Join(utils.FunctionsDir, slug)
+	relayFuncPath := filepath.Join(relayFuncDir, slug, "index.ts")
```
soedirgo, line 93: spitballing: wouldn't this use the `\` separator on Windows instead of `/`?

sweatybridge: good catch. changed to unix path separator

soedirgo (approved):

Issue #470, opened 2022-09-23
Large SQL files skipped without error on `supabase db reset`


FelixZY:
# Bug report

## Describe the bug

My database should contain geographical data for Sweden. This data is inserted using 327 `INSERT` statements and the SQL file takes up ~4MB. When I attempt to run this file (using `supabase db reset`) as part of either my migrations (`supabase/migrations/*.sql`) or my initial seed (`supabase/seed.sql`)  some data is skipped over and the reset operation is reported as a success.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Place the [provided SQL file](https://gist.github.com/FelixZY/051f4079258c84368481367d6272d2ee) in `supabase/migrations/`
2. Run `supabase db reset`
3. Go to studio -> `locations` table

## Expected behavior

* `locations` table should contain 327 records - the full SQL file should be executed.

## Screenshots

In this case, the actual count was 206 records:

![Screenshot from 2022-09-23 21-50-00](https://user-images.githubusercontent.com/2656517/192046514-08a85721-d03c-4e61-80a4-3e969d10b809.png)

## System information

- Host OS: Ubuntu 20.04.4 LTS (focal) (64-bit)
- Devcontainer OS: Debian 11 (bullseye)
- Docker (Running on host): 20.10.17, build 100c701
- Version of supabase cli (running in devcontainer): 1.4.8

```
CONTAINER ID   IMAGE                                                   COMMAND                  CREATED             STATUS                       PORTS                                                                                                                                   NAMES
0c91c366d071   public.ecr.aws/t3w2s2c9/studio:v0.1.0                   "docker-entrypoint.s…"   About an hour ago   Up About an hour             0.0.0.0:54323->3000/tcp, :::54323->3000/tcp                                                                                             supabase_studio_portal
ed3f8ddddef6   public.ecr.aws/t3w2s2c9/postgres-meta:v0.45.0           "docker-entrypoint.s…"   About an hour ago   Up About an hour             8080/tcp                                                                                                                                supabase_pg_meta_portal
0c37d35fc865   public.ecr.aws/t3w2s2c9/pgadmin-schema-diff:cli-0.0.5   "sleep infinity"         About an hour ago   Up About an hour                                                                                                                                                     supabase_differ_portal
[...]

PR #472, opened 2022-09-24
fix: allow for long lines in SQL migrations

FelixZY:

Bug fix

When lines that are too long are encountered, the scanner in `internal/utils/parser/token.go#Split` fails, but the `.Err()` is not checked, causing the error to be suppressed and the `internal/utils/parser/token.go#Split` func to return partial data.

See #470 

## What is the new behavior?

* Errors produced by the scanner used in `internal/utils/parser/token.go#Split` are propagated to the caller.
* Lines up to 256K can now be read by `internal/utils/parser/token.go#Split`.

### DIFF
--- a/internal/utils/parser/token.go
+++ b/internal/utils/parser/token.go
@@ -6,6 +6,15 @@ import (
 	"unicode/utf8"
 )
 
+const (
+	// Default max capacity is 64 * 1024 which is not enough for certain lines
+	// containing e.g. geographical data.
+	// 256K ought to be enough for anybody...
+	maxScannerCapacity = 256 * 1024
+	// Equal to `startBufSize` from `bufio/scan.go`
+	startBufSize = 4096
+)
+
 // State transition table for tokenizer:
 //
 //   Ready -> Ready (default)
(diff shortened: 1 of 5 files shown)


internal/utils/parser/token.go
```diff
@@ -67,13 +72,18 @@ func (t *tokenizer) ScanToken(data []byte, atEOF bool) (advance int, token []byt
 // token can be parsed as statement separator.
 //
 // Each statement is split as it is, without removing comments or white spaces.
-func Split(sql io.Reader) (stats []string) {
+func Split(sql io.Reader) (stats []string, err error) {
 	t := tokenizer{state: &ReadyState{}}
 	scanner := bufio.NewScanner(sql)
+
+	// Increase scanner capacity to support very long lines containing e.g. geodata
+	buf := make([]byte, maxScannerCapacity)
```
sweatybridge, line 80:
```suggestion
	buf := make([]byte, 64 * 1024)
```

Shall we keep the default capacity at 64KB and let the buffer grow to new max capacity as needed? https://stackoverflow.com/a/37455465

In the future, we might also consider making max capacity configurable via env var if more users run into this issue.

sweatybridge (commented):
Thanks a lot for looking into this issue and fixing. Just a minor suggestion before merging.

internal/utils/parser/token.go
FelixZY: Perhaps even

```go
buf := make([]byte, 4096)
```

given the following setup from `bufio/scan.go`:

```go
const (
	// MaxScanTokenSize is the maximum size used to buffer a token
	// unless the user provides an explicit buffer with Scanner.Buffer.
	// The actual maximum token size may be smaller as the buffer
	// may need to include, for instance, a newline.
	MaxScanTokenSize = 64 * 1024

	startBufSize = 4096 // Size of initial allocation for buffer.
)
```

I'll push a fix :+1:

FelixZY: Fix pushed via `--amend` on commit `fix: Increase buffer capacity (#470)`

sweatybridge (approved):

PR #473, opened 2022-09-25
feat: support configuring arbitrary OIDC provider url for gotrue

zlepper:

Bug fix/feature

You have to use multi-tenant azure application to enable sign-in with azure. which is impractical if you are making an org specific URL. 

## What is the new behavior?

This allows specifying a `url` property on external auth providers which will be passed along to GoTrue:
```toml
[auth.external.azure]
enabled = true
client_id = "****"
secret = "****"
url = "https://login.microsoftonline.com/****"
```

The property is optional and will not be passed along if not specified. 

## Additional context

Related ticket in gotrue: https://github.com/supabase/gotrue/issues/292

### DIFF
--- a/internal/start/start.go
+++ b/internal/start/start.go
@@ -535,6 +535,7 @@ EOSQL
 					"KONG_PLUGINS=request-transformer,cors,key-auth",
 					// Need to increase the nginx buffers in kong to avoid it rejecting the rather
 					// sizeable response headers azure can generate
+					// Ref: https://github.com/Kong/kong/issues/3974#issuecomment-482105126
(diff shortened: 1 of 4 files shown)


sweatybridge (approved):
Thanks for submitting this PR.

For documentation purposes, we export all configurable parameters even if they are optional. I've added the change accordingly to `init_config.toml`.

Since this field can be useful to other OIDC providers like self-hosted gitlab, I've also added a comment to clarify its use cases.

lgtm

sweatybridge (changes requested):
Let me also check the gotrue image version.

sweatybridge (approved):

PR #476, opened 2022-09-26
Chore: Remove seemingly redundant flag redefinition

loong:

Tiny nitpick. Found a repeated line that seems to be a duplicate. If intentional can just close the PR!


cmd/functions.go
```diff
@@ -87,7 +87,6 @@ var (
 func init() {
 	functionsDeleteCmd.Flags().StringVar(&projectRef, "project-ref", "", "Project ref of the Supabase project.")
 	functionsDeployCmd.Flags().BoolVar(&noVerifyJWT, "no-verify-jwt", false, "Disable JWT verification for the Function.")
-	functionsDeployCmd.Flags().StringVar(&projectRef, "project-ref", "", "Project ref of the Supabase project.")
```
loong, line 90: Repeat is here on line 88 and 90

loong:
False alert! Just realized one is for the delete command and one is for the deploy command!

PR #494, opened 2022-09-30
feat: add storage.file_size_limit_bytes config

### LINKED ISSUE #489: Configurable storage FILE_SIZE_LIMIT
# Feature request

I have customized the `FILE_SIZE_LIMIT` on my production supabase app, and it would be helpful to use the same `FILE_SIZE_LIMIT` value in my local development environment so that local testing matches the behavior of production.

I'd like the `config.toml` to include a new `storage` table with a `file_size_limit` key, as shown:

```
[storage]
# The maximum file size allowed, in bytes.
file_size_limit = 52428800
```

I haven't dug into config deserialization, but hopefully it would be possible to make the `storage` table optional for backwards-compatibility.

The config value for `file_size_limit` would then be passed to the environment of the storage-api container when it is started, similar to how other config values are used.

None come to mind, other than modifying my local supabase CLI to do the above or to hardcode a different file size limit.

## Additional context

If this is an acceptable suggestion, I can likely submit a PR.

grschafer:

Feature

Adds [storage] table to config.toml with a file_size_limit_bytes value that defaults to 50 MiB. This config value becomes the FILE_SIZE_LIMIT env var for the storage docker container.

The storage-api's file size limit is hard-coded to 50 MiB and is not configurable.

This PR fixes #489

## What is the new behavior?

The storage-api's file size limit can be configured in config.toml like below:

```
[storage]
# The maximum file size allowed, in bytes.
file_size_limit_bytes = 52428800
```

If the storage table or the file_size_limit_bytes key are not specified, the limit defaults to 50 MiB.

## Additional context

I'm unfamiliar with Go, so please let me know if I did anything wrong! :grinning:

### DIFF
--- a/internal/utils/config.go
+++ b/internal/utils/config.go
@@ -60,6 +60,17 @@ var (
 	testInitConfigTemplate, _ = template.New("initConfig.test").Parse(testInitConfigEmbed)
 )
 
+// Type for turning human-friendly bytes string ("5MB", "32kB") into an int64 during toml decoding.
+type sizeInBytes struct {
+	Value int64
+}
+
+func (a *sizeInBytes) UnmarshalText(text []byte) error {
+	var err error
+	a.Value, err = units.FromHumanSize(string(text))
+	return err
+}
+
 var Config config
 
 type (
@@ -98,7 +109,7 @@ type (
 	}
 
 	storage struct {
-		FileSizeLimit string `toml:"file_size_limit"`
+		FileSizeLimit sizeInBytes `toml:"file_size_limit"`
 	}
 
 	auth struct {
@@ -189,8 +200,8 @@ func LoadConfigFS(fsys afero.Fs) error {
 		if Config.Inbucket.Port == 0 {
 			return errors.New("Missing required field in config: inbucket.port")
 		}
-		if Config.Storage.FileSizeLimit == "" {
-			Config.Storage.FileSizeLimit = 5 * units.MB
+		if Config.Storage.FileSizeLimit.Value == 0 {
+			Config.Storage.FileSizeLimit.Value = 5 * units.MB
 		}
 		if Config.Auth.SiteUrl == "" {
 			return errors.New("Missing required field in config: auth.site_url")
--- a/internal/utils/config.go
+++ b/internal/utils/config.go
@@ -61,13 +61,13 @@ var (
 )
 
 // Type for turning human-friendly bytes string ("5MB", "32kB") into an int64 during toml decoding.
-type sizeInBytes struct {
-	Value int64
-}
+type sizeInBytes int64
 
-func (a *sizeInBytes) UnmarshalText(text []byte) error {
-	var err error
-	a.Value, err = units.FromHumanSize(string(text))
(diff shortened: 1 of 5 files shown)


internal/utils/config.go
```diff
@@ -95,6 +96,10 @@ type (
 		Pop3Port uint `toml:"pop3_port"`
 	}
 
+	storage struct {
+		FileSizeLimit uint `toml:"file_size_limit_bytes"`
```
sweatybridge, line 100:
```suggestion
		FileSizeLimit string `toml:"file_size_limit"`
```


For better human readability, may I suggest typing this field as string and use docker's [go-units](https://pkg.go.dev/github.com/docker/go-units) package to parse acceptable values like `50MiB` or `100KB`? Small usability improvements like this can add up to make a difference in terms of developer experience.

sweatybridge (changes requested):
Just a small suggestion.

internal/utils/config.go
grschafer: I like the idea, and took a stab at implementing it. I broke my work into 2 commits:
- 8278165 - This is a minimal change that follows your suggestion of using a string field. However, this approach means we're not validating the string until it is used later (when starting containers), which feels like a bad experience.
- 012e632 - So, I took a stab at validating `file_size_limit` when parsing the toml ([here's an example I referenced in the toml package docs](https://github.com/BurntSushi/toml#using-the-marshaler-and-encodingtextunmarshaler-interfaces)). This means there's an extra struct and `Value` field -- but, I'm a novice with golang so perhaps you know how to make this cleaner.

Thoughts/suggestions?

grschafer: A few more things I wanted to point out:

1. The go-units package doesn't correctly handle binary sizes (https://github.com/docker/go-units/issues/31), but it's also not very actively maintained so this likely won't be fixed.
3. Supabase studio handles binary vs decimal byte sizes incorrectly (5MB is treated as 5MiB): https://github.com/supabase/supabase/blob/867fc0d739c145e8b48388ef2456157930e171d4/studio/tests/components/Storage/StorageSettings.utils.test.js#L13
4. The combination of the above 2 items means that if someone puts "5MB" into go-units (i.e. into the CLI config.toml) and "5MB" into supabase studio they'll get different values (5,000,000 vs 5,242,880).

sweatybridge (approved):

internal/utils/config.go
sweatybridge: > So, I took a stab at validating file_size_limit when parsing the toml

This is a great idea. We can go a step further by using a type alias instead of a struct.

sweatybridge: > As a result, if someone puts "5MB" into go-units (i.e. into the CLI config.toml) and "5MB" into supabase studio they'll get different values (5,000,000 vs 5,242,880), which seems unideal.

Thanks for pointing out this difference. For compatibility with studio, we can use [RAMInBytes](https://pkg.go.dev/github.com/docker/go-units#RAMInBytes) for now.

grschafer:
Thanks for pushing this over the finish line! 🎉

Issue #404, opened 2022-08-22
Migration statements not running in order?


maxkostow:
# Bug report

## Describe the bug

Getting `42P01` (table does not exist)  when running migrations with `insert` statements

## To Reproduce

1. Use supabase CLI 1.1.1 (#396 seems related)
2. Create the migration file
```sql
CREATE TABLE IF NOT EXISTS public.example
(
    id bigint NOT NULL,
    name text COLLATE pg_catalog."default" NOT NULL,
    CONSTRAINT example_pkey PRIMARY KEY (id)
)

TABLESPACE pg_default;
ALTER TABLE IF EXISTS public.example
    OWNER to postgres;
GRANT ALL ON TABLE public.example TO anon;
GRANT ALL ON TABLE public.example TO authenticated;
GRANT ALL ON TABLE public.example TO postgres;
GRANT ALL ON TABLE public.example TO service_role;

INSERT INTO public.example VALUES (1, 'a');
```
3. Run
```sh
supabase stop
supabase start
supabase db reset --debug
```
5. See error
```
PG Recv: {"Type":"ErrorResponse","Severity":"ERROR","SeverityUnlocalized":"ERROR","Code":"42P01","Message":"relation \"public.example\" does not exist","Detail":"","Hint":"","Position":13,"InternalPosition":0,"InternalQuery":"","Where":"","SchemaName":"","TableName":"","ColumnName":"","DataTypeName":"","ConstraintName":"","File":"parse_relation.c","Line":1360,"Routine":"parserOpenTable","UnknownFields":null}
``` 

## Expected behavior

The migration should run without issue as it did in `v0.38.4`

## Additional Information

If you move the `insert` to a separate file, the `db reset` completes successfully, however this will cause issues with our remote database as the migrations are already live.

sweatybridge:
Hi @maxkostow , thanks for the detailed report. We are able to narrow down the problem to pgx statement cache and implement a workaround.

Feel free to open this issue again if you still face the same problem in `v1.1.3`.

PR #510, opened 2022-10-12
fix: remove session config from dumped schema

### LINKED ISSUE #507: supabase start error with generated migration without schema of the table
# Bug report

## Describe the bug

`supabase start` can't accept generated migrations in local dev

## To Reproduce
1. init new supabase local project `supabase init`
2. Run `supabase start`
3. use GUI(Supabase studio) and create two tables with foreign key.
4. create a migration using CLI `supabase db diff --use-migra -f test`
5. Run `supabase stop`
6. Run `supabase start`
7. See error
```
Applying migration 20221011193213_test.sql...
Error: ERROR: relation "[YOUR_TABLE_NAME]" does not exist (SQLSTATE 42P01)
```

I noticed that the generated migration doesn't always add a schema to the table name.

## Example generated migration
```
...
alter table "public"."test" add constraint "test_board_id_fkey" FOREIGN KEY (board_id) REFERENCES [!!!missing schema-->]boards(board_id) not valid;
...
```
in this example is missing the schema name in the boards table
if I edit the file and add the schema manually, then the error disappears.
```
...
alter table "public"."test" add constraint "test_board_id_fkey" FOREIGN KEY (board_id) REFERENCES public.boards(board_id) not valid;
...
```

sweatybridge:

Bug fix #507 

Session configs are persistent per connection. When using pgx, those emitted from `pg_dump` are carried over to subsequent migrations, resulting in unintended side effects.

## What is the new behavior?

- use [local config](https://www.postgresql.org/docs/13/sql-set.html) where possible
- updates migra to latest version

## Additional context


soedirgo:
Does [`RESET`](https://www.postgresql.org/docs/current/sql-reset.html) work better? Have not tested though

maxkostow:
Could this also be similar to #407 ?

soedirgo (approved):

soedirgo:
@maxkostow might be a different issue I think. In https://github.com/supabase/cli/issues/404 the tables are schema-qualified but it's still failing - in this case the table isn't schema-qualified (`boards` instead of `public.boards`), which failed because of the `SET`s by pg_dump.

sweatybridge:
> Does [RESET ALL](https://www.postgresql.org/docs/current/sql-reset.html) work better? Have not tested though

Changed to `reset all` since it works and is simpler.

PR #512, opened 2022-10-12
fix: improper url-escaping of db password

soedirgo:

Bug fix

db password `pass word` is parsed as `pass+word` because of `url.QueryEscape()`

## What is the new behavior?

db password `pass word` is parsed as `pass word`


sweatybridge (approved):
Nice catch. I wish pgx supports using config object directly instead of always parsing strings.

soedirgo:
+1 to that! 😄

PR #521, opened 2022-10-12
fix: add backup flag to stop command

### LINKED ISSUE #150: supabase start/stop no longer restores development database
# Bug report
       
## Describe the bug

Prior to supabase cli 0.16.0/1 when you started and stopped the supabase service it would dump and restore your database so that any data and schema changes would get restored.  It appears that supabase stop no longer writes to the `dump.sql` file when you stop the service.

## To Reproduce

1. Run `supabase start`
2. Make changes to the database
3. Run `supabase stop`
4. Run `supabase start`
5. Check that changes from step 2 are still in the database.

## Expected behavior
 
I would expect the data and schema changes to still exist after restarting the service.

## Additional context

I believe this worked by writing to a file named `dump.sql`.  Another solution could be to mount a local folder in the hose system as a volume in the container for the postgresql data folder so the database isn't ephemeral.

sweatybridge:

Bug fix #150 

on supabase stop, `supabase/.branches` directory is deleted. existing data in your local database is lost
on supabase start, empty directories in `supabase/.branches` are deleted

## What is the new behavior?

- supabase stop with `--backup` flag dumps your local database to `supabase/.branches/<current_branch>/dump.sql`
- using backup flag also keeps `supabase/.branches` directory in tact. but on supabase start, empty dirs are still deleted

## Additional context

- tested locally with 18 common extensions
"pg_cron" 
"pg_graphql" 
"pg_net" 
"pgsodium" 
"plv8" 
"fuzzystrmatch" 
"http" 
"moddatetime" 
"pg_jsonschema" 
"pg_stat_statements" 
"pgcrypto" 
"pgjwt" 
"pgtap" 
"plpgsql_check" 
"postgis" 
"postgis_raster" 
"uuid-ossp"


soedirgo:
No need for the flag, we should make this the default behavior.

sweatybridge:
Let's make it opt-in before changing the default behaviour. It's tricky to restore the dump for every case given the number of extensions we have. Better to give users a fallback behaviour in case the dump doesn't work.

internal/stop/stop.go
```diff
@@ -23,21 +32,49 @@ func Run(ctx context.Context, fsys afero.Fs) error {
...
-	branchDir := filepath.Dir(utils.CurrBranchPath)
-	if err := fsys.RemoveAll(branchDir); err != nil {
-		return err
+	if !backup {
+		// Remove other branches
+		branchDir := filepath.Dir(utils.CurrBranchPath)
+		if err := fsys.RemoveAll(branchDir); err != nil {
+			fmt.Fprintln(os.Stderr, err)
+		}
 	}
```
soedirgo, line 53: Let's just keep other branches? Otherwise people can't `supabase stop` without backing up without also removing other branches.

sweatybridge: Current behaviour is

- Without `backup` flag: all branch dirs are deleted
- With `backup` flag: current branch is dumped, other branches are kept as empty dir 

When running `supabase start`, empty branch dirs are deleted.

soedirgo: OK, let's just punt the local branching situation

soedirgo (approved):

PR #538, opened 2022-10-18
feat: add `--experimental`

soedirgo:

Feature

No way to specify which commands are experimental.

## What is the new behavior?

Experimental commands can do:

```go
myExperimentalCmd.MarkFlagRequired("experimental")
```

to make sure the flag is explicitly set

## Additional context

To be used for https://github.com/supabase/cli/pull/534

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -12,8 +12,11 @@ import (
 	"github.com/supabase/cli/internal/utils"
 )
 
-// Passed from `-ldflags`: https://stackoverflow.com/q/11354518.
-var version string
+var (
+	// Passed from `-ldflags`: https://stackoverflow.com/q/11354518.
+	version      string
+	experimental bool
+)
 
 var rootCmd = &cobra.Command{
 	Use:           "supabase",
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -64,6 +67,9 @@ func init() {
 		key := strings.ReplaceAll(f.Name, "-", "_")
 		cobra.CheckErr(viper.BindPFlag(key, flags.Lookup(f.Name)))
 	})
+
+	flags.BoolVar(&experimental, "experimental", false, "enable experimental features")
```
sweatybridge, line 71:
```suggestion
	flags.BoolVar(&experimental, "experimental", false, "enable experimental features (commands that require this flag may break semver)")
```

Shall we explain semver rules for this flag or do you think it's ok to be implicit?

sweatybridge (approved):

cmd/root.go
soedirgo: I think it's OK - should be clear from the flag name

Issue #578, opened 2022-11-03
db test command does not print failing test results


bhaan:
# Bug report

## Describe the bug

When running the command `supabase db test` or `supabase test db`, if any of the tests fail, the CLI does not print the `pg_prove` test results to the terminal window. Instead it only prints a generic `Error: error executing command` message.

## To Reproduce

Create a file like `supabase/tests/failure.sql` with the following contents:

```sql
begin;
select * from plan(1);

select fail('this is expected to fail');

select * from finish();
rollback;
```

Then run:
```
$ supabase db test
```

## Expected behavior

I expect to see the failing test results as described by the internally invoked `pg_prove` command. For example:

```
$ supabase db test
/tmp/supabase/tests/failure.sql ..
# Failed test 1: "this is expected to fail"
# Looks like you failed 1 test of 1
Failed 1/1 subtests

Test Summary Report
-------------------
/tmp/supabase/tests/failure.sql (Wstat: 0 Tests: 1 Failed: 1)
  Failed test:  1
Files=1, Tests=1,  0 wallclock secs ( 0.01 usr  0.00 sys +  0.01 cusr  0.00 csys =  0.02 CPU)
Result: FAIL

Error: error executing command
Try rerunning the command with --debug to troubleshoot the error.
```

## Actual behavior

Instead I only see an error message suggesting that the internally invoked command has failed.

```
$ supabase db test
Error: error executing command
Try rerunning the command with --debug to troubleshoot the error.
```

## Additional context

If all the tests pass, the CLI outputs the `pg_prove` test results as I would expect, suggesting that the CLI is not failing for a different reason:

```
$ supabase db test
/tmp/supabase/tests/success.sql .. ok
All tests successful.
Files=1, Tests=1,  0 wallclock secs ( 0.02 usr  0.00 sys +  0.02 cusr  0.00 csys =  0.04 CPU)
Result: PASS
```

## System information

- OS: Ubuntu 22.04
- Version of supabase/cli: 1.11.4

PR #579, opened 2022-11-03
fix: show pg_prove failing test results

bhaan:

Bug fix / enhancement

See #578 

The `supabase db test` command does not print the `pg_prove` test results if any of the tests fail. This of course makes it difficult to identify the failing test.

## What is the new behavior?

The command will now print the stdout results from the `pg_prove` command if it exits with a non-zero exit code.

The new output looks like what I shared in the "Expected behavior" section of the referenced issue.

## Additional context

This makes a small change to a shared function in `internal/utils/docker.go`. It's not clear to me if this will have undesired side effects for other commands. Let me know if you think this could be handled differently.

Thanks!

### DIFF
--- a/internal/db/test/templates/test.sh
+++ b/internal/db/test/templates/test.sh
@@ -3,7 +3,8 @@ set -euo pipefail
 
 # TODO: move this dependency to base image
 if ! command -v pg_prove &> /dev/null; then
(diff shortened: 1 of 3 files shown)


bhaan:
Thank you for the quick response and merge! :beers:

PR #601, opened 2022-11-13
chore: Use ESZIP format to deploy functions

laktek:

This changes `supabase function deploy` command to use ESZIP format.

* ESZIP building is done via a Deno script that's copied to `$HOME/.supabase/eszip`

* The older JS bundling is still supported and can be used via `--legacy-bundle` flag.

* With ESZIP, we no longer need to keep the Deno version pinned to 1.20.X. This commit contains changes to update Deno to `1.27.X` and also use a latest image of `functions-relay` (1.4.0), for locally serving functions.

TODO:
[x] Update Tests


internal/utils/misc.go
```diff
@@ -118,6 +120,9 @@ var (
 	DenoPathOverride string
 )
 
+//go:embed eszip/*
+var eszipEmbedDir embed.FS
+
```
sweatybridge, line 125: nitpick: can we move this to below line 84?

internal/utils/misc.go
```diff
@@ -376,6 +381,80 @@ func InstallOrUpgradeDeno(ctx context.Context, fsys afero.Fs) error {
...
+// Copy ESZIP scripts needed for function deploy, returning the build script path or an error.
+func CopyEszipScripts(ctx context.Context, fsys afero.Fs) (string, error) {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+
+	supabasePath := filepath.Join(home, ".supabase")
+	scriptDirPath := filepath.Join(supabasePath, "eszip")
+	buildScriptPath := filepath.Join(scriptDirPath, "build.ts")
```
sweatybridge, line 412: nitpick: can we make this path relative to `GetDenoPath`? slightly easier to keep track

cmd/functions.go
```diff
@@ -89,6 +90,7 @@ func init() {
 	functionsDeleteCmd.Flags().StringVar(&projectRef, "project-ref", "", "Project ref of the Supabase project.")
 	functionsDeployCmd.Flags().BoolVar(&noVerifyJWT, "no-verify-jwt", false, "Disable JWT verification for the Function.")
 	functionsDeployCmd.Flags().StringVar(&projectRef, "project-ref", "", "Project ref of the Supabase project.")
+	functionsDeployCmd.Flags().BoolVar(&useLegacyBundle, "legacy-bundle", false, "Use legacy bundling mechanism.")
```
sweatybridge, line 93: Do you think it's worth defaulting to `true` initially and flip it may be next week? I assume we haven't tested on windows?

internal/utils/eszip/build.ts
```diff
@@ -0,0 +1,44 @@
+import * as path from "https://deno.land/std@0.127.0/path/mod.ts";
```
sweatybridge, line 1: Is there any documentation on when and how to update these scripts?

sweatybridge (approved):
lgtm with some minor comments

internal/utils/misc.go
laktek: Good idea! fixed in 2189242

internal/utils/misc.go
laktek: fixed in 2189242

cmd/functions.go
laktek: Yup, good call on testing on Windows. Can we do a pre-release for testing? I have set the flag to true in 6603bde

internal/utils/eszip/build.ts
laktek: I simplified this to only have `eszip/build.ts` as an embed in f01cb72.  That should make updating modules easier.

laktek:
@sweatybridge I've addressed the PR feedback. Can you take another look before I merge?

internal/functions/deploy/deploy.go
```diff
@@ -64,29 +78,57 @@ func Run(ctx context.Context, slug string, projectRefArg string, verifyJWT bool,
...
+					Name:      slug,
+					Slug:      slug,
+					VerifyJwt: &verifyJWT,
+				})
+			} else {
+				resp, err = utils.GetSupabase().CreateFunctionWithBodyWithResponse(ctx, projectRef, &api.CreateFunctionParams{
+					Slug:      &slug,
+					Name:      &slug,
+					VerifyJwt: &verifyJWT,
+				}, "application/vnd.denoland.eszip", functionBody)
```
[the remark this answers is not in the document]
laktek: fixed and squashed.

PR #606, opened 2022-11-15
fix: error handling for CAA records during custom domains setup

darora:


sweatybridge (approved):

internal/hostnames/common.go
```diff
@@ -105,6 +109,17 @@ Please ensure that your custom domain is set up as a CNAME record to your Supaba
 		if res.Result.Ssl.Status == "initializing" {
 			return appendRawOutputIfNeeded("Custom hostname setup is being initialized; please request re-verification in a few seconds.\n", response, includeRawOutput), nil
 		}
+		if res.Result.Ssl.ValidationErrors != nil && len(res.Result.Ssl.ValidationErrors) > 0 {
```
sweatybridge, line 112:
```suggestion
		if len(res.Result.Ssl.ValidationErrors) > 0 {
```
nitpick: this won't be nil because the json tag doesn't have `omitempty`
nitpick: can this still be nil without an explicit `omitempty` tag?

darora: dunno - being explicit is simple for me but happy to accept it at face value

PR #614, opened 2022-11-16
feat: enable ESZIPs for functions

laktek:

Enable ESZIP format for deploying functions by default. Also, updates Deno to 1.28.0.

Currently, ESZIP format is disabled and can be enabled via `--legacy-bundle=false` flag.

## What is the new behavior?

ESZIPs are enabled by default and if you need legacy bundling you will need to add `--legacy-bundle=true` flag.

### DIFF
--- a/internal/utils/misc.go
+++ b/internal/utils/misc.go
@@ -34,7 +34,7 @@ const (
 	MigraImage     = "djrobstep/migra:3.0.1621480950"
 	PgmetaImage    = "supabase/postgres-meta:v0.50.2"
 	StudioImage    = "supabase/studio:0.22.08"
-	DenoRelayImage = "supabase/deno-relay:v1.4.0"
+	DenoRelayImage = "supabase/deno-relay:v1.5.0"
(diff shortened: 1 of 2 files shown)


internal/utils/misc.go
```diff
@@ -304,7 +304,7 @@ func InstallOrUpgradeDeno(ctx context.Context, fsys afero.Fs) error {
 
 	if _, err := fsys.Stat(denoPath); err == nil {
 		// Upgrade Deno.
-		cmd := exec.CommandContext(ctx, denoPath, "upgrade", "--version", "1.27.1")
+		cmd := exec.CommandContext(ctx, denoPath, "upgrade", "--version", "1.28.0")
```
soedirgo, line 307: Should we just remove the `--version` flag? That way it stays recent (i.e. always on stable)

laktek: we'd probably want to keep it the same version as `functions serve` (set via Docker image). I think it's better to set the version.

sweatybridge (approved):

PR #634, opened 2022-11-24
Switch npm hook to postinstall

### LINKED ISSUE #620: preinstall doesn't add to path when using yarn add
# Bug report

## To Reproduce

1. Go to '…'
2. Click on '…'
3. Scroll down to '…'

## System information

- OS: [e.g. macOS, Windows]
- Browser (if applies) [e.g. chrome, safari]
- Version of supabase-js: [e.g. 6.0.2]
- Version of Node.js: [e.g. 10.10.0]

ZackDeRose:

Bug fix #620 

When you install e.g.: `yarn add -D supabase` this does not install the supabase cli to the node_modules/bin directory. This is because is is being run on preinstall, and it doesn't yet have access to the script it wants to run as it hasn't been downloaded yet. So you end up having to run `yarn add supabase` (or equivalent) a second time.

## What is the new behavior?

Installs on the postinstall hook.

## Additional context

Tested with Verdaccio quick to verify, it attempted to install on the `yarn add -D supabase`, as it attempted to download the gz file from here: https://github.com/supabase/cli/releases/download/v999.9.9/supabase_999.9.9_darwin_arm64.tar.gz - note the mocked version number I published to my local registry to test. I'm assuming you're running simple version mapping here, so I assume that failure is actually a signal that this did work properly the first time.


sweatybridge:
Turns out `preinstall` works with npm but not yarn. Also, vice versa `postinstall` doesn't work with npm but works with yarn.

So I'm afraid we need to test this a bit more.

sweatybridge:
Closing as I have a separate fix open.

PR #648, opened 2022-11-29
chore: remove all tabs & trailing spaces from SQL files

kbsali:

docs update

When dumping an entire project, the SQL sometimes has tabs instead of 2 spaces (which seems the default)

### DIFF
--- a/internal/utils/templates/initial_schemas/13.sql
+++ b/internal/utils/templates/initial_schemas/13.sql
@@ -80,7 +80,7 @@ CREATE EXTENSION IF NOT EXISTS pgcrypto WITH SCHEMA extensions;
 
 
 --
--- Name: EXTENSION pgcrypto; Type: COMMENT; Schema: -; Owner: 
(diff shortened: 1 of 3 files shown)


sweatybridge:
> When dumping an entire project, the SQL sometimes has tabs

@kbsali are you referring to running pg_dump on your local dev database? If so, that is not really within the scope for CLI. What we do support is dumping remote project with `db remote commit`.

That said, we are looking into generating these initial schemas in a fully automated way. So our preference is to fix the schemas in the respective services upstream. In this case, it is https://github.com/supabase/gotrue/pull/843.

kbsali:
ok, perfect. I wasn't sure how these sql files were generated.
That being said, there are more pieces of SQL with "wrong" indent types like `storage.extension`... where is that coming from ? (not Gotrue as far as I can see)

sweatybridge:
Probably from either of these repos

- https://github.com/supabase/storage-api/tree/master/migrations/tenant
- https://github.com/supabase/postgres/tree/develop/migrations

PR #662, opened 2022-12-02
feat: support import maps in functions serve and deploy

laktek:

This adds support for import maps in `function serve` and `function deploy`. 

### What are Import Maps?

https://deno.land/manual@v1.28.3/basics/modules/import_maps

For deployed functions, `import_map.json` should always be placed in function's top-level directory.

### DIFF
--- a/internal/functions/serve/serve.go
+++ b/internal/functions/serve/serve.go
@@ -21,7 +21,7 @@ const (
 	relayFuncDir = "/home/deno/functions"
 )
 
-func Run(ctx context.Context, slug string, envFilePath string, verifyJWT bool, fsys afero.Fs) error {
+func Run(ctx context.Context, slug string, envFilePath string, verifyJWT bool, importMap bool, fsys afero.Fs) error {
 	// 1. Sanity checks.
 	{
 		if err := utils.AssertSupabaseCliIsSetUpFS(fsys); err != nil {
@@ -95,9 +95,17 @@ func Run(ctx context.Context, slug string, envFilePath string, verifyJWT bool, f
 	// 4. Start Function.
 	localFuncDir := filepath.Join(utils.FunctionsDir, slug)
 	dockerFuncPath := relayFuncDir + "/" + slug + "/index.ts"
+
+	// set import map
+	importMapFlag := "--"
+	if importMap {
+		importMapFullPath := filepath.Join(relayFuncDir, slug, "import_map.json")
+		importMapFlag = "--import-map=" + importMapFullPath
+	}
+
 	fmt.Println("Starting " + utils.Bold(localFuncDir))
 	if _, err := utils.DockerExecOnce(ctx, utils.DenoRelayId, nil, []string{
-		"deno", "cache", dockerFuncPath,
+		"deno", "cache", importMapFlag, dockerFuncPath,
 	}); err != nil {
 		return err
 	}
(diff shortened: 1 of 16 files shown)


soedirgo:
Can we make the `import_map.json` path non-configurable? It's a bit odd that it can be set for `serve` but not `deploy`

laktek:
> Can we make the import_map.json path non-configurable? It's a bit odd that it can be set for serve but not deploy

Yup, I had the same thought after making the PR :) let's do that. So a `--import-map` flag on serve would look for the file on default location.

laktek:
@soedirgo I've updated the PR to make the `import_map.json` path to be non-configurable in serve command.

soedirgo:
Added a commit to:
- remove the option altogether and just omit the option if there's no `import_map.json`
- fix import map path to always use Unix path separator, since the container image is always assumed to be Linux

internal/functions/serve/serve.go
```diff
@@ -94,11 +94,26 @@ func Run(ctx context.Context, slug string, envFilePath string, verifyJWT bool, f
...
 	localFuncDir := filepath.Join(utils.FunctionsDir, slug)
+	localImportMapPath := filepath.Join(localFuncDir, "import_map.json")
+	// We assume the image is always Linux, so path separator must always be `/`.
+	// We can't use filepath.Join because it uses the path separator for the host system, which is `\` for Windows.
 	dockerFuncPath := relayFuncDir + "/" + slug + "/index.ts"
+	dockerImportMapPath := relayFuncDir + "/" + slug + "/import_map.json"
+
+	denoCacheCmd := []string{"deno", "cache"}
+	{
+		if _, err := os.Stat(localImportMapPath); err == nil {
```
sweatybridge, line 105:
```suggestion
		if _, err := fsys.Stat(localImportMapPath); err == nil {
```

internal/functions/serve/serve.go
```diff
@@ -125,14 +140,24 @@ func Run(ctx context.Context, slug string, envFilePath string, verifyJWT bool, f
 			}
 		}
 
+		denoRunCmd := []string{"deno", "run", "--no-check=remote", "--allow-all", "--watch", "--no-clear-screen", "--no-npm"}
+		{
+			if _, err := os.Stat(localImportMapPath); err == nil {
```
sweatybridge, line 145:
```suggestion
			if _, err := fsys.Stat(localImportMapPath); err == nil {
```

sweatybridge (changes requested):

internal/utils/eszip/build.ts
```diff
@@ -8,27 +8,42 @@ import { load } from "https://deno.land/x/eszip@v0.30.0/loader.ts";
...
-        actualPath = specifier.replace(virtualBasePath, `./${path.dirname(p)}/`);
+        actualPath = specifier.replace(virtualBasePath, `./${funcDirPath}/`);
       }
-      const content = await Deno.readTextFile((actualPath));
-      return {
-        kind: "module",
-        specifier,
-        content
+      try {
+        const content = await Deno.readTextFile((actualPath));
```
sweatybridge, line 25: Is the inner bracket redundant?
```suggestion
        const content = await Deno.readTextFile(actualPath);
```

internal/functions/serve/serve.go
soedirgo: Oops, thanks!

sweatybridge (approved):

laktek:
@sweatybridge @soedirgo Just merge this PR yet. Noticed an issue with Deno Deploy config (which would affect older functions), we probably need to do a small change before releasing this.

laktek:
@soedirgo @sweatybridge I had to add another commit to the PR to send `import_map` param to the API when creating or updating a function. https://github.com/supabase/cli/pull/662/commits/f8279b71cc7ae0961bfbbb483a09fc44f99d4ea5 

This ensures we don't break old functions without an import map.

Issue #624, opened 2022-11-19
exclude parameter for `supabase start`


samuba:
# Feature request

## Is your feature request related to a problem? Please describe.

Starting supabase in CI environments (Github Actions) takes a long time because all containers are freshly pulled every time, but you might not even need all of them. 

## Describe the solution you'd like

It would be lovely to have something like an exclude parameter for `supabase start` where you can specify the containers that you do not need and those will not be started/pulled.
Like this for example: `supabase start --exclude inbucket --exclude studio`

Would you be open for a PR on this?

sweatybridge:
> Like this for example: supabase start --exclude inbucket --exclude studio

Hi @samuba, that's a great idea. I think Postgres is the only required container for basic CI. Feel free to tag me in a draft PR if you run into any difficulties.

PR #672, opened 2022-12-06
feat: exclude flag for start command

samuba:

It introduces this feature: https://github.com/supabase/cli/issues/624

## What is the new behavior?

Containers can be prevented from being pulled/started with a new `--exclude` flag

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -1,20 +1,35 @@
 package cmd
 
 import (
+	"strings"
+
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


samuba:
@sweatybridge I just realised that you already acted on this feature with https://github.com/supabase/cli/pull/642.
I still think this PR is useful because in my CI and dev environment I actually need most of the containers, except for some selected one.

sweatybridge:
> I still think this PR is useful because in my CI and dev environment I actually need most of the containers

Yup, understand. I will make some stylistic changes before merging. LGTM

PR #674, opened 2022-12-07
fix: show script size after function deploy

laktek:

Currently, users don't have an easy way to find their function's size when deploying. Supabase checks and returns an error in the API layer if the script is above the permitted limit. However, the error is not very user-friendly since users don't know the size of their script.

This PR prints another line when deploying a function (ie. `deploying {name} (script size: X.XMB)`)

Screen Shot 2022-12-07 at 9 25 42 pm

This should also work when `--legacy-bundle` flag is set.

### DIFF
--- a/internal/functions/deploy/deploy.go
+++ b/internal/functions/deploy/deploy.go
@@ -12,12 +12,12 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/docker/go-units"
 	"github.com/spf13/afero"
 	"github.com/supabase/cli/internal/utils"
 	"github.com/supabase/cli/pkg/api"
 )
 
-const MB float64 = 1 << (10 * 2)
 const EszipContentType = "application/vnd.denoland.eszip"
 
 func Run(ctx context.Context, slug string, projectRefArg string, verifyJWT bool, useLegacyBundle bool, fsys afero.Fs) error {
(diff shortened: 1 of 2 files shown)


internal/functions/deploy/deploy.go
```diff
@@ -95,7 +98,10 @@ func makeLegacyFunctionBody(functionBody io.Reader) (string, error) {
 	return buf.String(), nil
 }
 
-func deployFunction(ctx context.Context, projectRef, slug string, functionBody io.Reader, verifyJWT, useLegacyBundle bool) error {
+func deployFunction(ctx context.Context, projectRef, slug string, functionBody io.Reader, verifyJWT, useLegacyBundle bool, functionSize int) error {
+	functionSizeFmt := fmt.Sprintf("%.2fMB", float64(functionSize)/MB)
```
sweatybridge, line 102:
```suggestion
	functionSizeFmt := units.HumanSize(functionSize)
```

internal/functions/deploy/deploy.go
```diff
@@ -95,7 +98,10 @@ func makeLegacyFunctionBody(functionBody io.Reader) (string, error) {
...
+	fmt.Println("Deploying " + utils.Bold(slug) + " (script size: " + utils.Bold(functionSizeFmt) + ")")
```
sweatybridge, line 103: How about moving this print statement to line 87? then we dont need to change this function signature.

internal/functions/deploy/deploy.go
laktek: Updated in a4390a3

sweatybridge (approved):

PR #677, opened 2022-12-08
feat: adds support for vanity subdomains

darora:


cmd/vanitySubdomains.go
```diff
@@ -0,0 +1,95 @@
...
+		Use:     "vanity-subdomains",
+		Short:   "Manage vanity subdomains for Supabase projects",
+		Long: `Manage vanity subdomains for Supabase projects.
+
+Usage of vanity subdomains and custom domains is mutually exclusive.`,
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			if !experimental {
+				return errors.New("must set the --experimental flag to run this command")
+			}
+			return nil
```
sweatybridge, line 28:
```suggestion
			return cmd.Root().PersistentPreRunE(cmd, args)
```

I just realised that cobra [PersistentPreRun is not chained](https://github.com/spf13/cobra/issues/252) from root cmd. So we need to call it explicitly when overriding

sweatybridge (approved):
The rest lgtm

PR #678, opened 2022-12-08
feat: allow providing custom import-map paths to function serve/deploy

laktek:

This PR allows users to set a custom `--import-map` flag with relative path to an import map. 

This covers following use cases:
* Creating a project-level import_map.json outside of `supabase/functions` to be shared across all functions in the project.
* If you run `deno vendor`, those vendored dependencies needs to be linked using `vendor/import_map.json`

### DIFF
--- a/internal/utils/eszip/build.ts
+++ b/internal/utils/eszip/build.ts
@@ -7,7 +7,7 @@ import { load } from "https://deno.land/x/eszip@v0.30.0/loader.ts";
 
 const virtualBasePath = "file:///src/";
 
-async function buildAndWrite(p: string) {
+async function buildAndWrite(p: string, importMapPath: string) {
   const funcDirPath = path.dirname(p);
   const entrypoint = new URL("index.ts", virtualBasePath).href;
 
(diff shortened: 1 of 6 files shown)


cmd/root.go
```diff
@@ -31,6 +31,7 @@ var rootCmd = &cobra.Command{
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 		cmd.SilenceUsage = true
 		if viper.GetBool("DEBUG") {
+			fmt.Println("CLI version: " + version)
```
sweatybridge, line 34:
```suggestion
			fmt.Fprintln(os.Stderr, "CLI version: " + version)
```

Debug logs typically go to stderr because we don't want it to interfere with actual command output.

internal/utils/eszip/build.ts
```diff
@@ -21,6 +21,12 @@ async function buildAndWrite(p: string) {
       if (specifier.startsWith(virtualBasePath)) {
         actualPath = specifier.replace(virtualBasePath, `./${funcDirPath}/`);
       }
+
+      // If an import map path is set read file from the given path.
+      // Otherwise default to `import_map.json` in functions directory.
+      if (specifier.endsWith('import_map.json') && importMapPath != "") {
```
sweatybridge, line 27:
```suggestion
      if (specifier.endsWith('import_map.json') && importMapPath) {
```

`Deno.args[1]` may be undefined if called with only 1 arg. Safer to handle both cases

cmd/root.go
laktek: ah wasn't intending to actually add this log to the PR. will remove

laktek: removed and rebased

sweatybridge (approved):

PR #684, opened 2022-12-11
fix: regex pattern for migration file

### LINKED ISSUE #683: migrations list picks up hidden files and those not ending in .sql, such as vim swapfiles
# Bug report

## Describe the bug

Commands that interact with migrations pick up any file that contains the given regex, *not* those that end in `.sql` or that aren't hidden files.

tl;dr

https://github.com/supabase/cli/blob/72f79c76c477525f7f86d703848ce67daf960cdd/internal/utils/misc.go#L106 

should be changed to something like

```go
MigrateFilePattern = regexp.MustCompile(`^[^\.].*([0-9]+)_.*\.sql$`)
```

(note: I don’t know golang really)

## To Reproduce

```
> supabase migration new test
Created new migration at supabase/migrations/20221210224402_test.sql.
> touch supabase/migrations/.20221210224402_test.sql.swp
> supabase migration list

  
        LOCAL      │     REMOTE     │     TIME (UTC)
  ─────────────────┼────────────────┼──────────────────────
    20221210224402 │                │ 2022-12-10 22:44:02
    20221210224402 │                │ 2022-12-10 22:44:02
```

If you `supabase db push`, it will push both files.  If the second file was created because you ran vim, you get all kinds of weird errors because Postrgres tries to execute the contents of the vim swapfile.

## Expected behavior

Would expect it to ignore hidden files for sure, and also would only expect it to use files ending in `.sql`

## Screenshots

N/A

## System information

```
> supabase --version
1.22.1
```

installed on macos with homebrw

sweatybridge:

Bug fix #683

regex wasn't strict enough

## What is the new behavior?

add starting and termination characters

## Additional context

### DIFF
--- a/internal/utils/misc.go
+++ b/internal/utils/misc.go
@@ -103,7 +103,7 @@ var (
 	AccessTokenPattern = regexp.MustCompile(`^sbp_[a-f0-9]{40}$`)
 	ProjectRefPattern  = regexp.MustCompile(`^[a-z]{20}$`)
 	PostgresUrlPattern = regexp.MustCompile(`^postgres(?:ql)?:\/\/postgres:(.*)@(.+)\/postgres$`)
-	MigrateFilePattern = regexp.MustCompile(`([0-9]+)_.*\.sql`)
+	MigrateFilePattern = regexp.MustCompile(`^([0-9]+)_.*\.sql$`)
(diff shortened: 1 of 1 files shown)


davetron5000:
Thanks for the fast fix - looks like the issue is addressed!

PR #696, opened 2022-12-14
feat: implement db dump command

sweatybridge:

feature

difficult to dump the remote project schema with pg_dump as non-superuser

## What is the new behavior?

- Adds `db dump` command, defaults to stdout
- Adds `db dump -f schema.sql` flag to exporting to path
- For baselining, chain with pipe `supabase db dump | supabase migration new baseline`

## Additional context


soedirgo:
> supabase migration pull
supabase migration repair --status applied <version>

I would argue this is even more confusing because this is a potential footgun:
- user runs `supabase migration pull`
- user forgets to run `supabase migration repair`
- user runs `supabase migration push`

Let's just rename `db remote commit` but keep the old behavior.

sweatybridge:
> potential footgun

Can you think of case where a non-idempotent migration will succeed `db push` and cause harm?

soedirgo:
Something like creating indexes and sequences - you'd end up having duplicates.

We should also ask the flip side of this: is there a case where we want `migration pull` to *not* insert a migration version? And is that use case common enough that we want that to be the default? IMO `migration repair` should only be run if something is wrong with the migration history, whereas in the common case of `migration pull` things are working normally.

sweatybridge:
> creating indexes and sequences - you'd end up having duplicates.

I don't think that's the case for create sequence statements generated by pg_dump/migra? Each sequence should have a unique name https://www.postgresql.org/docs/current/sql-createsequence.html

> is there a case where we want migration pull to not insert a migration version? And is that use case common enough that we want that to be the default?

I believe a pull operation should not change the remote state. This is what I find most confusing about remote commit as we often have to explain to users that it is safe.

There are many reasons a user may choose to dump the remote schemas repeatedly. For example, backups, integration with 3rd party services, moving to/from self-hosted, checking schemas are correct, etc. I think it's nice to make all our fixes in dump_initial_migration.sh available to users, especially since we are removing superuser access. Having a more general purpose pull command helps with these use cases.

soedirgo:
> I don't think that's the case for create sequence statements generated by pg_dump/migra?

Maybe not, but users may edit the generated migration. But generally I think asserting there's no migration that's non-idempotent and doesn't fail to execute on subsequent runs is a big bet against unknown unknowns.

> This is what I find most confusing about remote commit as we often have to explain to users that it is safe.

But that's why it's called `db remote commit` in the first place instead of `db pull`: it *does* change state on the remote database - it "commits" the changes you made on the remote database as a migration.

> There are many reasons a user may choose to dump the remote schemas repeatedly

OK I didn't notice what this command was doing at first - I thought it was doing schema diffing. Then it doesn't look like it involves migrations at all? Shouldn't this be called `db dump` instead, and print to stdout instead of adding a migration? I can't think of a case where you already have some migrations, and then want to run this command, which saves as a migration some DDLs, some of which are already included in previous migrations.

soedirgo:
The "safe" alternative to `db remote commit` would be `db remote changes` (now `db diff --linked`) - that seems to capture some of the use cases there, like checking schemas are consistent.

So basically where I'm coming from is: I find that it's very common to be in a situation where you made some schema changes directly to the remote db and want to save it as a migration. I think there should be a single command for that, and it shouldn't be gated by scary-sounding commands like `repair`, because it's not in a bad state per se, we just need a way to get back to a state where you can start making schema changes locally.

sweatybridge:
For now I will remove the deprecation notice.

The pull/dump command is still useful for squashing migrations. By dumping into a new migration file, users can delete previous migration files which might be old and irrelevant. This process is also called baselining: https://flywaydb.org/documentation/command/baseline

I will also add a flag to support dumping schema to a user defined path. This will unlock many use cases for 3rd party tools.

soedirgo (approved):

PR #700, opened 2022-12-15
fix: mirror local DENO_* variables on function serve container

laktek:

This will mirror the following DENO env variables to the `functions serve` container.

`DENO_TLS_CA_STORE`
`DENO_CERT`
`DENO_AUTH_TOKENS`

https://deno.land/manual@v1.29.0/getting_started/setup_your_environment#environment-variables

### DIFF
--- a/internal/functions/serve/serve.go
+++ b/internal/functions/serve/serve.go
@@ -66,6 +66,13 @@ func Run(ctx context.Context, slug string, envFilePath string, verifyJWT bool, i
 			env = append(env, "VERIFY_JWT=false")
 		}
 
+		// mirror local DENO_* env variables to the Docker container.
+		env = append(env,
+			"DENO_TLS_CA_STORE="+os.Getenv("DENO_TLS_CA_STORE"),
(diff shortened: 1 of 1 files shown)


internal/functions/serve/serve.go
```diff
@@ -66,6 +66,13 @@ func Run(ctx context.Context, slug string, envFilePath string, verifyJWT bool, i
 			env = append(env, "VERIFY_JWT=false")
 		}
 
+		// mirror local DENO_* env variables to the Docker container.
+		env = append(env,
+			"DENO_TLS_CA_STORE="+os.Getenv("DENO_TLS_CA_STORE"),
+			"DENO_CERT="+os.Getenv("DENO_CERT"),
```
sweatybridge, line 72: don't we need to mount the cert into the deno container as well?

laktek: hmm..yeah you're right. env variable alone won't be enough as the container has no access to cert store. Wonder if this can be solved with a network config on docker?

laktek:
Closing this as this approach wouldn't work.

PR #722, opened 2022-12-19
Pass deno CLI args to functions serve

kruttik-lab49:

feature

Currently, there is no way to pass deno cli arguments to docker containers.

See discussion thread [here](https://github.com/supabase/supabase/discussions/10989).

## What is the new behavior?

This PR adds an ability to pass deno cli arguments to `supabase functions serve` command

## Additional context

Most of the details are in the discussion forum above.


.tool-versions
```diff
@@ -0,0 +1 @@
+golang 1.19.3
```
sweatybridge, line 1: Is there any preference for this over the current version specified in https://github.com/supabase/cli/blob/main/go.mod#L3?

.gitignore
```diff
@@ -20,3 +20,9 @@
 # NPM
 node_modules
 package-lock.json
+
+# Ignore built artifact - go build
+cli
+
+#Ignore IDE
+.idea
```
sweatybridge, line 28: let's group this under line 18 please

sweatybridge (commented):
A few nitpicks and general comments:

Is exposing all deno args necessary? As the CLI frequently upgrades deno version to match that used by the platform, I think it will be harder to maintain compatibility if we allow arbitrary deno flags.

How about exposing only `SUPABASE_DENO_CERT` env var which will be injected as `DENO_CERT` inside the container? This won't conflict with host system deno envs and minimizes the compatibility surface.

sweatybridge:
Anyway thanks for the thorough investigation. I've released a fix to pass user defined env vars from `--env-file` flag to all deno commands. Let me know if it solves your issue.

kruttik-lab49:
> Anyway thanks for the thorough investigation. I've released a fix to pass user defined env vars from `--env-file` flag to all deno commands. Let me know if it solves your issue.

Great!! I will give it a try and report back.

kruttik-lab49:
I can confirm that it works!! Thank you so much.

PR #734, opened 2022-12-22
fix: inject user envs to all deno commands

sweatybridge:

supersedes #722 
supersedes #700 

user defined envs are only available to functions

## What is the new behavior?

pass user defined envs to all deno commands

## Additional context

### DIFF
--- a/internal/functions/serve/serve.go
+++ b/internal/functions/serve/serve.go
@@ -22,6 +22,24 @@ const (
 	customDockerImportMapPath = "/home/deno/import_map.json"
 )
 
+func ParseEnvFile(envFilePath string) ([]string, error) {
+	env := []string{}
+	if len(envFilePath) == 0 {
+		return env, nil
+	}
+	envMap, err := godotenv.Read(envFilePath)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


internal/functions/serve/serve.go
```diff
@@ -22,6 +22,24 @@ const (
...
+	for name, value := range envMap {
+		if strings.HasPrefix(name, "SUPABASE_") {
+			return env, errors.New("Invalid env name: " + name + ". Env names cannot start with SUPABASE_.")
+		}
+		env = append(env, name+"="+value)
+	}
+	return env, nil
+}
+
 func Run(ctx context.Context, slug string, envFilePath string, verifyJWT bool, importMapPath string, fsys afero.Fs) error {
```
laktek, line 43: Looks like we missed to update `cmd/functions.go` with the new flag?

sweatybridge: It's using the same `--env-file` flag that we added previously :) Previously it was only exposed to one exec command, but now it's for the whole container so deno env vars can be defined inside too.

PR #773, opened 2023-01-07
fix: update gotrue to match platform

sweatybridge:

deps update

## What is the new behavior?

update pg13-pg15 with latest auth and realtime schemas

## Additional context

tested locally with db reset


soedirgo:
Do we also need to update initial schemas for PG13 & PG14?

PR #781, opened 2023-01-10
chore: add command examples to docgen

sweatybridge:

docs update https://github.com/supabase/supabase/pull/11565

## What is the new behavior?

more detailed generator for reference docs.
- merges with extended description
- merges with extended examples

```bash
go run docs/main.go > cli_v1_commands.yaml
```

## Additional context

Screenshot 2023-01-19 at 6 28 28 PM


docs/README.md
```diff
@@ -0,0 +1 @@
+## Extended man pages for CLI commands
```
soedirgo, line 1: These aren't for public use right? Should we move these to tools/docgen?

sweatybridge: hmm I'm considering moving docgen/main.go to `docs` actually, but no strong opinion here. Main motivation is to follow this [folder structure](https://github.com/golang-standards/project-layout/tree/master/docs). Also see this openshift example https://github.com/openshift/origin/blob/master/docs/cli.md.

soedirgo: Sgtm as long as it's clear that it's not user-facing

soedirgo (approved):

docs/README.md
sweatybridge: We can allow users to contribute to these markdown docs. This will be dumped to our docs site eventually. See this screenshot https://user-images.githubusercontent.com/1639722/213419223-3fc58030-19ec-4396-849e-6d68f2a3cad6.png

Issue #280, opened 2022-06-15
Allow auto detection of --no-verify-jwt


riderx:
# Feature request
When you update your functions, you have to do folder by folder and most of the time the only function who need `--no-verify-jwt` you forgot to do it.

This will never change, it can be cool to have a way to set for once.
I have made a script who looks into the function folder and deploy all of them
To handle my `--no-verify-jwt` issue, I added an empty file `.no-verify-jwt` into the folder `stripe_event` who is the one i need to have this.
When my script sees this, it runs the command with the option.
Then it's never forgotten.

## Describe the solution you'd like

It could be cool to integrate this directly in the CLI

## Describe alternatives you've considered

continuing to use my script

soedirgo:
Thanks for the suggestion! This is likely going to be a config eventually, e.g.:
```toml
[functions.my_function]
no_verify_jwt = true
```

riderx:
For now what i'm doing to handle this is a script:
```
import { existsSync, readdirSync } from 'fs'
import { promisify } from 'util'
import { exec as execCb } from 'child_process'
import { exit } from 'process'
import { outputFile } from 'fs-extra'
import { supa_url } from './utils.mjs'

const exec = promisify(execCb)
const folders = readdirSync('./supabase/functions')
  .filter(file => !file.startsWith('_'))

const projectRef = supa_url.split('.')[0].replace('https://', '')

try {
  console.log('projectRef', projectRef)
  await outputFile('./supabase/.temp/project-ref', projectRef)
  // for in folders
  for (const folder of folders) {
    const fileNoJWT = `./supabase/functions/${folder}/.no_verify_jwt`
    const fileNoDeploy = `./supabase/functions/${folder}/.no_deploy`
    let command = `supabase functions deploy ${folder}`
    let no_verify_jwt = false
    if (existsSync(fileNoJWT)) {
      command += ' --no-verify-jwt'
      no_verify_jwt = true
    }
    if (!existsSync(fileNoDeploy)) {
      console.log(`Upload ${folder}${no_verify_jwt ? ' no_verify_jwt' : ''}`)
      await exec(command).then((r) => {
        if (r.stderr) {
          console.error(folder, r.stderr)
          exit(1)
        }
        return r
      })
      console.log('Done ✅')
    }
    else {
      console.log('Ignored ⏭')
    }
  }
}
catch (e) {
  console.error(e) // should contain code (exit code) and signal (that caused the termination).
  exit(1)
}
```
[...]

PR #786, opened 2023-01-11
feat: support `verify_jwt` functions config

soedirgo:

Feature

Need to set `--no-verify-jwt` flag every time `functions serve` and `functions deploy` commands are run.

## What is the new behavior?

Add per-function config to allow setting `verify_jwt`. `verify_jwt` value falls back to the config is the flag is not set - if it's set, it overrides the config.

## Additional context

Closes https://github.com/supabase/cli/issues/280

### DIFF
--- a/cmd/functions.go
+++ b/cmd/functions.go
@@ -12,7 +12,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/supabase/cli/internal/functions/delete"
 	"github.com/supabase/cli/internal/functions/deploy"
-	"github.com/supabase/cli/internal/functions/new"
+	new_ "github.com/supabase/cli/internal/functions/new"
 	"github.com/supabase/cli/internal/functions/serve"
 	"github.com/supabase/cli/internal/login"
 	"github.com/supabase/cli/internal/utils"
@@ -42,7 +42,7 @@ var (
 		},
 	}
 
-	noVerifyJWT     bool
+	noVerifyJWT     = new(bool)
 	useLegacyBundle bool
 	importMapPath   string
 
@@ -60,7 +60,11 @@ var (
 				return err
 			}
 			ctx, _ := signal.NotifyContext(cmd.Context(), os.Interrupt)
-			return deploy.Run(ctx, args[0], projectRef, !noVerifyJWT, useLegacyBundle, importMapPath, fsys)
+			// Fallback to config if user did not set the flag.
+			if !cmd.Flags().Changed("no-verify-jwt") {
+				noVerifyJWT = nil
+			}
+			return deploy.Run(ctx, args[0], projectRef, noVerifyJWT, useLegacyBundle, importMapPath, fsys)
 		},
 	}
 
@@ -70,7 +74,7 @@ var (
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx, _ := signal.NotifyContext(cmd.Context(), os.Interrupt)
-			return new.Run(ctx, args[0], afero.NewOsFs())
+			return new_.Run(ctx, args[0], afero.NewOsFs())
 		},
 	}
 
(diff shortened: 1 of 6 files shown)


cmd/functions.go
```diff
@@ -42,7 +42,7 @@ var (
 		},
 	}
 
-	noVerifyJWT     bool
+	noVerifyJWT     = new(bool)
```
sweatybridge, line 45: Isn't this unnecessary? because users can never pass in `noVerifyJWT = false` explicitly via cli flag. So if the flag is false, we can safely default to load from config file.

sweatybridge: ok I guess they could do `--no-verify-jwt=false`... but it's kind of a rare case because `--no-verify-jwt false` will throw an error.

```bash
$ supabase functions deploy hello_world --no-verify-jwt false
Error: accepts 1 arg(s), received 2
```

soedirgo: That sounds unintuitive - flags should always override config file if specified. Alternatively we could throw an error if a user does `--no-verify-jwt=false` and ask them to drop the flag.

sweatybridge: Seems like having boolean flags accept true/false args has caused confusion before with cobra users https://github.com/spf13/cobra/issues/1657.

```bash
$ man --ascii=true
man: option '--ascii' doesn't allow an argument
```

I'd prefer rejecting `--no-verify-jwt=false` for clarity coz it's double negation. But may be we can improve handling in a separate PR that deals with all our boolean flags.

sweatybridge (approved):

cmd/functions.go
soedirgo: Yeah, I'd prefer disallowing arguments to bool flags altogether, but alas...

internal/functions/deploy/deploy.go
```diff
@@ -34,6 +34,17 @@ func Run(ctx context.Context, slug string, projectRefArg string, verifyJWT bool,
 		} else if !utils.ProjectRefPattern.MatchString(projectRef) {
 			return errors.New("Invalid project ref format. Must be like `abcdefghijklmnopqrst`.")
 		}
+		// Ensure noVerifyJWT is not nil.
+		if noVerifyJWT == nil {
+			x := false
+			// Load function config if any, but continue on error.
+			if err := utils.LoadConfigFS(fsys); err == nil {
+				if functionConfig, ok := utils.Config.Functions[slug]; ok && !*functionConfig.VerifyJWT {
```
laktek, line 42: Should we show an error message here if the config file cannot be loaded in here (parsing issues, etc) or is it handled elsewhere?

Otherwise users may wonder why their config isn't get picked up.

laktek (commented):
Nice! probably better to do `import-map` flag for consistency (in a separate PR).

internal/functions/deploy/deploy.go
sweatybridge: Yup `LoadConfigFS` will do a syntax check on the config. But we need to write additional logic for handling semantic error like misspelt function name that doesn't exist in supabase/functions for eg.

baptistesx:
Hi there,

is there any way to handle jwt verification per function? I see in supabase/docker/docker-compose.yml this :
      # TODO: Allow configuring VERIFY_JWT per function. This PR might help: https://github.com/supabase/cli/pull/786

https://github.com/supabase/supabase/blob/b65b6c60735fcb3dfdc3bd91ee437e550a7cba93/docker/docker-compose.yml#L330

sweatybridge:
> is there any way to handle jwt verification per function?

Yup, you can customise it via config.toml

```toml
[functions.hello]
verify_jwt = true
```

baptistesx:
Thanks for your answer @sweatybridge 
  But this works only with supabase cli with local and Supabase Cloud, but not with self host (with docker), right?

w0rmh0l3:
> Thanks for your answer @sweatybridge But this works only with supabase cli with local and Supabase Cloud, but not with self host (with docker), right?

This is my first project with Supabase (self-hosted), and I was able to make modifications to allow per function jwt verification. Not sure what's Supabase's way forward strategy for edge function capability in Studio and how that might impact what I have modified. I can raise a PR if someone wants to take a look at it and review if it's worth adding?

sweatybridge:
> I can raise a PR if someone wants to take a look at it and review if it's worth adding?

Yup, that would be a nice add. Feel free to ask @laktek for review when you've opened the PR.

PR #790, opened 2023-01-13
feat(functions): fallback import_map.json to FunctionsDir

soedirgo:

Feature

For each function, `import_map.json` is implicitly used from `supabase/functions/my_func/import_map.json`.

## What is the new behavior?

Implicitly fall back to `supabase/functions/import_map.json` if the per-function import map doesn't exist. The priority becomes (in decreasing order):
- `--import-map` flag
- `import_map` per-function config
- `supabase/functions/<function name>/import_map.json`
- `supabase/functions/import_map.json`

## Additional context

[Context](https://supabase.slack.com/archives/C02KMRX22NR/p1673582728382819?thread_ts=1673233215.735009&cid=C02KMRX22NR)

### DIFF
--- a/internal/functions/deploy/deploy.go
+++ b/internal/functions/deploy/deploy.go
@@ -44,11 +44,20 @@ func Run(ctx context.Context, slug string, projectRefArg string, noVerifyJWT *bo
 			}
 			noVerifyJWT = &x
 		}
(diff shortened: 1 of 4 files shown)


sweatybridge (approved):
would be nice to include some tests

PR #794, opened 2023-01-18
feat: added functions download command

laktek:

This adds support for downloading the existing Edge Functions from the CLI.

Example:

`supabase functions download my-function --project-ref abcd`

No support for downloading existing edge functions. Users have to contact support to get their source code.

## What is the new behavior?

Screen Shot 2023-01-18 at 11 00 59 am

## Additional context

NA


internal/functions/download/download.go
```diff
@@ -0,0 +1,87 @@
...
+	{
+		fmt.Println("Downloading " + utils.Bold(slug))
+		denoPath, err := utils.GetDenoPath()
+		if err != nil {
+			return err
+		}
+
+		resp, err := utils.GetSupabase().GetFunctionBodyWithResponse(ctx, projectRef, slug, func(ctx context.Context, req *http.Request) error {
+			// set encoding header
+			req.Header.Set("Accept-Encoding", "br")
```
sweatybridge, line 54: Possible to put this into the openapi spec for codegen? If not, let's add a todo for now
```suggestion
			// TODO: set encoding header in openapi spec
			req.Header.Set("Accept-Encoding", "br")
```

internal/functions/download/download.go
```diff
@@ -0,0 +1,87 @@
...
+		case http.StatusOK: // Function exists
+			resBuf := bytes.NewReader(resp.Body)
+
+			extractScriptPath := filepath.Join(scriptDirPath, "extract.ts")
+			funcDir := filepath.Join(utils.FunctionsDir, slug)
+			var outBuf, errBuf bytes.Buffer
+			args := []string{"run", "-A", extractScriptPath, funcDir}
+			cmd := exec.CommandContext(ctx, denoPath, args...)
+			cmd.Stdin = resBuf
+			cmd.Stdout = &outBuf
```
sweatybridge, line 73:
```suggestion
			cmd.Stdout = os.Stdout
```

Shall we simply stream the output to stdout? If we encounter an error half way, it would be helpful to know what was the last successful write.

internal/functions/deploy/deploy.go
```diff
@@ -94,6 +94,7 @@ func Run(ctx context.Context, slug string, projectRefArg string, noVerifyJWT *bo
 			}
 		}
 
+		buildScriptPath := filepath.Join(scriptDirPath, "build.ts")
```
sweatybridge, line 97: Seems a little error prone to assume that the caller knows what files are inside `scriptDirPath`. Shall we make `CopyDenoScripts` return a struct with explicit path fields?

```go
type ScriptDir struct {
	ExtractPath string
	BuildPath   string
}
```

laktek: updated in https://github.com/supabase/cli/pull/794/commits/5c95cb9d57de7d137ec88e1b9e07240e9b0a9c20

internal/functions/download/download.go
laktek: done https://github.com/supabase/cli/pull/794/commits/5c95cb9d57de7d137ec88e1b9e07240e9b0a9c20

internal/functions/download/download.go
laktek: this is actually not needed. Golang client doesn't do uncompress for Brotli, so I will remove this header.

internal/utils/denos/extract.ts
```diff
@@ -0,0 +1,44 @@
+import * as path from "https://deno.land/std@0.127.0/path/mod.ts";
+import { readAll } from "https://deno.land/std@0.162.0/streams/conversion.ts";
+import { decompress } from "https://deno.land/x/brotli@0.1.7/mod.ts";
+
+import { Parser } from "https://deno.land/x/eszip@v0.30.0/mod.ts";
+
+function url2path(url: string) {
+  return path.join(...(new URL(url).pathname.replace("/src", "").split("/").filter(Boolean)));
```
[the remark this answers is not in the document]
laktek: done https://github.com/supabase/cli/pull/794/commits/5c95cb9d57de7d137ec88e1b9e07240e9b0a9c20

sweatybridge (approved):

Issue #797, opened 2023-01-19
[Error] supabase gen types typescript produces error "error running container: exit 1"


abarcod:
# Bug report

## Describe the bug

Running supabase gen types typescript --local produces error

## To Reproduce

1. run: supabase init
2. run: supabase start
3. run: supabase gen types typescript --local > Database.ts

## Expected behavior
Creates file Database.ts

## Screenshots
![Unbenannt2](https://user-images.githubusercontent.com/96572110/213404559-70d4d092-aa67-4f23-a3f5-5f7f6415bf8a.PNG)
![Unbenannt3](https://user-images.githubusercontent.com/96572110/213404563-708933c8-5fa2-42c5-b0d7-59b365adcac3.PNG)

## System information

- OS: Debian 11
- CLI Version: 1.34.1
- Version of Node.js: 18.13.0

PR #798, opened 2023-01-19
fix: Correct type generation script path.

kouwasi:

Bug fix.

Get an error. #797 

## What is the new behavior?

No changes.

## Additional context

The path was changed since `supabase/postgres-meta:v0.56.0`

Current version is v0.58.0, So that makes an error.
https://github.com/supabase/cli/blob/main/internal/utils/misc.go#L36


sweatybridge (approved):

sweatybridge:
Thank you for the fix!

Issue #778, opened 2023-01-09
'supabase start' frequently fails with 'service not healthy'


floitsch:
# Bug report

## Describe the bug

Running `supabase start` on Github builders frequently fails with 'service not healthy'.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Create a Github action workflow with `supabase/setup-cli@v1` and `version: latest`.
2. In the workflow start the supabase setup with `supabase start`.
3. Optionally, for good measure, do this with 2 other supabase configurations (using different ports).

## Expected behavior

The builder starts cleanly without errors.

## Github action log

```
Run supabase start
  supabase start
  supabase status
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
Pulling images... (1/13)
Pulling images... (1/13)
Pulling images... (2/13)
Pulling images... (3/13)
Pulling images... (4/13)
Pulling images... (5/13)
Pulling images... (6/13)
Pulling images... (7/13)
Pulling images... (8/13)
Pulling images... (9/13)
Pulling images... (10/13)
Pulling images... (11/13)
Pulling images... (12/13)
Starting database...
Restoring branches...
Setting up initial schema...
Applying migration 20230105212858_initial.sql...
Seeding data supabase/seed.sql...
Starting containers...
Error: service not healthy: [supabase_storage_supabase_test supabase_pg_meta_supabase_test supabase_studio_supabase_test]
Try rerunning the command with --debug to troubleshoot the error.
Error: Process completed with exit code 1.
```

Unfortunately, I wasn't able to get a better log with `--debug`. With the debug flag, the action didn't exhibit the problem.

## System information

Github builder, ubuntu-latest (22.04).
See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

## Additional context

Older versions don't show the error.

This was most likely introduced by the "fix" to https://github.com/supabase/cli/issues/146
https://github.com/supabase/cli/pull/770

sweatybridge:
I can increase the wait time from 10s to 20s and see how it goes. You can also exclude services from starting if they are not needed, for eg.

```bash
supabase start -x storage-api,postgres-meta,studio
```

If you are testing migrations only, just the database needs to be started `supabase db start`

shindradavid:
Am experiencing a similar error, albeit on my local machine

## System info
```
  System:
    OS: Linux 6.1 Manjaro Linux
[...]

PR #803, opened 2023-01-20
fix: bypass non-fatal errors when running supabase start

### LINKED ISSUE #787: Flag to allow `supabase start` to return code 0 if another container is already running
# Feature request

For various reasons, it's helpful to be able to be able to run `supabase start` and have it start the containers if there are no containers running, perform no-op and return success if containers are already running, and fail if it's failing for some other reason (cannot connect to docker daemon).

Either:
- Different return code if it's exited because containers are already running (instead of returning exit code 1)
- Flag to make the CLI return code 0 even if containers are already running

- Running `supabase stop` every time before start. This is not a good solution for the case where we have a `yarn dev` command that spins up all the services, since if there is already a supabase instance, it might have some changes that the developer hasn't written to a permanent migration script yet and we don't want to just kill and delete all the changes.

sweatybridge:

Closes #787
relates to #778

If supabase is already started, throws error.
If any service is unhealthy, throws error.

## What is the new behavior?

- exit 0 instead (consistent with `db start`)
- adds `--ignore-health-check` flag to exit 0 when services are unhealthy

## Additional context

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -11,13 +11,14 @@ import (
 var (
 	allowedContainers  = start.ExcludableContainers()
 	excludedContainers []string
+	ignoreHealthCheck  bool
 
 	startCmd = &cobra.Command{
 		GroupID: groupLocalDev,
 		Use:     "start",
 		Short:   "Start containers for Supabase local development",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return start.Run(cmd.Context(), afero.NewOsFs(), excludedContainers)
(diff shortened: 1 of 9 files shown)


soedirgo:
Can we omit the flag for now until there's a feature request for it?

sweatybridge:
I've linked the [ticket](https://github.com/supabase/cli/issues/778) which "requests" for it.

Adding this flag allows users to bypass the error and run `supabase start`. We still need to figure out why those containers don't start as normal by getting the logs from the those containers.

soedirgo (approved):

PR #821, opened 2023-01-30
fix: allow deno functions to reach host OS

sweatybridge:

feature

no consistent way for locally served functions to reach host OS

## What is the new behavior?

Starting from Docker Engine `20.04`, it's now possible to use `host.docker.internal` consistently across platforms https://stackoverflow.com/a/61424570

## Additional context

https://github.com/supabase/supabase/discussions/9837#discussioncomment-4805370

- [x] tested on codespaces

### DIFF
--- a/internal/functions/serve/serve.go
+++ b/internal/functions/serve/serve.go
@@ -120,6 +120,8 @@ func Run(ctx context.Context, slug string, envFilePath string, noVerifyJWT *bool
 			},
 			container.HostConfig{
 				Binds: binds,
+				// Allows containerized functions on Linux to reach host OS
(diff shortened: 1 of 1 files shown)


soedirgo:
Does this also work with other daemons? I don't think we've used Docker Desktop-isms so far.

sweatybridge:
Haven't tested on other daemons but it should be safe to add because this config only updates the [`/etc/hosts` file](https://github.com/docker/cli/blob/850fb6921cf7a208fd47d83074f27d0eb7ae268e/opts/hosts.go#L27). The extra entry wouldn't break existing functions and we can look into supporting other daemons in the future.

soedirgo (approved):

PR #826, opened 2023-01-31
fix: close migration file immediately after executing

### LINKED ISSUE #822: `Too many open files`
# Bug report

## Describe the bug

With the 245th migration file, we are not able to start the local database anymore. Postgres complaints about `Too many open files`. Since this essentially halts development for us, it would be great to prioritise this accordingly. AFAIK there is a pg setting `max_files_per_process` which can be increased, but I do not know if there are any side effects.

## To Reproduce

1. add 245 sql files in `migrations/`
2. run `supababase start`

## Expected behavior

The number of migration files should be unlimited.

sweatybridge:

Bug fix #822 

Deferred statements are run before function returns, not when out of scope.

## What is the new behavior?

Close migration files immediately to release resources back to the host OS.

## Additional context

Tested locally with

```bash
ulimit -n 200
for i in {1..250}; do supabase migration new "$i"; done
supabase db start
```

Checked all other usage of `defer` keyword and found an issue where postgREST may be restarted before database is fully up. Fixed it in this PR as well.

### DIFF
--- a/internal/db/diff/migra.go
+++ b/internal/db/diff/migra.go
@@ -201,17 +201,21 @@ func MigrateDatabase(ctx context.Context, conn *pgx.Conn, fsys afero.Fs) error {
 	}
 	// Apply migrations
 	for _, filename := range migrations {
-		fmt.Fprintln(os.Stderr, "Applying migration "+utils.Bold(filename)+"...")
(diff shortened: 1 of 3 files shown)


soedirgo (approved):

psteinroe:
that was fast, thanks!! ❤️

PR #828, opened 2023-01-31
feat: Allow serving all edge functions using edge runtime

laktek:

This PR introduces **experimental** `functions serve --all` option, which would use edge-runtime to serve all edge functions in a project locally.

Please note this is still experimental and missing several key features:
- `verify-jwt` option doesn't work
- `import-maps` doesn't work

TODO:

[x] Update the edge-runtime image tag

### DIFF
--- a/internal/functions/serve/serve.go
+++ b/internal/functions/serve/serve.go
@@ -285,12 +285,8 @@ func runServeAll(ctx context.Context, envFilePath string, noVerifyJWT *bool, imp
 			binds = append(binds, filepath.Join(cwd, importMapPath)+":"+customDockerImportMapPath+":ro,z")
 		}
 
-		// bind deno cache directory
-		cachePath, err := utils.GetDenoCachePath()
-		if err != nil {
-			return err
-		}
-		binds = append(binds, cachePath+":/root/.cache/deno:rw,z")
+		// bind deno cache volume
+		binds = append(binds, utils.DenoRelayId+":/root/.cache/deno:rw,z")
 
 		containerID, err := utils.DockerStart(
 			ctx,
--- a/internal/functions/serve/serve.go
+++ b/internal/functions/serve/serve.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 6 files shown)


internal/functions/serve/serve.go
```diff
@@ -216,3 +221,134 @@ func Run(ctx context.Context, slug string, envFilePath string, noVerifyJWT *bool
...
+		if importMapPath != "" {
+			binds = append(binds, filepath.Join(cwd, importMapPath)+":"+customDockerImportMapPath+":ro,z")
+		}
+
+		// bind deno cache directory
+		cachePath, err := utils.GetDenoCachePath()
+		if err != nil {
+			return err
+		}
+		binds = append(binds, cachePath+":/root/.cache/deno:rw,z")
```
sweatybridge, line 293: Does deno cache work consistently across host and container OS? ie. if my host is windows, and the cache is mounted into a linux container, would it work?

laktek: We provide a container-specific cache (defaults to `~/.supabase/deno_cache`). This is intended to be used only by the container and acts independent to user's local Deno installation.

sweatybridge: ic, it sounds like this cache folder doesn't need to be exposed to the host OS then. Wdyt about using docker named volumes to manage it? We started using it for backing up the `PGDATA` directory between cli restarts. Example here https://github.com/supabase/cli/blob/main/internal/db/start/start.go#L77 and https://github.com/supabase/cli/blob/main/internal/stop/stop.go#L52

laktek: yep, sounds like that's a good option

laktek: Updated in [c5b3291](https://github.com/supabase/cli/pull/828/commits/c5b32917b0fdc35041161119d1fbd78f79b76f1e) `functions serve` forcefully removes the container and its volumes. So didn't add to remove volumes in `supabase stop`.

internal/utils/misc.go
```diff
@@ -28,15 +28,16 @@ const (
...
+	KongImage        = "library/kong:2.8.1"
+	InbucketImage    = "inbucket/inbucket:3.0.3"
+	PostgrestImage   = "postgrest/postgrest:v10.1.1.20221215"
+	DifferImage      = "supabase/pgadmin-schema-diff:cli-0.0.5"
+	MigraImage       = "djrobstep/migra:3.0.1621480950"
+	PgmetaImage      = "supabase/postgres-meta:v0.60.3"
+	StudioImage      = "supabase/studio:20230127-6bfd87b"
+	DenoRelayImage   = "supabase/deno-relay:v1.5.0"
+	ImageProxyImage  = "darthsim/imgproxy:v3.8.0"
+	EdgeRuntimeImage = "supabase/edge-runtime:latest"
```
sweatybridge, line 40:
```suggestion
	EdgeRuntimeImage = "supabase/edge-runtime:v1.0.6"
```

laktek: Updated to 1.0.7

sweatybridge (approved):

PR #831, opened 2023-02-01
chore: support binds in run with stream

sweatybridge:
minor refactor


laktek (approved):

internal/db/diff/diff.go
```diff
@@ -82,6 +82,7 @@ func DiffSchema(ctx context.Context, source, target string, schema []string, p u
 			utils.DifferImage,
 			nil,
 			args,
+			nil,
```
laktek, line 85: these changes are intentional?

sweatybridge: yup, due to introducing a new `binds` arg to the function signature.

PR #834, opened 2023-02-01
fix: match db locale of Supabase cloud in local development

### LINKED ISSUE #825: Local environment uses en_US.UTF-8 locale, cloud uses C.UTF-8
# Bug report

## Describe the bug

The local environment created by `supabase start` creates a database with a locale of `en_US.UTF-8` whereas the Supabase cloud database is created with a locale of `C.UTF-8`.

## To Reproduce

1. Run `supabase start`
2. Dump local database.
3. Create a Supabase cloud project.
4. Dump cloud database.
5. Compare the difference.

## Expected behavior

I would expect them to be the same because a different locale means that Postgres sorting, string functions, pattern matching and other aspects of the database can differ in varying degrees.

References: https://www.postgresql.org/docs/current/locale.html#id-1.6.11.3.5

## Screenshots

None.

## System information

- OS: Linux
- Browser: none
- Version of supabase-cli: 1.35.0
- Version of Node.js: 16.18.0

## Additional context

I am betting this is caused by the locale of my desktop Linux system being set to `en_US.UTF-8`. However, I don't think that changing my system's locale should be the solution.

waynebloss:

Fixes #825 

The locale of the Postgres database is inherited from the host system.

## What is the new behavior?

The locale of the Postgres database matches that of Supabase cloud.

## Additional context

None.

### DIFF
--- a/internal/db/start/start.go
+++ b/internal/db/start/start.go
@@ -45,6 +45,7 @@ func StartDatabase(ctx context.Context, fsys afero.Fs, w io.Writer, options ...f
 		Env: []string{
 			"POSTGRES_PASSWORD=postgres",
 			"POSTGRES_HOST=/var/run/postgresql",
+			"LC_ALL=C.UTF-8",
 		},
 		Healthcheck: &container.HealthConfig{
 			Test:     []string{"CMD", "pg_isready", "-U", "postgres", "-h", "localhost", "-p", "5432"},


sweatybridge (approved):
Thanks for contributing this fix! The other services should just follow the locale used by the database so I think this is fine for now.

In the future, we can look into a way to read the locale from hosted project.

Issue #827, opened 2023-01-31
Updating to 1.35.2+ breaks local supabase start for ARM-based Mac


zhawtof:
# Bug report

## Describe the bug

After v1.35.2 was introduced, my Mac M1 struggles to start the `supabase_pg_meta_*` image.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. On a Mac M1 (or possibly any ARM-based Mac), set up a new Supabase CLI with version 1.35.2 (also happens with latest 1.35.3)
2. Run `supabase start`
3. Notice the following error logs in the `supabase_pg_meta_*` image:
```
(node:19) ExperimentalWarning: Importing JSON modules is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
node:internal/modules/cjs/loader:1249
  return process.dlopen(module, path.toNamespacedPath(filename));
                 ^

Error: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /usr/src/app/node_modules/libpg-query/build/Release/queryparser.node)
    at Object.Module._extensions..node (node:internal/modules/cjs/loader:1249:18)
    at Module.load (node:internal/modules/cjs/loader:1043:32)
    at Function.Module._load (node:internal/modules/cjs/loader:878:12)
    at Module.require (node:internal/modules/cjs/loader:1067:19)
    at require (node:internal/modules/cjs/helpers:103:18)
    at Object.<anonymous> (/usr/src/app/node_modules/libpg-query/index.js:1:17)
    at Module._compile (node:internal/modules/cjs/loader:1165:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1219:10)
    at Module.load (node:internal/modules/cjs/loader:1043:32)
    at Function.Module._load (node:internal/modules/cjs/loader:878:12) {
  code: 'ERR_DLOPEN_FAILED'
}
```

## Expected behavior

I expected to be able to start Supabase locally. After downgrading to v1.35.1, I am able to get the expected behavior which leads me to believe there is a regression in 1.35.2 when pgmeta was [updated in this PR](https://github.com/supabase/cli/commit/eee4c50ac5fa390069e0e23c15cb343314417019)

## System information

- OS: macOS Ventura 13.1 M1
- Browser (if applies) N/A
- Version of supabase-js: v2.4.1
- Version of Node.js: v19.5.0

sweatybridge:
[...]

PR #830, opened 2023-01-31
update pgmeta to latest version

kartikk-k:
## What kind of change does this PR introduce?

pgmeta update

## What is the new behavior?
[supabase/postgress-meta#483](https://github.com/supabase/postgres-meta/pull/483#issuecomment-1409849142)

## Additional context

Add any other context or screenshots.

PR #838, opened 2023-02-02
Update postgres-meta to 0.60.6

TomasHubelbauer:
Hi! This is a follow-up to this comment:
https://github.com/supabase/postgres-meta/pull/483#issuecomment-1409562603

I am proposing an update of postgres-meta to version 0.60.6. I am hoping to make use of https://github.com/supabase/postgres-meta/pull/492 through the Supabase CLI.

As per the above linked comment, I am aware that in general, the postgres-meta version is kept at the same version as what is used on the hosted platform. This is a minor changes from the current version 0.60.3 so perhaps it will quality for an exception? I am not sure how I can check what version is live on the hosted platform.

Thanks for considering!

### DIFF
--- a/internal/utils/misc.go
+++ b/internal/utils/misc.go
@@ -33,7 +33,7 @@ const (
 	PostgrestImage   = "postgrest/postgrest:v10.1.1.20221215"
 	DifferImage      = "supabase/pgadmin-schema-diff:cli-0.0.5"
 	MigraImage       = "djrobstep/migra:3.0.1621480950"
-	PgmetaImage      = "supabase/postgres-meta:v0.60.3"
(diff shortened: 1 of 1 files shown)


TomasHubelbauer:
I noticed https://github.com/supabase/cli/pull/824 but it seems it got reverted as we're on 0.60.3 not 0.60.5 now.
Reverted in https://github.com/supabase/cli/pull/829 to fix https://github.com/supabase/cli/issues/827.
I am not sure if 0.60.6 suffers from the same issue as 0.60.5.
It might be necessary to hold the update.

TomasHubelbauer:
Oh this is a dupe of https://github.com/supabase/cli/pull/830

PR #846, opened 2023-02-06
feat: adds support for enabling SSL enforcement on DB conns

darora:
Experimental commands for enabling/disabling SSL enforcement for
connections to the DB (PG+pooler).

### DIFF
--- a/cmd/sslEnforcement.go
+++ b/cmd/sslEnforcement.go
@@ -1,0 +1,63 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"os/signal"
+
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+	"github.com/supabase/cli/internal/ssl_enforcement/get"
+	"github.com/supabase/cli/internal/ssl_enforcement/update"
+)
+
+var (
+	sslEnforcementCmd = &cobra.Command{
+		GroupID: groupManagementAPI,
+		Use:     "ssl-enforcement",
+		Short:   "Manage SSL enforcement configuration",
+	}
+
+	dbEnforceSsl bool
+	dbDisableSsl bool
+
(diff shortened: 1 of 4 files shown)


internal/ssl_enforcement/get/get.go
```diff
@@ -0,0 +1,47 @@
...
+			return fmt.Errorf("failed to retrieve SSL enforcement config: %w", err)
+		}
+		if resp.JSON200 == nil {
+			return errors.New("failed to retrieve SSL enforcement config; received: " + string(resp.Body))
+		}
+
+		if err != nil {
+			return err
+		}
+		if resp.JSON200.CurrentConfig.Database && resp.JSON200.AppliedSuccessfully {
```
sweatybridge, line 40: Why would get request need to check if `AppliedSuccessfully`?

sweatybridge (approved):
Lgtm in general

cmd/sslEnforcement.go
```diff
@@ -0,0 +1,63 @@
...
+				return err
+			}
+			ctx, _ := signal.NotifyContext(cmd.Context(), os.Interrupt)
+			return get.Run(ctx, projectRef, fsys)
+		},
+	}
+)
+
+func init() {
+	sslEnforcementCmd.PersistentFlags().StringVar(&projectRef, "project-ref", "", "Project ref of the Supabase project.")
```
soedirgo, line 55: Nit: use `--project-id` instead - we're deprecating "project ref" terminology across API + CLI

internal/ssl_enforcement/get/get.go
darora: The request is stored first (so that it can be retried internally if needed), and an attempt is made to apply it afterwards. If the latter fails (for w/e reason), this indicates to the user that the request might've partially succeeded.

PR #879, opened 2023-02-21
feat: add `sso` commands

hf:
Commands to manage Single Sign-On authentication on projects.


cmd/gen.go
```diff
@@ -22,23 +22,23 @@ var (
...
 	}
 
-	local     bool
-	linked    bool
-	projectId string
-	dbUrl     string
-	schemas   []string
+	local        bool
+	linked       bool
+	genProjectId string
```
J0, line 27: Are these changes related to SSO?

cmd/sso.go
```diff
@@ -0,0 +1,128 @@
...
+	ssoAttributeMappingFile string
+	ssoDomains              []string
+	ssoAddDomains           []string
+	ssoRemoveDomains        []string
+
+	ssoAddCmd = &cobra.Command{
+		Use:     "add <type> [flags]",
+		Short:   "Add a new SSO identity provider",
+		Args:    cobra.ExactArgs(1),
+		Example: `  supabase sso add saml --project-ref abcdefghijklmn --metadata-file ~/SAMLMetadata.xml`,
```
J0, line 26: Would <type> here refer to the SSO method? (e.g. SAML, OIDC, password SSO etc) ?

cmd/sso.go
```diff
@@ -0,0 +1,128 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return nil
+		},
+	}
+
+	ssoGetCmd = &cobra.Command{
+		Use:     "get <provider-id> [flags]",
+		Short:   "Get information about an SSO identity provider",
+		Args:    cobra.ExactArgs(1),
+		Example: `  supabase sso get b5ae62f9-ef1d-4f11-a02b-731c8bbb11e8 --project-ref abcdefghijklmn`,
```
J0, line 65: What information would we be returning here? Perhaps can consider allowing a filter by specific attributes in the future since metadata files (and other xml files like attribute mapping) tend to be long and might flood the screen

cmd/sso.go
hf: Sorry what's "here" 🙃

cmd/sso.go
hf: All of it, probably we can also do something to fetch the metadata only, let's see.

cmd/sso.go
J0: Oops that should be `type`, but looks like it's been added with a default of saml.

internal/sso/update/update.go
```diff
@@ -0,0 +1,45 @@
...
+	"context"
+	"errors"
+	"os"
+
+	"github.com/supabase/cli/internal/sso/internal/render"
+	"github.com/supabase/cli/internal/utils"
+	"github.com/supabase/cli/pkg/api"
+)
+
+func Run(ctx context.Context, ref, providerId string, params api.UpdateProviderByIdJSONRequestBody, addDomains, removeDomains []string, format string) error {
```
sweatybridge, line 13: Let's validate the project ref here as a sanity check. for eg. https://github.com/supabase/cli/blob/main/internal/ssl_enforcement/get/get.go#L13

internal/sso/internal/render/render.go
sweatybridge, line 1: Shall we name this as `internal/sso/render/render.go`? I think it's fine to have some internal packages not exposed as cmds.

cmd/status.go
```diff
@@ -12,19 +12,9 @@ import (
...
-			status.OutputPretty,
-			status.OutputToml,
-			status.OutputYaml,
-		},
-		Value: status.OutputPretty,
-	}
-
+	override  []string
+	names     status.CustomName
+	output    = utils.OutputFlag(true /* allow env output */)
```
sweatybridge, line 17:
```suggestion
	override []string
	names    status.CustomName
	output   = utils.EnumFlag{
		Allowed: append([]string{status.OutputEnv, utils.OutputDefaultAllowed}),
		Value: status.OutputPretty,
	}
```

Can we do something like this ^? I think it reads more explicit than wrapping a function.

internal/utils/output.go
```diff
@@ -0,0 +1,75 @@
...
+		return yaml.NewEncoder(w).Encode(value)
+
+	case OutputToml:
+		return toml.NewEncoder(w).Encode(value)
+
+	default:
+		return fmt.Errorf("Unsupported output encoding %q", format)
+	}
+
+	return nil
```
sweatybridge, line 74:
```suggestion
```

unnecessary return

cmd/sso.go
```diff
@@ -0,0 +1,259 @@
...
+		PreRun: func(cmd *cobra.Command, args []string) {
+			cobra.CheckErr(cmd.MarkFlagRequired("project-ref"))
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			var params api.UpdateProviderByIdJSONRequestBody
+
+			fsys := afero.NewOsFs()
+
+			if ssoMetadataFile != "" {
+				data, err := readMetadataFile(fsys, ssoMetadataFile)
```
sweatybridge, line 117: Can we move file reading into internals? I think it's fine to build the full `api.UpdateProviderByIdJSONRequestBody` object inside internals too.

cmd/sso.go
```diff
@@ -0,0 +1,259 @@
...
+				if err != nil {
+					return err
+				}
+
+				params.AttributeMapping = mapping
+			}
+
+			params.Domains = &ssoDomains
+
+			return update.Run(cmd.Context(), ssoProjectRef, args[0], params, ssoAddDomains, ssoRemoveDomains, ssoOutput.Value)
```
sweatybridge, line 139: seems like `ssoAddDomains` and `ssoRemoveDomains` are unused by `update.Run`.

internal/sso/internal/render/render.go
hf: Fine with me, just doesn't make sense to be usable outside sso.

internal/utils/output.go
hf: AH THIS IS THE UNREACHABLE CODE! I kept looking for it since golangci doesn't tell you where it is!

cmd/sso.go
hf: Ah forgot about that, there should be a todo on it tho. Will use them.

cmd/sso.go
```diff
@@ -0,0 +1,259 @@
...
+				if err != nil {
+					return err
+				}
+
+				params.MetadataXml = &data
+			} else if ssoMetadataURL != "" {
+				// TODO: fetch and validate Metadata
+				params.MetadataUrl = &ssoMetadataURL
+			} else {
+				return errors.New("--metadata-file or --metadata-xml must be provided")
```
kangmingtay, line 72:
```suggestion
				return errors.New("--metadata-file or --metadata-url must be provided")
```

cmd/sso.go
```diff
@@ -0,0 +1,259 @@
...
+
+			if ssoMetadataFile != "" {
+				data, err := readMetadataFile(fsys, ssoMetadataFile)
+				if err != nil {
+					return err
+				}
+
+				params.MetadataXml = &data
+			} else if ssoMetadataURL != "" {
+				// TODO: fetch and validate Metadata
```
kangmingtay, line 69: hmm do we need to validate this or can we just leave it to gotrue?

cmd/sso.go
kangmingtay: i dont think this defaults to saml - if it's excluded, i get the following error: `Error: accepts 1 arg(s), received 0`

kangmingtay (changes requested):

hf:
@kangmingtay can't reply to that comment not sure why, so here it is:
> i dont think this defaults to saml - if it's excluded, i get the following error: Error: accepts 1 arg(s), received 0

This is intentional, the user has to enter `saml` or in the future `oidc` to explicitly specify the type.

cmd/sso.go
hf: Nothing too complex, but fetching the URL, checking it's returning 200 and XML should be enough. Also that it's HTTPS.

.gitignore
```diff
@@ -4,6 +4,7 @@
 *.dll
 *.so
 *.dylib
+/cli
```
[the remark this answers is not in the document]
hf: Actually this should be `/cli` as that means ignore the `cli` entry at the root of the git repo.

.gitignore
```diff
@@ -20,3 +21,4 @@
 # NPM
 node_modules
 package-lock.json
+
```
sweatybridge, line 24:
```suggestion
```

cmd/functions.go
```diff
@@ -156,7 +158,7 @@ Enter your project ref: `, utils.GetSupabaseDashboardURL())
 
 		scanner := bufio.NewScanner(os.Stdin)
 		if !scanner.Scan() {
-			return errors.New("Cancelled " + utils.Aqua("supabase functions deploy") + ".")
+			return errors.New("Cancelled " + utils.Aqua(cmd.UseLine()) + ".")
```
sweatybridge, line 161: We will probably refactor this separately as the Scan() may terminate due to either error and cancellation.

cmd/sso.go
```diff
@@ -0,0 +1,202 @@
...
+		},
+	}
+
+	ssoRemoveCmd = &cobra.Command{
+		Use:     "remove <provider-id> [flags]",
+		Short:   "Remove an existing SSO identity provider",
+		Args:    cobra.ExactArgs(1),
+		Example: `  supabase sso remove b5ae62f9-ef1d-4f11-a02b-731c8bbb11e8 --project-ref mwjylndxudmiehsxhmmz`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return remove.Run(cmd.Context(), projectRef, args[0], ssoOutput.Value)
```
sweatybridge, line 88: Is it possible to validate the providerId with regex?, ie. `args[0]`

cmd/sso.go
```diff
@@ -0,0 +1,202 @@
...
+	ssoDomains              []string
+	ssoAddDomains           []string
+	ssoRemoveDomains        []string
+	ssoOutput               = utils.EnumFlag{
+		Allowed: utils.OutputDefaultAllowed,
+		Value:   utils.OutputPretty,
+	}
+
+	ssoAddCmd = &cobra.Command{
+		Use:     "add <type = saml> [flags]",
```
sweatybridge, line 60: Might be a bit late to ask but can we make this an enum flag? ie. `sso add --type saml`

```go
	ssoType = utils.EnumFlag{
		Allowed: []string{"saml"},
		Value:   "saml",
	}
```

This would give us automatic validation and default value support.

Also, the `[flags]` part of UseLine will be added automatically by cobra

```suggestion
		Use:     "add",
```

cmd/sso.go
```diff
@@ -0,0 +1,202 @@
...
+				MetadataURL:       ssoMetadataURL,
+				SkipURLValidation: ssoSkipURLValidation,
+				AttributeMapping:  ssoAttributeMappingFile,
+				Domains:           ssoDomains,
+			})
+		},
+	}
+
+	ssoRemoveCmd = &cobra.Command{
+		Use:     "remove <provider-id> [flags]",
```
sweatybridge, line 83:
```suggestion
		Use:     "remove <provider-id>",
```

ditto

cmd/sso.go
```diff
@@ -0,0 +1,202 @@
...
+		Short:   "Remove an existing SSO identity provider",
+		Args:    cobra.ExactArgs(1),
+		Example: `  supabase sso remove b5ae62f9-ef1d-4f11-a02b-731c8bbb11e8 --project-ref mwjylndxudmiehsxhmmz`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return remove.Run(cmd.Context(), projectRef, args[0], ssoOutput.Value)
+		},
+	}
+
+	ssoUpdateCmd = &cobra.Command{
+		Use:     "update <provider-id> [flags]",
```
sweatybridge, line 93:
```suggestion
		Use:     "update <provider-id>",
```

cmd/sso.go
```diff
@@ -0,0 +1,202 @@
...
+				AttributeMapping:  ssoAttributeMappingFile,
+				Domains:           ssoDomains,
+				AddDomains:        ssoAddDomains,
+				RemoveDomains:     ssoRemoveDomains,
+			})
+		},
+	}
+
+	ssoShowCmd = &cobra.Command{
+		Use:     "show <provider-id> [flags]",
```
sweatybridge, line 115:
```suggestion
		Use:     "show <provider-id>",
```

internal/sso/get/get.go
```diff
@@ -0,0 +1,46 @@
...
+	if resp.JSON200 == nil {
+		if resp.StatusCode() == http.StatusNotFound {
+			return fmt.Errorf("An identity provider with ID %q could not be found.", providerId)
+		}
+
+		return errors.New("Unexpected error fetching identity provider: " + string(resp.Body))
+	}
+
+	switch format {
+	case "metadata":
```
sweatybridge, line 30: Let's make a new constant for this special case in this file, ie. `const OutputMetadata = "metadata"`

sweatybridge (commented):
Just a couple of comments.

cmd/sso.go
kangmingtay: yeah should be possible since it follows the uuid_v4 format

kangmingtay (approved):

cmd/sso.go
hf: Sure no problem. I was a bit skeptical of the positional argument myself.

sweatybridge (approved):

PR #892, opened 2023-02-28
fix: exclude temporary schemas from db diff

sweatybridge:

Bug fix https://gist.github.com/tmcw/6f43e433bae38fffacec6d2aa5c9345a

## What is the new behavior?

- removes all temporary schemas from being diffed by default
- uses `pg_*` wildcard pattern to support both diff and dump

## Additional context


soedirgo (approved):
We can also just exclude `pg_%` since the `pg_` prefix is reserved:
```
postgres=# create schema pg_foo;
ERROR:  unacceptable schema name "pg_foo"
DETAIL:  The prefix "pg_" is reserved for system schemas.
```

PR #938, opened 2023-03-17
feat: removes harded coded value for jwt secret, service role key, and anon key

ryankazokas:
Pulls values from a .env file similar to how docker works. Otherwise defaults to values that are hardcoded as to not introduce any breaking changes.

I used a .env file took example from the docker configuration however if you think it suites it better to use the config.toml i can update to reflect that.


internal/utils/misc.go
```diff
@@ -82,10 +97,10 @@ SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname = '%[1]s';
 DO 'BEGIN WHILE (
 	SELECT COUNT(*) FROM pg_replication_slots WHERE database = ''%[1]s''
 ) > 0 LOOP END LOOP; END';`
-	AnonKey        = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZS1kZW1vIiwicm9sZSI6ImFub24iLCJleHAiOjE5ODM4MTI5OTZ9.CRXP1A7WOeoJeXxjNni43kdQwgnWNReilDMblYTn_I0"
-	ServiceRoleKey = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZS1kZW1vIiwicm9sZSI6InNlcnZpY2Vfcm9sZSIsImV4cCI6MTk4MzgxMjk5Nn0.EGIM96RAZx35lJzdJsyH-qQwv8Hdp7fsn3W0YpN81IU"
-	JWTSecret      = "super-secret-jwt-token-with-at-least-32-characters-long"
+	AnonKey        = getEnvValueOrDefault("ANON_KEY", "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZS1kZW1vIiwicm9sZSI6ImFub24iLCJleHAiOjE5ODM4MTI5OTZ9.CRXP1A7WOeoJeXxjNni43kdQwgnWNReilDMblYTn_I0")
```
sweatybridge, line 100: Let's prefix these with `SUPABASE_AUTH_` since we might be adding other secrets to .env file in the future. 

It would also be ideal to account for `--workdir` flag when searching for .env file. You can do that by loading these keys after os.Chdir in [root cmd](https://github.com/supabase/cli/blob/main/cmd/root.go).

sweatybridge (changes requested):
Thank you for the contribution. Just a few minor suggestions. Let me know if you have time to make them

internal/utils/misc.go
ryankazokas: Thanks for the feedback. I will have time later today or tomorrow to make these changes. I'll comment back once finished.

ryankazokas: Alright, i updated this a little more than i had expected. 

1. I changed to use viper, since it was already imported into the project and being used for CLI args.
2. Had to change the env var setting to it's own function since utils is used before the call to os.ChangeDir 
3. The LoadEnvVars which will be done after working directory is determined i am just passing the workdirectory plus .env. I think in future if it warrants it we can probably just take in the .env from a CLI arg as well. I think having this default is probably a good start though.

sweatybridge: Thanks for updating your PR. I've refactored these secrets to be nested under the Config object and loaded via viper. This allows both env var and .env file to work.

sweatybridge (approved):

PR #961, opened 2023-03-26
chore: run builtin migrations on database start

sweatybridge:

chore

## What is the new behavior?

## Additional context


soedirgo:
This only applies for PG15 though right? Do we need to do something special for PG13 & PG14?

sweatybridge:
> This only applies for PG15 though right? Do we need to do something special for PG13 & PG14?

Yup, I was thinking to publish a new pg14 image which also benefits from smaller image size. But pg13 will probably need a flag.

soedirgo (approved):

PR #964, opened 2023-03-27
feat: add sso info cmd

kangmingtay:
* Adds the info cmd which returns the service provider (gotrue) SAML settings to be added to the identity provider 

## Example

* `supabase sso info --project-ref <project_ref>`

```
            PROPERTY             │                               VALUE
  ───────────────────────────────┼─────────────────────────────────────────────────────────────────────
    Single sign-on URL (ACS URL) │ https://projectref.supabase.co/auth/v1/sso/saml/acs
    Audience URI (SP Entity ID)  │ https://projectref.supabase.co/auth/v1/sso/saml/metadata
    Default Relaystate           │ https://projectref.supabase.co

```

### DIFF
--- a/cmd/functions.go
+++ b/cmd/functions.go
@@ -72,7 +72,7 @@ var (
 			if err := PromptLogin(fsys); err != nil {
 				return err
 			}
-			if err := PromptProjectRef(fsys); err != nil {
+			if err := PromptProjectRef(fsys, cmd); err != nil {
 				return err
 			}
 			ctx, _ := signal.NotifyContext(cmd.Context(), os.Interrupt)
@@ -145,7 +145,9 @@ func PromptLogin(fsys afero.Fs) error {
 	}
 }
 
-func PromptProjectRef(fsys afero.Fs) error {
+// PromptProjectRef prompts for the project ref if there is no project ref provided
+// or if there is no linked project
+func PromptProjectRef(fsys afero.Fs, cmd *cobra.Command) error {
 	if len(projectRef) > 0 {
 		return nil
 	} else if err := utils.AssertIsLinkedFS(fsys); err == nil {
@@ -156,7 +158,7 @@ Enter your project ref: `, utils.GetSupabaseDashboardURL())
 
 		scanner := bufio.NewScanner(os.Stdin)
 		if !scanner.Scan() {
-			return errors.New("Cancelled " + utils.Aqua("supabase functions deploy") + ".")
+			return errors.New("Cancelled " + utils.Aqua(cmd.UseLine()) + ".")
 		}
 
 		projectRef = strings.TrimSpace(scanner.Text())
--- a/cmd/sso.go
+++ b/cmd/sso.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
+	"os/signal"
 	"strings"
 	"unicode/utf8"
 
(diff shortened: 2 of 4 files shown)


internal/sso/info/info.go
```diff
@@ -0,0 +1,24 @@
...
+	switch format {
+	case utils.OutputPretty:
+		return render.InfoMarkdown(ref)
+
+	default:
+		return utils.EncodeOutput(format, os.Stdout, map[string]interface{}{
+			"Single sign-on URL (ACS URL)": fmt.Sprintf("https://%s.supabase.co/auth/v1/sso/saml/acs", ref),
+			"Audience URI (SP Entity ID)":  fmt.Sprintf("https://%s.supabase.co/auth/v1/sso/saml/metadata", ref),
+			"Default Relaystate":           fmt.Sprintf("https://%s.supabase.co", ref),
+		})
```
hf, line 22: Other than the "pretty" output, the others are actually technical encodings like JSON / YAML / TOML. So these can be technical like `acs_url`, `audience_uri`, etc.

hf (approved):

kangmingtay:
Thanks for the suggestions @sweatybridge - right now, it seems like we have a bunch of global flags / utility functions declared in the `cmd` package across several different files. I'm thinking of moving all the global flags / utility functions like `projectRef` and `PromptProjectRef()`  to a separate file in a FLUP PR for better code organisation - lmk if this is something you'd like or if you have some ideas of how to improve this.

cmd/sso.go
```diff
@@ -25,9 +28,22 @@ var (
 		GroupID: groupManagementAPI,
 		Use:     "sso",
 		Short:   "Manage Single Sign-On (SSO) authentication for projects",
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
```
kangmingtay, line 31: using `PersistentPreRunE` so that the validity checks are being enforced across all subcommands

cmd/functions.go
```diff
@@ -156,7 +158,7 @@ Enter your project ref: `, utils.GetSupabaseDashboardURL())
 
 		scanner := bufio.NewScanner(os.Stdin)
 		if !scanner.Scan() {
-			return errors.New("Cancelled " + utils.Aqua("supabase functions deploy") + ".")
+			return errors.New("Cancelled " + utils.Aqua(cmd.UseLine()) + ".")
```
kangmingtay, line 161: this will print out the `Usage` information of the command - for example, cancelling the `supabase sso info ...` will print `Cancelled supabase sso info [flags]`

cmd/sso.go
sweatybridge: One caveat about `PersistentPreRunE` is that it doesn't inherit from parent command properly. So you also need sth [like this](https://github.com/supabase/cli/blob/main/cmd/db.go#L134) 
```suggestion
		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
			if err := cmd.Root().PersistentPreRunE(cmd, args); err != nil {
				return err
			}
```

cmd/sso.go
```diff
@@ -25,9 +28,22 @@ var (
...
+			signal.NotifyContext(cmd.Context(), os.Interrupt)
```
sweatybridge, line 32: also need to set context so that subsequent calls to `cmd.Context()` use the wrapped context

```suggestion
			ctx, _ := signal.NotifyContext(cmd.Context(), os.Interrupt)
			cmd.SetContext(ctx)
```

sweatybridge:
> I'm thinking of moving all the global flags / utility functions like projectRef and PromptProjectRef() to a separate file in a FLUP PR for better code organisation - lmk if this is something you'd like or if you have some ideas of how to improve this.

Yup, this is something I've been wanting to do as well. I would love to have a separate package for flag parsing that can be unit tested and reused by all commands.

J0 (approved):

PR #965, opened 2023-03-29
fix: Provide a better error message for non-existing function

laktek:

Show a more descriptive and easy-to-understand error when a user tries to deploy a function that doesn't exist.

## Current behavior

image

## New behavior

Screen Shot 2023-03-29 at 11 22 20 am

## Additional details

Ideally, this should be fixed in the Go code that handles the deploy command. However, we currently do some deduction of paths when a directory doesn't exist (eg: traversing down `supabase/functions`). We can refactor that logic and account for true non-existing functions in a future iteration.

### DIFF
--- a/internal/utils/denos/build.ts
+++ b/internal/utils/denos/build.ts
@@ -9,6 +9,13 @@ const virtualBasePath = "file:///src/";
 
 async function buildAndWrite(p: string, importMapPath: string) {
   const funcDirPath = path.dirname(p);
+  try {
+    await Deno.lstat(funcDirPath);
+  } catch (e) {
+    console.error(`Error: Cannot access "${funcDirPath}". Check if directory exists or has read permissions.`);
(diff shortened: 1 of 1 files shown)


internal/utils/denos/build.ts
```diff
@@ -9,6 +9,13 @@ const virtualBasePath = "file:///src/";
 
 async function buildAndWrite(p: string, importMapPath: string) {
   const funcDirPath = path.dirname(p);
+  try {
+    await Deno.lstat(funcDirPath);
+  } catch (e) {
+    console.error(`Error: Cannot access "${funcDirPath}". Check if directory exists or has read permissions.`);
```
sweatybridge, line 15:
```suggestion
    console.error(`Error: Cannot access "${funcDirPath}". Check if directory exists and has read permissions.`);
```

sweatybridge (approved):

Issue #959, opened 2023-03-25
Edge function bundling fails if code is loaded from private source (DENO_AUTH_TOKENS)


evelant:
# Bug report

- [X ] I confirm this is a bug with Supabase, not with my own application. 
- [ X] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Loading dependencies from a private source fails when bundling edge functions.  I've got an auth token for the private code in env.DENO_AUTH_TOKENS (per deno docs) and Deno can download and cache the code without issue. The function runs fine locally but bunding fails as soon as I attempt to deploy.

```
error: Uncaught (in promise) Error: Module not found "https://raw.githubusercontent.com/path/to/my/private/code/privateModule.ts".
      const ret = new Error(getStringFromWasm0(arg0, arg1));
                  ^
    at __wbg_new_8d2af00bc1e329ee (https://deno.land/x/eszip@v0.30.0/eszip_wasm.generated.js:312:19)
    at <anonymous> (https://deno.land/x/eszip@v0.30.0/eszip_wasm_bg.wasm:1:79439)
    at <anonymous> (https://deno.land/x/eszip@v0.30.0/eszip_wasm_bg.wasm:1:1388039)
    at <anonymous> (https://deno.land/x/eszip@v0.30.0/eszip_wasm_bg.wasm:1:1862894)
    at __wbg_adapter_18 (https://deno.land/x/eszip@v0.30.0/eszip_wasm.generated.js:146:6)
    at real (https://deno.land/x/eszip@v0.30.0/eszip_wasm.generated.js:130:14)
```

## To Reproduce

1. Create a new edge function
2. Add a dependency to the edge function that requires DENO_AUTH_TOKENS to access (such as a private github repo)
3. Verify that DENO_AUTH_TOKENS is correct then observe that `deno cache -r index.ts` caches the files without problem. Also observe that `supabase functions serve my-function` works without issue.
4. Run `supabase functions deploy my-function` and observe the crash during bundling when it cannot access the files in the private repository

## Expected behavior

Anything that works locally should successfully bundle and deploy. 

## Screenshots

N/A

## System information

- OS: macOS
- Browser (if applies) [e.g. chrome, safari]
- Version of supabase-js: 1.45.2
- Version of Node.js: 16.19.0

## Additional context

[...]

PR #968, opened 2023-03-29
fix(functions): respect `DENO_AUTH_TOKENS`

soedirgo:
Fixes https://github.com/supabase/cli/issues/959

### DIFF
--- a/cmd/link.go
+++ b/cmd/link.go
@@ -22,7 +22,7 @@ var (
 			}
 			dbPassword = viper.GetString("DB_PASSWORD")
 			if dbPassword == "" {
-				dbPassword = link.PromptPassword(os.Stdin)
+				dbPassword = link.PromptPasswordAllowBlank(os.Stdin)
 			}
 			ctx, _ := signal.NotifyContext(cmd.Context(), os.Interrupt)
 			return link.Run(ctx, projectRef, dbPassword, fsys)
--- a/docs/templates/examples.yaml
+++ b/docs/templates/examples.yaml
@@ -20,19 +20,19 @@ supabase-link:
     name: Basic usage
     code: supabase link --project-ref ********************
     response: |
-      Enter your database password: ********
+      Enter your database password (or leave blank to skip): ********
       Finished supabase link.
   - id: without-password
     name: Link without database password
     code: supabase link --project-ref ******************** <<< ""
     response: |
-      Enter your database password:
+      Enter your database password (or leave blank to skip):
       Finished supabase link.
   - id: using-alternate-dns
     name: Link using DNS-over-HTTPS resolver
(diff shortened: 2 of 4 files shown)


internal/utils/misc.go
```diff
@@ -524,6 +524,10 @@ func CopyDenoScripts(ctx context.Context, fsys afero.Fs) (*DenoScriptDir, error)
 	return &sd, nil
 }
 
+func GetDenoAuthTokensEnv() string {
+	return "DENO_AUTH_TOKENS="+os.Getenv("DENO_AUTH_TOKENS")
```
sweatybridge, line 528: Can't users already specify this in the function's env map file?

internal/utils/denos/build.ts
```diff
@@ -72,3 +76,127 @@ async function buildAndWrite(p: string, importMapPath: string) {
...
+  /** The loaded module is either _external_ or _built-in_ to the runtime. */
+  kind: "external" | "builtIn";
+  /** The string URL of the resource. If there were redirects, the final
+   * specifier should be set here, otherwise the requested specifier. */
+  specifier: string;
+}
+
+type LoadResponse = LoadResponseModule | LoadResponseExternalBuiltIn;
+
+async function load(
```
laktek, line 146: We should probably continue to use `load` function from eszip module in the long-term. Otherwise we'd miss any updates/security fixes. Don't mind we reimplement it for the short-term. 

Maybe we can propose a change to eszip package to accept a custom function that provides request headers? eg: `load(specifier: string, headerFunc: (url: URL) => Headers)`

laktek (approved):

internal/utils/misc.go
sweatybridge: People already use the env map file for `DENO_CERT` so it would be cleaner to keep all deno related env vars inside it.

sweatybridge (changes requested):
suggest using env map file instead of forwarding cli env var

internal/utils/misc.go
soedirgo: Good point, will update

internal/utils/denos/build.ts
soedirgo: Yeah makes sense, I'll add a task to upstream this as a follow up

internal/utils/misc.go
soedirgo: One caveat with this is the `--env-file` doesn't work for `deploy` since it doesn't have that option. So the user still needs to use `DENO_AUTH_TOKENS` as an env to `functions deploy`.

sweatybridge (approved):

internal/utils/misc.go
sweatybridge: yup, good to know. may be bundle inside a container in the future?

PR #971, opened 2023-03-30
fix: sso add cmd

kangmingtay:
* The value of the flag `--type saml` is not being passed down to the underlying API call
* Fix a bug with the metadata url scheme validation
* Fix staticcheck errors

### DIFF
--- a/internal/sso/internal/saml/files.go
+++ b/internal/sso/internal/saml/files.go
@@ -69,8 +69,8 @@ func ValidateMetadataURL(ctx context.Context, metadataURL string) error {
 		return err
 	}
 
-	if strings.ToLower(parsed.Scheme) != "https:" {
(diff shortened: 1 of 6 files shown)


sweatybridge (approved):

internal/sso/internal/saml/files.go
```diff
@@ -69,8 +69,8 @@ func ValidateMetadataURL(ctx context.Context, metadataURL string) error {
 		return err
 	}
 
-	if strings.ToLower(parsed.Scheme) != "https:" {
-		return errors.New("Only HTTPS Metadata URLs are supported.")
+	if strings.ToLower(parsed.Scheme) != "https" {
```
hf, line 72: Why didn't the test catch this?

hf:
Can you please leave the capitalization in the errors as they are printed to the user?

PR #972, opened 2023-03-30
fix: make add-domains and remove-domains work

kangmingtay:
* `--add-domains` wasn't adding the domain because `params.Domains` always defaults to an empty slice and will never by `nil`
* `--remove-domains` wasn't removing the domain because `var domains []string` defaults to nil instead of an empty slice. When the `domainsSet` is empty, `domains` will remain `nil` and thus, no domains will be removed.


hf:
Let's merge once we've identified all the problems.

Issue #980, opened 2023-04-03
Could not diff schema on local development setup due to non-existence function error


vhpx:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Running `supabase start` work as expected. After making any changes to the local database via the hosted Studio interface, then running `pnpm supabase db diff -f new_migration` results in the following error:

![image](https://user-images.githubusercontent.com/81307469/229537590-a85ca30b-6f65-41e5-887d-f9e265f44dd3.png)

If I run `supabase stop`, then `supabase start` again, it launched as if there was no error at all. This bug prevent me from making changes using the Studio Dashboard and automatically diff-ing it against old schema.

## Expected behavior

It should be able to diff the current schema without giving any warning, as `supabase start` did not give any errors/warnings.

## System information

- OS: Windows 11 (22H2), Build 22621.1485
- Version of supabase-js: v2.13.1
- Version of supabase cli: v1.47.2
- Version of Node.js: v19.8.1

vhpx:
Running with `--debug` tag provides the following output:

```bash
Applying migration 20230209135251_add_wallets.sql...
2023/04/03 21:25:00 PG Send: {"Type":"Parse","Name":"","Query":"create table \"public\".\"project_wallets\" (\r\n    \"id\" uuid not null default uuid_generate_v4(),\r\n    \"name\" text,\r\n    \"balance\" numeric default '0'::numeric,\r\n    \"currency\" text default 'VND'::text,\r\n    \"description\" text,\r\n    \"created_at\" timestamp with time zone default now(),\r\n    \"project_id\" uuid not null\r\n)","ParameterOIDs":null}
2023/04/03 21:25:00 PG Send: {"Type":"Bind","DestinationPortal":"","PreparedStatement":"","ParameterFormatCodes":null,"Parameters":[],"ResultFormatCodes":[]}
2023/04/03 21:25:00 PG Send: {"Type":"Describe","ObjectType":"P","Name":""}
2023/04/03 21:25:00 PG Send: {"Type":"Execute","Portal":"","MaxRows":0}
[...]

PR #992, opened 2023-04-05
fix: Update to pass postgres startup options through entrypoint

kouwasi:

Bug fix.

Passing arguments through Cmd option.
But it is not works.

## What is the new behavior?

Pass arguments through Entrypoint options.

## Additional context

related #980

### DIFF
--- a/internal/db/start/start.go
+++ b/internal/db/start/start.go
@@ -43,7 +43,25 @@ func Run(ctx context.Context, fsys afero.Fs) error {
 	return err
 }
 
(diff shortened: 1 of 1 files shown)


kouwasi:
Looks like this PR is duplicate of #991. 

I don't know, Which one is better for Supabase. If this fix is unnecessary, please close it.

sweatybridge:
Yup thanks for looking into this. The cmd override is not always necessary so I merged the other PR.

Issue #995, opened 2023-04-05
 Bump version of Storage-API


gcrespi:
# Chore

## Describe the chore

Bump the version of storage-api to at least `v0.30.0`

## Additional context

Storage API is currently on `v0.29.1`, I would like to have available on my supabase local environment [this feature](https://github.com/supabase/storage-api/pull/282) that was introduced on version 0.30.0.

PR #999, opened 2023-04-06
fix: update storage image and migrations

sweatybridge:

Bug fix https://github.com/supabase/cli/issues/995

## What is the new behavior?

## Additional context

### DIFF
--- a/internal/db/start/start.go
+++ b/internal/db/start/start.go
@@ -62,7 +62,7 @@ func NewContainerConfig() container.Config {
 EOF
 `},
 	}
-	if utils.Config.Db.MajorVersion >= 13 {
+	if utils.Config.Db.MajorVersion >= 14 {
 		config.Cmd = []string{"postgres",
 			"-c", "config_file=/etc/postgresql/postgresql.conf",
 			// Ref: https://postgrespro.com/list/thread-id/2448092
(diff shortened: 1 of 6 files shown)


soedirgo (approved):

internal/utils/templates/initial_schemas/13.sql
```diff
@@ -2906,4 +2942,5 @@ ALTER EVENT TRIGGER pgrst_drop_watch OWNER TO supabase_admin;
 -- PostgreSQL database dump complete
 --
 
+DROP EXTENSION pg_graphql; CREATE EXTENSION pg_graphql WITH SCHEMA graphql;
```
soedirgo, line 2945: pg_graphql is only available on PG14+

internal/db/start/start.go
```diff
@@ -62,7 +62,7 @@ func NewContainerConfig() container.Config {
 EOF
 `},
 	}
-	if utils.Config.Db.MajorVersion >= 14 {
+	if utils.Config.Db.MajorVersion >= 13 {
```
soedirgo, line 65: We only support PG13+ for the CLI, so this is always `true`. But if it ain't broke... Note that for PG13 the default `config_file` is different: `/var/lib/postgresql/data/postgresql.conf`

sweatybridge: Hmm I probably broke it after resetting entrypoint. Let me revert this and make another fix

soedirgo (approved):

PR #1002, opened 2023-04-07
fix: update storage api image

grschafer:

Bump storage image version from 0.34.0 to 0.35.1

I recently ran into this issue https://github.com/supabase/storage-api/issues/255 (can't read a 2+GB file from disk), which is fixed in storage-api version 0.35.0

## What is the new behavior?

I'm curious to try out the new version so I can read large files from disk.


sweatybridge:
I'm hesitant to merge this because there's a lot of changes from https://github.com/supabase/storage-api/compare/v0.34.0...v0.35.1 and we haven't fully tested and released 0.35.1 on hosted platform.

Generally CLI tries to match the docker image with what we have in prod for feature parity. If this specific issue is blocking you, one alternative is to build the binary locally for now.

grschafer:
Gotcha, thanks for the explanation! Feel free to close this MR if you prefer.

sweatybridge (approved):

Issue #942, opened 2023-03-18
supabase serve uses "--no-npm" mode for Deno. Why? This breaks code that needs npm modules.


evelant:
# Bug report

I ported my firebase functions to supabase edge functions. In doing so I had to use some npm modules, which I thought was fine since deno supports npm. Upon trying to run my code however I was suprised to be hit with 

```
worker thread panicked Unsupported scheme "npm" for module "npm:/@effect/io@0.7.0/Exit". Supported schemes: [
    "data",
    "blob",
    "file",
    "http",
    "https",
]
```

## Describe the bug

The supabase cli passes "--no-npm" when invoking Deno. 
Code is here https://github.com/supabase/cli/blob/a30210717bdb1f3c9d3d5ce940d000971fc1a6da/internal/functions/serve/serve.go#L190

As far as I can tell this isn't documented anywhere by Supabase resulting in a surprise for users. Since Supabase targets people who want to switch from Firebase this is likely to affect almost everyone who attempts to switch as most people use npm modules in their firebase functions.

## To Reproduce

1. Add an npm specifier to your import_map.json such as `"@effect/io/": "npm:/@effect/io@0.7.0/",`
2. run `supabase functions serve`
3. Observe that function crashes at runtime due to `--no-npm` option passed to Deno by supabase cli.

## Expected behavior

All Deno features should work. If there's a specific reason why `--no-npm` is needed it should be clearly documented along with an explanation of how to work around being unable to use npm import specifiers.

## Screenshots

N/A

## System information

- OS: macOS
- Browser N/A
- Version of supabase-js: N/A
- Version of Node.js: N/A

evelant:
For additional context -- I need to use npm resolutions in Deno because there are some modules (like `@effect/io`) that tools like `esm.sh` choke on. The only way to load them in deno is via npm.

sweatybridge:
Thank you for clarifying your use case. I've pinged our functions team to help take a look.

laktek:
@evelant [Deno Deploy](https://deno.com/deploy), the hosting platform we use for functions doesn't support npm specifiers yet. To keep dev/prod parity we disable the npm resolution when serving functions locally.

We will update the docs to mention this limitation.

evelant:
[...]

PR #1005, opened 2023-04-10
Enable JWT verification when serving all functions

### LINKED ISSUE #873: Serving multiple functions (with edge-runtime) removes the authorization header requirement
# Bug report

## Describe the bug

When serving edge functions locally using the CLI (without specifying `--no-verify-jwt`), they enforce that the request has an Authorization header containing a valid JWT. When serving all edge functions (i.e. `supabase functions serve` rather than `supabase functions serve funcName`), this authentication requirement is not being enforced -- requests without an Authorization header at all are accepted and processed.

This issue was originally pointed out in https://github.com/supabase/cli/issues/869, but that issue focused on a different bug.

## To Reproduce

Set up a new project:

```
mkdir supabase-cli-playground2
cd supabase-cli-playground2
supabase init
supabase start
supabase functions new foo
```

Edit `supabase/functions/foo/index.ts` to have the following content:

```
import { serve } from "https://deno.land/std@0.168.0/http/server.ts";
import { createClient } from "https://esm.sh/@supabase/supabase-js@2.8.0";

console.log("Hello from Functions!");

serve(async (req) => {
  const supabaseAdmin = createClient(
    Deno.env.get("SUPABASE_URL") ?? "",
    Deno.env.get("SUPABASE_SERVICE_ROLE_KEY") ?? ""
  );

  return new Response("ok");
});
```

Serve all functions:

```
supabase functions serve
```

In a separate terminal, curl the function and get a 200 OK response:

```
$ curl -v -X POST 'http://localhost:54321/functions/v1/foo'
*   Trying 127.0.0.1:54321...
* Connected to localhost (127.0.0.1) port 54321 (#0)
> POST /functions/v1/foo HTTP/1.1
> Host: localhost:54321
> User-Agent: curl/7.81.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Content-Type: text/plain;charset=UTF-8
< Content-Length: 2
< Connection: keep-alive
< vary: Accept-Encoding
< date: Fri, 17 Feb 2023 16:40:07 GMT
< X-Kong-Upstream-Latency: 41
< X-Kong-Proxy-Latency: 1
< Via: kong/2.8.1
< 
* Connection #0 to host localhost left intact
ok
```

If I instead serve a single function:

```
supabase functions serve foo
```

And make the curl request, then I get the expected 401 response with body `Missing authorization header`:

```
at 09:40:07 ❯ curl -v -X POST 'http://localhost:54321/functions/v1/foo'
*   Trying 127.0.0.1:54321...
* Connected to localhost (127.0.0.1) port 54321 (#0)
> POST /functions/v1/foo HTTP/1.1
> Host: localhost:54321
> User-Agent: curl/7.81.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 401 Unauthorized
< Content-Type: text/plain; charset=utf-8
< Content-Length: 28
< Connection: keep-alive
< x-relay-error: true
< vary: Accept-Encoding
< date: Fri, 17 Feb 2023 16:40:35 GMT
< X-Kong-Upstream-Latency: 5
< X-Kong-Proxy-Latency: 2
< Via: kong/2.8.1
< 
* Connection #0 to host localhost left intact
Missing authorization header
```

## Expected behavior

I would expect edge functions to behave the same whether they're served via `supabase functions serve foo` or `supabase functions serve`.

## System information

- OS: Ubuntu 22.04
- Version of supabase-js: 2.8.0
- Version of supabase-cli: 1.38.4
  - supabase/edge-runtime:v1.0.15
  - supabase/kong:2.8.1
  - supabase/deno-relay:v1.5.0
- Version of docker: 23.0.1

laktek:

* Updates edge-runtime to 1.2.12 (the latest release).
* Provide a main function script as required by the newer release of edge-runtime.
* Enable JWT verification for functions via the main function.

No JWT verification when serving all functions via `supabase function serve`.

## What is the new behavior?

Serving all functions will default to verify JWT unless `--no-verify-jwt` flag is passed.

## Additional context

Fixes #873. 

Also, the new version of edge-runtime should include `Deno.errors` (as mentioned in the ticket #942)

Tested `supabase start` and it also seems to work.


internal/functions/serve/serve.go
```diff
@@ -297,21 +311,39 @@ func ServeFunctions(ctx context.Context, envFilePath string, noVerifyJWT *bool,
...
 	fmt.Println("Serving " + utils.Bold(utils.FunctionsDir))
-	cmd := []string{"start", "--dir", relayFuncDir, "-p", "8081"}
+	cmd := "edge-runtime start --main-service /home/deno/main -p 8081"
 	if importMapPath != "" {
-		cmd = append(cmd, "--import-map", dockerImportMapPath)
+		cmd = cmd + " --import-map " + dockerImportMapPath
 	}
 	if viper.GetBool("DEBUG") {
-		cmd = append(cmd, "--verbose")
+		cmd = cmd + " --verbose"
```
laktek, line 334: this concats feel hackish. Feel free to suggest a cleaner way to do it.

internal/functions/serve/templates/main.ts
```diff
@@ -0,0 +1,92 @@
...
+  try {
+    const worker = await EdgeRuntime.userWorkers.create({
+      servicePath,
+      memoryLimitMb,
+      workerTimeoutMs,
+      noModuleCache,
+      importMapPath: importMapPath && importMapPath.length ? importMapPath : null,
+      envVars
+    });
+    return worker.fetch(req);
```
sweatybridge, line 83: The worker constructor is async but the actual fetch call isn't?

internal/functions/serve/templates/main.ts
```diff
@@ -0,0 +1,92 @@
...
+    )
+  }
+
+  const servicePath = `{{ .FuncDir }}/${service_name}`;
+  console.error(`serving the request with ${servicePath}`);
+
+  const memoryLimitMb = 150;
+  const workerTimeoutMs = 1 * 60 * 1000;
+  const noModuleCache = false;
+  const importMapPath = "{{ .ImportMapPath }}";
```
sweatybridge, line 71: Shall we make these environment variables instead?

internal/functions/serve/serve.go
sweatybridge: I would prefer keeping cmd as an array of strings then have another variable `cmdString = strings.Join(cmd, " ")`

internal/functions/serve/templates/main.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  if (!service_name || service_name === "") {
+    const error = { msg: "missing function name in request" }
+    return new Response(
+        JSON.stringify(error),
+        { status: 400, headers: { "Content-Type": "application/json" } },
+    )
+  }
+
+  const servicePath = `{{ .FuncDir }}/${service_name}`;
```
sweatybridge, line 65: Since we know the `service_name`, shall we add per function verify-jwt option to config.toml?

internal/functions/serve/templates/main.ts
laktek: fetch is also async. we returns its promise (since serve is an async function)

internal/functions/serve/templates/main.ts
laktek: yep, we can do a per function config for both import map and verify jwt. I'm not familiar with parsing toml files, so if you can help with an update, that'd be great.

internal/functions/serve/templates/main.ts
sweatybridge: > we returns its promise

That would return a `Promise<Promise<Response>>` since serve itself is async and we don't await on the promise returned by worker.fetch? Even if it works because of futures continuation, I think it's more consistent to use await.

internal/functions/serve/templates/main.ts
sweatybridge: My bad. Seems like we already support per function [VerifyJWT](https://github.com/supabase/cli/blob/main/internal/utils/config.go#L153) so no need to update config toml.

We just need to update main.ts and [ServeFunctions](https://github.com/supabase/cli/blob/main/internal/functions/serve/serve.go#L282) to use the configured value.

sweatybridge: Perhaps a simple way is to use per function env var such as `VERIFY_JWT_<SERVICE_NAME>`

internal/functions/serve/templates/main.ts
soedirgo: I don't think this will be an issue, but made the change anyway.

internal/functions/serve/templates/main.ts
laktek: Cool, we can do it as a follow up task. You can provide it as environment variables or as a template variable to the main function.

laktek:
Thanks for fixing those @soedirgo. Shall we merge this and do a follow up PR for per function configs?

soedirgo:
Yea sounds good.

internal/functions/serve/templates/main.ts
sweatybridge: Currently this will introduce yet another different behaviour for `supabase start` and `functions serve`, in addition to the differences with legacy functions serve. I think it's better to complete it in this PR.

soedirgo: Is this different from how `serve --all` works currently? Otherwise I can work on supporting this today/tomorrow.

sweatybridge: On second thought, users can now specify `VERIFY_JWT=false` in `supabase/functions/.env` to support the use case for `supabase start`. So I think it's not really a blocker.

sweatybridge (approved):

Issue #1017, opened 2023-04-16
import_maps.json throws error in version 1.50.4  Downgraded to 1.49.4 and issue resolved.


shrmaky:
# Bug report

- [X] I confirm this is a bug with Supabase, not with my own application. 
- [X] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Import_maps doesnt work in latest 1.50.4 version

Throws following error whenever we try to serve edge function:
<img width="395" alt="image" src="https://user-images.githubusercontent.com/39021707/232347309-17e35409-ba79-4a13-95ae-79d5c1b0eb7b.png">

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Add import_map.json in version 1.50.4 version of supabase-cli
2. Run - supabase functions serve <edge_func_name>
3. See error

## Screenshots

<img width="395" alt="image" src="https://user-images.githubusercontent.com/39021707/232347465-01be13fe-62c9-45f9-b6eb-2a8ac14ec3d3.png">

## System information

- OS: Mac 
- Browser (if applies) [e.g. chrome, safari] - NA
- Version of supabase-js: [e.g. 6.0.2] - 2.11.0
- Version of Node.js: [e.g. 10.10.0] Node 18.12.0

PR #1018, opened 2023-04-17
fix(functions): use absolute path on fallback import map when using relay

soedirgo:
Fixes https://github.com/supabase/cli/issues/1017


sweatybridge (approved):

shrmaky:
Thanks @soedirgo

PR #1023, opened 2023-04-18
fix: remove key auth from kong config

sweatybridge:

Bug fix https://www.notion.so/supabase/Remove-Kong-Token-AuthN-28782c22a29a45b5a1317cb795d98df5?pvs=4

Kong validates jwt which means tokens must be predefined before starting local dev stack

## What is the new behavior?

Allows arbitrary jwt to pass through kong. 

## Additional context

### DIFF
--- a/internal/start/start.go
+++ b/internal/start/start.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strconv"
 	"strings"
 	"text/template"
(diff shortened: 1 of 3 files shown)


soedirgo:
> Allows arbitrary jwt to pass through kong

This seems more like outright omitting AuthN :) Asking for more details on the linked task

sweatybridge:
Discussed with @inian before. I think we intend to remove jwt validation from CF on platform as well. It's more flexible to have individual services to decode the token with mounted secret.

soedirgo (approved):
Sgtm - just not sure if anyone's depending on the current behavior

PR #1032, opened 2023-04-24
fix: update global roles for pg14

sweatybridge:

Bug fix https://github.com/supabase/storage-api/issues/315

New roles were added to [postgres/init-scripts](https://github.com/supabase/postgres/blob/develop/migrations/db/init-scripts/00000000000000-initial-schema.sql) but they are not available in pg14 docker image.

## What is the new behavior?

- add those new roles to globals

## Additional context


sweatybridge:
Thanks for flagging this issue. I've created a PR to fix it in the next release, ie. v1.52.2.

PR #1036, opened 2023-04-25
feat: implement test new command

sweatybridge:

feature

## What is the new behavior?

```bash
$ supabase test new employees
```

## Additional context


soedirgo:
Should we scope this like `supabase test new --type=db foo` so we can add tests for other services in the future?

sweatybridge:
Agree, added a `--template` flag that defaults to pgtap.

soedirgo (approved):

PR #1038, opened 2023-04-26
feat: add vscode template on supabase init

thorwebdev:

- Generate vscode deno settings on new function creation if none exist.

- [x] Implement `Generate workspace settings for VS Code? [y/N]`


internal/functions/new/new.go
```diff
@@ -39,6 +43,23 @@ func Run(ctx context.Context, slug string, fsys afero.Fs) error {
...
+		if _, err := fsys.Stat(vscodeDir); !errors.Is(err, os.ErrNotExist) {
+			return nil
+		}
+		if err := utils.MkdirIfNotExistFS(fsys, vscodeDir); err != nil {
+			return err
+		}
+		if err := afero.WriteFile(fsys, filepath.Join(vscodeDir, "extensions.json"), []byte(vscodeExtensions), 0644); err != nil {
+			return err
+		}
+		if err := afero.WriteFile(fsys, filepath.Join(vscodeDir, "settings.json"), []byte(vscodeSettings), 0644); err != nil {
```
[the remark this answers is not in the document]
thorwebdev: Currently this line is only reached if no `.vscode` directory exists at this location (see line 49 above), so overwriting files shouldn't be possible.

Agreed that appending the settings would be ideal, but it's a bit more complicated than the gitignore since the content is json. 

but since this is only the settings for the functions folder, I think we can start with this as an initial low hanging fruit, wdyt?

internal/functions/new/new.go
```diff
@@ -39,6 +43,23 @@ func Run(ctx context.Context, slug string, fsys afero.Fs) error {
 		}
 	}
 
+	// 3. Set up VS Code Settings
+	{
+		vscodeDir := filepath.Join(utils.FunctionsDir, ".vscode")
+		if _, err := fsys.Stat(vscodeDir); !errors.Is(err, os.ErrNotExist) {
+			return nil
+		}
```
sweatybridge, line 51:
```suggestion
		if _, err := fsys.Stat(vscodeDir); !errors.Is(err, os.ErrNotExist) {
			return err
		}
```

thorwebdev: This is intentional. If the file exists we just silently exit.

thorwebdev:
@sweatybridge any hint how I would implement a `Generate VS Code workspace settings? [y/N]` prompt?

Or would you recommend making it a flag instead?

internal/functions/new/new.go
sweatybridge: hmm i think there are 3 cases here
1. file exists, err == nil, return err will silently exit
2. file not exists, continue with creating
3. other type of os error, eg. no permission to open directory, err != nil, we want to bubble this up

thorwebdev:
Okay, I've added a Yes/No prompt to supabase init now. Wdyt?

sweatybridge (approved):

sweatybridge:
lgtm! made some minor changes and merged

PR #1039, opened 2023-04-26
feat: bump gotrue to v2.60.7

J0:
see title. Not sure if we should keep all the realtime schema changes - lmk

### DIFF
--- a/internal/utils/templates/initial_schemas/15.sql
+++ b/internal/utils/templates/initial_schemas/15.sql
@@ -1791,22 +1791,22 @@ INSERT INTO auth.schema_migrations (version) VALUES ('20230411005111');
 -- Data for Name: migrations; Type: TABLE DATA; Schema: storage; Owner: supabase_storage_admin
 --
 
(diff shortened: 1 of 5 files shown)


internal/utils/templates/initial_schemas/15.sql
```diff
@@ -2,8 +2,8 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 15.1 (Debian 15.1-1.pgdg110+1)
--- Dumped by pg_dump version 15.0
+-- Dumped from database version 14.5 (Debian 14.5-2.pgdg110+2)
+-- Dumped by pg_dump version 14.7 (Homebrew)
```
sweatybridge, line 6: Seems like the dump is created from pg14 database. Could you double check the major version in `supabase/config.toml`?

J0: Ah yeah my bad, have just changed it, pushing up the changes now

sweatybridge (approved):

PR #1070, opened 2023-05-05
feat: implement usage commands for db stats

sweatybridge:

feature

## What is the new behavior?

```bash
$ supabase usage cache

         NAME      │  RATIO
  ─────────────────┼───────────
    index hit rate │ 0.999393
    table hit rate │ 0.974684
```

## Additional context


soedirgo:
Would prefer to make this `supabase db stats cache-hit <--local|--linked|--db-url>`:
- "usage" is less generic
- "usage cache" doesn't specify which service it applies to
- it's ambiguous which DB the stats are for

sweatybridge:
Yup our [current thought](https://supabase.slack.com/archives/C01KSDNKNCW/p1683335532299209?thread_ts=1683169165.684179&channel=C01KSDNKNCW&message_ts=1683335532.299209) is to move these queries into supabase schema. The cli can take the sql function name as command arg to fetch respective stats. For eg,

```bash
supabase db inspect cache-hit
```

Will just do a `select supabase.cache-hit()`

KevinBrolly:
@sweatybridge I put in a new commit and changed the command to a (suggested) style of `supabase inspect db:cache-hit`

I like the `inspect` top level command, then I figured we can use the `db:` prefix because in the future we might want to add ability to inspect other products, like `supabase inspect realtime:connections` or `supabase inspect auth:something` etc.

I also added two new commands, one that shows the status of replication slots on the database, and another that shows the index usage in queries across tables useful for debugging performance problems.

sweatybridge:
Thank you! This DX is much better than before @soedirgo wdyt?

soedirgo:
I kinda prefer to have consistent naming, i.e. either use subcommands everywhere (e.g. `supabase inspect db cache-hit`) or use `:` everywhere (e.g. `supabase inspect db:cache-hit`), and we already use the former for other commands.

Otherwise nice work! 🚀

KevinBrolly:
> I kinda prefer to have consistent naming

@soedirgo @sweatybridge I agree. If you are already using subcommands `supabase inspect db cache-hit` throughout the rest of the CLI I will update the commands under `supabase inspect` to be the same.

I will change in a separate PR when I go to add more commands 👍

PR #1126, opened 2023-05-19
feat: added refresh token options to config

paolodesa:

Feature

The value of the "GOTRUE_SECURITY_REFRESH_TOKEN_ROTATION_ENABLED" environment variable for the gotrue container was hardcoded and set to false.

## What is the new behavior?

Both "GOTRUE_SECURITY_REFRESH_TOKEN_ROTATION_ENABLED" and "GOTRUE_SECURITY_REFRESH_TOKEN_REUSE_INTERVAL" environment variables are parsed from the "config.toml" file when starting the local development containers.


soedirgo:
Thanks for the PR!

Can you outline why these settings need to be exposed? I'm not sure what the attack vector is when developing locally.

paolodesa:
> Thanks for the PR!
> 
> Can you outline why these settings need to be exposed? I'm not sure what the attack vector is when developing locally.

In my particular use case, I create 2 Supabase clients (with the Next.js auth helpers library): one on the browser side and one on the server side. The problem is that sometimes they go out of sync and the server-side client tries to reuse the same refresh token which has already been used on the browser-side, resulting in a forced logout for the user.
By setting the refresh token reuse interval to 10 seconds (just like in the default setup of the hosted version), the server can refresh the token without resulting in an error.
The other environment variable also has to be exposed because the reuse interval is only valid if token rotation is enabled.

Issue #751, opened 2022-12-29
Official Puppeteer example does not work locally


StefanVDWeide:
# Bug report

## Describe the bug

I'm trying to follow the [official puppeteer example](https://github.com/supabase/supabase/tree/master/examples/edge-functions/supabase/functions/puppeteer) but Puppeteer requires Deno to be run with the --unstable flag. See the documentation [here](https://deno.land/x/puppeteer@16.2.0):

```
deno run -A --unstable example.js
```

However, it seems that the Supabase CLI does not support passing additional parameters so running the official example locally does not work.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Init a new Supabase project
2. Follow the steps as described here: [official puppeteer example](https://github.com/supabase/supabase/tree/master/examples/edge-functions/supabase/functions/puppeteer)
3. Run the example locally using the `supabase functions serve` command

## Expected behavior

For the example to work locally so that it can be tested before deploying.

sweatybridge:
Transferring to edge runtime repo since @supabase/edge-functions is maintaining the example code.

PR #1139, opened 2023-05-23
feat: allow --unstable flag in functions serve

sammccord:

feature

Currently, `supabase functions serve` doesn't allow passing any custom deno flags into its underling `deno run` command. This change adds an `--unstable` flag.

fixes https://github.com/supabase/cli/issues/751

## What is the new behavior?

If `--unstable` flag is explicitly added to `supabase functions serve --unstable`, the flag is appended to the `deno run` string before the function path.

Screenshot 2023-05-23 at 9 27 41 AM


cmd/functions.go
```diff
@@ -75,6 +75,7 @@ var (
 	}
 
 	envFilePath string
+	unstable    = new(bool)
```
sweatybridge, line 78: Thank you for contributing. We probably don't want to expose this as a flag because we want maximum parity between serving locally and deploying to prod. And afaik, this is a runtime config and not part of deno bundle.

internal/functions/serve/serve.go
```diff
@@ -174,6 +174,11 @@ func Run(ctx context.Context, slug string, envFilePath string, noVerifyJWT *bool
 			} else {
 				return fmt.Errorf("failed to check index.ts for function %s: %w", slug, err)
 			}
+
+			if *unstable {
+				denoRunCmd = append(denoRunCmd, "--unstable")
```
sweatybridge, line 179: Have you tried deploying the puppeteer example with `supabase functions deploy`? If that works, we can always add the `--unstable` flag to deno run by default.

Ideally the same code should work in both local serving and production deployment.

cmd/functions.go
sammccord: I believe `deno bundle` does expose the unstable flag:
Screenshot 2023-05-24 at 10 41 50 AM

sammccord: > Thank you for contributing. We probably don't want to expose this as a flag because we want maximum parity between serving locally and deploying to prod. And afaik, this is a runtime config and not part of deno bundle.

I believe `deno bundle` does expose the unstable flag:
Screenshot 2023-05-24 at 10 41 50 AM

internal/functions/serve/serve.go
sammccord: For probably unrelated reasons, the puppeteer example doesn't work locally all for me (even on official cli), there's some issue with readable streams that doesn't occur when running the same with `deno run` -
```
Error: channel closed
    at async UserWorker.fetch (ext:sb_user_workers/user_workers.js:54:21)
    at async Server.<anonymous> (file:///home/deno/main/index.ts:104:16)
    at async Server.#respond (https://deno.land/std@0.182.0/http/server.ts:220:24)
worker thread panicked Uncaught SyntaxError: The requested module '/v124/jszip@3.5.0/esnext/lib/readable-stream-browser.js' does not provide an export named 'default'
    at https://esm.sh/v124/jszip@3.5.0/esnext/jszip.mjs:2:291
worker "/home/deno/functions/puppeteer" returned an error: Uncaught SyntaxError: The requested module '/v124/jszip@3.5.0/esnext/lib/readable-stream-browser.js' does not provide an export named 'default'
    at https://esm.sh/v124/jszip@3.5.0/esnext/jszip.mjs:2:291
```

... but since it works great with `deno deploy` - https://gqlatxhpnxwuzspuqqvh.functions.supabase.co/puppeteer, and `deno bundle` and `deno run` both support `--unstable`, maybe it could be the default behavior?

sweatybridge: I'm afraid Deno KV is not currently supported by supabase functions because we are switching to [edge-runtime](https://github.com/supabase/edge-runtime) for both serving locally and prod deployment. The `--unstable` flag added in this PR targets the legacy deno relay server which we have deprecated.

Please consider opening a [github issue](https://github.com/supabase/edge-runtime/issues/new/choose) with edge runtime repo to help our functions team prioritise Deno KV support.

sweatybridge:
I will close this PR for now until upstream support is added to edge runtime.

tgdn:
Hi @sweatybridge, what do you mean by "we can always add the `--unstable` flag to deno run by default."? 

I'm unable to access the `unstable` Deno API when running supabase functions locally. I'd like to make a request using a certificate and key using the unstable `fetch` ([docs](https://deno.land/api@v1.37.1?s=fetch&unstable=)). Thanks!

Issue #265, opened 2022-05-07
Podman support


probablykasper:
# Feature request

Would be really nice to be able to use [Podman](https://podman.io/) instead of Docker. Podman is more lightweight, and from my understanding can run containers rootless

zakaria-chahboun:
+1

zakaria-chahboun:
Yes. You can! Just follow these instructions:

Install `podman` `podman-docker` and `docker-compose`

*i'm using linux pop_os*

```bash
sudo apt-get -y install podman
sudo apt-get -y install podman-docker
sudo apt-get -y install docker-compose
```

Now. You have to clone the supabase repo:
```bash
git clone --depth 1 https://github.com/supabase/supabase
```

Then:
```bash
# Get the code
git clone --depth 1 https://github.com/supabase/supabase

# Go to the docker folder
cd supabase/docker

# Copy the fake env vars
cp .env.example .env
```

Now you have to configure Docker Hub mirror in /etc/containers/registries.conf:

```bash
sudo nano /etc/containers/registries.conf
```

Add these lines to the end of file:
```toml
unqualified-search-registries = ['docker.io']

[[registry]]
prefix = "docker.io"
location = "docker.io"

[[registry.mirror]]
prefix = "docker.io"
location = "docker.mirrors.ustc.edu.cn"
```

Now. You are ready to install supabase studio:
```bash
sudo docker-compose up
```

Finally. visit http://localhost:3000 to start using Supabase Studio.

In case you face this error:

> ERROR: for db  error preparing container 22e6838a1fd10267d63f69ef400000a84a1a857a5e46718003f8a5f63388bb0f for attach: cannot listen on the TCP port: listen tcp4 :5432: bind: address already in use

Just change the `POSTGRES_PORT` in `.env` file.

zakaria-chahboun:
Yes. You can!
https://dev.to/zakariachahboun/supabase-self-hosting-with-podman-1mhl

probablykasper:
That's great, though personally not looking to self host, just running Supabase for local development

zakaria-chahboun:
The local development consume a lot of resources! I don't know why but it seems like there is an issue with the "authentication" part
The issue executed in a loop ..

Don't forget to tell us your experience with it!

GZGavinZhao:
I think the OP meant that when running a local dev instance using `supabase start`, `cli` can use `podman` to spin up the containers instead of using `docker`.

[...]

PR #1179, opened 2023-06-04
fix: handle podman network error type

Wakeful-Cloud:

Check for an existing Docker network via the `NetworkInspect` method instead of `errdefs.IsConflict`.

See [this](https://github.com/supabase/cli/issues/265#issuecomment-1430213314) comment.

## What is the new behavior?

Uses the `NetworkInspect` method to check for existing Docker networks, increasing the compatibility with Podman.

## Additional context


sweatybridge:
I'm inclined not to change this logic because currently the request to create network is atomic. If we make separate api calls to check before create, supabase start may randomly fail due to unrelated docker activity.

My first preference would be to send a PR to podman to address the differences in http status code. 

If that doesn't work out, we can revisit this PR.

internal/utils/docker.go
```diff
@@ -52,7 +51,18 @@ func AssertDockerIsRunning(ctx context.Context) error {
...
+		ctx,
+		networkId,
+		types.NetworkInspectOptions{},
+	)
+
+	// if network already exists, abort
+	if existing.ID != "" && err == nil {
+		return nil
+	}
+
```
sweatybridge, line 64: Please also distinguish network not exist error from other errors like daemon unreachable. We want to proceed with creating the network only in former case and bubble up other errors.

sweatybridge (changes requested):

sweatybridge (approved):

PR #1186, opened 2023-06-06
fix: storage resumable uploads

fenos:

Bug fix

## What is the new behavior?

Add forwarded header in order to generate the correct upload URL for resumable upload

### DIFF
--- a/internal/start/start.go
+++ b/internal/start/start.go
@@ -98,6 +98,7 @@ type kongConfig struct {
 	PgmetaId    string
 	DenoRelayId string
 	LogflareId  string
+	ApiPort     uint
 }
 
 var (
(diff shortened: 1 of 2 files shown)


sweatybridge (approved):

internal/start/start.go
```diff
@@ -98,6 +98,7 @@ type kongConfig struct {
 	PgmetaId    string
 	DenoRelayId string
 	LogflareId  string
+	ApiPort     uint
```
sweatybridge, line 101: This should really be `uint16` but we can probably punt on it to a later time.

fenos: Yhea, the base type is currently `uint` 
https://github.com/supabase/cli/blob/main/internal/utils/config.go#L102

PR #1199, opened 2023-06-11
feat: add experimental command to specify custom Postgres config

darora:
Docs at https://github.com/supabase/supabase/pull/14963

### DIFF
--- a/cmd/postgres.go
+++ b/cmd/postgres.go
@@ -1,0 +1,54 @@
+package cmd
+
+import (
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+	"github.com/supabase/cli/internal/postgresConfig/get"
+	"github.com/supabase/cli/internal/postgresConfig/update"
+	"github.com/supabase/cli/internal/utils/flags"
+)
+
+var (
+	postgresCmd = &cobra.Command{
+		GroupID: groupManagementAPI,
+		Use:     "postgres",
+		Short:   "Manage Postgres database",
+	}
+	postgresConfigCmd = &cobra.Command{
+		Use:   "config",
+		Short: "Manage Postgres database config",
+	}
+
+	postgresConfigGetCmd = &cobra.Command{
+		Use:   "get",
+		Short: "Update network restrictions",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return get.Run(cmd.Context(), flags.ProjectRef, afero.NewOsFs())
+		},
+	}
+
+	postgresConfigUpdateCmd = &cobra.Command{
+		Use:   "update",
+		Short: "Update Postgres database config.",
+		Long: `Overriding the default Postgres config could result in unstable database behaviour.
+Custom configuration also overrides the optimizations generated based on the compute add-ons in use.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return update.Run(cmd.Context(), flags.ProjectRef, postgresConfigValues, postgresConfigUpdateReplaceMode, afero.NewOsFs())
+		},
+	}
+
+	postgresConfigValues            []string
+	postgresConfigUpdateReplaceMode bool
+)
+
+func init() {
+	postgresCmd.PersistentFlags().StringVar(&flags.ProjectRef, "project-ref", "", "Project ref of the Supabase project.")
+	postgresCmd.AddCommand(postgresConfigCmd)
+	postgresConfigCmd.AddCommand(postgresConfigGetCmd)
+	postgresConfigCmd.AddCommand(postgresConfigUpdateCmd)
+
+	postgresConfigUpdateCmd.Flags().StringSliceVar(&postgresConfigValues, "config", []string{}, "Config overrides specified as a 'key:value' pair")
+	postgresConfigUpdateCmd.Flags().BoolVar(&postgresConfigUpdateReplaceMode, "replace-existing-overrides", false, "If true, replaces all existing overrides with the ones provided. If false (default), merges existing overrides with the ones provided.")
+
+	rootCmd.AddCommand(postgresCmd)
+}
--- a/cmd/postgres.go
+++ b/cmd/postgres.go
@@ -11,12 +11,8 @@ import (
 var (
 	postgresCmd = &cobra.Command{
 		GroupID: groupManagementAPI,
-		Use:     "postgres",
-		Short:   "Manage Postgres database",
-	}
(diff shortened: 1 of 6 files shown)


cmd/postgres.go
```diff
@@ -0,0 +1,54 @@
...
+		Short:   "Manage Postgres database",
+	}
+	postgresConfigCmd = &cobra.Command{
+		Use:   "config",
+		Short: "Manage Postgres database config",
+	}
+
+	postgresConfigGetCmd = &cobra.Command{
+		Use:   "get",
+		Short: "Update network restrictions",
```
sweatybridge, line 24:
```suggestion
		Short: "Get the current Postgres database config overrides",
```

cmd/postgres.go
```diff
@@ -0,0 +1,54 @@
...
+	postgresConfigUpdateReplaceMode bool
+)
+
+func init() {
+	postgresCmd.PersistentFlags().StringVar(&flags.ProjectRef, "project-ref", "", "Project ref of the Supabase project.")
+	postgresCmd.AddCommand(postgresConfigCmd)
+	postgresConfigCmd.AddCommand(postgresConfigGetCmd)
+	postgresConfigCmd.AddCommand(postgresConfigUpdateCmd)
+
+	postgresConfigUpdateCmd.Flags().StringSliceVar(&postgresConfigValues, "config", []string{}, "Config overrides specified as a 'key:value' pair")
```
sweatybridge, line 50:
```suggestion
	postgresConfigUpdateCmd.Flags().StringSliceVar(&postgresConfigValues, "config", []string{}, "Config overrides specified as a 'key=value' pair")
```

Can we use `=` to avoid confusion with other commands? For eg. `supabase inspect db:cache-hit` command already uses colon as a namespace separator.

internal/postgresConfig/update/update.go
```diff
@@ -0,0 +1,67 @@
...
+			}
+			finalOverrides = config
+		}
+	}
+	// 3. Create the list of final overrides
+	{
+		for k, v := range newConfigOverrides {
+			// this is hacky - if we're able to convert the value to an integer, we do so
+			// if we start supporting config fields with e.g. floating pt overrides this'll need to be updated
+			attemptedConvert, err := strconv.Atoi(v)
```
sweatybridge, line 41: This can be removed after next prod deploy.

cmd/postgres.go
```diff
@@ -0,0 +1,54 @@
...
+var (
+	postgresCmd = &cobra.Command{
+		GroupID: groupManagementAPI,
+		Use:     "postgres",
+		Short:   "Manage Postgres database",
+	}
+	postgresConfigCmd = &cobra.Command{
+		Use:   "config",
+		Short: "Manage Postgres database config",
+	}
```
sweatybridge, line 20: Can we use `postgres-config` instead for consistency with other commands? For eg. we currently have `network-bans` and `network-restrictions`.

cmd/postgres.go
```diff
@@ -0,0 +1,54 @@
...
+		Use:   "get",
+		Short: "Update network restrictions",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return get.Run(cmd.Context(), flags.ProjectRef, afero.NewOsFs())
+		},
+	}
+
+	postgresConfigUpdateCmd = &cobra.Command{
+		Use:   "update",
+		Short: "Update Postgres database config.",
```
sweatybridge, line 32:
```suggestion
		Short: "Update Postgres database config",
```

internal/postgresConfig/get/get.go
```diff
@@ -0,0 +1,57 @@
...
+func PrintOutPostgresConfigOverrides(config map[string]interface{}) {
+	fmt.Println("- Custom Postgres Config -")
+	const padding = 4
+	w := tabwriter.NewWriter(os.Stdout, 0, 0, padding, ' ', tabwriter.Debug)
+	fmt.Fprintf(w, "Config\tValue\t\n")
+	for k, v := range config {
+		fmt.Fprintf(w, "%s\t%+v\t\n", k, v)
+	}
+	w.Flush()
+	fmt.Println("- End of Custom Postgres Config -")
```
sweatybridge, line 35: We've been rendering markdown tables in most other commands, ie. `migration list`, `sso list`, etc. Shall we do the same here for consistency?

internal/postgresConfig/update/update.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+		resp, err := utils.GetSupabase().UpdateConfigWithBodyWithResponse(ctx, projectRef, "application/json", bytes.NewReader(bts))
+		if err != nil {
+			return err
+		}
+		if resp.JSON200 == nil {
+			if resp.StatusCode() == 400 {
+				return fmt.Errorf("failed to update config overrides: %s (%s). This usually indicates that an unsupported or invalid config override was attempted. Please refer to https://supabase.com/docs/guides/platform/custom-postgres-config", resp.Status(), string(resp.Body))
+			}
+			return fmt.Errorf("failed to update config overrides: %s", resp.Status())
```
sweatybridge, line 63:
```suggestion
			return fmt.Errorf("failed to update config overrides: %s (%s)", resp.Status(), string(resp.Body))
```

It would be helpful to include response body in case we get 4xx/5xx error.

cmd/postgres.go
soedirgo: > db:cache-hit

I thought we wanted to change that? https://github.com/supabase/cli/pull/1070#issuecomment-1582623850

Anyway yeah, we use `foo=bar` for setting secrets

cmd/postgres.go
darora: It was probably a mistake to have them be top-level commands - I can make it postgres-config, but you'll likely end up with a explosion of postgres related commands in the future

internal/postgresConfig/update/update.go
darora: no - there's still ints that will throw an error

internal/postgresConfig/get/get.go
darora: done

cmd/postgres.go
sweatybridge: Hmm there's also a discoverability issue if we nest the commands too deep. I'd prefer keeping it top level for now and refractor later when we add more. Either way not a deal breaker since it's marked experimental.

sweatybridge (approved):

PR #1210, opened 2023-06-14
fix: stop creating shadow database from initial schema

sweatybridge:

Bug fix https://github.com/supabase/cli/pull/1066

## What is the new behavior?

- creates shadow database by running service migrations (this is the same sequence as supabase start)

## Additional context


soedirgo:
Have we resolved whatever was causing the issue here? https://github.com/supabase/cli/pull/1074

sweatybridge:
Yes, what broke db diff was the use of initial schema when creating shadow db. Here we fixed it by running gotrue migrate and storage migrate on the shadow container instead.

soedirgo (approved):

PR #1218, opened 2023-06-17
fix: update postgrest to v11.1.0

samuba:

Updates postgrest container to 11.1.0 (latest)


samuba:
v11.1.0 was not yet published to https://ghcr.io/supabase/postgrest, hence the `CI / START` action fails.
Is there any reason it is not published yet?

sweatybridge:
Yup, the publish action is only triggered on PR approval.

Is there a PostgREST v11 feature you would like to use? Currently our hosted platform still runs [10.2.0](https://github.com/supabase/postgres/blob/develop/ansible/vars.yml#LL13C6-L13C10) release. As much as possible, we would like to keep the local dev version the same as hosted. But if you have a specific use case in mind, we can work on prioritising the upgrade on hosted platform too.

samuel-bach:
Ah I see. Im keen to use this feature: https://postgrest.org/en/stable/references/api/tables_views.html#bulk-insert-with-default-values

samuba:
Ah I see. I thought It's already deployed on hosted version.
Im keen to use this feature: https://postgrest.org/en/stable/references/api/tables_views.html#bulk-insert-with-default-values

Btw. would be nice to have the container/service versions visible in the dashboard somewhere. I was not able to determine for sure which version is deployed

Kamahl19:
I personally would like to use ordering based on parent foreign table. 

Also I have created #1230 for printing the versions

steve-chavez:
New projects on Supabase cloud now have PostgREST v11.1. I think this should be fine to merge.

sweatybridge (approved):

soedirgo (approved):

PR #1226, opened 2023-06-21
feat(api): add api-keys command for projects

Ananya2001-an:

Created `projects api-keys` sub-command. Added necessary test as well.  

## Additional context

I have improved the table rendering using charm/bubbles pkg. In terms of UX, it's better now and maybe we can implement the same everywhere else by adding a common styles file. Also here the keys are not getting word wrapped and I wasn't able to find a solution anywhere. Wdyt? Should we go with this or the same old glamour rendering?

Screenshot 2023-06-21 210627


sweatybridge:
> Also here the keys are not getting word wrapped and I wasn't able to find a solution anywhere. Wdyt? Should we go with this or the same old glamour rendering?

Bubbles doesn't seem to support word wrapping as it currently [truncates](https://github.com/charmbracelet/bubbles/blob/master/table/table.go#L396) cell content to `...` when it overflows. Not able to show the full key would make this command quite useless...

I do appreciate the new look though. Perhaps we can work on the styling on a separate PR.

sweatybridge:
Thank you! The rest of the changes LGTM.

Ananya2001-an:
Sorry for the unnecessary commits... I think we need a squash and merge here haha :)

sweatybridge (approved):

PR #1233, opened 2023-06-22
feat(api): add functions list command

Ananya2001-an:

Added functions list subcommand. Also did some small refactors.


sweatybridge (approved):
Thank you for helping to cleanup the code too!

PR #1234, opened 2023-06-22
chore: improve error style

Ananya2001-an:

Using lipgloss style to format error message. Created `utils.Red` for errors....implemented in only a few places for now.


sweatybridge:
Before you dive too deep into this, you can change the styling of all error messages in [one place](https://github.com/supabase/cli/blob/main/cmd/root.go#L96).

```go
func Execute() {
	if err := rootCmd.Execute(); err != nil {
		fmt.Fprintln(os.Stderr, utils.Red(err.Error()))
...
```

To prevent printing duplicated errors, you can also set `SilenceErrors: true` for rootCmd.

Ananya2001-an:
Oh that would be nice... since I have never worked on a Go cli before I was not aware of this! Thanks for the heads up :)

internal/utils/colors.go
```diff
@@ -13,6 +15,11 @@ func Yellow(str string) string {
 	return lipgloss.NewStyle().Foreground(lipgloss.Color("11")).Render(str)
 }
 
+// For errors.
+func Red(str string) error {
+	return errors.New(lipgloss.NewStyle().Foreground(lipgloss.Color("9")).Render(str))
```
sweatybridge, line 20:
```suggestion
func Red(str string) string {
	return lipgloss.NewStyle().Foreground(lipgloss.Color("9")).Render(str)
```

internal/utils/colors.go
```diff
@@ -1,6 +1,8 @@
 package utils
 
 import (
+	"errors"
+
```
sweatybridge, line 5:
```suggestion
```

Ananya2001-an:
Sure that sounds good!

sweatybridge (approved):

PR #1243, opened 2023-06-25
chore(docs): update stop.md to match new default

tiniscule:
Documentation update for stop command

Incorrectly shows the previous functionality where supabase stop would drop all data by default

## What is the new behavior?

Updated the docs to show the new behavior, but I'm not sure I worded it very well.  I tried to maintain the original docs style.


sweatybridge (approved):
Thank you for spotting.

PR #1271, opened 2023-07-01
feat: add kong config with nginx_worker_processes option

Wakeful-Cloud:

This PR adds the ability to configure [Kong's `nginx_worker_processes`](https://docs.konghq.com/gateway/latest/reference/configuration/#nginx_worker_processes) value via the below TOML config:
```toml
[kong]
# Number of Nginx workers. Set this to 1 to minimize memory usage.
nginx_worker_processes = 1
```

When running in memory-constrained environments, Kong/Nginx may spawn an excessive number of workers, drastically increasing memory usage.

## What is the new behavior?

This PR allows users to fine-tune the number of Kong/Nginx worker processes when desired. If no configuration is provided, the previous behavior (`auto`) is used.

## Additional context

For my use case, this allowed me to cut memory usage of the entire Supabase Docker compose by 79%! (1487 MiB -> 312 MiB) by setting it to just `1` (I'm not especially concerned about the reduced routing performance).

### DIFF
--- a/internal/start/start.go
+++ b/internal/start/start.go
@@ -268,6 +268,22 @@ EOF
 	// Start Kong.
 	p.Send(utils.StatusMsg("Starting containers..."))
 	if !isContainerExcluded(utils.KongImage, excluded) {
+		env := []string{
+			"KONG_DATABASE=off",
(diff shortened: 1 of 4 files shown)


sweatybridge (approved):
This is great change! I think it's safe to default to 1 for local development.

PR #1274, opened 2023-07-03
feat: support sms authentication in local dev

### LINKED ISSUE #180: Support for auth.phone in the config file
# Feature request

Yes, supabase cli doens't allow to config the phone authentication 

Currently the config.toml allows the configuration of many things related to email auth but none for the phone auth although the docker gotrue has those env vars
```env
GOTRUE_SMS_AUTOCONFIRM="false"
GOTRUE_SMS_MAX_FREQUENCY="5s"
GOTRUE_SMS_OTP_EXP="6000"
GOTRUE_SMS_OTP_LENGTH="6"
GOTRUE_SMS_PROVIDER="twilio"
GOTRUE_SMS_TWILIO_ACCOUNT_SID=""
GOTRUE_SMS_TWILIO_AUTH_TOKEN=""
GOTRUE_SMS_TWILIO_MESSAGE_SERVICE_SID=""
GOTRUE_SMS_TEMPLATE="This is from supabase. Your code is {{ .Code }} ."
GOTRUE_SMS_MESSAGEBIRD_ACCESS_KEY=""
GOTRUE_SMS_MESSAGEBIRD_ORIGINATOR=""
GOTRUE_SMS_TEXTLOCAL_API_KEY=""
GOTRUE_SMS_TEXTLOCAL_SENDER=""
GOTRUE_SMS_VONAGE_API_KEY=""
GOTRUE_SMS_VONAGE_API_SECRET=""
GOTRUE_SMS_VONAGE_FROM=""
```

You can run a docker-compose file shared in the supabase/supabase repository with all those settings but it is heavily overwhelming configure everything. The supabase cli makes it soooo much faster to init a supabase dev environment.

### LINKED ISSUE #270: Allow configuration of SUPABASE_REST_URL when using Supabase Studio from CLI
# Feature request
I'd like to be able to set SUPABASE_REST_URL when using Supabase Studio from CLI.

At the moment, I believe I can configure the port using the config.toml file but there doesn't seem to be a way to change the server address.

I think it has something to do with this line in https://github.com/supabase/cli/blob/main/internal/start/start.go:

fmt.Sprintf("SUPABASE_REST_URL=http://localhost:%v/rest/v1/", utils.Config.Api.Port),

CLI seems to assume that REST API is accessible from local host.

I installed Supabase via CLI on a linux server at work, which I can only access from shell.

When accessing Supabase studio at the network address, Supabase Studio can't connect to database.

It gets stuck at "Connecting to default project".

Looking in network tab of chrome developer tools, I can see Studio is sending a request to the localhost but does not receive a reply from:
http://localhost:54321/rest/v1/

I believe this has something to do with the way the server/docker itself is set up on my work network, so there may be another solution, but ideally I would like to be able to configure Studio to use a different server than localhost and to do this by editing the config.toml file.

It seems like it might be possible to change the .env variable for SUPABASE_REST_URL for the docker container but I have no idea how.

sweatybridge:

feature

## What is the new behavior?

- support sms authentication in local dev (closes #180)
- support rest api url config (closes #270)
- removes kong worker config as 1 worker is good enough (relates to #1271)
- initialise default values from init_config.toml

## Additional context


internal/utils/config.go
```diff
@@ -83,13 +84,21 @@ func (s *sizeInBytes) UnmarshalText(text []byte) error {
...
+//     externalising to toml but allows reading from an env var named SUPABASE_AUTH_ANON_KEY.
+//  2. Default values should be added to LoadConfigFS function, after checking for empty value.
 type (
 	config struct {
 		ProjectId string              `toml:"project_id"`
 		Api       api                 `toml:"api"`
-		Db        db                  `toml:"db" mapstructure:"db"`
+		Db        db                  `toml:"db"`
 		Studio    studio              `toml:"studio"`
-		Kong      kong                `toml:"kong"`
```
soedirgo, line 92: This is a breaking change :( Is this already released on stable?

sweatybridge: Nah it's still on beta so I think should be ok

soedirgo: I propose we take out this config and set it to 1, and if people want it to be reverted we can revisit adding this config.

internal/utils/templates/init_config.toml
```diff
@@ -63,13 +72,29 @@ double_confirm_changes = true
...
+enable_confirmations = false
+
+# Configure one of the supported SMS providers: `twilio`, `messagebird`, `textlocal`, `vonage`.
+[auth.sms.twilio]
+enabled = false
+account_sid = ""
+message_service_sid = ""
+# DO NOT commit your Twilio auth token to git. You can leave this field blank and define the
+# following environment variable instead: SUPABASE_AUTH_SMS_TWILIO_AUTH_TOKEN
+auth_token = ""
```
soedirgo, line 88: Why not use `env(SUPABASE_AUTH_SMS_TWILIO_AUTH_TOKEN)`?

internal/utils/config.go
```diff
@@ -229,21 +276,12 @@ func LoadConfigFS(fsys afero.Fs) error {
 		if Config.Api.Port == 0 {
 			return errors.New("Missing required field in config: api.port")
 		}
-		if Config.Api.MaxRows == 0 {
-			Config.Api.MaxRows = 1000
-		}
```
soedirgo, line 234: These need to be kept - setting the field in init_config.toml doesn't change the fact that these are optional and have default values.

soedirgo: Also, we can't detect if boolean values are/aren't supplied if we use `bool` instead of `*bool`.

soedirgo (changes requested):

internal/utils/config.go
sweatybridge: I made a change on L234 to load init_config before user's config.toml. This allows us to define default values in one place.

soedirgo: Ah, missed that

internal/utils/config.go
sweatybridge: I think one use case for overriding the default value is CI runner. Users may have many typescript integration tests hitting the api in parallel. Setting it to a higher value could make their tests run faster.

internal/utils/templates/init_config.toml
sweatybridge: Sure, that sounds better because it's more explicit for the user.

internal/utils/config.go
soedirgo: Sure, we can revisit this if/when a user complains that their tests are slow. Otherwise we risk adding a config that might become irrelevant in the future.

Consider e.g. when we replace Kong with Envoy/Edge Runtime - there may not be an equivalent config in those services.

soedirgo (approved):

Issue #1277, opened 2023-07-04
`ERROR: column "public" of relation "buckets" does not exist (SQLSTATE 42703)` During `supabase db reset`


nbarrow-inspire-labs:
**Describe the bug**
I was previously using v1.68.6 of the cli, and had no problem running migrations that included, among other things:
```plpgsql
-- supabase/migrations/20230420235920_buckets.sql
insert into storage.buckets (id, name, public)
  values
    ('public', 'public', 'TRUE'),
    ('private', 'private', 'FALSE');
```

I then upgraded to v1.75.3 and the following error began presenting itself:
```
Applying migration 20230420235920_buckets.sql...
ERROR: column "public" of relation "buckets" does not exist (SQLSTATE 42703)
At statement 0: insert into storage.buckets (id, name, public)              
  values                                                                    
    ('public', 'public', 'TRUE'),                                           
    ('private', 'private', 'FALSE')                                         
Try rerunning the command with --debug to troubleshoot the error.
```

I tried running the exact same query in an empty project, using v1.75.3, without any migrations, once `supabase db reset` finished (from the web UI's SQL Editor) and get no errors (`Success. 0 rows returned`). 

**To Reproduce**
1. Create a new supabase project without any migrations
```shell
cd ~ && mkdir supabase-bug-test
cd supabase-bug-test
supabase init && supabase start
```
2. Create a new migration: 
```shell
supabase migration new buckets
```
3. Add the following to the newly created migration:
```plpgsql
-- supabase/migrations/..._buckets.sql
insert into storage.buckets (id, name, public)
  values
    ('public', 'public', 'TRUE'),
    ('private', 'private', 'FALSE');
```
4. Attempt to reset the database (`supabase db reset`)

**Expected behavior**
The database should not be throwing the error (indeed, it worked all the way to, at least, v1.68.6 and broke somewhere between v1.75.3)

**Desktop (please complete the following information):**
 - OS: Mac OS 13.3 (22E252)
 - Version of CLI: 
   - v1.68.6 (no error)
   - v1.75.3 (error)

Edit: current workaround is downgrading back to 1.68.6:
```shell
brew tap-new supabase/local-tap                                      
brew extract --version=1.68.6 supabase/tap/supabase supabase/local-tap
brew install supabase@1.68.6
supabase -v
```

sweatybridge:
Can you try `supabase stop --no-backup` before running `supabase start`? This flag cleans up any db volumes leftover from previous runs to ensure a clean start.

[...]

PR #1278, opened 2023-07-04
fix: wait for containers to be ready after db reset

sweatybridge:

Bug fix https://github.com/supabase/cli/issues/1277

## What is the new behavior?

Wait for storage migrations to finish running before applying user migrations.

## Additional context


MichaelDeBoey:
@sweatybridge It seems that somehow this PR got reverted in v2.x as I currently have the same problem as in #1277 with the newest version of the CLI

I've seen @AbdullahAhmadAAK had the same problem a few months ago with #29804, but that one got closed by @Hallidayo due to inactivity

I would be happy to contribute a fix, but I'm not familiar with Go as a language unfortunately, so I'm hoping this one can get resolved by somebody from the supabase team 🙏

PR #1282, opened 2023-07-05
feat(db diff): add --local flag

soedirgo:

feature

`supabase db diff` defaults to diffing against the local database, but there's no `--local` flag, which makes it inconsistent with other `supabase db` subcommands.

## What is the new behavior?

Add `--local` flag (basically a noop vs. the existing behavior).

### DIFF
--- a/cmd/db.go
+++ b/cmd/db.go
@@ -88,7 +88,7 @@ var (
 				if err := parseDatabaseConfig(fsys); err != nil {
 					return err
 				}
-			}
+			} // else use --local, which is the default
 			ctx, _ := signal.NotifyContext(cmd.Context(), os.Interrupt)
 			if usePgAdmin {
 				return diff.Run(ctx, schema, file, dbConfig, fsys)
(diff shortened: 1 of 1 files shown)


thorwebdev (approved):
LGTM, thanks

cmd/db.go
```diff
@@ -237,7 +237,10 @@ func init() {
 	diffFlags.BoolVar(&useMigra, "use-migra", true, "Use migra to generate schema diff.")
 	diffFlags.BoolVar(&usePgAdmin, "use-pgadmin", false, "Use pgAdmin to generate schema diff.")
 	dbDiffCmd.MarkFlagsMutuallyExclusive("use-migra", "use-pgadmin")
-	diffFlags.BoolVar(&linked, "linked", false, "Diffs local schema against the linked project.")
+	diffFlags.StringVar(&dbUrl, "db-url", "", "Diffs local migration files against the database specified by the connection string (must be percent-encoded).")
```
sweatybridge, line 240: This is already defined as a persistent flag on L228.

cmd/db.go
```diff
@@ -237,7 +237,10 @@ func init() {
...
+	diffFlags.BoolVar(&linked, "linked", false, "Diffs local migration files against the linked project.")
+	diffFlags.BoolVar(&local, "local", false, "Diffs local migration files against the local database.")
```
sweatybridge, line 242:
```suggestion
	diffFlags.BoolVar(&local, "local", true, "Diffs local migration files against the local database.")
```

cmd/db.go
soedirgo: Yeah, but I couldn't use it as part of MarkFlagsMutuallyExclusive with the global flag

cmd/db.go
soedirgo: Ah true - will push a fix

PR #1297, opened 2023-07-10
Initial setup for email template server

silentworks:

Add email template feature. This is a PoC, the code should be usable in the CLI but I think it requires some attention from someone in the CLI team.

You can test this out locally but be sure to set the `SUPABASE_INTERNAL_IMAGE_REGISTRY=docker.io` environment variable in order for it to pull down the `gotrue-email` image from my docker hub account.

I think we need to add a few more things to this, maybe the port for the email template server should be configurable inside the config.toml (in this current setup it is not, its serving on a harcoded port of 54330) and also we need to expose the subject environment variables in the `config.toml` file too for the email sending mechanism.

Environment variable that needs exposing:

```
GOTRUE_MAILER_SUBJECTS_INVITE
GOTRUE_MAILER_SUBJECTS_CONFIRMATION
GOTRUE_MAILER_SUBJECTS_RECOVERY
GOTRUE_MAILER_SUBJECTS_MAGIC_LINK
GOTRUE_MAILER_SUBJECTS_EMAIL_CHANGE
```

We will either need to document or create stub files on `supabase init` for the available templates. This would probably be best documented (I'll hande documentation) rather than stub files as its not mandatory to modify your email templates in the local setup but rather an optional step.

The filenames along with their paths are as follows:

```
supabase/templates/invite.html
supabase/templates/confirmation.html
supabase/templates/recovery.html
supabase/templates/magic-link.html
supabase/templates/email-change.html
```

The path could potentially be changed to a nested directory incase in the future we would use `supabase/templates` for more than email templates.

We can't currently edit email templates in local development

## What is the new behavior?

We can now edit email templates in local development

## Additional context


sweatybridge:
Is it possible to avoid introducing another container? Possible alternatives
1. pass static template string to gotrue?
2. let studio api serve these templates

kiwicopple:
>avoid introducing another container

💯 - I think this is important given the number of containers we're already running. Passing these as static strings to GoTrue on startup should be (relatively) simple

> The path could potentially be changed to a nested directory incase in the future

Perhaps we just make it exactly that way from the start? The user can give a path to any file that they want to use in `config.toml`? That saves us from hardcoding some folder structure which we might regret later

silentworks:
> avoid introducing another container

Is this much of a negative if the additional image is only 6mb in size?

> 1. pass static template string to gotrue?

Not sure this is possible at the moment as gotrue requires a file served via http. Also if this was possible would it mean that if I need to make changes to my email template I would have to stop and start the container per change made to the templates?

> 2. let studio api serve these templates

This could potentially work except now studio itself is given a responsibility it didn't have before and would probably require a change in how things work.

silentworks:
A suggestion by @soedirgo was to let Kong serve the static files, but I couldn't find enough information on Kong to serve files while serving the API at the same time.

sweatybridge:
> A suggestion by @soedirgo was to let Kong serve the static files, but I couldn't find enough information on Kong to serve files while serving the API at the same time.

There's a section on custom nginx template that we can refer to for serving static files https://docs.konghq.com/gateway-oss/2.1.x/configuration/#custom-nginx-templates

silentworks:
Closing this out as #1339 is using just kong with a custom nginx config instead of introducing another container.

PR #1339, opened 2023-07-28
feat(auth): initial setup for email template server using kon

silentworks:

Add email template feature. This is a PoC, the code should be usable in the CLI but I think it requires some attention from someone in the CLI team.

I think we need to expose the subject environment variables in the `config.toml` file for the email sending mechanism.

We will either need to document or create stub files on `supabase init` for the available templates. This would probably be best documented (I'll hande documentation) rather than stub files as its not mandatory to modify your email templates in the local setup but rather an optional step.

The filenames along with their paths are as follows:

```
supabase/templates/email/invite.html
supabase/templates/email/confirmation.html
supabase/templates/email/recovery.html
supabase/templates/email/magic-link.html
supabase/templates/email/email-change.html
```

This PR is a replacement for #1297 

We can't currently edit email templates in local development

## What is the new behavior?

We can now edit email templates in local development

## Additional context


silentworks:
@sweatybridge I currently have one issue in this setup where the directories `supabase/templates/email` are being created with the wrong owner. It seems to be assigning `root` along with the `root` group as the owner for just these directories. I can't see what is different in how it's being done for the `functions` and `migrations` directory.

silentworks:
@sweatybridge I tested this again in MacOS and the owner/group of the directories are being set correctly, however in WSL2 on Windows this is still setting the wrong owner/group.

internal/utils/misc.go
```diff
@@ -170,6 +170,7 @@ var (
 	CurrBranchPath        = filepath.Join(SupabaseDirPath, ".branches", "_current_branch")
 	MigrationsDir         = filepath.Join(SupabaseDirPath, "migrations")
 	FunctionsDir          = filepath.Join(SupabaseDirPath, "functions")
+	EmailTemplatesDir     = filepath.Join(SupabaseDirPath, "templates", "email")
```
sweatybridge, line 173:
```suggestion
	EmailTemplatesDir     = filepath.Join(SupabaseDirPath, "auth", "email")
```

Seems more appropriate to nest under `auth` since the config object is also nested under auth.

internal/start/templates/custom_nginx.template
```diff
@@ -0,0 +1,29 @@
...
+
+        location /email {
+            autoindex on;
+            root /home/kong/templates;
+        }
+    }
+
+    # include default Kong Nginx config
+    include 'nginx-kong.conf';
+}
```
sweatybridge, line 29:
```suggestion
}

```

sweatybridge (approved):

PR #1340, opened 2023-07-28
fix(db): bump pg15 release to 15.1.0.103

chaoky:
I don't know if it is that simple, but tests seem to pass and nothing showed up during casual usage ¯\_(ツ)_/¯ 

This release includes pg_graphql v1.2.3 which enables computed fields to return `setof`, that is a essential feature for one to many relations in general (we're using it for i18n)

https://supabase.github.io/pg_graphql/computed_fields/#to-many


sweatybridge (approved):
Looks good! Sometimes we need to add new env vars to the Postgres container. Otherwise just bumping the version is fine.

PR #1342, opened 2023-07-29
add - Postgres Backend for Analytics

filipecabaco:

Bumps version to be used by analytics to allow Postgres ingestion and changes configuration to allow both Big Query or Postgres

Only allowed for Big Query backend for analytics

## What is the new behavior?

Ingests events using Postgres

Screenshot 2023-07-29 at 02 05 51


internal/utils/docker.go
```diff
@@ -184,7 +184,7 @@ func GetRegistryAuth() string {
 }
 
 // Defaults to Supabase public ECR for faster image pull
-const defaultRegistry = "public.ecr.aws"
+const defaultRegistry = "docker.io"
```
filipecabaco, line 187: To be changed. Currently this version of logflare isn't available in ecr

silentworks: This needs to remain the previous value, we need to publish a mirror image of logflare to ecr. For testing you can use this environment variable at runtime to change the registry `SUPABASE_INTERNAL_IMAGE_REGISTRY=docker.io`.

example:
```sh 
SUPABASE_INTERNAL_IMAGE_REGISTRY=docker.io supabase start
```

filipecabaco: Ah nice one! Will change to use the env var for now until we push the image 🙏

Ziinc:
Screenshot 2023-08-07 at 12 13 11 AM
v.1.3.28 contains query translation logic, though slightly buggy. 
final cut should be v1.4.0

studio will need a version bump as well

sweatybridge (approved):

Issue #1370, opened 2023-08-10
Migration squash does not include vault data


sweatybridge:
**Describe the bug**
Up till this point, we have assumed all migrations are schema migrations. However, there are cases where it's useful to insert specific data into the db using a migration. For example, you may want to initialise a lookup table, or update existing rows with a default value before adding a not-null constraint.

Not sure what is the best way to support data migrations, but here are a few ideas to start with:

1. Run data migration from application bootstrap code
2. Run sql manually or via cli hook
3. Interleave data migration in existing schema migration file

**To Reproduce**

```bash
echo 'create table test(name text)' | supabase migration new
echo 'insert into test(name) values("Ant")' | supabase migration new
supabase db push
supabase migration squash
```

**Expected behavior**

Squashed migration should contain inserted data, ie. `insert into test(name) values("Ant")`

**Screenshots**

**Desktop (please complete the following information):**
 - Browser (if applicable) [e.g. chrome, safari]
 - Version of CLI [e.g. v1.60.0]
 - Version of supabase-js (if applicable) [e.g. v2.22.0]
 - Version of Node.js (if applicable) [e.g. v16.20.0]

**Additional context**

simpros:
I think I can add something to this Issue:

I have a migration setting up the stripe wrapper like this:

```
SELECT
    pgsodium.create_key(name := 'stripe');

DO $$
DECLARE
    key_id text;
BEGIN
    SELECT
        id INTO key_id
    FROM
        pgsodium.valid_key
    WHERE
        name = 'stripe'
    LIMIT 1;
    EXECUTE format(E'create server stripe_server foreign data wrapper stripe_wrapper options (api_key_id ''%s'');', key_id);
END
$$;
```
This part is gone after doing `supabase migration squash` but without it I get an error that there is no `stripe_server`

Maybe there is just a better way to setup wrappers in a migration? That was the only way I was able to do it in a reproduceable way.

sweatybridge:
merging this with another related issue https://github.com/supabase/cli/issues/1618

PR #1371, opened 2023-08-10
WIP: dump data when squashing migration

sweatybridge:

Bug fix https://github.com/supabase/cli/issues/1370

Migration files often contain data insert / delete statements. Squashing only dumps the schema.

## What is the new behavior?

Also dumps data.

TODO:
- [ ] probably also need to decrypt columns

## Additional context

### DIFF
--- a/internal/db/dump/dump.go
+++ b/internal/db/dump/dump.go
@@ -42,7 +42,7 @@ func Run(ctx context.Context, path string, config pgconn.Config, schema []string
 	}
 	if dataOnly {
 		fmt.Fprintln(os.Stderr, "Dumping data from remote database...")
-		return dumpData(ctx, config, schema, useCopy, dryRun, outStream)
+		return DumpData(ctx, config, schema, useCopy, dryRun, outStream)
 	} else if roleOnly {
 		fmt.Fprintln(os.Stderr, "Dumping roles from remote database...")
 		return dumpRole(ctx, config, keepComments, dryRun, outStream)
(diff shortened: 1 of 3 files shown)


silentworks:
I don't think data is a part of the schema, so this looks wrong to me especially when running a squashing command.

sweatybridge:
I agree that up till this point, we have assumed all migrations are schema migrations. However, there are cases where it's useful to insert specific data into the db using a migration. For example, you may want to initialise a lookup table, or insert a default value to existing rows before adding a not-null constraint. 

I think we have to support data migration in cli somehow. Although I'm not sure what's the best approach, here are a few ideas to start with:

1. Run data migration from application bootstrap code
2. Run sql manually or via cli hook
3. Interleave data migration in existing schema migration file

This PR assumes the third approach but it's quite problematic like you mentioned.

soedirgo:
This PR doesn't solve the linked issue, at least when I tried it locally. AFAICT, `CREATE SERVER` is omitted because `--schema-only`/`--schema`/`--extension` are set - when you do that pg_dump expects the dump to be schema-local/extension-local, and foreign servers are global across schemas (they're database-level objects).

Simply setting these flags also isn't enough, since you'd be hardcoding pgsodium keys in the migration. This is one of many cases where hand massaging the pg_dump output is unavoidable. Another case where this happens is when we include `CREATE ROLE`s in migrations (maybe conditionally to make it idempotent).

FWIW I haven't seen a migration tool autogenerating a migration that includes data - I figure most people don't expect that either.

sweatybridge:
Closing in favour of a more detailed RFC.

Issue #1367, opened 2023-08-10
Reduce CPU usage by reducing/improving health checks


probablykasper:
**Is your feature request related to a problem? Please describe.**
High CPU usage while idle (~25%).

**Describe the solution you'd like**
There is a `--ignore-health-check` option, but I assume health checks are there for a reason. Maybe there could be some way to improve CPU usage while keeping health checks.

Or other ways to reduce CPU usage

sweatybridge:
Our current health check frequency is set to 2s because we wanted the CLI to respond quicker on `supabase start`. It needs to know when services are up before running migrations or other operations.

I think the ideal fix would be to upgrade docker engine once `start-interval` flag is supported https://github.com/moby/moby/issues/45897. It would allow us to use a high frequency when starting containers, and a low frequency after services become healthy to reduce idle cpu usage.

Alternatively, we can tune each service individually to balance the tradeoff between startup time and idle cpu usage. But it's gonna be a lot of work that I'm not sure if it's worthwhile.

sweatybridge:
I just found out that querying the container health api from cli directly can serve as an alternative to `start-interval` flag. I've increased health check interval to 10s and tested on intel mac with `docker stats`. Idle cpu usage should be around ~5% now.

Let me know if it works for you too `npx supabase@beta start`.

PR #1373, opened 2023-08-11
chore: add health check for edge runtime

sweatybridge:

chore

## What is the new behavior?

Add a health check for edge runtime.

## Additional context

### DIFF
--- a/internal/functions/serve/serve.go
+++ b/internal/functions/serve/serve.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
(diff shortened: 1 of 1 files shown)


internal/functions/serve/serve.go
```diff
@@ -155,6 +156,12 @@ EOF
 			Env:          append(env, userEnv...),
 			Entrypoint:   entrypoint,
 			ExposedPorts: nat.PortSet{"8081/tcp": {}},
+			Healthcheck: &container.HealthConfig{
+				Test:     []string{"CMD", "bash", "-c", "printf \\0 > /dev/tcp/localhost/8081"},
+				Interval: 2 * time.Second,
+				Timeout:  2 * time.Second,
+				Retries:  10,
```
soedirgo, line 163: Let's use the default healthcheck params, in case they're consuming a lot of resources: https://github.com/supabase/cli/issues/1367

Though it's more applicable to `db` or `postges-meta` which have costlier checks

sweatybridge: There's no default health check defined for edge runtime container.

```bash
qiao@Qiaos-MacBook-Pro cli % docker ps -a
CONTAINER ID   IMAGE                            COMMAND                  CREATED          STATUS                    PORTS                                              NAMES
484b18c6f575   supabase/edge-runtime:v1.8.1     "sh -c 'mkdir -p /ho…"   48 minutes ago   Up 48 minutes             8081/tcp                                           supabase_edge_runtime_cli
```

sweatybridge: Anyway I've responded to the other ticket. I don't think the fix is as straightforward as increasing the interval universally to a high value.

soedirgo: TIL, didn't know the flag was a recent addition.

soedirgo (approved):

Issue #1326, opened 2023-07-22
"supabase db reset --linked" can corrupt data in production environment..


DialBird:
Hi.

I'm wondering if I would inadvertently run "supabase db reset --linked" and remove all data in remote DB.

There does not seem to be a permission setting that allows you to execute specific commands, nor is there any kind of "warning" when executing them.

I don't think I am the only one concerned about this issue, but I would like to know how you all are dealing with this risk!

Best Regards.

PR #1392, opened 2023-08-21
fix(db): confirm before resetting remote database

sweatybridge:

Bug fix https://github.com/supabase/cli/issues/1326

Some commands are missing confirmation when updating remote.

## What is the new behavior?

In general, we should prompt for confirmation before updating remote database. The exception is when the command name is explicit, for eg. db push.

## Additional context


soedirgo (approved):
> In general, we should prompt for confirmation before updating remote database

Agree 👍

PR #1395, opened 2023-08-21
feat: add checksum verification in postinstall script

### LINKED ISSUE #1391: Which checksum needs to be verified in postinstall script ?
I was going throigh the repo of the cli and i saw this TODO comment, so i thought of contributing to the repo,
https://github.com/supabase/cli/blob/59b71ba3bc473d3f0e07a546cb2bb9e5a5acf1d0/scripts/postinstall.js#L106

but please correct me if i am wrong, diesnt the package manager checks for the integrity checksum automatically after install or is it something else we are trying to acheive?

If you could provide me an insight of the goal , I would love to contribur to the repo

code-withAshish:

fixes #1391

### DIFF
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -132,17 +132,19 @@ async function main() {
 
   try {
     await pipeline(resp.body, hash);
+    await pipeline(resp.body, ungz);
+    await pipeline(resp.body, untar);
   } catch (error) {
     console.error("Error:", error);
     return;
   }
 
   const calculatedChecksum = hash.digest("hex");
-  if (calculatedChecksum !== expectedChecksum) {
+  if (calculatedChecksum === expectedChecksum) {
+    console.info("Checksum verified");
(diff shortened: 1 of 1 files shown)


code-withAshish:
@sweatybridge can you tell me what's wrong with the npm command in tests?

scripts/postinstall.js
```diff
@@ -91,7 +128,23 @@ async function main() {
...
+      if (calculatedChecksum !== expectedChecksum) {
+        throw checksumError;
+      }
+    });
+
   resp.body.pipe(ungz).pipe(untar);
+
   await new Promise((resolve, reject) => {
     untar.on("error", reject);
     untar.on("end", () => resolve());
```
sweatybridge, line 150: Sorry it's not immediately obvious to me why there's an error. Perhaps you can try registering the callbacks in this promise instead, similar to what we've been doing for untar.

PR #1396, opened 2023-08-21
feat: persist inbucket emails using named volume

kinolaev:

This PR adds named volumes for Inbucket container.

Currently supabase creates 2 new volumes for Inbucket (config and storage) on every `start`.

## What is the new behavior?

Inbucket container uses the same volumes after a restart. These volumes are removed when `--no-backup` is passed to `stop` command.

## Screenshots

Before
Screenshot 2023-08-21 at 13 57 34

After
Screenshot 2023-08-21 at 14 00 31

### DIFF
--- a/internal/utils/config.go
+++ b/internal/utils/config.go
@@ -18,24 +18,26 @@ import (
 )
 
 var (
-	DbImage       string
-	NetId         string
-	DbId          string
-	ConfigId      string
-	KongId        string
-	GotrueId      string
-	InbucketId    string
-	RealtimeId    string
-	RestId        string
-	StorageId     string
-	ImgProxyId    string
-	DifferId      string
-	PgmetaId      string
-	StudioId      string
-	EdgeRuntimeId string
-	LogflareId    string
-	VectorId      string
-	PoolerId      string
+	DbImage                 string
+	NetId                   string
+	DbId                    string
+	ConfigId                string
+	KongId                  string
+	GotrueId                string
+	InbucketId              string
+	InbucketConfigVolumeId  string
+	InbucketStorageVolumeId string
+	RealtimeId              string
+	RestId                  string
+	StorageId               string
+	ImgProxyId              string
+	DifferId                string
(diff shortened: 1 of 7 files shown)


internal/utils/config.go
```diff
@@ -329,6 +331,8 @@ func LoadConfigFS(fsys afero.Fs) error {
 			KongId = "supabase_kong_" + Config.ProjectId
 			GotrueId = "supabase_auth_" + Config.ProjectId
 			InbucketId = "supabase_inbucket_" + Config.ProjectId
+			InbucketConfigVolumeId = "supabase_inbucket_config_" + Config.ProjectId
```
sweatybridge, line 334: What's the use case for persisting the config volume? There's only a single file in `/config/greeting.html` which is created by default. Other configs should be added via [env vars](https://github.com/inbucket/inbucket/blob/main/doc/config.md) instead.

sweatybridge (approved):

kinolaev:
Hello, @sweatybridge! Thank you for your time, but I think you made from my PR something different) The problem I've tried to solve is not persisting emails but preventing `supabase` from creating 2 new unnamed volume every time you call "start". You can see 4 unnamed volumes in the "before" screenshot, 2 of the "in use" and 2 other I should remove manually. This is because `Dockerfile` of Inbucket specifies 2 volumes. That's why docker creates unnamed volumes we don't give them names or mount points.
So, yes, I don't think that we really need to persist `greeting.html`, but we definitely don't need to persist an infinite number of its copies) What do you think? I can create another PR that adds a name to the config volume if you don't mind.

sweatybridge:
Oh I see, thanks for clarifying. I think a more targeted solution here would be to not mount the config volume for inbucket at all. I will give it a try now to see if it's possible.

sweatybridge:
This should be fixed in the latest beta release by mounting the same volume to both `/config` and `/storage`. Thanks for highlighting this issue and suggesting the fix.

PR #1399, opened 2023-08-21
feat: add verifying checksums for downloaded binary

### LINKED ISSUE #1391: Which checksum needs to be verified in postinstall script ?

code-withAshish:

fixes #1391

### DIFF
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -62,34 +62,26 @@ const parsePackageJson = (packageJson) => {
   return { binPath, url };
 };
 
-const parseCheckSumFile = async (packageJson) => {
+const fetchAndParseCheckSumFile = async (packageJson) => {
   const version = packageJson.version;
   const pkgName = packageJson.name;
-  const pkgNameWithPlatform = `${pkgName}_${platform}_${arch}.tar.gz`;
   const checksumFileUrl = `https://github.com/supabase/cli/releases/download/v${version}/${pkgName}_${version}_checksums.txt`;
 
   // Fetch the checksum file
+  console.info("Downloading", checksumFileUrl);
   const response = await fetch(checksumFileUrl);
   if (response.ok) {
     const checkSumContent = await response.text();
     const lines = checkSumContent.split("\n");
 
-    let checksumForRequiredPackage = null;
+    let checksums = {};
 
-    // Find the checksum for the required package
     for (const line of lines) {
       const [checksum, packageName] = line.split(/\s+/);
-      if (packageName === pkgNameWithPlatform) {
-        checksumForRequiredPackage = checksum;
-        break;
-      }
+      checksums[packageName] = checksum;
     }
 
-    if (checksumForRequiredPackage) {
-      return checksumForRequiredPackage;
-    } else {
-      console.error("Error finding checksum for package", pkgNameWithPlatform);
-    }
+    return checksums;
   } else {
     console.error(
       "Could not fetch checksum file",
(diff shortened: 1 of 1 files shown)


code-withAshish:
@sweatybridge Finally i figured it out!!!

scripts/postinstall.js
```diff
@@ -91,7 +131,24 @@ async function main() {
...
+    })
+    .pipe(ungz);
+
+  ungz.on("end", () => {
+    const calculatedChecksum = hash.digest("hex");
+
+    if (calculatedChecksum === expectedChecksum) {
+      console.info("Checksum verified.");
+    } else {
+      console.error("Checksum mismatch. Downloaded data might be corrupted.");
```
sweatybridge, line 146: Don't you need to throw an error here?

code-withAshish: @sweatybridge I updated the PR

scripts/postinstall.js
```diff
@@ -80,6 +119,8 @@ async function main() {
   }
 
   const pkg = await readPackageJson();
+  const hash = createHash("sha256");
+  const expectedChecksum = await parseCheckSumFile(pkg);
```
sweatybridge, line 123: If checksum for a binary is not found, `expectedChecksum` will be undefined and the subsequent comparison with calculated checksum will also fail. I think it's a bit harsh to fail the installation completely if no checksum is found.

How about we change `parseCheckSumFile` to return a dictionary of filename => checksum, and only compares the checksum if the filename exists in the dictionary?

code-withAshish: Not finding the checksum for a file would be very unlike as it's all generated by the GitHub releases artifacts and is available in the checksum text file.

So, don't you think that the checksum would always be found?
Also I generated a dictionary earlier but thought keeping that much data is useles as we will only be comparing checksum for one file.
And about failing the installation I would say isn't that the whole point if veryfying the checksums? 
To fail the installation if verification fails

code-withAshish: I don't mind making the changes you said,
It's just I am thinking that it would be very abnormal situation when we can't find the checksum in the checksum file from GitHub artifacts...
So final decision is on you
I'll make the changes right away if you say so

sweatybridge: There are two cases here that I think would be nice to handle separately

1. If checksum doesn't exist => log a warning and continue installation
2. if checksum exists but mismatch => fail the installation

The current implementation handles the 2nd case well. The 1st case could happen when checksum fails to download due to network issue, or may be goreleaser fails to upload the checksum for a specific platform, or some other unknown scenario.

I agree those scenarios don't happen often. What I'm hoping to avoid though is a hard dependency on `checksum.txt` before downloading any binary via npm.

code-withAshish: ahhh huh i see your point now.
Fair enough
give me 2 mins i'll update the PR

code-withAshish: The changes have been done and now if the file is not fetched then it looks like this :- 

Screenshot from 2023-08-22 15-49-55

and when all works out this :-

image

code-withAshish: and yes for testing purposes and simulating a scenario where file is not fetched i intentionally wrote the wrong url for checksums in first image. 😃

sweatybridge: Nice, this is greatly appreciated!

code-withAshish: Thank you😁
So it's mergeable now ig

sweatybridge (approved):

PR #1407, opened 2023-08-23
fix: update edge runtime to 1.11.1

laktek:

This PR updates the edge runtime to 1.11.1 image.

Notable changes:
* Uses Deno 1.36
* Experimental support for Node built-ins 
* Enabled `Deno.stdout` or `Deno.isatty`

### DIFF
--- a/internal/utils/misc.go
+++ b/internal/utils/misc.go
@@ -33,7 +33,7 @@ const (
 	PgmetaImage      = "supabase/postgres-meta:v0.68.0"
 	StudioImage      = "supabase/studio:20230803-15c6762"
 	ImageProxyImage  = "darthsim/imgproxy:v3.8.0"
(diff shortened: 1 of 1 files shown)


sweatybridge (commented):
Should we update DenoVersion in misc.go as well?

laktek:
> Should we update DenoVersion in misc.go as well?

That's only used to create the eszips right (since we deprecated legacy bundling)? I think it's ok to keep it unchanged. I haven't tested the eszip script with newer versions of Deno.

sweatybridge (approved):
> That's only used to create the eszips right (since we deprecated legacy bundling)? I think it's ok to keep it unchanged. I haven't tested the eszip script with newer versions of Deno.

Yup, sounds good to me.

PR #1427, opened 2023-08-26
Print stopped services only for pretty output

davemurphysf:
Presumably, using the `-o env | json` option for the `status` command is to allow the connection parameters to be machine readable/parsable.  However, the line listing the stopped services breaks this functionality.  This PR changes the status command so that line is only printed on the pretty print (i.e. human readable) format.

### DIFF
--- a/internal/status/status.go
+++ b/internal/status/status.go
@@ -70,10 +70,10 @@ func Run(ctx context.Context, names CustomName, format string, fsys afero.Fs) er
 		utils.LogflareId,
 	}
 	stopped := checkServiceHealth(ctx, services, os.Stderr)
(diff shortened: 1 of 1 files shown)


sweatybridge:
Hmm the stopped services are printed to stderr so they shouldn't break any redirect of command output to file.

Do you mind sharing the full command which breaks?

davemurphysf:
Hey @sweatybridge , you're response made my realize I'm an idiot :-).  I was using `execa` with the [all option](https://www.npmjs.com/package/execa#all-2).  I can switch that to just be `stdout` and it should give me the intended behavior.  Thanks for the quick response!!

Issue #1198, opened 2023-06-10
Enable IP version specification for realtime


Silur:
https://github.com/supabase/realtime/pull/253 got merged into the upstream realtime container and works as expected if local development is happening purely in a docker-compose file but as the CLI does not support the `REALTIME_IP_VERSION` setting, having ipv6 disabled on a local system results in `supabase start` crashing the realtime component

A solution would be to add:
```toml
[realtime]
ip_version="ipv4"
```

Support for the __config.toml__ file

sweatybridge:
This should be possible now with `npx supabase@beta start`

```
[realtime]
enabled = true
ip_version = "IPv4"
```

Silur:
any news on this?

sweatybridge:
@Silur are you still running into the same error? CLI now defaults to IPv4 because docker support for IPv6 is not very good.

Silur:
yes the issue still persists on latest cli versions when ipv6 is disabled

lolcabanon:
I still get this problem today with CLI v2.26.9.

I have this in my `config.toml` :

```toml
[realtime]
enabled = true
ip_version = "IPv4"
```

Running `pnpx supabase start --debug` leave me with this crashlog :

```log
Supabase CLI 2.26.9
2025/06/21 18:37:20 PG Send: {"Type":"StartupMessage","ProtocolVersion":196608,"Parameters":{"database":"postgres","user":"postgres"}}
2025/06/21 18:37:20 PG Recv: {"Type":"AuthenticationSASL","AuthMechanisms":["SCRAM-SHA-256"]}
2025/06/21 18:37:20 PG Send: {"Type":"SASLInitialResponse","AuthMechanism":"SCRAM-SHA-256","Data":"n,,n=,r=rfVsozjWzvH1h5WinLqGNLRl"}
2025/06/21 18:37:20 PG Recv: {"Type":"AuthenticationSASLContinue","Data":"r=rfVsozjWzvH1h5WinLqGNLRlD2+Y/rOxWkO0dI3vtamhZBWf,s=xUvRWlzwhNgwBedJwyIxtA==,i=4096"}
2025/06/21 18:37:20 PG Send: {"Type":"SASLResponse","Data":"c=biws,r=rfVsozjWzvH1h5WinLqGNLRlD2+Y/rOxWkO0dI3vtamhZBWf,p=x2a9oAvzKr90rI8ZclFeUQR1CtGRFgzDPlGwomtgYVA="}
[...]

PR #1430, opened 2023-08-28
feat(realtime): support configuring realtime ip version

sweatybridge:

closes https://github.com/supabase/cli/issues/1198

## What is the new behavior?

Adds the following config
- `realtime.enabled` allow disabling realtime service
- `realtime.ip_version` allow binding realtime to IPv4

## Additional context

### DIFF
--- a/internal/utils/templates/init_config.test.toml
+++ b/internal/utils/templates/init_config.test.toml
@@ -39,7 +39,7 @@ max_client_conn = 100
 [realtime]
 enabled = true
 # Bind realtime via either IPv4 or IPv6. (default: IPv6)
-ip_version = "IPv6"
+ip_version = "IPv4"
 
 [studio]
 enabled = true
--- a/internal/utils/templates/init_config.test.toml
+++ b/internal/utils/templates/init_config.test.toml
@@ -31,6 +31,11 @@ pool_mode = "transaction"
 default_pool_size = 20
 max_client_conn = 100
 
+[realtime]
+enabled = true
+# Bind realtime via either IPv4 or IPv6. (default: IPv6)
+ip_version = "IPv6"
+
 [studio]
 enabled = true
 # Port to use for Supabase Studio.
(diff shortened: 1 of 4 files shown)


internal/utils/templates/init_config.test.toml
```diff
@@ -26,11 +26,21 @@ major_version = 15
...
 pool_mode = "transaction"
+# How many server connections to allow per user/database pair.
 default_pool_size = 20
+# Maximum number of client connections allowed.
 max_client_conn = 100
 
+[realtime]
+enabled = true
+# Bind realtime via either IPv4 or IPv6. (default: IPv6)
+ip_version = "IPv6"
```
soedirgo, line 42: How about only supporting IPv4 for now? Unless a user requests IPv6, this violates "don't make me think"

sweatybridge: I think it's better to have explicit configs. Alternatively we can comment out this line if it reduces user's cognitive load.

soedirgo: There's a reason we're not exposing all configs available in Realtime, GoTrue etc. - this seems like something that's unnecessary to configure unless a user's machine doesn't support IPv4.

sweatybridge: As mentioned in the linked issue, realtime defaults to ipv6. User is requesting to use ipv4 because their computer doesn't support ipv6.

soedirgo: Yes, what I mean is let’s hardcode it to use IPv4, and if it breaks some setup and an IPv6 option is needed, then we make it configurable

sweatybridge: I think commenting out the default config is a better option rather than changing the default and risking breaking setups. Also once users rely on IPv4 as the default, it will be much harder to plan an upgrade path as that's potentially another breaking change. 

The concern that the config is irrelevant also applies to other multitenant services. Given that the cli not only serves platform users, but also self hosted, it's inevitable that we need to support a wider range of configs than we previously planned.

soedirgo (approved):

PR #1432, opened 2023-08-28
fix: pass down jwt secret as database config

sweatybridge:

## What is the new behavior?

## Additional context

### DIFF
--- a/internal/db/start/templates/schema.sql
+++ b/internal/db/start/templates/schema.sql
@@ -1,10 +1,14 @@
 \set pgpass `echo "$PGPASSWORD"`
+\set jwt_secret `echo "$JWT_SECRET"`
+
+ALTER DATABASE postgres SET "app.settings.jwt_secret" TO :'jwt_secret';
 
 ALTER USER authenticator WITH PASSWORD :'pgpass';
 ALTER USER pgbouncer WITH PASSWORD :'pgpass';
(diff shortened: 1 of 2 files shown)


internal/db/start/templates/schema.sql
```diff
@@ -1,10 +1,14 @@
 \set pgpass `echo "$PGPASSWORD"`
+\set jwt_secret `echo "$JWT_SECRET"`
+
+ALTER DATABASE postgres SET "app.settings.jwt_secret" TO :'jwt_secret';
```
steve-chavez, line 4: In theory this should offer the same result as https://github.com/supabase/supabase/pull/16681 but it requires SUPERUSER. Setting any unknown custom option at the db level requires superuser. On constrast, on https://github.com/supabase/supabase/pull/16681 postgREST just sets the custom option at the transaction level (doesn't require superuser).

steve-chavez: If the above is not a concern then yes it would be an alternative solution with the advantage of being able to access the JWT secret in any transaction like the one mentioned on https://github.com/supabase/supabase/pull/16681#pullrequestreview-1598681970

soedirgo (approved):

internal/db/start/templates/schema.sql
sweatybridge: Yup, for feature parity with platform, I will set it at the database level for now. We may change it in the future if this causes problem.

PR #1451, opened 2023-09-04
feat: add proxy support to npm install

arika0093:

Enabled `node-fetch` proxy support by using `proxy-agent`.

`npm install supabase` fails to run in environments where a proxy is required.

## What is the new behavior?

It works fine under proxy environment.

## Additional context

None.

### DIFF
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -8,6 +8,7 @@ import binLinks from "bin-links";
 import { createHash } from "crypto";
 import fs from "fs";
 import fetch from "node-fetch";
+import { ProxyAgent } from "proxy-agent";
 import path from "path";
 import tar from "tar";
 import zlib from "zlib";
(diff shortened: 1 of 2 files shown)


scripts/postinstall.js
```diff
@@ -71,7 +72,7 @@ const fetchAndParseCheckSumFile = async (packageJson) => {
 
   // Fetch the checksum file
   console.info("Downloading", checksumFileUrl);
-  const response = await fetch(checksumFileUrl);
+  const response = await fetch(checksumFileUrl, {agent: new ProxyAgent()});
```
sweatybridge, line 75: Thanks for the contrib. We need to pass down the ProxyAgent to all places where fetch is called. See https://github.com/supabase/cli/pull/1451/files#diff-8d291e232c16ef847e6e526d55be7a3c2ec355821af7da7b948077163769eb70L126

arika0093: Oops, I inadvertently forgot to do that. Adding.

sweatybridge (approved):

PR #1473, opened 2023-09-09
Update new function template

Aschen:

Update the CURL command used to invoke the function.

Reference documentation https://supabase.com/docs/guides/functions/quickstart#invoking-remote-functions

### DIFF
--- a/internal/functions/new/templates/index.ts
+++ b/internal/functions/new/templates/index.ts
@@ -19,7 +19,8 @@ serve(async (req) => {
 })
 
 // To invoke:
-// curl -i --location --request POST 'http://localhost:54321/functions/v1/' \
-//   --header 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZS1kZW1vIiwicm9sZSI6ImFub24iLCJleHAiOjE5ODM4MTI5OTZ9.CRXP1A7WOeoJeXxjNni43kdQwgnWNReilDMblYTn_I0' \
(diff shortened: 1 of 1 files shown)


sweatybridge:
The comments in template were meant for invoking locally served functions. Could you elaborate on the steps you found difficult so we can find a better place to fix the docs?

internal/functions/new/templates/index.ts
```diff
@@ -19,7 +19,8 @@ serve(async (req) => {
 })
 
 // To invoke:
-// curl -i --location --request POST 'http://localhost:54321/functions/v1/' \
-//   --header 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZS1kZW1vIiwicm9sZSI6ImFub24iLCJleHAiOjE5ODM4MTI5OTZ9.CRXP1A7WOeoJeXxjNni43kdQwgnWNReilDMblYTn_I0' \
+// curl -i --location --request POST 'https://<project reference>.functions.supabase.co/<function name>' \
```
sweatybridge, line 22: The comments in template were meant for invoking locally served functions. Could you elaborate on the steps you found difficult so we can find a better place to fix the docs?

sweatybridge (changes requested):

internal/functions/new/templates/index.ts
Aschen: I'm not sure it would work even for locally served function since the function name is missing.

The doc was very clear, I just wanted to fix the CURL command in the template to works with the cloud version but I understand if you prefer to keep the local version.

sweatybridge:
Superseded by a new PR

PR #1476, opened 2023-09-12
fix: abort edge function requests after 200s

laktek:

* Bump to edge-runtime 1.16.0
*  abort edge function requests after 200s (this is the same timeout value as prod)


internal/functions/serve/templates/main.ts
```diff
@@ -126,7 +126,13 @@ serve(async (req: Request) => {
       cpuBurstIntervalMs,
       maxCpuBursts,
     });
-    return await worker.fetch(req);
+    const controller = new AbortController();
+
+    const signal = controller.signal;
```
sweatybridge, line 131:
```suggestion
    const { signal } = controller;
```
nitpick

sweatybridge (approved):

Issue #389, opened 2022-08-18
Update local postgrest container from `9.0.0.20220211` to `9.0.0.20220531`


Marviel:
https://github.com/supabase/cli/blob/f71be9e378b44043e96b0283e221831e722fe7d9/internal/utils/misc.go#L36

When you hit swagger on Supabase hosted, you see the following output:

```
"info": {
        "description": "@graphql({\"inflect_names\": true})",
        "title": "PostgREST API",
        "version": "**9.0.0.20220531** (pre-release) (d14f745)"
},
```

However, locally you see:

```
"info": {
        "version": "**9.0.0.20220211** (pre-release)",
        "title": "PostgREST API",
        "description": "standard public schema"
},
```

sweatybridge:
Hi @Marviel , I'm getting different versions for my hosted project.
```json
{
  "swagger": "2.0",
  "info": {
    "description": "standard public schema",
    "title": "PostgREST API",
    "version": "9.0.1.20220717 (pre-release) (d2df289)"
```

and locally
```json
{
  "swagger": "2.0",
  "info": {
    "description": "standard public schema",
    "title": "PostgREST API",
    "version": "9.0.1.20220802 (pre-release)"
```

The local version should be on the [latest release](https://hub.docker.com/r/postgrest/postgrest/tags), ie. `9.0.1.20220802`. Perhaps you can look into clearing local docker image cache?

The hosted postgREST version is currently on [v9.0.1.20220717](https://github.com/supabase/kps/blob/master/ansible/arm64_vars.yml#L51). But we do batched upgrades, so your project may not have been included in the upgrade yet. If this is blocking something, we can upgrade manually for you too.

We will look into a more automated way of synching these image versions as an enhancement.

PR #1481, opened 2023-09-12
feat: sync postgrest version from linked project

sweatybridge:

closes https://github.com/supabase/cli/issues/389

## What is the new behavior?

- fetch PostgREST version from linked project

## Additional context


internal/utils/tenant/postgrest.go
```diff
@@ -0,0 +1,38 @@
...
+	Description string `json:"description"`
+	Version     string `json:"version"`
+}
+
+type SwaggerResponse struct {
+	Swagger string      `json:"swagger"`
+	Info    SwaggerInfo `json:"info"`
+}
+
+func GetPostgrestVersion(ctx context.Context, projectRef string) (string, error) {
```
soedirgo, line 24: Have you tested this? The PostgREST version in the OpenAPI can't be used directly as Docker image versions - it looks like `11.2.0 (c820efb)`

sweatybridge: Is it causing problems for you? I prefixed with `v` on line 37

The linked project I tested was on 11.1.0 though

soedirgo: Yeah, just created a new project and the `.info.version` is the one above

soedirgo: > I prefixed with v

The problem is the commit hash - the whitespace and everything after needs to be omitted

sweatybridge: Updated code to handle both cases.

soedirgo (approved):

Issue #1506, opened 2023-09-19
Failed to save database password: failed to unlock correct collection '/org/freedesktop/secrets/collection/login'


DYAKOOO:
### Describe the bug

Installed supabase using wsl2 and I'm trying to link my local project. After inserting the database password and hitting enter, I get the output error of not being able to "Failed to save database password: failed to unlock correct collection '/org/freedesktop/secrets/collection/login'". 

### To Reproduce

  `  SUPABASE_DB_PASSWORD="ABC" npx supabase link --project-ref ebxvtsvwyzmohvkhiime`
    
    

### Expected behavior

Command should continue and show error/sucess.

### Screenshots

![image](https://github.com/supabase/cli/assets/94733813/3a21e65f-135e-4b36-b69b-bbbef2174d0b)

### System information

    OS: Windows WSL2
    Browser (if applies) -
    Version of supabase-js: ^2.33.2"
    Version of supabase-cli: 1.93.0
    Version of Node.js: 18.17.0

### Additional context

https://github.com/supabase/cli/issues/899#issue-1604841778 - This is the same problem as this link. I followed the latest solutions but none of them worked.

Issue #899, opened 2023-03-01
`supabase link` hangs after adding database password


arpadgabor:
# Bug report

## Describe the bug

Installed `supabase` using **homebrew** and I'm trying to link my local project. After inserting the database password and hitting enter, the process hangs and I can't do anything else. Ctrl + C does not work either. Any password I add and it just hangs, even if it's correct or not. I also tried restarting the project but got the same behavior.

## To Reproduce

1. Install with homebrew
2. `supabase link --project-ref=abc`
3. Insert db password, hit enter

## Expected behavior

Command should continue and show error/sucess.

## Screenshots

![Code_K7xkf6n8SR](https://user-images.githubusercontent.com/9270494/222131902-c8dd1569-04d4-408e-b339-43e4508192e2.png)

It just hangs like this forever.

## System information

- OS: Windows WSL2
- Browser (if applies) -
- Version of supabase-js: 2.8.0 (but I don't think it affects the issue)
- Version of supabase-cli: 1.40.4
- Version of Node.js: 16.19 or 18.14

## Additional context

n/a

sweatybridge:
This is a duplicate of https://github.com/supabase/cli/issues/819

sweatybridge:
All our envs are prefixed with `SUPABASE`. Can you give `SUPABASE_DB_PASSWORD` a try? I will reopen this for now.

sweatybridge:
It would also be helpful if you can share with us the logs after running with `--debug` flag

arpadgabor:
> All our envs are prefixed with SUPABASE

Ah okay. Tried now also with `--debug` but I'm getting nothing. However what I did find out is that if I run it outside my project it continues the process and then fails with

```
Error: cannot read config in /home/arpad/work: open supabase/config.toml: no such file or directory
Have you set up the project with supabase init?
```

Which is expected. But when running again in the project, it hangs with no logs.

```
✦ ❯ SUPABASE_DB_PASSWORD=abc supabase link --project-ref=xyz --debug

# nothing happening
```

djordje:
I had same issue in WSL2. Most likely you have WSLg (some GUI linux app) installed, which leads to some Gnome password (gnome keyring) manager which is actually leading to this command hanging.
What happens actually is UI popup to unlock vault is not appearing in Windows and it looks like command is hanging and you are not able to kill it.

In order to avoid this I actually use one distro without GUI apps which leads to supabase to use dotfiles to store credentials and everything works like on OSX or real Linux.

[...]

PR #1509, opened 2023-09-20
fix: disable keyring on wsl

sweatybridge:

fixes https://github.com/supabase/cli/issues/1506
fixes https://github.com/supabase/cli/issues/1483
fixes https://github.com/supabase/cli/issues/899

## What is the new behavior?

This means that db password won't be saved. Access token will be written to a file.

## Additional context

### DIFF
--- a/internal/utils/credentials/store.go
+++ b/internal/utils/credentials/store.go
@@ -1,6 +1,10 @@
 package credentials
 
 import (
+	"bytes"
+	"errors"
+	"os"
+
 	"github.com/zalando/go-keyring"
 )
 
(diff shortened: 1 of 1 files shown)


minism:
@sweatybridge This doesn't seem to work on WSL1:

     cat /proc/sys/kernel/osrelease
     4.4.0-19041-Microsoft

Would the supabase team be willing to support an environment variable to force disable keyring?

sweatybridge:
> Would the supabase team be willing to support an environment variable to force disable keyring?

That could be a bit risky on CI because the said env var may be injected by an untrusted source to potentially leak passwords or tokens. I'd rather make a fix for WSL1 in this case.

PR #1510, opened 2023-09-20
feat: container network aliases

gregnr:
Adds network aliases to each container. Makes networking between containers easier during local development.

For example, to invoke an edge function from Postgres using `pg_net`, now all you need is:

```sql
select
    net.http_post(
      url := 'http://api.supabase.internal:8000/functions/v1/my-function',
      headers := ('{"Content-Type": "application/json", "Authorization": "Bearer <anon_key>"}')::jsonb,
      body := ('{"input": "Hello world"}')::jsonb
    );
```

instead of trying to determine the container host name `http://supabase_kong_my-supabase-project:8000`.

## Alternatives
Another option would be to use `http://host.docker.internal:54321` to route packets through the host, though unsure if this DNS host is supported by all container runtimes.

## Long term solution
These are all more short-term solutions - the best [solution](https://notion.so/d9d7e2c7bf7a4e6983fddb877588c0c6) would be to provide something like:
```
current_setting('supabase.project.url')
```

within Postgres to grab the project URL dynamically, regardless of environment (local, preview, prod) like we do with Edge Function [default secrets](https://supabase.com/docs/guides/functions/secrets#default-secrets).


internal/utils/config.go
```diff
@@ -36,6 +36,21 @@ var (
 	VectorId      string
 	PoolerId      string
 
+	DbAliases          = []string{"db"}
+	KongAliases        = []string{"kong", "api"}
```
gregnr, line 40: Alternatively we could make this something like:
```suggestion
	KongAliases        = []string{"kong", "api.supabase.internal"}
```

internal/utils/docker.go
```diff
@@ -281,7 +282,13 @@ func DockerStart(ctx context.Context, config container.Config, hostConfig contai
 		hostConfig.Binds = binds
 	}
 	// Create container from image
-	resp, err := Docker.ContainerCreate(ctx, &config, &hostConfig, nil, nil, containerName)
+	resp, err := Docker.ContainerCreate(ctx, &config, &hostConfig, &network.NetworkingConfig{
+		EndpointsConfig: map[string]*network.EndpointSettings{
+			NetId: {
+				Aliases: aliases,
```
sweatybridge, line 288: I wonder if we can simplify the implementation by automatically adding a network alias based on the container name. For eg.

```go
parts := strings.Split(strings.TrimSuffix(containerName, utils.Config.ProjectId), "_")
alias := strings.Join(parts[1:len(parts)-1], "_")
```

Would give us the following transformation
```
supabase_pooler_cli                => pooler
supabase_studio_cli                => studio
supabase_pg_meta_cli               => pg_meta
supabase_edge_runtime_cli          => edge_runtime
storage_imgproxy_cli               => imgproxy
supabase_storage_cli               => storage
supabase_rest_cli                  => rest
realtime-dev.supabase_realtime_cli => realtime
supabase_inbucket_cli              => inbucket
supabase_auth_cli                  => auth
supabase_kong_cli                  => kong
supabase_db_cli                    => db
```

gregnr: Yeah sure - originally I felt this approach could be a bit brittle, but it's definitely simpler and removes the need to change function signatures 👍

Only potential downside is we would be limited to a single alias for each container, in case we wanted to add more for some containers, like `api` or `api.supabase.internal` on Kong.

sweatybridge: > this approach could be a bit brittle

That's true. Perhaps it would be less brittle if we generate the container names from the list of aliases instead. That should give us more flexibility as we only need to guarantee that the aliases don't change.

internal/utils/docker.go
```diff
@@ -244,7 +245,7 @@ func DockerPullImageIfNotCached(ctx context.Context, imageName string) error {
 	return DockerImagePullWithRetry(ctx, imageUrl, 2)
 }
 
-func DockerStart(ctx context.Context, config container.Config, hostConfig container.HostConfig, containerName string) (string, error) {
+func DockerStart(ctx context.Context, config container.Config, hostConfig container.HostConfig, containerName string, aliases []string) (string, error) {
```
sweatybridge, line 248:
```suggestion
func DockerStart(ctx context.Context, config container.Config, hostConfig container.HostConfig, networkingConfig *network.NetworkingConfig) (string, error) {
```

Would be nice to eventually arrive at this signature

gregnr:
@sweatybridge I've refactored the code to:
1. Generate IDs based on the primary alias
  - I kept the variables for each service ID (since they're used in many places, including tests), and dynamically set them on load based on the primary alias. Let me know if you have any thoughts on this approach.
2. Add `networkingConfig` to the `DockerStart` call signature
  - Aliases are now set one level upstream through this parameter

internal/utils/config.go
```diff
@@ -36,6 +36,21 @@ var (
 	VectorId      string
 	PoolerId      string
 
+	DbAliases          = []string{"db"}
+	KongAliases        = []string{"kong", "api.supabase.internal"}
```
gregnr, line 40: re: `api.supabase.internal`:

Personally I like this as a FQDN, feels more approachable for users. Also similar Docker's own DNS aliases, ie. `host.docker.internal`. Thoughts? Any downsides?

sweatybridge: Sounds good to me. Are we going to update the other services as well? ie. `db.supabase.internal` etc.

sweatybridge (approved):

internal/utils/config.go
sweatybridge: Otherwise feel free to merge this PR as it is.

gregnr: Added `db.supabase.internal`, but leaving the rest until a use case arises (most are proxied through kong).

Issue #1508, opened 2023-09-20
Local `pg-bouncer` container startup auth error


bombillazo:
**Describe the bug**

We are having trouble with the local pg-bouncer image when running on a linux machine.

**To Reproduce**
1. Run `supabase start`
2. Go to the `supabase_pooler_app` container logs to see error

**Expected behavior**

**Screenshots**

![image](https://github.com/supabase/cli/assets/44002321/690dc80e-c172-4f36-8147-ffa2fdd4d4da)

**Desktop (please complete the following information):**
 - OS: Linux (Fedora 38)
 - Version of CLI: 1.97.1

**Additional context**

sweatybridge:
Kinda crazy how this bug trickled down from upstream [bitnami/pgbouncer repo](https://github.com/bitnami/containers/issues/48959). I've updated our ECR mirror with their latest fixes for this tag. Could you run `docker pull public.ecr.aws/supabase/pgbouncer:1.20.1` and restart the local stack to see if it helps?

I've also made a PR to use their immutable tag to prevent such cases from happening in the future.

bombillazo:
@sweatybridge we just confirmed it works, thanks!

PR #1511, opened 2023-09-20
fix(pooler): pin pgbouncer image to an immutable tag

sweatybridge:

Bug fix https://github.com/supabase/cli/issues/1508

Turns out bitnami updated their docker hub image tag 1.20.1 which broke the generated userlist.txt https://github.com/bitnami/containers/issues/48264#issuecomment-1727534704

We mirrored the image to our ECR repo and propagated the bug to cli stack.

## What is the new behavior?

To prevent such breaking changes, we pin bitnami/pgbouncer to an immutable tag.
Also synced [pgbouncer config](https://github.com/supabase/postgres/blob/develop/docker/all-in-one/etc/pgbouncer/pgbouncer.ini#L123) to support logging in as non-postgres user.

## Additional context


soedirgo (approved):

internal/start/start.go
```diff
@@ -770,10 +770,12 @@ EOF
 				Image: utils.PgbouncerImage,
 				Env: []string{
 					"POSTGRESQL_HOST=" + dbConfig.Host,
-					"POSTGRESQL_NAME=" + dbConfig.Database,
-					"POSTGRESQL_USER=" + dbConfig.User,
 					fmt.Sprintf("POSTGRESQL_PORT=%d", dbConfig.Port),
+					"POSTGRESQL_USERNAME=pgbouncer",
```
soedirgo, line 774: Not `dbConfig.User`?

sweatybridge: Yup, this is for pgbouncer to connect and fetch the actual user password from postgres.

PR #1545, opened 2023-10-04
chore(docs): clarify need for commas in schema flag of db commands

mikelhamer:

docs update

## Additional context

As a first time user and CLI noob, it took me a bit to figure out the list of schema had to be comma separated, so I thought it would be useful to add this bit to the docs. This is consistent with other parts of the docs such as the description of domains flag in the sso add command.


sweatybridge (approved):
lgtm! definitely clearer than before

mikelhamer:
Awesome :) Thank you!

PR #1581, opened 2023-10-16
feat: allow network to be defined

samlown:

Simple change to allow a custom docker network name to be used instead of the default.

Default and only available network is `supabase_network_project_id` (where `project_id` is that of the current project).

## What is the new behavior?

If the configuration toml defines this at the top-level:

```
network = my-custom-network
```

The network called `my-custom-network` will be used instead of the default.

## Additional context

This is useful for running supabase inside an existing local development network so that other docker images can access supabase resources without going through the host. Given the simple names of supabase services, like `db`, `auth`, `rest` etc. there is a chance of name collisions. #1510 helps reduce complications with longer aliases.

### DIFF
--- a/internal/utils/config.go
+++ b/internal/utils/config.go
@@ -185,6 +185,7 @@ var Config = config{
 type (
 	config struct {
 		ProjectId string              `toml:"project_id"`
+		Network   string              `toml:"network"`
 		Api       api                 `toml:"api"`
 		Db        db                  `toml:"db" mapstructure:"db"`
 		Realtime  realtime            `toml:"realtime"`
@@ -395,6 +396,9 @@ func LoadConfigFS(fsys afero.Fs) error {
 			LogflareId = GetId(LogflareAliases[0])
 			VectorId = GetId(VectorAliases[0])
 			PoolerId = GetId(PoolerAliases[0])
+		}
+		if Config.Network != "" {
+			NetId = Config.Network
 		}
 		// Validate api config
 		if Config.Api.Port == 0 {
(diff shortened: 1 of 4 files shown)


sweatybridge:
> This is useful for running supabase inside an existing local development network so that other docker images can access supabase resources without going through the host.

If `supabase start` errors due to unhealthy container, the current implementation will try to [delete the network by id](https://github.com/supabase/cli/blob/main/internal/start/start.go#L82). This means any custom network may be deleted and recreated by cli if there are no resources using it. The recreated network will likely have different properties, like additional labels. Would this be a concern for you?

If so, I think a more desirable behaviour might be to leave the existing custom network alone. This can be implemented by pruning the network only if it matches a [filter by label](https://github.com/supabase/cli/blob/main/internal/stop/stop.go#L92).

samlown:
Thanks @sweatybridge! Normally docker wouldn't let you delete a network that is in active use, which is our case when using `docker compose down`, it'll just raise an error which is currently ignored as per the link you posted. Even if deleted, running `docker compose up` again will just re-create it from scratch.

In our use-case, this wouldn't be an issue, but I can see the potential for problems. Filtering by label for networks that were created by supabase could be a reasonable solution.

sweatybridge:
I thought about this more and find that an extra flag like `supabase start --use-network <id>` might be more suitable here. Reason is network id doesn't seem to be a project level config. This concept doesn't apply to remote project, it may change between different local dev machines, and doesn't need to be committed to git.

I would be happy to add the start flag if you are ok with it too.

sweatybridge:
closing due to staleness

Issue #1388, opened 2023-08-18
Downloading Storage Bucket Contents


Angus-Moore-Dev:
**Is your feature request related to a problem? Please describe.**
I would love if Supabase had the functionality to be able to download contents from the storage buckets. I have a number of projects that require referencing to files in buckets. This works fine if I am storing the URL of the object (for public files), but this creates real problems when fetching relative files from buckets for existing data.

**Describe the solution you'd like**
The ability to download the contents from storage buckets into my local development environment. Maybe as a flag for the `db dump` command?

**Describe alternatives you've considered**
None, since this becomes really problematic when you have nested folders and files to deal with.

**Additional context**
N/A

If you guys are able to add this, that would be fantastic. Might be a problem for users who have sensitive information in their databases, but for those of us storing relative paths for files in the database (because of RLS), it'd be nice to be able to actually have those files locally for when we run our development environment. Thanks :D

sweatybridge:
This is a great idea. We have an internal RFC to add the following commands to sync storage bucket contents between remote and local. The files are expected to be saved to `supabase/storage` directory while entries in `storage.buckets` and `storage.objects` table will probably be pulled as a new migration file in `supabase/migrations`.

- `supabase storage ls`

```bash
$ supabase storage ls my_bucket
test1.csv test2.txt

$ supabase storage ls *.csv
my_bucket/test1.csv
```

- `supabase storage cp`

```bash
$ supabase storage cp ss://my_bucket/test3.png .
Downloaded file: ./my_bucket/test3.png

$ supabase storage cp my_bucket ss://my_bucket
Uploaded file: my_bucket/test1.csv
Uploaded file: my_bucket/test2.txt
Overwritten file: my_bucket/test3.png

$ supabase storage cp my_bucket/test3.png ss://public/
Confirm uploading files from my_bucket to public? [y/N]
Uploaded file: public/test3.png
```

- `supabase storage rm`

```bash
$ supabase storage rm my_bucket/test3.png
Deleted file: my_bucket/test3.png

$ supabase storage rm -r my_bucket
Confirm deleting all files in my_bucket? [y/N]
Deleted file: my_bucket/test1.csv
Deleted file: my_bucket/test2.txt
```

These commands should feel familiar to bash users but we are also open to your feedback if there are use cases not covered here.

gruckion:
Please implement this

rdewolff:
Hey there! 👋 

[...]

PR #1587, opened 2023-10-17
feat: implement storage management commands

sweatybridge:

feature
closes https://github.com/supabase/cli/issues/1388

## What is the new behavior?

Implements storage `ls` / `cp` / `mv` / `rm` as proposed.

## Additional context

Not implemented in this PR:
- [ ] support cp between buckets 
- [ ] support batch cp and mv
- [ ] support self hosted 
- [x] support cp file metadata
- [ ] support file explorer for large buckets
- [ ] support resumable TUS upload

### DIFF
--- a/cmd/storage.go
+++ b/cmd/storage.go
@@ -35,7 +35,7 @@ var (
 
 	cpCmd = &cobra.Command{
 		Use: "cp <src> <dst>",
-		Example: `cp readme.md ss:///bucket
+		Example: `cp readme.md ss:///bucket/readme.md
 cp -r docs ss:///bucket/docs
 cp -r ss:///bucket/docs .
 `,
--- a/cmd/storage.go
+++ b/cmd/storage.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
+	"github.com/supabase/cli/internal/storage"
 	"github.com/supabase/cli/internal/storage/cp"
 	"github.com/supabase/cli/internal/storage/ls"
 	"github.com/supabase/cli/internal/storage/mv"
@@ -24,7 +25,7 @@ var (
 		Short:   "List objects by path prefix",
 		Args:    cobra.MaximumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			objectPath := ls.STORAGE_SCHEME + ":///"
+			objectPath := storage.STORAGE_SCHEME + ":///"
 			if len(args) > 0 {
 				objectPath = args[0]
 			}
--- a/cmd/storage.go
+++ b/cmd/storage.go
@@ -19,7 +19,7 @@ var (
 	recursive bool
 
 	lsCmd = &cobra.Command{
-		Use:     "ls [prefix]",
+		Use:     "ls [path]",
 		Example: "ls ss:///bucket/docs",
 		Short:   "List objects by path prefix",
 		Args:    cobra.MaximumNArgs(1),
@@ -38,7 +38,7 @@ var (
 cp -r docs ss:///bucket/docs
 cp -r ss:///bucket/docs .
 `,
-		Short: "Copy objects by from src to dst path",
+		Short: "Copy objects from src to dst path",
 		Args:  cobra.ExactArgs(2),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return cp.Run(cmd.Context(), args[0], args[1], recursive, afero.NewOsFs())
@@ -47,7 +47,7 @@ cp -r ss:///bucket/docs .
 
 	mvCmd = &cobra.Command{
 		Use:     "mv <src> <dst>",
-		Short:   "Move objects by from src to dst path",
+		Short:   "Move objects from src to dst path",
 		Example: "mv -r ss:///bucket/docs ss:///bucket/www/docs",
 		Args:    cobra.ExactArgs(2),
 		RunE: func(cmd *cobra.Command, args []string) error {
--- a/cmd/storage.go
+++ b/cmd/storage.go
@@ -1,0 +1,81 @@
+package cmd
+
+import (
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+	"github.com/supabase/cli/internal/storage/cp"
+	"github.com/supabase/cli/internal/storage/ls"
+	"github.com/supabase/cli/internal/storage/mv"
+	"github.com/supabase/cli/internal/storage/rm"
+)
+
+var (
+	storageCmd = &cobra.Command{
+		GroupID: groupManagementAPI,
+		Use:     "storage",
+		Short:   "Manage Supabase Storage objects",
+	}
+
+	recursive bool
+
+	lsCmd = &cobra.Command{
+		Use:     "ls [prefix]",
+		Example: "ls ss:///bucket/docs",
+		Short:   "List objects by path prefix",
+		Args:    cobra.MaximumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			objectPath := ls.STORAGE_SCHEME + ":///"
+			if len(args) > 0 {
+				objectPath = args[0]
+			}
+			return ls.Run(cmd.Context(), objectPath, recursive, afero.NewOsFs())
+		},
+	}
+
+	cpCmd = &cobra.Command{
+		Use: "cp <src> <dst>",
+		Example: `cp readme.md ss:///bucket
+cp -r docs ss:///bucket/docs
+cp -r ss:///bucket/docs .
+`,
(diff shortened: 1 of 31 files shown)


cmd/storage.go
```diff
@@ -0,0 +1,81 @@
...
+		},
+	}
+
+	cpCmd = &cobra.Command{
+		Use: "cp <src> <dst>",
+		Example: `cp readme.md ss:///bucket
+cp -r docs ss:///bucket/docs
+cp -r ss:///bucket/docs .
+`,
+		Short: "Copy objects by from src to dst path",
```
kamilogorek, line 41:
```suggestion
		Short: "Copy objects from src to dst path",
```

cmd/storage.go
```diff
@@ -0,0 +1,81 @@
...
+		Args:  cobra.ExactArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return cp.Run(cmd.Context(), args[0], args[1], recursive, afero.NewOsFs())
+		},
+	}
+
+	mvCmd = &cobra.Command{
+		Use:     "mv <src> <dst>",
+		Short:   "Move objects by from src to dst path",
```
kamilogorek, line 50:
```suggestion
		Short:   "Move objects from src to dst path",
```

cmd/storage.go
```diff
@@ -0,0 +1,81 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			objectPath := ls.STORAGE_SCHEME + ":///"
+			if len(args) > 0 {
+				objectPath = args[0]
+			}
+			return ls.Run(cmd.Context(), objectPath, recursive, afero.NewOsFs())
+		},
+	}
+
+	cpCmd = &cobra.Command{
```
kamilogorek, line 35: As an improvement, we could extend `cp` and `mv` commands to have a shape of `<scr> [src...] <dst>`, just like regular unix tools. This way you can move/cp multiple files at once.

cmd/storage.go
```diff
@@ -0,0 +1,81 @@
...
+rm ss:///bucket/docs/example.md ss:///bucket/readme.md
+`,
+		Args: cobra.MinimumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return rm.Run(cmd.Context(), args, recursive, afero.NewOsFs())
+		},
+	}
+)
+
+func init() {
```
kamilogorek, line 71: I'm not sure if moving directory shouldn require a recursive call tbh, as this is not the behavior people are used to. Doing `mv foo bar` if `foo` is a directory will work in any unix environment. Only `cp` and `rm` require this flag.

internal/storage/cp/cp.go
```diff
@@ -0,0 +1,138 @@
...
+			}
+		} else {
+			if baseName != "." && (dirExists || len(noSlash) == 0) {
+				dstPath = path.Join(dstPath, baseName)
+			}
+			dstPath = path.Join(dstPath, relPath)
+		}
+		fmt.Fprintln(os.Stderr, "Uploading:", filePath, "=>", dstPath)
+		err = client.UploadStorageObject(ctx, projectRef, dstPath, filePath, fsys)
+		if err != nil && strings.Contains(err.Error(), `"error":"Bucket not found"`) {
```
kamilogorek, line 123: Tying ourselves to the exact error message is not great imo. Can we at least assert on http code instead? It's also available from what I see. Same for 2 different places.

kamilogorek, line 123: I'm not sure if we should always create a bucket without any additional flag, but from UX perspective it sounds ok.

internal/storage/cp/cp_test.go
```diff
@@ -0,0 +1,584 @@
...
+	t.Run("throws error on missing project", func(t *testing.T) {
+		// Setup in-memory fs
+		fsys := afero.NewMemMapFs()
+		// Run test
+		err := Run(context.Background(), ".", ".", false, fsys)
+		// Check error
+		assert.ErrorIs(t, err, utils.ErrNotLinked)
+	})
+
+	t.Run("throws error on missing project", func(t *testing.T) {
```
kamilogorek, line 171: This is testing permissions, not presence.

internal/storage/ls/ls.go
```diff
@@ -0,0 +1,167 @@
...
+	"os"
+	"path"
+	"strings"
+
+	"github.com/spf13/afero"
+	"github.com/supabase/cli/internal/storage/client"
+	"github.com/supabase/cli/internal/utils"
+)
+
+const STORAGE_SCHEME = "ss"
```
kamilogorek, line 17: This should live in global scope of the package, as it's tied to the whole `storage` concept, not just `ls`.

cmd/storage.go
```diff
@@ -0,0 +1,81 @@
...
+	storageCmd = &cobra.Command{
+		GroupID: groupManagementAPI,
+		Use:     "storage",
+		Short:   "Manage Supabase Storage objects",
+	}
+
+	recursive bool
+
+	lsCmd = &cobra.Command{
+		Use:     "ls [prefix]",
```
kamilogorek, line 22:
```suggestion
		Use:     "ls [path]",
```

As this argument overrides whole `objectPath`, not appends it

cmd/storage.go
```diff
@@ -0,0 +1,81 @@
...
+		Example: "ls ss:///bucket/docs",
+		Short:   "List objects by path prefix",
+		Args:    cobra.MaximumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			objectPath := ls.STORAGE_SCHEME + ":///"
+			if len(args) > 0 {
+				objectPath = args[0]
```
kamilogorek, line 29: Not sure what's the point of having a default `objectPath` here, as it will fail anyway due to `ParseStorageURL` check.

internal/storage/ls/ls.go
```diff
@@ -0,0 +1,167 @@
...
+		fmt.Println(objectPath)
+		return nil
+	}
+	if recursive {
+		return IterateStoragePathsAll(ctx, projectRef, remotePath, callback)
+	}
+	return IterateStoragePaths(ctx, projectRef, remotePath, callback)
+}
+
+var errInvalidURL = errors.New("URL must match pattern ss:///bucket/prefix")
```
kamilogorek, line 38: Prefix is optional, so the pattern is `ss:///bucket/<prefix>`

cmd/storage.go
kamilogorek: Unless this is actual "root" bucket and `ParseStorageURL` will treat it as literal `/`

internal/storage/cp/cp.go
sweatybridge: 404 could either be thrown from http path not found or bucket doesn't exist. We can add the status code as an additional check but eventually we need a more specific error code to distinguish the two.

cmd/storage.go
sweatybridge: hmm I named this `prefix` because it can match a partial path prefix. But I guess `path` is fine.

cmd/storage.go
sweatybridge: This is meant to be a convenience for listing all buckets or objects, for eg.
```
$ supabase storage ls
test/
nothing/
private/
```

kamilogorek (approved):
LGTM, although next time maybe let's try merge it command by command, as 2.5k LoC is difficult to fit in the head at once reviewing 😅 Especially that `ls` is the base, on which remaining 3 commands are built on.

cmd/storage.go
sweatybridge: Agree on the difference. I was a bit reluctant to make `mv` recursive by default because it would mean at least 2 network calls to copy a file, ie. first call tries to move as a file and then second call to move as a directory.

I find adding a `-r` flag isn't too bad in this case because it's more consistent with other commands. Also we are only supporting moving within the same bucket (no local -> remote) so I think it's ok to have some minor differences.

If it turns out to be a big inconvenience, reverting to unix behaviour would not be a breaking change for those who already use the `-r` flag. So this is probably ok for now.

cmd/storage.go
sweatybridge: Yup, good suggestion. I've added as a TODO.

internal/storage/rm/rm.go
```diff
@@ -49,6 +49,10 @@ func Run(ctx context.Context, paths []string, recursive bool, fsys afero.Fs) err
 		return err
 	}
 	for bucket, prefixes := range groups {
+		confirm := fmt.Sprintf("Confirm deleting files in bucket %v?", utils.Bold(bucket))
```
kamilogorek, line 52: I like that it handles non-TTY by default 😌

homerjam:
This is just what I need (I think), are there any docs?

Issue #1574, opened 2023-10-12
supabase db diff fails with: psycopg2.OperationalError: could not translate host name "localhost" to address: Try again


vixalien:
**Describe the bug**
I'm trying to do `supabase db diff` but it fails with `psycopg2.OperationalError: could not translate host name "localhost" to address: Try again`. I even tried changing the DNS resolver from native to https, but it doesn't work

**To Reproduce**
1. Try to do a supabase db diff

**Expected behavior**
Normal diff to get written

**Screenshots**

Logs:

```
$ supabase db diff --dns-resolver https
Connecting to local database...
Creating shadow database...
Setting up initial schema...
Applying migration 20220822165112_xyz.sql...
Applying migration 20220822165112_xyo.sql...
Diffing schemas: extensions,public
error diffing schema: error running container: exit 1:                                                                                                 
WARNING: Running pip as root will break packages and permissions. You should install packages reliably by using venv: https://pip.pypa.io/warnings/venv
WARNING: You are using pip version 21.1.1; however, version 23.2.1 is available.                                                                       
You should consider upgrading via the '/usr/local/bin/python -m pip install --upgrade pip' command.                                                    
Traceback (most recent call last):                                                                                                                     
  File "/usr/local/lib/python3.9/site-packages/schemainspect/misc.py", line 9, in connection_from_s_or_c                                               
    s_or_c.engine                                                                                                                                      
AttributeError: 'Session' object has no attribute 'engine'                                                                                             
                                                                                                                                                       
During handling of the above exception, another exception occurred:                                                                                    
                                                                                                                                                       
Traceback (most recent call last):                                                                                                                     
  File "/usr/local/lib/python3.9/site-packages/sqlalchemy/engine/base.py", line 3212, in _wrap_pool_connect                                            
    return fn()                                                                                                                                        
  File "/usr/local/lib/python3.9/site-packages/sqlalchemy/pool/base.py", line 301, in connect                                                          
    return _ConnectionFairy._checkout(self)                                                                                                            
  File "/usr/local/lib/python3.9/site-packages/sqlalchemy/pool/base.py", line 761, in _checkout                                                        
    fairy = _ConnectionRecord.checkout(pool)                                                                                                           
  File "/usr/local/lib/python3.9/site-packages/sqlalchemy/pool/base.py", line 419, in checkout                                                         
    rec = pool._do_get()                                                                                                                               
[...]

PR #1595, opened 2023-10-20
fix: replace all localhost instances with 127.0.0.1

kamilogorek:
Essentially `s/localhost/127.0.0.1/g`. Using localhost can be flaky due to cli running in various environments and this will potentially fix some random network issues that has been reported.

We don't want to perform a lookup on `localhost` as it will go to IPv6 first, where we always want to use IPv4 instead.


sweatybridge (approved):

sweatybridge:
Thanks! This could potentially resolve https://github.com/supabase/cli/issues/1574

PR #1619, opened 2023-10-26
fix: no need to decompress buffer when downloading functions

laktek:

This PR fixes the issue with downloading edge functions. Looks like the function payload is now automatically decompressed (from brtoli), so no longer need to decompress payload manually (couldn't find the change that caused the behavior change though)


sweatybridge (approved):

sweatybridge:
This change broke deploying new functions. I've reverted it for it so we can investigate further.

laktek:
> This change broke deploying new functions. I've reverted it for now so we can investigate further.

Do you have an error report? This code path is different from deploying new functions.

sweatybridge:
Sorry I wasn't clear. It broke downloading of newly deployed functions. Steps to reproduce:

```
npx supabase@1.110.2 functions deploy hello_world
npx supabase@1.110.2 functions download hello_world
```

Error is the same as reported by Inian
image

PR #1630, opened 2023-11-01
feat: Add automatic login flow support

kamilogorek:
Adds automated login flow, which uses the browser session to authenticate and generate a new token, which is used in the cli `login` command.

```
Usage:
  supabase login [flags]

Flags:
  -h, --help           help for login
      --name string    Name that will be used to store token in your settings, defaults to built-in token name generator
      --no-browser     Do not open browser automatically
      --token string   Use provided token instead of automatic login flow
```

RFC: https://www.notion.so/supabase/RFC-Automatic-browser-based-CLI-login-flow-fc4f91fc7e7f4f168d2da6192841dcec?d=d8c9bcbe826c4665a89f9240d0854a36

API changes: https://github.com/supabase/infrastructure/pull/15529
Studio changes: https://github.com/supabase/supabase/pull/18641

### DIFF
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/mattn/go-isatty"
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
@@ -64,6 +65,17 @@ var (
 				params.Token = token
 			}
 
+			// Login encryption and Session ID are only required for end-to-end communication.
+			// We can skip it if token is already provided by user.
+			if params.Token == "" {
+				enc, err := login.NewLoginEncryption()
+				if err != nil {
+					return err
+				}
+				params.Encryption = enc
+				params.SessionId = uuid.New().String()
+			}
+
 			if !term.IsTerminal(int(os.Stdin.Fd())) && params.Token == "" {
 				return ErrMissingToken
 			}
@@ -73,13 +85,13 @@ var (
 				if err != nil {
 					return fmt.Errorf("cannot parse 'name' flag: %w", err)
 				}
-				params.Name = name
+				params.TokenName = name
 			} else {
 				name, err := generateTokenName()
 				if err != nil {
-					params.Name = fmt.Sprintf("cli_%d", time.Now().Unix())
+					params.TokenName = fmt.Sprintf("cli_%d", time.Now().Unix())
 				} else {
-					params.Name = name
+					params.TokenName = name
 				}
 			}
 
@@ -90,7 +102,7 @@ var (
 				params.OpenBrowser = isatty.IsTerminal(os.Stdin.Fd()) && isatty.IsTerminal(os.Stdout.Fd())
 			}
 
-			return login.Run(cmd.Context(), os.Stdin, params)
+			return login.Run(cmd.Context(), os.Stdout, params)
 		},
 	}
 )
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -1,14 +1,40 @@
 package cmd
 
 import (
+	"bytes"
+	"errors"
 	"fmt"
+	"io"
 	"os"
+	"os/user"
+	"strings"
+	"time"
 
+	"github.com/mattn/go-isatty"
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 	"github.com/supabase/cli/internal/login"
 	"github.com/supabase/cli/internal/utils"
+	"golang.org/x/term"
 )
+
+var (
+	ErrMissingToken = errors.New("Cannot use automatic login flow inside non-TTY environments. Please provide " + utils.Aqua("--token") + " flag or set the " + utils.Aqua("SUPABASE_ACCESS_TOKEN") + " environment variable.")
+)
+
+func generateTokenName() (string, error) {
+	user, err := user.Current()
+	if err != nil {
+		return "", fmt.Errorf("cannot retrieve username: %w", err)
+	}
+
+	hostname, err := os.Hostname()
+	if err != nil {
+		return "", fmt.Errorf("cannot retrieve hostname: %w", err)
+	}
+
+	return fmt.Sprintf("cli_%s@%s_%d", user.Username, hostname, time.Now().Unix()), nil
+}
 
 var (
 	loginCmd = &cobra.Command{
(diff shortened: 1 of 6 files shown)


internal/login/login.go
```diff
@@ -1,18 +1,185 @@
...
+	createLoginSessionQuery := "?session_id=" + sessionId + "&token_name=" + tokenName + "&public_key=" + encodedPublicKey
+	createLoginSessionUrl := utils.GetSupabaseDashboardURL() + createLoginSessionPath + createLoginSessionQuery
+
+	if params.OpenBrowser {
+		fmt.Printf("Here is your login link in case browser did not open: %s\n\n", utils.Bold(createLoginSessionUrl))
+	} else {
+		fmt.Printf("Here is your login link: %s\n\n", utils.Bold(createLoginSessionUrl))
+	}
+
+	spinner := spinner.New(spinner.CharSets[11], 100*time.Millisecond)
```
sweatybridge, line 154: We have been using bubbletea/spinner for existing commands like [supabase stop](https://github.com/supabase/cli/blob/main/internal/stop/stop.go#L26). Could you try if that works?

internal/login/login.go
```diff
@@ -1,18 +1,185 @@
...
+	}
+
+	return string(decryptedAccessToken), nil
+}
+
+func pollForAccessToken(url string) (AccessTokenResponse, error) {
+	var accessTokenResponse AccessTokenResponse
+
+	// We fully control the url here, so it's safe to perform the request and ignore the G107 gosec rule.
+	resp, err := http.Get(url) //nolint:gosec
```
sweatybridge, line 92: Once backend moves to [v1 route](https://github.com/supabase/infrastructure/pull/15529/files#r1381779789), it would be nice to replace this with the http client generated from openapi spec.

internal/login/login.go
```diff
@@ -1,18 +1,185 @@
...
 
+	"github.com/briandowns/spinner"
+	"github.com/google/uuid"
 	"github.com/spf13/afero"
 	"github.com/supabase/cli/internal/utils"
 	"github.com/supabase/cli/internal/utils/credentials"
 )
 
-func Run(stdin *os.File, fsys afero.Fs) error {
-	accessToken := PromptAccessToken(stdin)
```
sweatybridge, line 14: Are we going to stop prompting for access token completely? This could be a breaking change for users who pipe in tokens from CI environments, for eg. `supabase login <<< "token"`.

This is probably a rare case but I'd prefer keeping backwards compatibility if we can.

cmd/login.go
```diff
@@ -3,27 +3,77 @@ package cmd
...
+			}
+
+			return login.Run(os.Stdin, params)
 		},
 	}
 )
 
 func init() {
+	loginFlags := loginCmd.Flags()
+	loginFlags.String("token", "", "Use provided token instead of automatic login flow")
```
sweatybridge, line 75: Passing in the token via cli flag seems insecure because it could be saved in shell history. Shall we only support `SUPABASE_ACCESS_TOKEN` env var?

internal/login/login.go
kamilogorek: Good catch, will update.

cmd/login.go
kamilogorek: I used `--auth-token` flags in the past extensively and not a single person reported this as an issue tbh

internal/login/login.go
kamilogorek: Updated. Although it's now quite obvious why I went with other library for such a simple UI thing :p

https://github.com/supabase/cli/pull/1630/commits/127b9e2723f7294a69e7a4db1892203639155ff3#diff-63f7df4adcd885e26af5017d4653902be40956c5d99878eb9795b1c406b16d77

kamilogorek: Tbh we can also remove it entirely, or redo `RunProgram` to return the value instead of going through channels

internal/login/login.go
```diff
@@ -1,18 +1,225 @@
...
+		decryptedAccessToken, err := decryptAccessToken(accessTokenResponse, curve, privateKey)
+		if err != nil {
+			return err
+		}
+
+		fmt.Printf("Token created successfully: %s\n\n", utils.Bold(tokenName))
+		fmt.Println(loggedInMsg)
+
+		return utils.SaveAccessToken(decryptedAccessToken, params.Fsys)
+	}
```
sweatybridge, line 222:
```suggestion
	return utils.RunProgram(ctx, func(p utils.Program, ctx context.Context) {
		p.Send(utils.StatusMessage("Your token is now being generated and securely encrypted. Waiting for it to arrive..."))

		sessionPollingUrl := utils.GetSupabaseAPIHost() + "/platform/cli/login/" + sessionId
		accessTokenResponse, err := pollForAccessToken(sessionPollingUrl)
		if err != nil {
			return err
		}
		decryptedAccessToken, err := decryptAccessToken(accessTokenResponse, curve, privateKey)
		if err != nil {
			return err
		}

		fmt.Printf("Token created successfully: %s\n\n", utils.Bold(tokenName))
		fmt.Println(loggedInMsg)

		return utils.SaveAccessToken(decryptedAccessToken, params.Fsys)
	})
```

I think something like this would work? Not much changes given our current refactoring.

internal/login/login.go
```diff
@@ -1,18 +1,225 @@
...
+	encodedPublicKey := hex.EncodeToString(publicKey.Bytes())
+
+	createLoginSessionPath := "/cli/login"
+	createLoginSessionQuery := "?session_id=" + sessionId + "&token_name=" + tokenName + "&public_key=" + encodedPublicKey
+	createLoginSessionUrl := utils.GetSupabaseDashboardURL() + createLoginSessionPath + createLoginSessionQuery
+
+	if params.OpenBrowser {
+		fmt.Printf("Here is your login link in case browser did not open: %s\n\n", utils.Bold(createLoginSessionUrl))
+
+		openCmd := exec.Command("open", createLoginSessionUrl)
```
sweatybridge, line 178:
```suggestion
		openCmd := exec.CommandContext(ctx,
"open", createLoginSessionUrl)
```

Use context to allow cancellation.

internal/login/login.go
kamilogorek: Perfection :chefs_kiss:

kamilogorek: Had to wait for program to end first, otherwise prints overlap, as bubbletea controls terminal state inside the program.

internal/login/login.go
```diff
@@ -1,18 +1,193 @@
...
+
+	if params.OpenBrowser {
+		fmt.Print("Hello from ", utils.Aqua("Supabase"), "! Press ", utils.Aqua("Enter"), " to open browser and login automatically.\n")
+		fmt.Scanln()
+	}
+
+	sessionId := uuid.New().String()
+	tokenName := params.Name
+	curve := ecdh.P256()
+	privateKey, err := curve.GenerateKey(rand.Reader)
```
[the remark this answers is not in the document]
kamilogorek: I talked with Stojan about it, and he said that it’s best to just let the session expire on the backend and force user to retry. It should rarely happen, and it will work just fine if they retry the command.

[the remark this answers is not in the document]
sweatybridge: Ok, I haven't read the dashboard code. How do we deal with dangling access tokens that were left behind from cancelled or interrupted login sessions?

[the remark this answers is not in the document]
kamilogorek: They stay assigned to your account and you can delete them manually.
I'll add a note about where they are stored alongside the link to settings page from the "success" login page, good idea.

kamilogorek:
Let's merge and wait for API changes deployment, and once it's done, I'll confirm that the CLI is working as expected, and we'll be able to throw it into the beta release newt.

sweatybridge:
Our beta release happens on merge. So the order should be first deploy api changes to staging, test with cli, then merge to cli.

kamilogorek:
That was exactly my thinking, bad wording, sorry :D

sweatybridge (approved):

internal/login/login.go
```diff
@@ -1,18 +1,193 @@
...
+
+	decryptedAccessToken, err := aesgcm.Open(nil, nonce, decodedAccessToken, nil)
+	if err != nil {
+		return "", fmt.Errorf("%s: %w", decryptionErrorMsg, err)
+	}
+
+	return string(decryptedAccessToken), nil
+}
+
+func pollForAccessToken(url string) (AccessTokenResponse, error) {
```
sweatybridge, line 90: Let's add more unit tests for this. It would help with future refactoring if we change the http client.

You can check the uncovered lines here https://coveralls.io/builds/63814279/source?filename=internal%2Flogin%2Flogin.go#L90

kamilogorek:
Added one test that covers the "happy path", as it's quite complicated to test this whole feature in the first place. Had to jump through some hoops 👀 
Also fixed broken `sigint` handling when spinner was in its pending state.

Issue #1627, opened 2023-10-31
Add support for SUPABASE_PROJECT_ID env variable in place of --project-ref flag


martypdx:
**Describe the bug**
According to the [docs](https://supabase.com/docs/guides/cli/managing-environments#configure-github-actions), the cli should be able to use the following env vars:

- SUPABASE_ACCESS_TOKEN is your personal access token
- SUPABASE_DB_PASSWORD is your project specific database password
- SUPABASE_PROJECT_ID is your project specific reference string

The first two work fine, but cli will fail if `--project-ref <project-id>` is not included. Tried SUPABASE_PROJECT_ID, PROJECT_ID, SUPABASE_PROJECT_REF

**To Reproduce**
1. Create an `.env` file or other means to pass env variables to a supabase cli command
2. Supabase should not be linked
3. Run a command that needs `--project-ref`
5. Fails with error like "Cannot find project ref. Have you run supabase link?"

**Expected behavior**
The cli uses the provided `SUPABASE_PROJECT_ID` env variable value when attempting the supabase cli command

**Desktop (please complete the following information):**
 - OS: MacOS
 - Version of CLI: 1.106.1
 - Version of Node.js: 21.1

kamilogorek:
This is somewhat of a wording issue in the docs. If you see the usage example, it's explicitly using those variables, that's why docs state you need to set them first before using examples, see: https://github.com/supabase/supabase-action-example/blob/d201824df60e6ee22c8d72788296e5cb919a2ee5/.github/workflows/production.yaml#L25C45-L25C45

CLI is not using them right now, but I agree it would be a nice touch.

martypdx:
I opened a [PR on docs](https://github.com/supabase/supabase/pull/18735) for this.

@kamilogorek The issue was never about the examples or whether I was able to get it to work. Call me old-fashion, but getting it to work on my machine is hardly the end of the sw dev process. Building out and refactoring the full pipeline for the project, it added a few extra hours spent hours tracking this down amongst the other factors like GH secrets or vars.

Issue #878, opened 2023-02-20
Supabase link should have better UX when --project-ref isn't provided


tmcw:
# Bug report

## Describe the bug

You look for supabase commands:

```
val.town git:(main) ✗ supabase
Supabase CLI 1.38.5

Usage:
  supabase [command]

Local Development:
  db                   Manage local Postgres databases
  gen                  Run code generation tools
  init                 Initialize a local project
  link                 Link to a Supabase project
```

You run supabase link, and it asks for a database password:

```
val.town git:(main) ✗ supabase link
Enter your database password:
```

You provide that password, and you get this error:

```
Error: Invalid project ref format. Must be like `abcdefghijklmnopqrst`.
```

What's a project ref? Where is it specified? How is it invalid, and what's its current value? The answer is that you need to do:

```
supabase link --project-ref=xxx
```

There is no suggestion that you need to do this and the error message does not guide you.

## Expected behavior

If `--project-ref` is a required parameter, which it appears to be, there should be an error message showing that it's missing and saying to provide it as `--project-ref`, rather than an error about it being in the wrong format.

- Supabase CLI version 1.38.5

sweatybridge:
Thank you for reporting this issue. It was introduced by a previous refactor which messed up the order of project-ref arg validation. I have addressed it in the linked PR.

neuquen:
It is still not clear to me what the project-ref is supposed to contain.  Can that be clarified here and updated in the documentation?

neuquen:
It is still not clear to me what the project-ref is supposed to contain. Can that be clarified here and updated in the documentation?

UPDATE: I found example usage [here](https://github.com/supabase/setup-cli#advanced-usage) which is great, but I still think the help text needs added clarification.

Jonnotie:
I just ran into this issue as well

kkukshtel:
I also missed this — the flag should be changed to project-id to match that it's the project ID, and more explicitly called out that it's just the string of letters in the project url on supabase (no idea if it's available elsewhere).

landsman:
> I also missed this — the flag should be changed to project-id to match that it's the project ID, and more explicitly called out that it's just the string of letters in the project url on supabase (no idea if it's available elsewhere).

exactly

sweatybridge:
In the beta release, we added an interactive prompt if project-ref flag is not specified when running supabase link.

Carles11:
Yes, very unclear what is expected. This is how it worked for me:
[...]

PR #1633, opened 2023-11-03
feat: support project id via env var

sweatybridge:

feature
closes https://github.com/supabase/cli/issues/1627
closes https://github.com/supabase/cli/issues/878

## What is the new behavior?

Commands which support `--project-ref` flag can now specify `SUPABASE_PROJECT_ID` env var as well. Env var takes lower precedence than flag but higher than linked file.

## Additional context

### DIFF
--- a/cmd/link.go
+++ b/cmd/link.go
@@ -1,6 +1,9 @@
 package cmd
 
 import (
+	"context"
+	"errors"
+	"fmt"
 	"os"
 	"os/signal"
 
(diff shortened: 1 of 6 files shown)


kamilogorek (changes requested):
Needs additional update for `LoadProjectRef`.

kamilogorek (approved):
`LoadProjectRef` is part of commands that utilize linking and we can skip it for now.

PR #1634, opened 2023-11-03
ref: Provide better missing project id error message

kamilogorek:
Fixes https://github.com/supabase/cli/issues/878
Depends on https://github.com/supabase/cli/pull/1633

### DIFF
--- a/internal/utils/misc.go
+++ b/internal/utils/misc.go
@@ -180,7 +180,7 @@ var (
 	CustomRolesPath       = filepath.Join(SupabaseDirPath, "roles.sql")
 
 	ErrNotLinked  = errors.New("Cannot find project ref. Have you run " + Aqua("supabase link") + "?")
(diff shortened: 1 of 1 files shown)


internal/utils/misc.go
```diff
@@ -180,7 +180,7 @@ var (
 	CustomRolesPath       = filepath.Join(SupabaseDirPath, "roles.sql")
 
 	ErrNotLinked  = errors.New("Cannot find project ref. Have you run " + Aqua("supabase link") + "?")
-	ErrInvalidRef = errors.New("Invalid project ref format. Must be like `abcdefghijklmnopqrst`.")
+	ErrInvalidRef = errors.New("Missing or invalid project ID. Use " + Aqua("--project-ref") + "or" + Aqua("SUPABASE_PROJECT_ID") + "env variable to set it. Project ID can be found on project settings page.")
```
sweatybridge, line 183: I think it's better to create a new error for missing ref. We can update `AssertProjectRefIsValid` method to check for empty string.

PR #1654, opened 2023-11-13
fix: adding gotrue expected format to GOTRUE_SMS_TEST_OTP env

LautaroJayat:
```
# original commit
fix: adding gotrue expected format to testotp env

* Removing previous json encoding
* Adding a new method that outputs map as key1:value1,key2:value2,...
* Adding new test for the method above
```

This PR aims to address what has been pointed in https://github.com/supabase/gotrue/issues/1293.
In short:
* the cli was encoding a `map[string]string` as a json string
* this string was the value for a env that gotrue was trying to use to generate the same map
* the library gotrue is using expected `key1:value1,key2:value2,...` instead of json

The current behavior generates a json string as the value for the `GOTRUE_SMS_TEST_OTP` environment variable that gotrue is supposed to use.

To attest this:
```bash
# init the configs
go run . init

# then add some key-values for the test otp in config.toml file
# [auth.sms.test_otp]
# 12345678 = "654987"
# 123456 = "6554987"
# 136789 = "665454987"
# 236789 = "65495487"

# now start as usual
go run . start

# store gotrue container id
export CONTAINER=$(docker container ls | grep gotrue | awk '{print $1}')

# check what is in the env
docker container exec $CONTAINER printenv | grep GOTRUE_SMS_TEST_OTP=

``` 
example output:
Screenshot from 2023-11-12 21-03-26

## What is the new behavior?
In the new behavior the string is formated as `key1:value1,key2:value2,...`

if we follow the same steps we should expect the following:
Screenshot from 2023-11-12 21-08-43

### DIFF
--- a/internal/start/start.go
+++ b/internal/start/start.go
@@ -369,7 +369,7 @@ EOF
 	if utils.Config.Auth.Enabled && !isContainerExcluded(utils.Config.Auth.Image, excluded) {
 		var testOTP bytes.Buffer
 		if len(utils.Config.Auth.Sms.TestOTP) > 0 {
-			formatMapForEnvconfig(utils.Config.Auth.Sms.TestOTP, &testOTP)
+			formatMapForEnvConfig(utils.Config.Auth.Sms.TestOTP, &testOTP)
 		}
 
 		env := []string{
@@ -415,7 +415,7 @@ EOF
 			"GOTRUE_SMS_OTP_EXP=6000",
 			"GOTRUE_SMS_OTP_LENGTH=6",
 			"GOTRUE_SMS_TEMPLATE=Your code is {{ .Code }}",
-			fmt.Sprintf("GOTRUE_SMS_TEST_OTP=%s", testOTP.String()),
+			"GOTRUE_SMS_TEST_OTP=" + testOTP.String(),
 
 			fmt.Sprintf("GOTRUE_SECURITY_REFRESH_TOKEN_ROTATION_ENABLED=%v", utils.Config.Auth.EnableRefreshTokenRotation),
 			fmt.Sprintf("GOTRUE_SECURITY_REFRESH_TOKEN_REUSE_INTERVAL=%v", utils.Config.Auth.RefreshTokenReuseInterval),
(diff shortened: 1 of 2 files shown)


LautaroJayat:
@J0
May I have your eye on this one?

internal/start/start.go
```diff
@@ -906,3 +903,17 @@ func ExcludableContainers() []string {
 	}
 	return names
 }
+
+func formatMapForEnvconfig(input map[string]string, output *bytes.Buffer) {
```
sweatybridge, line 907:
```suggestion
func formatMapForEnvConfig(input map[string]string, output *bytes.Buffer) {
```
 Nitpick

internal/start/start.go
```diff
@@ -418,7 +415,7 @@ EOF
 			"GOTRUE_SMS_OTP_EXP=6000",
 			"GOTRUE_SMS_OTP_LENGTH=6",
 			"GOTRUE_SMS_TEMPLATE=Your code is {{ .Code }}",
-			"GOTRUE_SMS_TEST_OTP=" + testOTP.String(),
+			fmt.Sprintf("GOTRUE_SMS_TEST_OTP=%s", testOTP.String()),
```
sweatybridge, line 418: Why do we need to change this line?

sweatybridge (approved):
Looks good. Just a few nitpicks

internal/start/start.go
LautaroJayat: Hi! @sweatybridge
I'm trying to see the difference between the line in the PR and the suggested change, but I cant find it.

internal/start/start.go
LautaroJayat: @sweatybridge
Oh, my bad. I think deleted that while testing, and then re wrote it following the same pattern used for other envs.
Shall I change it and left it like before?

internal/start/start.go
kamilogorek: env **C** onfig - capitalized `C` :)

LautaroJayat: It seems I need glasses, lol.

LautaroJayat: @sweatybridge @kamilogorek
I've just changed the function name and replaced all occurrences.

internal/start/start.go
LautaroJayat: @sweatybridge 
I have just went back to the previous implementation

PR #1655, opened 2023-11-13
feat: use sms template from config.toml

jibin2706:

Use SMS template from `config.toml` instead of hardcoded string

Uses the hardcoded template for sending SMS otp

## What is the new behavior?

Uses template from env for sending SMS OTP message

## Additional context

gotrue repo already has the env, let me know if any other changes are required
https://github.com/supabase/gotrue/blob/93e5f82ced83c08799ce99020be9dea82fc56d24/example.env#L183

### DIFF
--- a/internal/utils/config.go
+++ b/internal/utils/config.go
@@ -285,6 +285,7 @@ type (
 	sms struct {
 		EnableSignup        bool              `toml:"enable_signup"`
 		EnableConfirmations bool              `toml:"enable_confirmations"`
+		Template            string            `toml:"template"`
 		Twilio              twilioConfig      `toml:"twilio" mapstructure:"twilio"`
 		TwilioVerify        twilioConfig      `toml:"twilio_verify" mapstructure:"twilio_verify"`
 		Messagebird         messagebirdConfig `toml:"messagebird" mapstructure:"messagebird"`
--- a/internal/utils/config.go
+++ b/internal/utils/config.go
@@ -130,6 +130,9 @@ var Config = config{
 				"email_change": {},
 			},
 		},
+		Sms: sms{
+			Template: "Your code is {{ .Code }} .",
+		},
 		External: map[string]provider{
 			"apple":     {},
 			"azure":     {},
--- a/internal/utils/templates/init_config.test.toml
+++ b/internal/utils/templates/init_config.test.toml
@@ -99,6 +99,8 @@ content_path = "./supabase/templates/invite.html"
 enable_signup = true
 # If enabled, users need to confirm their phone number before signing in.
 enable_confirmations = false
+# Template for sending OTP to users
+template = "Your code is {{ .Code }} ."
 
 # Use pre-defined map of phone number to OTP for testing.
 [auth.sms.test_otp]
--- a/internal/utils/templates/init_config.test.toml
+++ b/internal/utils/templates/init_config.test.toml
@@ -100,7 +100,7 @@ enable_signup = true
 # If enabled, users need to confirm their phone number before signing in.
 enable_confirmations = false
 # Template for sending OTP to users
-template = "Your code is {{ `{{.Code }}` }} ."  
+template = "Your code is {{ `{{ .Code }}` }} ."  
 
 # Use pre-defined map of phone number to OTP for testing.
 [auth.sms.test_otp]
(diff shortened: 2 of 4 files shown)


hf (approved):
Looks good.

internal/utils/templates/init_config.test.toml
```diff
@@ -99,6 +99,8 @@ content_path = "./supabase/templates/invite.html"
 enable_signup = true
 # If enabled, users need to confirm their phone number before signing in.
 enable_confirmations = false
+# Template for sending OTP to users
+template = "Your code is {{ .Code }} ."
```
LautaroJayat, line 103: **suggestion:**
As this is parsed as if it was an actual go template, it is failing because it can't find the reference to ".Code".
I would change this to 
```toml
template = "Your code is {{ `{{.Code }}` }} ."
```
So the output in `config.yaml` is 
```toml
template = "Your code is {{ `{{.Code }}` }} ."
```

internal/utils/templates/init_config.toml
```diff
@@ -99,6 +99,8 @@ enable_confirmations = false
 enable_signup = true
 # If enabled, users need to confirm their phone number before signing in.
 enable_confirmations = false
+# Template for sending OTP to users
+template = "Your code is {{ .Code }} ."
```
LautaroJayat, line 103: **suggestion**:
As this is parsed as if it was an actual go template, it is failing because it can't find the reference to ".Code".
I would change this to

template = "Your code is {{ `{{.Code }}` }} ."
So the output in config.yaml is

template = "Your code is {{ `{{.Code }}` }} ."

LautaroJayat (commented):
I think the pipeline is failing because when generating `config.toml` files, the `init_config` templates are parsed as go templates, but there is nothing to replace `{{.Code}}`.
As we want to output an actual template, we should look for a strategy to "escape" what shouldn't be interpolated with something else.

internal/utils/templates/init_config.toml
jibin2706: Made the changes, not sure how I can test it out

LautaroJayat: @jibin2706 
To begin with, the command `go run . init` wont fail.
Then, to manually test the deployment, after `go run . start` you can do something like:

```bash
# store gotrue container id
export CONTAINER=$(docker container ls | grep gotrue | awk '{print $1}')

# check what is in the env
docker container exec $CONTAINER printenv | grep GOTRUE_SMS_TEMPLATE=
```
That will output the environment that was injected into the container. You can then check if the value is the one you are expecting.

jibin2706: thanks @LautaroJayat, working as expected

internal/utils/config.go
```diff
@@ -285,6 +285,7 @@ type (
 	sms struct {
 		EnableSignup        bool              `toml:"enable_signup"`
 		EnableConfirmations bool              `toml:"enable_confirmations"`
+		Template            string            `toml:"template"`
```
sweatybridge, line 288: We should ensure backwards compatibility with existing config.toml which doesn't define this field. One way is to add a default value after line 122.

sweatybridge (changes requested):

internal/utils/config.go
jibin2706: done @sweatybridge

internal/utils/templates/init_config.test.toml
```diff
@@ -99,6 +99,8 @@ content_path = "./supabase/templates/invite.html"
 enable_signup = true
 # If enabled, users need to confirm their phone number before signing in.
 enable_confirmations = false
+# Template for sending OTP to users
+template = "Your code is {{ `{{.Code }}` }} ."
```
sweatybridge, line 103:
```suggestion
template = "Your code is {{ `{{ .Code }}` }} ."  
```

internal/utils/templates/init_config.toml
```diff
@@ -99,6 +99,8 @@ enable_confirmations = false
 enable_signup = true
 # If enabled, users need to confirm their phone number before signing in.
 enable_confirmations = false
+# Template for sending OTP to users
+template = "Your code is {{ `{{.Code }}` }} ."
```
sweatybridge, line 103:
```suggestion
template = "Your code is {{ `{{ .Code }}` }} ."  
```

sweatybridge (approved):
Thanks, just a nitpick on spacing. The rest looks good!

PR #1661, opened 2023-11-15
chore: update templates to quote test_otp

J0:

Test otp phone number needs to be in quotes 

Relevant issue: https://github.com/supabase/gotrue/issues/1293


J0:
on hold, let me confirm why it's not working with double quotes first..

Issue #1688, opened 2023-11-27
How to signout from CLI


iammohan01:
How can i signout supabse account from cli ?

kamilogorek:
Remove access token from the default keyring of your operating system.
For OSX it's Keychain, for Windows it's "Accessing Credential Manager" and most linux distros have their own "Passwords and Keys" apps. See https://pkg.go.dev/github.com/zalando/go-keyring@v0.2.3#section-readme for implementation details.

If not supported, it fallbacks to `$HOME/.supabase/access-token` file.

iammohan01:
Thanks man, fixed. but why don't we have command like `supabase logout` ?

kamilogorek:
That sounds like a good addition. PRs welcome :)

iammohan01:
@kamilogorek Are we storing anything other tha access token in this diretory `~/.supabase/access-token` ?

iammohan01:
@kamilogorek is there anything contains other than access token in this file `~/.supabase/access-token` ?

Is it safe to delete this file while logout ?

kamilogorek:
Good catch, thanks! Fixed in #1700 

> @kamilogorek is there anything contains other than access token in this file ~/.supabase/access-token ?
> Is it safe to delete this file while logout ?

No, just a token. And in keyring its scoped to a predefined key, so you don't need to do any existence checks if someone calls logout multiple times (other than standard go errors for removing missing file).

PR #1696, opened 2023-11-29
feat: support logout cmd

S96EA:

feature

[1688](https://github.com/supabase/cli/issues/1688)


internal/logout/logout.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+type RunParams struct {
+	Fsys afero.Fs
+}
+
+var loggedOutMsg = "You are now logged out."
+
+func Run(ctx context.Context, stdout *os.File, params RunParams) error {
+	err := utils.RunProgram(ctx, func(p utils.Program, ctx context.Context) error {
```
kamilogorek, line 19: This operation should be instantaneous, I don't think we need to wrap it inside a separate program.

internal/logout/logout.go
```diff
@@ -0,0 +1,33 @@
...
+
+		return utils.DeleteAccessToken(params.Fsys)
+	})
+
+	if err != nil {
+		return err
+	}
+
+	fmt.Fprintf(stdout, "Token deleted successfully.\n\n")
+	fmt.Fprintln(stdout, loggedOutMsg)
```
kamilogorek, line 30:
```suggestion
	fmt.Fprintln(stdout, "Token deleted successfully. You are now logged out.")
```

This should be sufficient.

internal/logout/logout.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"github.com/spf13/afero"
+	"github.com/supabase/cli/internal/utils"
+)
+
+type RunParams struct {
+	Fsys afero.Fs
+}
+
+var loggedOutMsg = "You are now logged out."
```
kamilogorek, line 16:
```suggestion
```

internal/logout/logout.go
kamilogorek, line 19: I'd add a prompt asking users to confirm the logout saying that it will remove the token from their system.

S96EA: Yes. Thanks. I remove the "RunProgram" wrap.

internal/logout/logout.go
S96EA: Thanks~

internal/logout/logout.go
S96EA: Yes. Thanks. I added a PreRun stage to ask users.

internal/logout/logout.go
S96EA: Thanks~

internal/utils/access_token.go
```diff
@@ -84,6 +84,23 @@ func fallbackSaveToken(accessToken string, fsys afero.Fs) error {
...
+	// Delete from native credentials store
+	if err := credentials.Delete(AccessTokenKey); err == nil {
+		return nil
+	}
+	// Fallback to token file
+	return fallbackDeleteToken(fsys)
+}
+
+func fallbackDeleteToken(fsys afero.Fs) error {
+	path, err := getAccessTokenPath()
```
kamilogorek, line 97: It should not fail when there's no access token found. When `NotFound` error is returned, `logout` command should just print `You were not logged in, nothing to do.` or similar message. Now it'll fail with:
```
remove /Users/kamilogorek/.supabase/access-token: no such file or directory
```
When called multiple times.

internal/logout/logout.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/spf13/afero"
+	"github.com/supabase/cli/internal/utils"
+)
+
+type RunParams struct {
+	Fsys afero.Fs
+}
+
+func PreRun() error {
+	if !utils.PromptYesNo("Do you want to log out, which will remove the token from your system?", true, os.Stdin) {
```
kamilogorek, line 18: Canceling removal should still be exitcode `0`, as it's terminating the program correctly. Not sure why project deletion is throwing error here.
However because `PreRun` cannot handle cancellation other than via error, you can move this check into `Run` and write

```go
if !utils.PromptYesNo("Do you want to log out, which will remove the token from your system?", true, os.Stdin) {
  fmt.Fprintln(stdout, "Not deleting token")
  return nil
}
```

internal/utils/access_token.go
S96EA: Thanks. I checked if the error returned by `Remove` is ENOENT, This is OK for OS fs implementation in afero. But I didn't find a common error or check function in afero that would express different fs NotFound error.

internal/logout/logout.go
S96EA: Thanks. Done.

kamilogorek:
Lovely, thanks @S96EA. Added few changes myself so you don't have to go back and forth anymore :)
Please apply those to this PR and we're good to go - https://gist.github.com/kamilogorek/cf2cacf7ff0cce8b91d65734df776fad

S96EA:
> Lovely, thanks @S96EA. Added few changes myself so you don't have to go back and forth anymore :) Please apply those to this PR and we're good to go - https://gist.github.com/kamilogorek/cf2cacf7ff0cce8b91d65734df776fad
> 
> (changed default to `false` as its destructive operation, fixed `actual/expected` order in tests, simplified error check and used separate error value for not logged in)

I'm sincerely appreciative of your support. Thanks again.

kamilogorek (approved):

kamilogorek:
LGTM! @sweatybridge feel free to merge anytime.

PR #1700, opened 2023-11-29
fix: Ask users to log in before using management api

kamilogorek:
`supabase login` command now accepts multiple different flags so we simply run it as before. Prompt users to call it themselves first.

Fixes https://github.com/supabase/cli/issues/1688#issuecomment-1831111540


cmd/root.go
```diff
@@ -38,9 +37,7 @@ func IsManagementAPI(cmd *cobra.Command) bool {
 
 func PromptLogin(ctx context.Context, fsys afero.Fs) error {
 	if _, err := utils.LoadAccessTokenFS(fsys); err == utils.ErrMissingToken {
-		return login.Run(ctx, os.Stdout, login.RunParams{
-			Fsys: fsys,
-		})
+		return errors.New("You need to be logged-in in order to use Management API commands. Run `supabase login` first.")
```
sweatybridge, line 40:
```suggestion
		utils.CmdSuggestion = fmt.Sprintf("Run %s first.", utils.Aqua("supabase login"))
		return errors.New("You need to be logged-in in order to use Management API commands.")
```

Just a nitpick for consistency of error message style.

sweatybridge (approved):

Issue #1716, opened 2023-12-05
Column Level Security missing in diff


anarkrypto:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Postgres grant / revoke queries missing in diff

## To Reproduce

```sql
revoke update on public.profiles from authenticated;
grant update (avatar_url, name) on public.profiles to authenticated;
```

```bash
supabase db diff
```

## Expected behavior

Column privileges queries should be present in diff.
Therefore how can I save migrations ? Is this a expected behavior ? 

## System information

- OS: Linux
- Version of supabase-js: 1.115.4

anarkrypto:
garyaustin via Supabase Discord:

> Supabase I believe uses migra as part of diff...
> I don't know if this is the right migra but this is interesting...
> https://github.com/djrobstep/migra/blob/da6671acae0b4acebdec804ae25a0c65cf8561c2/docs/options.md?plain=1#L28
> 
```

> ## `--with-privileges`
> This tells migra to spit out permission-related change statements (grant/revoke). This is False by default: Often one is comparing databases from different environments, where the users and permissions are completely different and not something one would want to sync.
```

> It is the same migra they use... so dig into if there is a way to set it's parameters...

anarkrypto:
Since --with-privileges is false by default in migra, Supabase cli should allow us to pass this parameter.

Columns privileges are very important to ensure db security when allowing authenticated users to crud the DB directly.

The lack of a way to migrate these privileges it's not just a lack of functionality, but also a security issue as Supabase strongly encourages devs to adopt end-to-end user security from the browser to the database.

sweatybridge:
This sounds like a sensible default. But I'm not sure if it will handle column privileges as the upstream PR is still open https://github.com/djrobstep/schemainspect/pull/67. We will have to merge and test it out.

anarkrypto:
> This sounds like a sensible default. But I'm not sure if it will handle column privileges as the upstream PR is still open [djrobstep/schemainspect#67](https://github.com/djrobstep/schemainspect/pull/67). We will have to test it out after the next release.

Yes, I can confirm it's not working with column privileges yet. After adding --with-privileges, migra only returns table privileges.

I believe this issue should not be marked as completed since the issue persists.

[...]

PR #1728, opened 2023-12-09
fix: always diff privileges using migra

sweatybridge:

Bug fix https://github.com/supabase/cli/issues/1716

## What is the new behavior?

## Additional context

### DIFF
--- a/internal/db/diff/templates/migra.sh
+++ b/internal/db/diff/templates/migra.sh
@@ -10,7 +10,7 @@ trap 'kill -9 %1' TERM
 run_migra() {
     # additional flags for diffing extensions
     [ "$schema" = "extensions" ] && set -- --create-extensions-only --ignore-extension-versions "$@"
-    migra --unsafe --schema="$schema" "$@"
+    migra --with-privileges --unsafe --schema="$schema" "$@"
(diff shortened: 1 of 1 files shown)


anarkrypto:
@sweatybridge 

Migra docs explains why **--with-privileges** param is **false** by default:

> "This is False by default: Often one is comparing databases from different environments, where the users and permissions are completely different and not something one would want to sync."

See: https://github.com/djrobstep/migra/blob/da6671acae0b4acebdec804ae25a0c65cf8561c2/docs/options.md?plain=1#L28

For this reason I thought it would also be better to make the parameter optional in the Supabase CLI.

Making it default true won't have any negative impact?

sweatybridge:
> Making it default true won't have any negative impact?

Yup, we have been using default for our internal database migrations for over a year and haven't found any negative impact caused by this. The missing flag was an oversight when porting over to cli.

The way permissions are managed by supabase auth schemas via a static set of roles, namely `anon`, `authenticated`, and `service_role`. The privileges of these roles should be the same across hosted and local projects. If you need them to be different between environments, RLS policies are usually an easier option as it can use plain table lookups.

Issue #1315, opened 2023-07-17
Support for Windows ARM64


StrawMonkey:
Loving Supabase so far, but we've gotten a bit stuck trying to use the supabase-cli on a Surface 9. I think ARM64 is not supported? Would be great if it was or if there was an option to force install the AMD64 version.

````
Downloading https://github.com/supabase/cli/releases/download/v1.77.9/supabase_windows_arm64.tar.gz
npm ERR! node:events:489
npm ERR!       throw er; // Unhandled 'error' event
npm ERR!       ^
npm ERR!``

sweatybridge:
Yea, we haven't been compiling for windows arm64 as no one was requesting for it. Thanks for creating this issue and we will look into it.

sweatybridge:
We've added windows arm64 build to our releases since 1.126.1. We don't have a Surface 9 to verify but feel free to reopen this issue if it doesn't work for you.

PR #1746, opened 2023-12-12
feat: `--create-ticket` flag to send error reports to sentry

sweatybridge:

feature

## What is the new behavior?

Reports error to sentry when any commands are with `--create-ticket` flag. Helps to collect diagnostics information.
- [x] server side rate limiting
- [x] add stacktrace to all errors

Going forward, any errors from external packages should be wrapped in `errors.New(err)` using [go-errors](https://github.com/go-errors/errors).

## Additional context

Eg: https://supabase.sentry.io/issues/4734608486/events/e250b9cc83c245d1aa520f1e3ece5017

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -5,7 +5,7 @@ builds:
     flags:
       - -trimpath
     ldflags:
-      - -s -w -X github.com/supabase/cli/internal/utils.Version={{.Version}}
+      - -s -w -X github.com/supabase/cli/internal/utils.Version={{.Version}} -X github.com/supabase/cli/internal/utils.SentryDsn={{ .Env.SENTRY_DSN }}
     env:
       - CGO_ENABLED=0
     targets:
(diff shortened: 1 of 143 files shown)


soedirgo:
Telemetry is a [very](https://news.ycombinator.com/item?id=34707583) [sensitive](https://news.ycombinator.com/item?id=35974653) [topic](https://news.ycombinator.com/item?id=27727150), even for anonymous info. IMO we should make this opt-in by prompting on first run instead of opt-out.

.goreleaser.yml
```diff
@@ -14,6 +14,7 @@ builds:
       - linux_amd64
       - linux_arm64
       - windows_amd64
+      - windows_arm64
```
soedirgo, line 17: I'd rather not support another platform when there's no one in the team with a Windows ARM machine. I don't see anyone requesting for it either.

soedirgo (approved):
Just a note re: Windows ARM, otherwise lgtm 👍

.goreleaser.yml
sweatybridge: This is the ticket which requests for windows arm support https://github.com/supabase/cli/issues/1315. Will move to a separate PR though.

PR #1792, opened 2023-12-29
stop infinite loop

dsl400:
INSERT INTO _analytics is causing an infinite loop when logflare uses postgres as backend

PS: this is my first interaction with go ...

Bug fix


dsl400:
The problem is still there, this solution was filtering out all the logs.

PR #1795, opened 2023-12-29
stop infinite loop 

dsl400:
INSERT INTO _analytics is causing an infinite loop when logflare uses postgres as backend

### DIFF
--- a/internal/start/templates/vector.yaml
+++ b/internal/start/templates/vector.yaml
@@ -132,7 +132,7 @@ transforms:
     inputs:
       - router.db
     source: |-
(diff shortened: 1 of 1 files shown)


sweatybridge:
@Ziinc could you take a closer look at this to see if a logflare version bump is needed?

Ziinc:
cannot be replicated on `main`.

@dsl400 Are you enabling audit logs globally? If that is the case then it is expected to view these logs and the resultant loop. Suppressing these is not a good idea as they represent active transactions taking place on the db. 

Using session/object audit logging is advisable in such situations

Closing

PR #1804, opened 2024-01-02
feat: update auth version to v2.130.0

J0:

With the latest deploy, GoTrue is now on v2.130.0 

Looks like no sql statements are involved. Diff: https://github.com/supabase/gotrue/compare/v2.125.1...v2.130.0


sweatybridge (approved):
thanks for checking the sql statements

J0:
Reverting to draft, need to check the container name since we renamed to Auth

J0:
Ok it [seems fine](https://hub.docker.com/r/supabase/gotrue/tags) as expected

PR #1817, opened 2024-01-05
feat(typegen): support `--postgrest-v9-compat` for `--db-url`

soedirgo:

feature

When running `supabase gen types typescript` with `--db-url`, we can't check which PostgREST version the client lib will query against, and unfortunately PostgREST v10 has a breaking change which changes the format of a nested table for one-to-one relationships from `T[]` to `T | null`. The generated types can only support one or the other.

For the hosted platform and local dev this isn't an issue since we can detect the PostgREST version and switch between the 2 formats accordingly, but for `--db-url` we need this to be configurable.

## What is the new behavior?

Add `--postgrest-v9-compat` flag, which can only be used in conjunction with `--db-url`.

### DIFF
--- a/cmd/gen.go
+++ b/cmd/gen.go
@@ -56,23 +56,30 @@ var (
 		Short: "Generate types from Postgres schema",
 	}
 
-	local     bool
-	linked    bool
-	projectId string
-	dbUrl     string
-	schemas   []string
+	local             bool
+	linked            bool
+	projectId         string
+	dbUrl             string
+	schemas           []string
+	postgrestV9Compat bool
 
 	genTypesTypescriptCmd = &cobra.Command{
 		Use:   "typescript",
 		Short: "Generate types for TypeScript",
 		Long:  "Generate types for TypeScript. Must specify one of --local, --linked, --project-id, or --db-url",
+		PreRunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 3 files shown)


kevcodez (approved):

cmd/gen.go
```diff
@@ -88,6 +95,7 @@ func init() {
 	genFlags.StringVar(&projectId, "project-id", "", "Generate types from a project ID.")
 	genFlags.StringVar(&dbUrl, "db-url", "", "Generate types from a database url.")
 	genFlags.StringArrayVar(&schemas, "schema", []string{}, "Schemas to generate types for.")
+	genFlags.BoolVar(&detectO2oRelationships, "detect-o2o-relationships", true, "Detect one-to-one relationships compatible with PostgREST v10+. Only use in together with --db-url.")
```
sweatybridge, line 98: Shall we rename this to `--for-PostgREST-9`? I think it's more intuitive flag name because users may not be fully aware of the differences in features supported but they should know the major version of their self-hosted PostgREST.

Doing so also leaves us some room to add other backwards compatibility switches in the future under that flag. 

Note that the book condition is inverted if we rename like above, ie. default value is false.

soedirgo:
Updated to use `--postgrest-v9-compat`

PR #1856, opened 2024-01-17
feat: add auth hooks config

J0:

Add configuration for auth hooks so that developers can test their hooks locally. Key fields for each hook are:

Enabled - true/false
URI - this should be in the form `pg-functions://<database>/<schema>/<function-name>`

This was tested by locally building the cli `go build .` and then `../cli/init` + `../cli/start`. Check that Docker has `GOTRUE_HOOK_<HOOK_NAME>_ENABLED`


internal/utils/config.go
```diff
@@ -347,6 +349,17 @@ type (
 		TestOTP             map[string]string `toml:"test_otp"`
 	}
 
+	hook struct {
+		MFAVerificationAttempt      extensibilityPointConfig `toml:"mfa_verification_attempt" mapstructure:"mfa_verification_attempt"`
+		PasswordVerificationAttempt extensibilityPointConfig `toml:"password_verification_attempt" mapstructure:"password_verification_attempt"`
+		CustomAccessToken           extensibilityPointConfig `toml:"custom_access_token" mapstructure:"custom_access_token"`
+	}
+
+	extensibilityPointConfig struct {
```
sweatybridge, line 358: I'm tempted to name this struct `hookConfig` for simplicity. Wdyt?

internal/utils/config.go
```diff
@@ -311,6 +312,7 @@ type (
 		EnableRefreshTokenRotation bool `toml:"enable_refresh_token_rotation"`
 		RefreshTokenReuseInterval  uint `toml:"refresh_token_reuse_interval"`
 		EnableManualLinking        bool `toml:"enable_manual_linking"`
+		Hook                       hook `toml:"hook" mapstructure:"hook"`
```
sweatybridge, line 315:
```suggestion
		Hook                       hook `toml:"hook"`
```

internal/utils/config.go
```diff
@@ -347,6 +349,17 @@ type (
 		TestOTP             map[string]string `toml:"test_otp"`
 	}
 
+	hook struct {
+		MFAVerificationAttempt      extensibilityPointConfig `toml:"mfa_verification_attempt" mapstructure:"mfa_verification_attempt"`
+		PasswordVerificationAttempt extensibilityPointConfig `toml:"password_verification_attempt" mapstructure:"password_verification_attempt"`
+		CustomAccessToken           extensibilityPointConfig `toml:"custom_access_token" mapstructure:"custom_access_token"`
```
sweatybridge, line 355:
```suggestion
		MFAVerificationAttempt      extensibilityPointConfig `toml:"mfa_verification_attempt"`
		PasswordVerificationAttempt extensibilityPointConfig `toml:"password_verification_attempt"`
		CustomAccessToken           extensibilityPointConfig `toml:"custom_access_token"`
```

No need to use mapstructure. See comment here https://github.com/supabase/cli/blob/main/internal/utils/config.go#L230

internal/utils/config.go
J0: sounds good

internal/utils/templates/init_config.toml
```diff
@@ -110,6 +110,11 @@ template = "Your code is {{ `{{ .Code }}` }} ."
 [auth.sms.test_otp]
 # 4152127777 = "123456"
 
+[auth.hook.custom_access_token]
```
sweatybridge, line 113: Can we also add a comment describing this field?

sweatybridge (approved):

internal/utils/templates/init_config.toml
J0: Good catch, thanks!

PR #1858, opened 2024-01-18
fix: Remove GOTRUE_EXTERNAL_LINKEDIN_* and add GOTRUE_EXTERNAL_LINKEDIN_OIDC_*

J0:

Currently, when a container is started, the variable `GOTRUE_EXTERNAL_LINKEDIN_*` is created. With the introduction of [LinkedIn OIDC provider](https://github.com/supabase/gotrue/pull/1248/files#diff-4c28cb40881781a1067b3b3681c43f805dab629f31f3c7614b0f781ffa096505L198) this should be `GOTRUE_EXTERNAL_LINKEDIN_OIDC_*`


internal/utils/config.go
```diff
@@ -181,23 +181,23 @@ var Config = config{
...
-			"linkedin":  {},
-			"notion":    {},
-			"twitch":    {},
-			"twitter":   {},
-			"slack":     {},
-			"spotify":   {},
-			"workos":    {},
-			"zoom":      {},
+			"apple":         {},
+			"azure":         {},
```
J0, line 185: Rest of the changes result from runnning `gofmt`

sweatybridge:
Do we need to maintain backwards compatibility with older versions of gotrue?

J0:
Oh probably not since we no longer [expose the config on the dashboard](https://github.com/supabase/supabase/pull/20510). We can leave it there for good measure though

internal/utils/config.go
```diff
@@ -181,23 +181,23 @@ var Config = config{
...
+			"bitbucket":     {},
+			"discord":       {},
+			"facebook":      {},
+			"github":        {},
+			"gitlab":        {},
+			"google":        {},
+			"keycloak":      {},
+			"linkedin_oidc": {},
```
sweatybridge, line 193:
```suggestion
			"linkedin":          {}, // TODO:
			"linkedin_oidc": {},
```

Technically this is a breaking change for config.toml that already uses `linkedin` provider. Let's keep the old field but make it a noop so we don't break users existing ci pipelines.

sweatybridge (approved):

PR #1890, opened 2024-01-28
feat: add pg-schema as a new diff tool

sweatybridge:

PoC https://github.com/stripe/pg-schema-diff/issues/85

## What is the new behavior?

```bash
$ supabase db diff --schema public --use-pg-schema
Diffing schemas: public
Creating shadow database...
Setting up initial schema...
Creating custom roles supabase/roles.sql...
Finished supabase db diff on branch pg-diff.

-- HAS_UNTRACKABLE_DEPENDENCIES: This sequence has no owner, so it cannot be tracked. It may be in use by a table or function.
CREATE SEQUENCE "public"."employees_id_seq"
	AS integer
	INCREMENT BY 1
	MINVALUE 1 MAXVALUE 2147483647
	START WITH 1 CACHE 1 NO CYCLE
;
CREATE TABLE "employees" (
	"id" integer NOT NULL,
	"name" text COLLATE "pg_catalog"."default",
	"age" smallint NOT NULL
);
CREATE UNIQUE INDEX CONCURRENTLY employees_pkey ON public.employees USING btree (id);
ALTER TABLE "public"."employees" ADD CONSTRAINT "employees_pkey" PRIMARY KEY USING INDEX "employees_pkey";
```

TODO:
- [x] find out why shadow database is created twice
- [x] support remote -> local diff (currently impl requires a db factory)
- [ ] support more [pg entities](https://github.com/djrobstep/schemainspect/tree/master/schemainspect/pg/sql), including custom types, RLS policies, etc

## Additional context

1. create an empty database: `supabase db start`
2. create a schema file and save to `supabase/migrations/0_schema.sql`
```sql
create table public.employees (
    id integer primary key generated always as identity,
    name text,
    age smallint not null
);
```


internal/db/diff/diff.go
sweatybridge, line 1: renamed from migra.go

internal/db/diff/pgadmin.go
sweatybridge, line 1: renamed from diff.go

PR #1919, opened 2024-02-06
fix: enable analytics by default

Ziinc:
closes https://github.com/supabase/supabase/issues/18707#issuecomment-1928750962


sweatybridge (approved):

sweatybridge:
Some of the integration tests are failing due to starting this container. Do you plan to look into those or should I?

Probably the easiest fix is to initialise the [init_config.test.toml](https://github.com/supabase/cli/blob/b380dfbc53d494a7b2e0c0f55b9503c6fd7d9796/internal/utils/templates/init_config.test.toml) instead for `tests`.

Issue #1976, opened 2024-02-21
Prevent accidental `db push` deployments to remote


daniel-j-h:
A common use case is to develop locally, run and test database migrations, and having CI/CD such as github actions to deploy to production using e.g. https://github.com/supabase/setup-cli running

    supabase db push

in a release job.

At the moment it's very easy to accidentally run `supabase db push` from your local development environment deploying to production.

Ideally we'd have a way to prevent local database deployments to remote and restrict it to CI/CD.

What's the best way to achieve this? Can we have e.g. a config option in `supabase/config.toml` restricting db pushes to remote (with a `--force` flag to overwrite it) or how can we best make sure to not have developers pushing to remote by accident and instead rely on CI/CD?

sweatybridge:
I think the long term solution may be to use [branching](https://supabase.com/blog/supabase-branching) so that developers don't need the production password to develop locally.

Until then, I've implemented a stop gap solution in the linked PR by asking users to confirm pushing to remote. Do you think that's sufficient for your use case?

daniel-j-h:
Gotcha, yeah the stop gap solution looks good!

I'm a liittle bit worried tho by how easy it it to reset the remote database and push changes to it :see_no_evil: Ideally I want developers to have a read-only view of the remote database, only, and only CI/CD being able to write to it.

If branching gives us this in the long term then :+1:

sweatybridge:
The patch has been released in `npx supabase@beta db push`. It will go into stable channel in 2 weeks time.

PR #1978, opened 2024-02-21
fix: confirm pushing migrations to remote database

sweatybridge:

related https://github.com/supabase/cli/issues/1976

## What is the new behavior?

```bash
$ supabase db push
Connecting to remote database...
Do you want to push these migrations to the remote database?
 • 0_schema.sql
 • 1_test.sql

 [Y/n] n
context canceled
exit status 1
```

## Additional context

### DIFF
--- a/internal/db/push/push.go
+++ b/internal/db/push/push.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	"github.com/go-errors/errors"
 	"github.com/jackc/pgconn"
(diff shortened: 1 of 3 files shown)


daniel-j-h:
Now that we are getting prompted, will `supabase db push` still work as is in a CI/CD setting where you can't answer the prompt?

sweatybridge:
Yes, in non-interactive consoles the default answer is used. In this case, it's Yes.

Gipetto:
@sweatybridge I'm trying to write a GitHub action to push schema changes, and I'm using `act` (https://github.com/nektos/act) to test this locally but I'm being prompted for confirmation when pushing the changes.

Your comment above says that non-interactive shells should use the default, but my workflow is hanging at this step and the worker output is showing the question as it would be presented to the user. Do you have any experience with testing workflows locally using `act` and maybe why this wouldn't be detecting the non-interactive session?

I'm using this to test wether `act` is giving me a non-interactive session, and it appears that it does:

```
[Vercel Staging Deployment/migrate-staging] ⭐ Run Main echo "$([[ $- == *i* ]] && echo 'Interactive' || echo 'Not interactive')"
[Vercel Staging Deployment/migrate-staging]   🐳  docker exec cmd=[bash --noprofile --norc -e -o pipefail /var/run/act/workflow/4] user= workdir=client
| Not interactive
[Vercel Staging Deployment/migrate-staging]   ✅  Success - Main echo "$([[ $- == *i* ]] && echo 'Interactive' || echo 'Not interactive')"
[Vercel Staging Deployment/migrate-staging] ⭐ Run Main Push Supabase Schema
[Vercel Staging Deployment/migrate-staging]   🐳  docker exec cmd=[bash --noprofile --norc -e -o pipefail /var/run/act/workflow/5] user= workdir=client
| Connecting to remote database...
| Do you want to push these migrations to the remote database?
| ... a list of all my schema changes ...
```
I am not familiar enough with interactive & non-interactive shells to know for sure if I've got this right, and if it is the `act` runner or the supabase client that has it wrong.

Any thoughts?
Thanks!

Issue #1328, opened 2023-07-23
pg_graphql comment directives


TonsiTT:
Using the "supabase db remote commit" and then setting up the local environment, the pg_graphql comment directives on tables in the linked hosted database are not replicated.    
For example, a comment directive like `@graphql({"totalCount": {"enabled": true}})`. 
Even comments on the schema like `@graphql({"max_rows": 100})`

PR #1981, opened 2024-02-21
fix: dont loose comments when dumping schema

chreck:
Remove comments in extensions

Bug fix https://github.com/supabase/cli/issues/1328

Every dump does not include the comments for tables, columns, functions, etc.

## What is the new behavior?

The db dump includes the comments for tables, columns, functions, etc. but not for the extensions. The extension is enabled by dashboard user and does not need to change the comment about the extension itself.

## Additional context

If you use the supabase cli as a backup solution for schema, data and roles data like it is described in [Supabase Doc - Automated backups using GitHub Actions](https://supabase.com/docs/guides/cli/github-action/backups) and you have setup a lot of comments for tables, functions etc. You will loose these comments, they are never included in the schema.

I removed the `--no-comments` flag in template `dump_schema.sh` because of this.


sweatybridge (approved):
Thanks for figuring this out!

PR #1986, opened 2024-02-22
feat: enable Supabase AI during local development

felixgabler:

Allows users to bring their own OpenAI API key into the Supabase Studio so Supabase AI works when developing locally too.

When using the Supabase AI fields locally, one gets an error message which is just the HTML of the standard error page. Also, the container throws an error that the OpenAI key cannot be found.

Screenshot 2024-02-22 at 08 29 06

## What is the new behavior?

When specifying the `OPENAI_API_KEY` while starting Supabase locally, it should now pass it through the config into the container as an env variable.

## Additional context

Further ideas for improvement:
- If accepted, should add this to the docs somewhere so people know where and how to add their OpenAI keys
- It would really be great if `supabase start` would also read the `.env` file (also for social auth secrets)
- Show a better error message when using the AI without the key (maybe pointing to above docs)

### DIFF
--- a/internal/utils/templates/init_config.toml
+++ b/internal/utils/templates/init_config.toml
@@ -49,6 +49,8 @@ enabled = true
 port = 54323
 # External URL of the API server that frontend connects to.
 api_url = "http://127.0.0.1"
+# OpenAI API Key to use for Supabase AI in the Supabase Studio.
+openai_api_key = "env(OPENAI_API_KEY)"
 
 # Email testing server. Emails sent with the local dev setup are not actually sent - rather, they
 # are monitored, and you can view the emails that would have been sent from the web interface.
--- a/internal/start/start.go
+++ b/internal/start/start.go
@@ -843,6 +843,7 @@ EOF
 					"SUPABASE_ANON_KEY=" + utils.Config.Auth.AnonKey,
 					"SUPABASE_SERVICE_KEY=" + utils.Config.Auth.ServiceRoleKey,
 					"LOGFLARE_API_KEY=" + utils.Config.Analytics.ApiKey,
+					"OPENAI_KEY=" + utils.Config.Studio.OpenaiApiKey,
 					fmt.Sprintf("LOGFLARE_URL=http://%v:4000", utils.LogflareId),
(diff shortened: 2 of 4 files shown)


internal/utils/templates/init_config.toml
```diff
@@ -49,6 +49,8 @@ enabled = true
 port = 54323
 # External URL of the API server that frontend connects to.
 api_url = "http://127.0.0.1"
+# OpenAI API Key to use for Supabase AI in the Supabase Studio.
+openai_api_key = "env(OPENAI_API_KEY)"
```
sweatybridge, line 53: Let's also add it to `init_config.test.toml` for consistency.

sweatybridge (approved):
LGTM with a minor nitpick. Thanks for contributing!

felixgabler:
I tested it and now it works. It also shows a warning now when people use AI features without specifying the environment variables, so that is nice! However, it talks about `OPENAI_KEY` instead of `OPENAI_API_KEY` which I am using here because it is what OpenAI always uses. Should I change it here or make a PR in the studio to also use `OPENAI_API_KEY`?

Also, I'm wondering whether the config reads `.env` files automatically already?

sweatybridge:
> Also, I'm wondering whether the config reads .env files automatically already?

Yes it does, but only when `.env` is in your work dir, ie. sibling of `supabase` directory, not within. We might change this behaviour in the future to support child paths as well.

felixgabler:
> Yes it does, but only when `.env` is in your work dir, ie. sibling of `supabase` directory, not within. We might change this behaviour in the future to support child paths as well.

That's cool! I have my env file at `my-app/supabase/.env` and it did not pick up the `OPENAI_API_KEY=...` value that I specified in there. Could that be because I ran it from my fork `cli` folder with `--workdir`?

sweatybridge:
It needs to be `my-app/.env` at the time of writing. I'm sure we've mentioned this in our docs but it trips up people so often that it's worth supporting `supabase/.env` in the future as well.

sweatybridge:
It's all good. Our release process is currently broken. I will merge this PR once it's fixed.

sweatybridge (approved):

Issue #2040, opened 2024-03-13
CLI: how to login in non-browser/non-gui cloud vscode env?


gowy222:
online cloud vscode IDE cant login,

```
root@73a777eae898:/workspace# npx supabase login
Hello from Supabase! Press Enter to open browser and login automatically.

then:
Here is your login link in case browser did not open https://supabase.com/dashboard/cli/login?session_id=5c5.....
cannot open default browser: exec: "xdg-open": executable file not found in $PATH
Try rerunning the command with --debug to troubleshoot the error.

```
but the Linux online cloud vscode IDE dont has any browser.

sweatybridge:
Thanks for reporting this issue. I believe the CLI should ignore this error to allow user to open the browser manually.

PR #2042, opened 2024-03-13
fix: log and ignore error opening browser

sweatybridge:

Bug fix https://github.com/supabase/cli/issues/2040

## What is the new behavior?

Users can open the login link manually if browser fails to open automatically.

## Additional context

### DIFF
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/go-errors/errors"
 	"github.com/google/uuid"
-	"github.com/mattn/go-isatty"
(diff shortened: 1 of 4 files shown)


kamilogorek:
Should we also remove `isatty` usage from `tea.go` as well? One dependency less, and its using the very same `if isatty.IsTerminal(os.Stdin.Fd()) && isatty.IsTerminal(os.Stdout.Fd())` check.

kamilogorek (approved):

sweatybridge:
Yup I've removed it from tea.go but I think there's a transitive dependency somewhere.

PR #2061, opened 2024-03-18
feat(analytics): configure vector for docker logs instead of syslog driver

sweatybridge:

feature

## What is the new behavior?

Uses native docker logs driver instead of syslog.

## Additional context


endyjasmi:
anything we can do to get this merged?

internal/start/start.go
```diff
@@ -158,14 +161,27 @@ func run(p utils.Program, ctx context.Context, fsys afero.Fs, excludedContainers
...
+		}
+		// Special case for GitLab pipeline
+		host := utils.Docker.DaemonHost()
+		if parsed, err := client.ParseHostURL(host); err == nil {
+			if parsed.Scheme == "tcp" {
+				env = append(env, "DOCKER_HOST="+host)
+			} else {
+				// TODO: mount windows named pipe as unix domain socket?
+				// TODO: mount domain socket from docker rootless mode
+				binds = append(binds, "/var/run/docker.sock:/var/run/docker.sock:ro")
```
sweatybridge, line 177: Currently this relies on mounting root docker socket into container. Ideally we want to find a way to mount the user socket instead so it doesn't require root access.

Issue #2046, opened 2024-03-13
Generating TypeScript types fails in Bitbucket pipeline


hammerlscs:
**Describe the bug**
We are using

```
npx supabase gen types typescript --local
```

to generate the TypeScript types in our pipeline. Since upgrading from 1.131.5 to 1.148.6, generating types fails with the following error:

```
v0.79.0: Pulling from supabase/postgres-meta
Status: Downloaded newer image for public.ecr.aws/supabase/postgres-meta:v0.79.0
failed to create docker container: Error response from daemon: cannot share the host's network namespace when user namespaces are enabled
```

**To Reproduce**
Use `npx supabase gen types typescript --local` in the Bitbucket pipeline.

**System information**
 - Ticket ID: eb96d7a5b9de43a0bbfe5fca68213cf4

**Additional context**
Possibly related to https://github.com/supabase/cli/issues/1830

etzelc:
Seeing this error in our BitBucket Pipeline (including current pre-release version 1.144.4) when executing `npx supabase start`
```
Status: Downloaded newer image for public.ecr.aws/supabase/postgres-meta:v0.79.5
failed to create docker container: Error response from daemon: cannot share the host's network namespace when user namespaces are enabled
Try rerunning the command with --debug to troubleshoot the error.
```

swamidass:
Having this same problem with version of the cli 1.172.2. Is there a fix in the works? Hopefully some regression testing too, because this seems to be a recurring bug.

maximilian-hammerl:
We are also using CLI version 1.172.2 and we are not having this problem. Everything works fine on our end.

PR #2064, opened 2024-03-19
fix: use custom network for local database when starting pg-meta

etzelc:

Bug fix for #2046. Inspired by #1836. 

pg-meta is started using host network. This does not work using Bitbucket SaaS runners, because Bitbucket pipeline enables user namespace for docker daemon, which doesn't allow host network mode. https://community.atlassian.com/t5/Bitbucket-articles/Changes-to-make-your-containers-more-secure-on-Bitbucket/ba-p/998464

## What is the new behavior?

When connecting to local database, join the existing custom network instead.

## Additional 

1. #1836 also changed `config.Host = utils.DbAliases[0]` and `config.Port = 5432`. Is this also necessary for this fix here?
2. [internal/db/diff/migra.go ](https://github.com/supabase/cli/blob/develop/internal/db/diff/migra.go) and [internal/db/dump.go](https://github.com/supabase/cli/blob/develop/internal/db/dump/dump.go) also use host network.


internal/gen/types/typescript/typescript.go
```diff
@@ -47,6 +48,9 @@ func Run(ctx context.Context, projectId string, dbConfig pgconn.Config, schemas
 		if strings.Contains(utils.Config.Api.Image, "v9") {
 			postgrestV9Compat = true
 		}
+
+		// Use custom network when connecting to local database
+		networkID = utils.NetId
```
sweatybridge, line 53:
```suggestion
		dbConfig.Host = utils.DbAliases[0]
		dbConfig.Port = 5432
		networkID = utils.NetId
```

Yup, host and port need to be updated as well.

sweatybridge (approved):

etzelc:
Thanks for taking over. 👍

sweatybridge:
Thanks for the contribution. This should be available in v1.150.3.

I will also make a patch for db diff and db dump.

PR #2075, opened 2024-03-21
fix: pg-meta v0.79.9

soedirgo:
https://github.com/supabase/postgres-meta/compare/v0.79.5...v0.79.9

Includes https://github.com/supabase/postgres-meta/commit/3b6c4aaef77444bf947fb9a05a62f8a2bc756298 which affects typegen


sweatybridge:
Shall we make a rule to bump the minor version if it affects the generated type? It would be slightly clearer to CI users when they notice differences in generated types after upgrading.

soedirgo:
This fixes the generated types - previously TypeScript wouldn't even compile if the bug occurs

sweatybridge (approved):

Issue #1215, opened 2023-06-16
Auto-generate a schema.sql file when you run migrations


saltcod:
**Is your feature request related to a problem? Please describe.**
I often want to know what my table schema looks like. To do this, I need to open the Dashboard table editor for the table in question, then scroll to see the columns. 

**Describe the solution you'd like**
Rails has a `schema.rb` file that does this. (I think) When you run a migration, it updates  the `schema.rb` file with a snapshot of all current tables and their schema.

The accepted answer for this question gives a concise explanation of why this file is useful: 
https://stackoverflow.com/questions/9884429/rails-what-does-schema-rb-do

<img width="757" alt="CleanShot 2023-06-16 at 09 45 59@2x" src="https://github.com/supabase/cli/assets/105593/238ec1da-e4d4-4618-8a7c-ff523ff88b8b">

Here's a `schema.rb` example: 

<img width="790" alt="CleanShot 2023-06-16 at 09 48 16@2x" src="https://github.com/supabase/cli/assets/105593/7554e9c3-2514-43f8-895e-8e6a29d3ca71">

This file could live in `/supabase/schema.sql` 

**Describe alternatives you've considered**
Generating types to a types.ts file is also helpful, but its a little noisier than this `schema.sql` approach.

jonkurtis:
> doesnt `supabase db dump -f supabase/schema.sql` do this?

it does indeed. 

however, i think an alternative approach that would better accomplish the intended DX would be to create all schemas in a /schemas file, then be able to run db reset to run off of /schema files instead of /migrations. then run db diff --use-migra on the newly reset schema to generate migration files. 

when you want to change a schema file you change it in /schema instead of /migrations, editing the original schema SQL. db reset off of /schema. db diff off of newly reset local db to update migrations to push up to prod. 

this enables you to better organize files under /supabase/schema/users.sql, /supabase/schema/teams.sql. can also include comments and organize each file as desired. 

easier to reason about than one big dump into schema.sql

You would need to be able to tell supabase db reset to look at /supabase/schema/ instead of /migrations

xleddyl:
@SweatyBridge I tried creating a `supabase/schemas` folder and putting the definitions inside, but I can't manage to make the script work. It always diffs the local DB with the one generated from the migrations and not the schema.

Can you check if this still works or has been deprecated?

PR #2099, opened 2024-03-27
feat: support declarative schema with db diff

sweatybridge:

closes https://github.com/supabase/cli/issues/1215

## What is the new behavior?

- Add schema files to `supabase/schemas` (they will be run in lexical order)
  - Easiest way to create one is `supabase db dump --local -f supabase/schemas/base.sql`
- Then run `supabase db diff`

A new local database will be created from declared schemas for diffing.

```
$ supabase db diff
Creating local database from supabase/schemas...
Setting up initial schema...
Creating custom roles supabase/roles.sql...
Applying schema supabase/schemas/base.sql...
Connecting to local database...
Creating shadow database...
Setting up initial schema...
Creating custom roles supabase/roles.sql...
Applying migration 20231204107450_grant.sql...
Diffing schemas: extensions,private,public
Finished supabase db diff on branch dec-schema.

No schema changes found
```

## Additional context


xleddyl:
@SweatyBridge I tried creating a `supabase/schemas` folder and putting the definitions inside, but I can't manage to make the script work. It always diffs the local DB with the one generated from the migrations and not the schema.

Can you check if this still works or has been deprecated?

sweatybridge:
oh.. you need to stop the local stack first with `supabase stop` before running `supabase db diff`. Otherwise it will always connect to the locally running db instead.

xleddyl:
Thank you so much, I’ll give it a try! By the way, I think this feature should be highlighted in the official docs. It’s incredibly useful and essential for efficiently managing migrations.

PR #2112, opened 2024-04-02
fix: disable keyring on WSL1

sweatybridge:

Bug fix https://github.com/supabase/cli/pull/1509#issuecomment-2030225608

## What is the new behavior?

Check for string either `WSL` or `Microsoft`.

## Additional context


minism:
Thanks for the quick fix! ❤️

PR #2114, opened 2024-04-03
feat: bump auth version 2.145.0

J0:

Upgrade auth version to 2.145.0 as we are nearing the end of the deploy. A follow up PR will be made to add relevant env vars.


sweatybridge (approved):

J0:
I think we’re probably going to v2.146.0 soon but let’s go with this for now!

PR #2118, opened 2024-04-04
fix: handle expo env vars in bootstrap.

thorwebdev:

- Handles `EXPO_PUBLIC_` prefixed env vars.

See template project here: https://github.com/supabase/supabase/pull/22428/files#diff-c4e15451c416716b1e086348e04fd301438501474054fc6e6498fa27d65e7b2fR3

@sweatybridge you okay to do like this or prefer a more generic approach?


internal/bootstrap/bootstrap.go
```diff
@@ -246,8 +248,10 @@ func writeDotEnv(keys []api.ApiKeyResponse, config pgconn.Config, fsys afero.Fs)
 		case POSTGRES_DATABASE:
 			initial[k] = config.Database
 		case NEXT_PUBLIC_SUPABASE_ANON_KEY:
+		case EXPO_PUBLIC_SUPABASE_ANON_KEY:
```
sweatybridge, line 251:
```suggestion
			fallthrough
		case EXPO_PUBLIC_SUPABASE_ANON_KEY:
```

Crazy thing about Golang switch case is that it breaks by default. So we need an explicit fall through here

internal/bootstrap/bootstrap.go
```diff
@@ -246,8 +248,10 @@ func writeDotEnv(keys []api.ApiKeyResponse, config pgconn.Config, fsys afero.Fs)
...
 			initial[k] = initial[SUPABASE_ANON_KEY]
 		case NEXT_PUBLIC_SUPABASE_URL:
+		case EXPO_PUBLIC_SUPABASE_URL:
```
sweatybridge, line 254:
```suggestion
			fallthrough
		case EXPO_PUBLIC_SUPABASE_URL:
```

sweatybridge (approved):

PR #2126, opened 2024-04-05
feat: add anonymous sign-ins config

kangmingtay:
* Supports the config to enable anonymous sign-ins for auth locally as requested in https://github.com/supabase/auth/issues/68#issuecomment-2028439659


sweatybridge (commented):
should also add to init_config.toml with some explanation of this flag https://github.com/supabase/cli/blob/develop/internal/utils/config.go#L215

sweatybridge (approved):

PR #2129, opened 2024-04-06
feat: add HTTP hook Secret Configuration

J0:

Adds the corresponding configuration for the `SendSMS` and `SendEmail` HTTP Hooks. Also introduces Secrets for previous hooks - `custom_access_token`, `mfa_verification_attempt` , `password_verification_attempt`

Backend PR (Merged): https://github.com/supabase/infrastructure/pull/17848
Frontend PR (WIP): https://github.com/supabase/supabase/pull/22495

### DIFF
--- a/internal/utils/config.go
+++ b/internal/utils/config.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	_ "embed"
 	"fmt"
+	"net/url"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -387,11 +388,14 @@ type (
 		MFAVerificationAttempt      hookConfig `toml:"mfa_verification_attempt"`
 		PasswordVerificationAttempt hookConfig `toml:"password_verification_attempt"`
 		CustomAccessToken           hookConfig `toml:"custom_access_token"`
+		SendSMS                     hookConfig `toml:"send_sms"`
+		SendEmail                   hookConfig `toml:"send_email"`
 	}
 
 	hookConfig struct {
 		Enabled bool   `toml:"enabled"`
 		URI     string `toml:"uri"`
+		Secrets string `toml:"secrets"`
 	}
 
 	twilioConfig struct {
@@ -459,6 +463,24 @@ type (
 	// 	AfterMigrations  string `toml:"after_migrations"`
 	// }
 )
+
+func (h *hookConfig) HandleHook(hookType string) error {
+	// If not enabled do nothing
+	if !h.Enabled {
+		return nil
+	}
+	if h.URI == "" {
+		return fmt.Errorf("missing required field in config: auth.hook.%s.uri", hookType)
+	}
+	if err := validateHookURI(h.URI, hookType); err != nil {
+		return err
+	}
+	var err error
+	if h.Secrets, err = maybeLoadEnv(h.Secrets); err != nil {
+		return fmt.Errorf("missing required field in config: auth.hook.%s.secrets", hookType)
+	}
+	return nil
+}
 
 func LoadConfigFS(fsys afero.Fs) error {
 	// Load default values
@@ -687,25 +709,21 @@ func LoadConfigFS(fsys afero.Fs) error {
 					return err
 				}
 			}
-
-			if Config.Auth.Hook.MFAVerificationAttempt.Enabled {
-				if Config.Auth.Hook.MFAVerificationAttempt.URI == "" {
-					return errors.New("Missing required field in config: auth.hook.mfa_verification_atempt.uri")
-				}
+			if err := Config.Auth.Hook.MFAVerificationAttempt.HandleHook("mfa_verification_attempt"); err != nil {
+				return err
 			}
-
-			if Config.Auth.Hook.PasswordVerificationAttempt.Enabled {
-				if Config.Auth.Hook.PasswordVerificationAttempt.URI == "" {
-					return errors.New("Missing required field in config: auth.hook.password_verification_attempt.uri")
-				}
+			if err := Config.Auth.Hook.PasswordVerificationAttempt.HandleHook("password_verification_attempt"); err != nil {
+				return err
 			}
-
-			if Config.Auth.Hook.CustomAccessToken.Enabled {
-				if Config.Auth.Hook.CustomAccessToken.URI == "" {
-					return errors.New("Missing required field in config: auth.hook.custom_access_token.uri")
-				}
+			if err := Config.Auth.Hook.CustomAccessToken.HandleHook("custom_access_token"); err != nil {
+				return err
 			}
-
+			if err := Config.Auth.Hook.SendSMS.HandleHook("send_sms"); err != nil {
+				return err
+			}
+			if err := Config.Auth.Hook.SendEmail.HandleHook("send_email"); err != nil {
+				return err
+			}
 			// Validate oauth config
 			for ext, provider := range Config.Auth.External {
 				if !provider.Enabled {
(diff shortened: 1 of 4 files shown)


internal/utils/config.go
```diff
@@ -674,6 +677,29 @@ func LoadConfigFS(fsys afero.Fs) error {
...
+				if Config.Auth.Hook.SendSMS.HTTPHookSecrets == "" {
+					return errors.New("Missing required field in config: auth.hook.send_sms.secrets")
+				}
+			}
+			if Config.Auth.Hook.SendEmail.Enabled {
+				if Config.Auth.Hook.SendEmail.URI == "" {
+					return errors.New("Missing required field in config: auth.hook.send_email.uri")
+				}
+				if !strings.HasPrefix(strings.ToLower(Config.Auth.Hook.SendEmail.URI), "http") {
+					return errors.New("Invalid HTTP hook config: auth.hook.send_email.uri should be a http or https URL")
```
J0, line 696:
```suggestion
					return errors.New("Invalid HTTP hook config: auth.hook.send_email.uri should be a HTTP or HTTPS URL")
```

internal/utils/misc.go
```diff
@@ -40,7 +40,7 @@ const (
 	VectorImage      = "timberio/vector:0.28.1-alpine"
 	PgbouncerImage   = "bitnami/pgbouncer:1.20.1-debian-11-r39"
 	PgProveImage     = "supabase/pg_prove:3.36"
-	GotrueImage      = "supabase/gotrue:v2.145.0"
+	GotrueImage      = "supabase/gotrue:v2.147.1"
```
J0, line 43: This isn't out yet, but will be deployed soon. Currently, we are on v2.147.0 on middleware and v2.146 for all projects. Can wait to bump this too if that's better.

internal/utils/config.go
```diff
@@ -674,6 +677,29 @@ func LoadConfigFS(fsys afero.Fs) error {
 				}
 			}
 
+			if Config.Auth.Hook.SendSMS.Enabled {
+				if Config.Auth.Hook.SendSMS.URI == "" {
+					return errors.New("Missing required field in config: auth.hook.send_sms.uri")
+				}
+				if !strings.HasPrefix(strings.ToLower(Config.Auth.Hook.SendSMS.URI), "http") {
```
sweatybridge, line 684: Shall we parse the URL and check for scheme instead?

internal/utils/templates/init_config.toml
```diff
@@ -119,6 +119,20 @@ template = "Your code is {{ `{{ .Code }}` }} ."
 # enabled = true
 # uri = "pg-functions://<database>/<schema>/<hook_name>"
 
+# Use a custom sms sender
+[auth.hook.send_sms]
+# enabled = true
+# We use host.docker.internal to connect to port 54321 on the host. See: https://docs.docker.com/desktop/networking/
+# uri = "http://host.docker.internal:54321/functions/v1/sms_sender"
```
sweatybridge, line 126: Since this hook is called from inside auth container, it should be possible to use either `kong` or `edge_runtime` network alias, ie. `http://kong:8000/functions/v1/sms_sender`

J0: Sure let's use `kong`  so it goes through the gateway still.

J0: Ah sigh. This would require an auth level change

J0: Opened a pr, but for reference do you happen to know which platforms?

sweatybridge: Linux was known to have problems https://stackoverflow.com/a/62431165 so we added the default workaround with `host-gateway`. However the workaround doesn't work for older versions of podman https://github.com/containers/podman/issues/14390.

But it appears that all these issues are resolved now so may be it's ok. 😄

internal/utils/misc.go
J0, line 43:
```suggestion
	GotrueImage      = "supabase/gotrue:v2.148.0"
```

internal/utils/templates/init_config.toml
```diff
@@ -119,6 +119,20 @@ template = "Your code is {{ `{{ .Code }}` }} ."
 # enabled = true
 # uri = "pg-functions://<database>/<schema>/<hook_name>"
 
+# Use a custom sms sender
+[auth.hook.send_sms]
+# enabled = true
+# We use host.docker.internal to connect to port 54321 on the host. See: https://docs.docker.com/desktop/networking/
```
sweatybridge, line 125: This comment looks outdated.

internal/utils/templates/init_config.toml
```diff
@@ -119,6 +119,20 @@ template = "Your code is {{ `{{ .Code }}` }} ."
...
+# uri = "http://kong:8000/functions/v1/sms_sender"
+# Use this secret in the email_sender edge function as well.
+# secrets = "v1,whsec_c3VwYWJhc2Vpc3Zlcnlhd2Vzb21lYW5kaWxpa2VpdA=="
```
sweatybridge, line 128: How is this secret generated? If it's only a static value, may be we can hardcode it in config.go. Otherwise we want to use `env(SOME_SECRET)` substitution.

To avoid accidents and confusion, we don't want to add secret to config.toml even if they are only used locally.

J0:
Actually I think we’ll need to wait for the FE to go out before merging :/

internal/utils/config.go
```diff
@@ -357,11 +358,14 @@ type (
...
+		SendSMS                     hookConfig `toml:"send_sms"`
+		SendEmail                   hookConfig `toml:"send_email"`
 	}
 
 	hookConfig struct {
-		Enabled bool   `toml:"enabled"`
-		URI     string `toml:"uri"`
+		Enabled         bool   `toml:"enabled"`
+		URI             string `toml:"uri"`
+		HTTPHookSecrets string `toml:"secrets"`
```
sweatybridge, line 368:
```suggestion
		Secrets string `toml:"secrets"`
```

Shall we just name this field `Secrets` for consistency with toml tag?

J0: Yup. I’ll need to add an additional set of config vars with our latest changes as well.

J0:
New Changes:
- Removed the `init_config.toml` 
- Added tests 

The backend is now merged and I'm hoping to get this in so that we can allow [contributors to write examples](https://github.com/supabase/auth/pull/1018#issuecomment-2098612913)

Tested locally

internal/utils/config.go
```diff
@@ -460,6 +464,24 @@ type (
 	// }
 )
 
+func (h *hookConfig) HandleHook(hookType string) error {
+	// If not enabled do nothing
+	if !h.Enabled {
+		return nil
+	}
+	if h.URI == "" {
+		return fmt.Errorf("missing required field in config: auth.hook.%s.uri", hookType)
```
sweatybridge, line 473:
```suggestion
		return errors.Errorf("missing required field in config: auth.hook.%s.uri", hookType)
```

use `go-errors` package for auto stack trace.

internal/utils/config.go
```diff
@@ -460,6 +464,24 @@ type (
...
+	}
+	if err := validateHookURI(h.URI, hookType); err != nil {
+		return err
+	}
+	var err error
+	if h.Secrets, err = maybeLoadEnv(h.Secrets); err != nil {
+		return fmt.Errorf("missing required field in config: auth.hook.%s.secrets", hookType)
```
sweatybridge, line 480:
```suggestion
		return errors.Errorf("missing required field in config: auth.hook.%s.secrets", hookType)
```

ditto

sweatybridge (approved):
Might want to add a valid config to `init_config.test.toml` as a sanity check.

internal/utils/config.go
J0: Thanks! TIL

PR #2130, opened 2024-04-06
feat: fetch start command for bootstrap from samples.json

J0:

Currently, we  use a hardcoded start command (e.g. `npm run dev`) which might not generalize to non-javascript projects. This PR aims to fix this by fetching from `samples.json`


J0:
Switch to pass down `Start` from top level command instead.

PR #2132, opened 2024-04-08
feat: add seed command to generate test users

kangmingtay:
* Adds a `supabase seed auth` command that generates sql examples to add test
users and saves them in a `seed.example.sql` file

## What is the new behavior?

## Additional context

### DIFF
--- a/cmd/seed.go
+++ b/cmd/seed.go
@@ -1,0 +1,36 @@
+package cmd
+
+import (
+	"os"
+	"os/signal"
+
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+	"github.com/supabase/cli/internal/seed"
+)
+
+var (
+	seedCmd = &cobra.Command{
+		GroupID: groupLocalDev,
+		Use:     "seed",
+		Short:   "Generate seed data",
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			ctx, _ := signal.NotifyContext(cmd.Context(), os.Interrupt)
+			cmd.SetContext(ctx)
+			return cmd.Root().PersistentPreRunE(cmd, args)
+		},
+	}
+
+	seedAuthCmd = &cobra.Command{
+		Use:   "auth",
+		Short: "Generate seed data for the Auth schema",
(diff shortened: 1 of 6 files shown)


cmd/seed.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+	"github.com/supabase/cli/internal/seed"
+)
+
+var (
+	seedCmd = &cobra.Command{
+		GroupID: groupLocalDev,
+		Use:     "seed",
+		Short:   "Generate seed data",
```
sweatybridge, line 16: Shall we move this under `supabase gen seed` instead since it generates a seed file?

If we continue using `supabase seed`, my expectation would be to actually seed the local / remote database.

internal/seed/seed.go
```diff
@@ -0,0 +1,194 @@
...
+			ExposedPorts: nat.PortSet{"9999/tcp": {}},
+			Healthcheck: &container.HealthConfig{
+				Test:     []string{"CMD", "wget", "--no-verbose", "--tries=1", "--spider", "http://127.0.0.1:9999/health"},
+				Interval: 10 * time.Second,
+				Timeout:  2 * time.Second,
+				Retries:  3,
+			},
+		},
+		container.HostConfig{
+			PortBindings:  nat.PortMap{"9999/tcp": []nat.PortBinding{{HostPort: "54325"}}},
```
sweatybridge, line 94: This port is currently reserved as smtp port for inbucket.

internal/seed/seed.go
```diff
@@ -0,0 +1,194 @@
...
+		"sso_providers",
+		"sso_domains",
+		"saml_relay_state",
+		"flow_state",
+	}, false, false, outStream); err != nil {
+		return err
+	}
+
+	// use awk to remove unnecessary sections from the dump
+	awkCmd := exec.Command("awk", `
```
sweatybridge, line 153: We can't expect `awk` to always be installed on user's system, ie. windows?

We can either process the docker output in go or make a shell script that runs inside gotrue container.

PR #2138, opened 2024-04-10
fix: exclude supabase managed schemas from inspect subcommand

encima:

Feature/Fix: moving queries for inspect to the root `inspect` module and removing Supabase schemas from queries

Queries are within the command and show results from Supabase schemas that could cause concern or make the user take actions on schemas they do not own

## What is the new behavior?

- Queries moved into a single place (could be further developed to template similar queries)
- Supabase schemas removed from relevant queries
- Explanation of hit rates added to `cache-hit` command and a column explains whether action might be needed

## Additional context

image

### DIFF
--- a/internal/inspect/queries.go
+++ b/internal/inspect/queries.go
@@ -107,7 +107,7 @@ FROM pg_class c
 LEFT JOIN pg_namespace n ON (n.oid = c.relnamespace)
 WHERE NOT n.nspname LIKE ANY($1)
 AND c.relkind = 'i'
-GROUP BY c.relname
+GROUP BY n.nspname, c.relname
 ORDER BY sum(c.relpages) DESC`
 
 const INDEX_USAGE_QUERY = `SELECT
@@ -210,14 +210,16 @@ AND c.relkind = 'r'
 ORDER BY pg_indexes_size(c.oid) DESC`
 
 const TABLE_RECORD_COUNTS_QUERY = `SELECT
-  schemaname || '.' || relname AS name,
+  schemaname AS schema,
+  relname AS name,
   n_live_tup AS estimated_count
 FROM pg_stat_user_tables
 WHERE NOT schemaname LIKE ANY($1)
 ORDER BY n_live_tup DESC`
 
 const TABLE_SIZES_QUERY = `SELECT
-  n.nspname || '.' || c.relname AS name,
+  n.nspname AS schema,
+  c.relname AS name,
   pg_size_pretty(pg_table_size(c.oid)) AS size
 FROM pg_class c
 LEFT JOIN pg_namespace n ON (n.oid = c.relnamespace)
@@ -233,7 +235,8 @@ WHERE NOT n.nspname LIKE ANY($1)
 AND c.relkind = 'i'`
 
 const TOTAL_TABLE_SIZES_QUERY = `SELECT
-  n.nspname || '.' || c.relname AS name,
+  n.nspname AS schema,
+  c.relname AS name,
   pg_size_pretty(pg_total_relation_size(c.oid)) AS size
 FROM pg_class c
 LEFT JOIN pg_namespace n ON (n.oid = c.relnamespace)
--- a/internal/inspect/queries.go
+++ b/internal/inspect/queries.go
@@ -1,31 +1,20 @@
 package inspect
 
 import (
-	"fmt"
-	"strings"
-
 	"github.com/supabase/cli/internal/utils"
 )
 
-var excludedString string
-
-func init() {
-	var excluded = make([]string, len(utils.InternalSchemas))
-	for i, schema := range utils.InternalSchemas {
-		excluded[i] = fmt.Sprintf("'%s'", schema)
-	}
-	excluded = append(excluded, "'pg_catalog'")
-	excludedString = strings.Join(excluded, ", ")
+func GetExcludedSchemas() []string {
+	return append(utils.InternalSchemas, "pg_catalog")
 }
 
-func BloatQuery() string {
-	return fmt.Sprintf(`WITH constants AS (
+const BLOAT_QUERY = `WITH constants AS (
   SELECT current_setting('block_size')::numeric AS bs, 23 AS hdr, 4 AS ma
 ), bloat_info AS (
   SELECT
     ma,bs,schemaname,tablename,
-    (datawidth+(hdr+ma-(case when hdr%%ma=0 THEN ma ELSE hdr%%ma END)))::numeric AS datahdr,
-    (maxfracsum*(nullhdr+ma-(case when nullhdr%%ma=0 THEN ma ELSE nullhdr%%ma END))) AS nullhdr2
+    (datawidth+(hdr+ma-(case when hdr%ma=0 THEN ma ELSE hdr%ma END)))::numeric AS datahdr,
+    (maxfracsum*(nullhdr+ma-(case when nullhdr%ma=0 THEN ma ELSE nullhdr%ma END))) AS nullhdr2
   FROM (
     SELECT
       schemaname, tablename, hdr, ma, bs,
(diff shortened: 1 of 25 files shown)


internal/inspect/table_index_sizes/table_index_sizes.go
```diff
@@ -18,7 +18,7 @@ SELECT c.relname AS table,
   pg_size_pretty(pg_indexes_size(c.oid)) AS index_size
 FROM pg_class c
 LEFT JOIN pg_namespace n ON (n.oid = c.relnamespace)
-WHERE n.nspname NOT IN ('pg_catalog', 'information_schema')
+WHERE n.nspname NOT IN ('pg_catalog', 'information_schema', 'auth', 'realtime', 'pgsodium', 'supabase_migrations', 'vault', 'storage')
```
sweatybridge, line 21: Why are we moving some queries but not this one?

Can we also split the refactor into a separate commit so it's easier to review what's changed?

There's a full list of supabase managed schemas defined in misc.go. I think we should use it instead.

encima: missed the removal of that query 🙏

internal/inspect/queries.go
```diff
@@ -0,0 +1,334 @@
...
+SELECT
+  'index' as type,
+  schemaname,
+  tablename || '::' || iname as object_name,
+  ROUND(CASE WHEN iotta=0 OR ipages=0 THEN 0.0 ELSE ipages/iotta::numeric END,1) AS bloat,
+  CASE WHEN ipages < iotta THEN '0' ELSE (bs*(ipages-iotta))::bigint END AS raw_waste
+FROM
+  index_bloat) bloat_summary
+WHERE
+ schemaname NOT IN (%s)
```
sweatybridge, line 82:
```suggestion
 schemaname NOT IN ($1)
```

I would use a placeholder here and pass in an array of strings to pgx client, for eg. 
 `conn.Query(sql, excluded)`. It's more robust than golang format strings and will probably simplify the code a little bit.

sweatybridge (commented):
Just one more comment. Everything else looks good.

internal/inspect/table_sizes/table_sizes.go
```diff
@@ -18,7 +18,7 @@ SELECT c.relname AS name,
   pg_size_pretty(pg_table_size(c.oid)) AS size
 FROM pg_class c
 LEFT JOIN pg_namespace n ON (n.oid = c.relnamespace)
-WHERE n.nspname NOT IN ('pg_catalog', 'information_schema')
+WHERE n.nspname NOT IN ('pg_catalog', 'information_schema', 'auth', 'realtime', 'pgsodium', 'supabase_migrations', 'vault', 'storage')
```
sweatybridge, line 21: We missed out this one too.

internal/inspect/queries.go
```diff
@@ -0,0 +1,299 @@
...
+
+const CALLS_QUERY = `SELECT
+'index hit rate' AS name,
+(sum(idx_blks_hit)) / nullif(sum(idx_blks_hit + idx_blks_read),0) AS ratio
+FROM pg_statio_user_indexes
+UNION ALL
+SELECT
+'table hit rate' AS name,
+sum(heap_blks_hit) / nullif(sum(heap_blks_hit) + sum(heap_blks_read),0) AS ratio
+FROM pg_statio_user_tables`
```
sweatybridge, line 102:
```suggestion
const CALLS_QUERY = `SELECT
	query,
	interval '1 millisecond' * total_exec_time AS total_exec_time,
	to_char((total_exec_time/sum(total_exec_time) OVER()) * 100, 'FM90D0') || '%'  AS prop_exec_time,
	to_char(calls, 'FM999G999G990') AS ncalls,
	interval '1 millisecond' * (blk_read_time + blk_write_time) AS sync_io_time
FROM pg_stat_statements	
ORDER BY calls DESC
LIMIT 10`
```

internal/inspect/table_sizes/table_sizes.go
encima: that was probably my poor rebasing! fixed now!

encima: see that you fixed this already, removed my commit

internal/inspect/queries.go
```diff
@@ -0,0 +1,299 @@
...
+ORDER BY calls DESC
+LIMIT 10`
+
+const INDEX_SIZE_QUERY = `SELECT c.relname AS name,
+pg_size_pretty(sum(c.relpages::bigint*8192)::bigint) AS size
+FROM pg_class c
+LEFT JOIN pg_namespace n ON (n.oid = c.relnamespace)
+WHERE n.nspname NOT IN (%1)
+AND n.nspname !~ '^pg_toast'
+AND c.relkind='i'
```
sweatybridge, line 110:
```suggestion
AND c.relkind = 'i'
```

This exclude clause is now redundant

internal/inspect/queries.go
```diff
@@ -0,0 +1,299 @@
...
+pg_stat_user_tables
+ORDER BY seq_scan DESC`
+
+const TABLE_INDEX_SIZE_QUERY = `SELECT c.relname AS table,
+pg_size_pretty(pg_indexes_size(c.oid)) AS index_size
+FROM pg_class c
+LEFT JOIN pg_namespace n ON (n.oid = c.relnamespace)
+WHERE NOT n.nspname LIKE ANY($1)
+AND n.nspname !~ '^pg_toast'
+AND c.relkind='r'
```
sweatybridge, line 208:
```suggestion
AND c.relkind = 'r'
```

Ditto

internal/inspect/queries.go
```diff
@@ -0,0 +1,299 @@
...
+ORDER BY
+n_live_tup DESC`
+
+const TABLE_SIZE_QUERY = `SELECT c.relname AS name,
+pg_size_pretty(pg_table_size(c.oid)) AS size
+FROM pg_class c
+LEFT JOIN pg_namespace n ON (n.oid = c.relnamespace)
+WHERE NOT n.nspname LIKE ANY($1)
+AND n.nspname !~ '^pg_toast'
+AND c.relkind='r'
```
sweatybridge, line 225:
```suggestion
AND c.relkind = 'r'
```

Ditto

internal/inspect/queries.go
```diff
@@ -0,0 +1,299 @@
...
+ORDER BY pg_table_size(c.oid) DESC`
+
+const TOTAL_INDEX_SIZE_QUERY = `SELECT pg_size_pretty(sum(c.relpages::bigint*8192)::bigint) AS size
+FROM pg_class c
+LEFT JOIN pg_namespace n ON (n.oid = c.relnamespace)
+WHERE NOT n.nspname LIKE ANY($1)
+AND n.nspname !~ '^pg_toast'
+AND c.relkind='i'`
```
sweatybridge, line 233:
```suggestion
AND c.relkind = 'i'`
```

internal/inspect/queries.go
```diff
@@ -0,0 +1,299 @@
...
+
+const TOTAL_TABLE_SIZE_QUERY = `SELECT c.relname AS name,
+pg_size_pretty(pg_total_relation_size(c.oid)) AS size
+FROM pg_class c
+LEFT JOIN pg_namespace n ON (n.oid = c.relnamespace)
+WHERE NOT n.nspname LIKE ANY($1)
+AND n.nspname !~ '^pg_toast'
+AND c.relkind='r'
```
sweatybridge, line 241:
```suggestion
AND c.relkind = 'r'
```

encima: want me to remove these or do you want to @sweatybridge?

sweatybridge: I just committed them. Do you mind running through all the commands again to confirm everything works?

sweatybridge (approved):

sweatybridge (approved):
I've verified all commands locally and fixed the result types where appropriate.

A new change I made is to include managed schemas like auth, storage, extensions, etc. when inspecting table sizes, ie. `table-records-count`, `table-sizes`, `total-table-sizes`. Because records in these schemas are mostly inserted by the user, it would be a good idea to surface their sizes. Schemas with `pg_*` prefix and `information_schema` are still excluded from these queries.

In the future, we can perhaps move queries.go to individual `*.sql` files for better syntax checks.

PR #2145, opened 2024-04-12
fix: bump edge-runtime to 1.41.5

laktek:

Updates to edge-runtime 1.41.5, which fixes an oddity when 

Previously if a relative path import like `../shared/index.ts` was included in the function bundle, it would get unbundled in to something like `./home/deno/functions/shared/index.ts`

## What is the new behavior?

Now unbundle command would create a `function-name/src` directory for the source. Any relative imports will be placed in top-level `function-name/` directory.

### DIFF
--- a/internal/utils/misc.go
+++ b/internal/utils/misc.go
@@ -36,7 +36,7 @@ const (
 	PgmetaImage      = "supabase/postgres-meta:v0.80.0"
 	StudioImage      = "supabase/studio:20240326-5e5586d"
 	ImageProxyImage  = "darthsim/imgproxy:v3.8.0"
-	EdgeRuntimeImage = "supabase/edge-runtime:v1.41.3"
+	EdgeRuntimeImage = "supabase/edge-runtime:v1.41.5"
 	VectorImage      = "timberio/vector:0.28.1-alpine"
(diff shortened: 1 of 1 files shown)


sweatybridge:
> Now unbundle command would create a function-name/src directory for the source. Any relative imports will be placed in top-level function-name/ directory.

I don't think adding `src` is the right approach. It would be saner to unbundle the closest ancestor to the output directory. For example,

If we have these 2 files in eszip
```
file:///home/deno/functions/slug/index.ts
file:///home/deno/shared/index.ts
```

And I specify output directory as `/tmp`, the following files should be extracted
```
/tmp/functions/slug/index.ts
/tmp/shared/index.ts
```

Because their common ancestor begins at `/home/deno` and can be removed.

As another example, if I have these 2 files

```
file:///home/functions/hello/index.ts
file:///home/functions/shared/index.ts
```

With the same output directory `/tmp`, I would expect the extracted paths to be
```
/tmp/slug/index.ts
/tmp/shared/index.ts
```

laktek:
Closing this - This change was reverted in edge-runtime https://github.com/supabase/edge-runtime/releases/tag/v1.43.1

PR #2149, opened 2024-04-13
fix: migration repair example more understandable

adrientiburce:

 docs update

The example provided with the command `migragrion repair` was showing an output with no issue and thus the example was hard to understand

## What is the new behavior?

NA


sweatybridge (approved):
Thanks for the suggestion. I made a minor edit and approved.

PR #2203, opened 2024-04-23
fix: update to the latest API spec

darora:


sweatybridge (approved):
looks ok as long as tests pass

Issue #2207, opened 2024-04-24
Cannot reset database because "supabase db reset" tries to possibly drop types in wrong order


hammerlscs:
**Describe the bug**

We cannot reset the database of our Supabase **testing** instance:

```
npx supabase db reset --linked --debug --create-ticket
Using connection pooler: postgres://...
Supabase CLI 1.163.2
Do you want to reset the remote database? [y/N]
y
Resetting remote database...

...

2024/04/24 13:31:55 PG Recv: {"Severity":"NOTICE","SeverityUnlocalized":"NOTICE","Code":"00000","Message":"truncate cascades to table \"mfa_amr_claims\"","Detail":"","Hint":"","Position":0,"InternalPosition":0,"InternalQuery":"","Where":"SQL statement \"truncate auth.sessions restart identity cascade\"\nPL/pgSQL function inline_code_block line 44 at EXECUTE","SchemaName":"","TableName":"","ColumnName":"","DataTypeName":"","ConstraintName":"","File":"tablecmds.c","Line":1826,"Routine":"ExecuteTruncateGuts","UnknownFields":null}
2024/04/24 13:31:55 PG Recv: {"Type":"ErrorResponse","Severity":"ERROR","SeverityUnlocalized":"ERROR","Code":"2BP01","Message":"cannot drop type our_enum_type[] because type our_enum_type requires it","Detail":"","Hint":"You can drop type our_enum_type instead.","Position":0,"InternalPosition":0,"InternalQuery":"","Where":"SQL statement \"drop type if exists public._our_enum_type cascade\"\nPL/pgSQL function inline_code_block line 64 at EXECUTE","SchemaName":"","TableName":"","ColumnName":"","DataTypeName":"","ConstraintName":"","File":"dependency.c","Line":828,"Routine":"findDependentObjects","UnknownFields":null}
2024/04/24 13:31:55 PG Recv: {"Type":"ReadyForQuery","TxStatus":"I"}
2024/04/24 13:31:55 PG Send: {"Type":"Terminate"}
ERROR: cannot drop type our_enum_type[] because type our_enum_type requires it (SQLSTATE 2BP01)  
```

Our theory so far is, that the order in pg_type is responsible for this problem.

This is the part of the code which is executed by `supabase db reset` to drop all types:

```sql
for rec in
    select *
    from pg_type t
    where t.typnamespace::regnamespace::name = 'public'
loop
[...]

PR #2208, opened 2024-04-24
fix: add fixed order when selecting types to drop (fixes issue #2207)

hammerlscs:

Adds a fixed order when selecting the types to drop for the `supabase db reset`command.

Currently, the order is not defined and this can lead to errors when dropping types. (More details: https://github.com/supabase/cli/issues/2207)

## What is the new behavior?

The types are dropped in a fixed order. The order is the number of type dependencies ascending.

The idea is that "base" types are always dropped before e.g. "array" types which have at least 1 type dependency to its "base" type.

### DIFF
--- a/internal/db/reset/templates/drop.sql
+++ b/internal/db/reset/templates/drop.sql
@@ -59,7 +59,8 @@ begin
   for rec in
     select *
     from pg_type t
-    where t.typnamespace::regnamespace::name = 'public'
+    where
+      t.typnamespace::regnamespace::name = 'public'
       and typtype != 'b'
   loop
     execute format('drop type if exists %I.%I cascade', rec.typnamespace::regnamespace::name, rec.typname);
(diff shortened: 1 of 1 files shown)


internal/db/reset/templates/drop.sql
```diff
@@ -60,6 +60,13 @@ begin
     select *
     from pg_type t
     where t.typnamespace::regnamespace::name = 'public'
+    order by (
+      select count(d.objid)
+      from pg_depend d
+        left join pg_class c on d.refclassid = c.oid
+      where t.oid = d.objid
+        and c.relname = 'pg_type'
+    )
```
sweatybridge, line 69:
```suggestion
      and typtype != 'b'
```

Filter out any base types as discussed.

sweatybridge (approved):

PR #2246, opened 2024-05-06
feat: add `inspect report` subcommand and embed queries as SQL scripts

encima:

Chore: split queries out into embedded SQL scripts
Feature: add an `inspect db report` subcommand that runs all queries and saves their output to CSV files in the current directory

Queries currently stored inside a go submodule

## What is the new behavior?

Embedded SQL scripts and report subcommand for `inspect` that runs all queries and saves their output to the current directory as CSV files

## Additional context


KevinBrolly:
LGTM but still needs review from CLI team 👍

internal/inspect/utils.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/spf13/cobra"
+	"github.com/supabase/cli/internal/db/reset"
+	"github.com/supabase/cli/internal/utils"
+)
+
+const ROOT_DIR = "queries/"
+const CSV_QUERY = `COPY (%s) TO STDOUT WITH CSV HEADER`
+
+//go:embed queries/*.sql
+var queries embed.FS
```
sweatybridge, line 24: The problem with embed.FS is that misspelt paths become runtime errors. This would have been fine if we had unit test coverage for all inspect commands. Since we are doing reliability month, could you look into adding them?

Another alternative is to embed each file as a separate variable.

internal/inspect/utils.go
```diff
@@ -0,0 +1,68 @@
...
+	return string(queryString)
+}
+
+func Report(cmd *cobra.Command, config pgconn.Config, fsys afero.Fs, options ...func(*pgx.ConnConfig)) error {
+	queries, err := queries.ReadDir("queries")
+	if err != nil {
+		return err
+	}
+	date := time.Now().Format("2006-01-02")
+	out, err := cmd.Flags().GetString("out-path")
```
sweatybridge, line 42: Instead of supporting a flag, can we print everything to stdout and let users redirect to a file?

We can be more flexible with output format this way, as users can choose json / yaml / csv / etc.

internal/inspect/utils.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	if out != "" {
+		err := os.MkdirAll(out, os.ModePerm)
+		if err != nil {
+			return errors.Errorf("failed to create output directory: %w", err)
+		}
+	} else {
+		out = "./"
+	}
+	fmt.Println("Running queries...")
```
sweatybridge, line 54:
```suggestion
	fmt.Fprintln(os.Stderr, "Running queries...")
```

Progress logs should be printed to stderr by POSIX standard. https://pubs.opengroup.org/onlinepubs/9699919799/functions/stderr.html


> standard error (for writing diagnostic output)


This allows other programs to consume from piped stdout.

internal/inspect/utils.go
```diff
@@ -0,0 +1,68 @@
...
+	for _, v := range queries {
+		name := strings.Split(v.Name(), ".")[0]
+		query := ReadQuery(name)
+		fq := strings.Replace(query, "$1", "'{"+strings.Join(reset.LikeEscapeSchema(utils.InternalSchemas), ",")+"}'::text[]", -1)
+		copyCmd := fmt.Sprintf(`COPY (%s) TO STDOUT WITH CSV HEADER`, fq)
+		// gosec lint error is unavoidable here because of formatted params
+		cmd := exec.CommandContext(cmd.Context(), "psql", utils.ToPostgresURL(config), "-At", "-F\",\"", "-c", copyCmd, "-o", fmt.Sprintf("%s/%s_%s.csv", out, name, date)) //nolint:gosec
```
sweatybridge, line 61: Let's not introduce a binary dependency on psql. Ideally we want to use pgx library for consistency with all other commands.

sweatybridge (changes requested):
Adding a reports command is nice. Apart from those review comments, another suggestion is to wrap all queries in a transaction so they are measuring the same state of the db.

internal/inspect/utils.go
encima: I would advise against this because special chars can cause users to need to run `reset` and larger outputs would go beyond the scrollback history. 
If they want to do something with the output, then they could run each command in the current behaviour

internal/inspect/utils.go
encima: That was my first efforts as well. Unfortunately, it is not possible for `pgx` to do this. `COPY` is only supported TO the database and not from it.
Potentially the `database/sql` module supports it but I did not see it.

The other option, which I not keen on, is to convert the output from pgx to CSV and then direct to a file

internal/inspect/utils.go
encima: Will look into adding unit tests, yep!

encima:
> Adding a reports command is nice. Apart from those review comments, another suggestion is to wrap all queries in a transaction so they are measuring the same state of the db.

This is certainly possible, although the current state is not time critical in my mind. The report is useful for insights and debugging but not realtime insights. It could be useful to remove more time-sensitive queries (like `locks`) from the command

internal/inspect/utils.go
sweatybridge: > The other option, which I not keen on, is to convert the output from pgx to CSV and then direct to a file

What's your concern with this approach?

IMO, this is the nicer option as reports would run queries exactly the same way as other inspect commands. It's more testable and we can avoid doing our own string substitution for $1 template variable.

encima: Pros:
- type safety
- more control over output and errors
Cons:
- cost to implement is prohibitive (I am unlikely to do it 😅 )
- change to queries impacts output
- not the "postgres" way as we are wrapping query output that is already formatted

internal/inspect/utils.go
sweatybridge: Ok, we can keep this flag since it writes to multiple files. Shall we rename it as `output-dir` so it's clearer?

internal/inspect/utils.go
sweatybridge: I can look into implementing it using pgx since I'm more familiar with the codebase. Probably after reliability improvements.

encima: welp, I was wrong. I saw v4 dropped `CopyToWriter` so I assumed it was no longer supported. Pushed an update to drop `psql` requirement, sorry about that!

encima:
Failing only on tests about table sizes and record counts, oddly.

PR #2284, opened 2024-05-14
fix: bump edge-runtime to 1.51.0

nyannyacha:

Bump edge-runtime to 1.51.0


nyannyacha:
Will close due to stale.

PR #2339, opened 2024-05-29
fix: pass jwt secret to studio container

sweatybridge:

relates to https://github.com/supabase/supabase/pull/26858

## What is the new behavior?

- [ ] Bump studio version once upstream is merged.

## Additional context

### DIFF
--- a/internal/start/start.go
+++ b/internal/start/start.go
@@ -835,6 +835,7 @@ EOF
 					"PG_META_DB_USER=" + dbConfig.User,
 					fmt.Sprintf("PG_META_DB_PORT=%d", dbConfig.Port),
 					"PG_META_DB_PASSWORD=" + dbConfig.Password,
+					"AUTH_JWT_SECRET=" + utils.Config.Auth.JwtSecret,
 				},
 				Healthcheck: &container.HealthConfig{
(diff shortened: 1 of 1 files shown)


barrownicholas:
@sweatybridge I'm trying to debug an issue related to the endpoint `/api/projects/default/config`, but I am getting a `-` instead of a valid JWT. 

Screenshot 2024-07-01 at 8 53 07 PM


I stumbled across this PR while researching and debugging; I noticed that in this PR, you put `"AUTH_JWT_SECRET=" + utils.Config.Auth.JwtSecret` under `utils.Config.Studio.PgmetaImage` instead of under `utils.Config.Studio.Image`. 

I was slightly confused, only because pg_meta does not seem to rely on the env var `AUTH_JWT_SECRET`: https://github.com/search?q=repo%3Asupabase%2Fpostgres-meta%20AUTH_JWT_SECRET&type=code

I am using CLI `1.178.2` if that helps (which I believe wouldv'e included this PR, which seemed to ship around 1.174). 

I just have no idea where the `-` would be coming from for the `/api/projects/default/config` endpoint.

sweatybridge:
Ah yes, it was an oversight on my side. I will add the env var to studio instead.

barrownicholas:
awesome, thank you @sweatybridge !

PR #2353, opened 2024-05-31
feat: replace pgbouncer with supavisor

sweatybridge:

feature

## What is the new behavior?

Replaces deprecated pgbouncer with supavisor while maintaining the same config.

## Additional context

### DIFF
--- a/internal/utils/misc.go
+++ b/internal/utils/misc.go
@@ -118,6 +118,7 @@ var (
 	ManagedSchemas = append([]string{
 		"_analytics",
 		"_realtime",
+		"_supavisor",
 		"pgbouncer",
 		"pgsodium",
 		"pgtle",
@@ -125,13 +126,14 @@ var (
 		"vault",
 	}, PgSchemas...)
 	InternalSchemas = append([]string{
+		"_analytics",
+		"_realtime",
+		"_supavisor",
 		"auth",
 		"extensions",
 		"pgbouncer",
 		"realtime",
-		"_realtime",
 		"storage",
-		"_analytics",
 		"supabase_functions",
 		"supabase_migrations",
 		// Owned by extensions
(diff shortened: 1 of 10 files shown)


internal/utils/misc.go
```diff
@@ -38,7 +38,7 @@ const (
 	ImageProxyImage  = "darthsim/imgproxy:v3.8.0"
 	EdgeRuntimeImage = "supabase/edge-runtime:v1.53.3"
 	VectorImage      = "timberio/vector:0.28.1-alpine"
-	PgbouncerImage   = "bitnami/pgbouncer:1.20.1-debian-11-r39"
+	SupavisorImage   = "supabase/supavisor:1.1.56"
```
abc3, line 41: Out of curiosity, why not use the latest version here? Does it mean that the pooler team should constantly update the version in this repo? For example, the RT version is too old (April 3) 

cc @filipecabaco

sweatybridge: We can't use `latest` tag in our default registry ECR because it requires immutable tags.

Typically other services like auth and storage expose an endpoint that returns the currently deployed version string. CLI queries that endpoint to update user's local environment. This way pooler team doesn't need to constantly update the version here.

When there's major features / releases, it's typically better to bump the version manually so it's more explicit when we deprecate certain configs, like statement mode, or add new config options to support new features.

abc3: > Typically other services like auth and storage expose an endpoint that returns the currently deployed version string

Supavisor can provide this as well

```bash
$ curl -s -i http://aws-0-eu-west-3.pooler.supabase.com:5432 | grep 'x-app-version'
x-app-version: 1.1.49
```

sweatybridge: That's nice, I will make a separate PR to sync the pooler version.

bombillazo:
Seems like this feature is causing https://github.com/supabase/cli/issues/2440

PR #2418, opened 2024-06-12
fix: bump logflare version

Ziinc:
This will include bugfix for https://github.com/supabase/supabase/issues/22605, as well as many other improvements to the ingestion pipeline


sweatybridge (approved):

Ziinc:
on hold, need to do a hotfix.

PR #2444, opened 2024-06-19
fix(realtime): bump up version to 2.29.13

filipecabaco:

This removes some complexity from starting up realtime into realtime startup process. To know more please check https://github.com/supabase/realtime/pull/1081

### DIFF
--- a/internal/db/start/start.go
+++ b/internal/db/start/start.go
@@ -223,6 +223,31 @@ func initSchema14(ctx context.Context, conn *pgx.Conn) error {
 func initSchema15(ctx context.Context, host string) error {
 	// Apply service migrations
 	logger := utils.GetDebugLogger()
+	if err := utils.DockerRunOnceWithStream(ctx, utils.Config.Realtime.Image, []string{
+		"PORT=4000",
+		"DB_HOST=" + host,
+		"DB_PORT=5432",
+		"DB_USER=supabase_admin",
+		"DB_PASSWORD=" + utils.Config.Db.Password,
+		"DB_NAME=postgres",
+		"DB_AFTER_CONNECT_QUERY=SET search_path TO _realtime",
+		"DB_ENC_KEY=" + utils.Config.Realtime.EncryptionKey,
+		"API_JWT_SECRET=" + utils.Config.Auth.JwtSecret,
+		"METRICS_JWT_SECRET=" + utils.Config.Auth.JwtSecret,
+		"APP_NAME=realtime",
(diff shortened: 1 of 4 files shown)


filipecabaco:
Indeed it seems that schema init is blocked due to Realtime, will better investigate tomorrow but I suspect that since we did some changes to how the application starts that is interfering with the seeding

internal/db/start/start.go
```diff
@@ -223,30 +223,6 @@ func initSchema14(ctx context.Context, conn *pgx.Conn) error {
...
-		"ENABLE_TAILSCALE=false",
-		"DNS_NODES=''",
-		"RLIMIT_NOFILE=",
-		fmt.Sprintf("MAX_HEADER_LENGTH=%d", utils.Config.Realtime.MaxHeaderLength),
-	}, []string{"/bin/sh", "-c", fmt.Sprintf(
-		`/app/bin/migrate && /app/bin/realtime eval 'Realtime.Release.seeds(Realtime.Repo); Application.ensure_all_started(:realtime); Realtime.Tenants.health_check("%s")'`,
-		utils.Config.Realtime.TenantId,
-	)}, io.Discard, logger); err != nil {
-		return err
-	}
```
sweatybridge, line 249: We want to keep this docker run for initialising realtime schema, similar to auth and storage. This is so that user migrations can reference the realtime schema to add custom RLS policies. See also https://supabase.slack.com/archives/C03S3SW9485/p1713365735529599?thread_ts=1711469985.749159&channel=C03S3SW9485&message_ts=1713365735.529599

filipecabaco: since now we migrate automatically on realtime start isn't this covered? or is it that question of being able migrate without starting realtime?

sweatybridge: > or is it that question of being able migrate without starting realtime?

Yes mostly that. Users can exclude some services from starting locally via config or `--exclude` flag but we still want to run all service migrations so that the database ends up with the same schema as hosted.

filipecabaco: sounds good will add back 👍

filipecabaco: I'm unable to trigger this path of the code, how do we run it normally?

I was trying with `go run . start` as that was one of the things "freezing" when realtime had a broken migration 🤔

sweatybridge: Using `go run . db start` would be faster. But `go run . start` should also work.

sweatybridge (approved):

PR #2460, opened 2024-06-26
feat(storage): seed buckets locally from config file

### LINKED ISSUE #2413: Support declaring storage buckets in config.toml
Currently the only reliable way to create a bucket is by calling storage-api locally, either via storage-js script or cli storage commands.

Allow adding buckets via config.toml

```toml
[storage.buckets.name]
public = true
```

Inserting bucket via migration file is prone to errors as storage-api could change the buckets schema in the future.

**Additional context**

sweatybridge:

closes #2413 

## What is the new behavior?

TODO:
- [x] push buckets on `supabase seed buckets [--local|--linked]`
- [x] overwrite existing bucket properties
- [ ] sync buckets on `supabase link`?

## Additional context

### DIFF
--- a/cmd/seed.go
+++ b/cmd/seed.go
@@ -1,0 +1,41 @@
+package cmd
+
+import (
+	"os"
+	"os/signal"
+
+	"github.com/spf13/cobra"
+	"github.com/supabase/cli/internal/seed/buckets"
+	"github.com/supabase/cli/internal/utils"
+)
+
+var (
+	seedCmd = &cobra.Command{
+		GroupID: groupLocalDev,
+		Use:     "seed",
+		Short:   "Seed a Supabase project from " + utils.ConfigPath,
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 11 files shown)


lightstrike:
@sweatybridge I needed this today and just upgraded to the latest CLI version to use it. The core functionality worked like a charm, thanks! 🎉 

One thing I noticed is that the `S3_*` variables in `[experimental]` are now required, throwing the following error when not included in `config.toml`:

`Error evaluating "env(S3_HOST)": environment variable S3_HOST is unset.`

I presume that this is unintended and that all experimental variables should be considered optional? In my case, I added the following to `config.toml` to get everything working locally:
```
[experimental]
S3_HOST="http://127.0.0.1:54321/storage/v1/s3"
S3_REGION=""
S3_ACCESS_KEY=""
S3_SECRET_KEY=""
```

knksmith57:
^ Confirming that this same `S3_*` breaking change bit us today and that adding the snippet from above to our `config.toml` also resolved the problem.

Thanks for the assist, @lightstrike 🙏

PR #2482, opened 2024-07-02
fix: bump studio to latest release

saltcod:
bump studio version


sweatybridge (approved):
Looks ok on my local machine. I remember there were some issues with viewing logs but they seem resolved.

PR #2529, opened 2024-07-18
fix: import edge runtime types from jsr

thorwebdev:
https://supabase.com/blog/supabase-js-on-jsr

### DIFF
--- a/internal/functions/new/templates/index.ts
+++ b/internal/functions/new/templates/index.ts
@@ -3,7 +3,7 @@
 // This enables autocomplete, go to definition, etc.
 
 // Setup type definitions for built-in Supabase Runtime APIs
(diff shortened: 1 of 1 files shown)


internal/functions/new/templates/index.ts
```diff
@@ -3,7 +3,7 @@
 // This enables autocomplete, go to definition, etc.
 
 // Setup type definitions for built-in Supabase Runtime APIs
-import "https://esm.sh/@supabase/functions-js/src/edge-runtime.d.ts"
+import "jsr:@supabase/functions-js/edge-runtime.d.ts";
```
thorwebdev, line 6:
```suggestion
import "jsr:@supabase/functions-js/edge-runtime.d.ts"
```

PR #2582, opened 2024-08-01
feat: add MFA (Phone) related config

J0:

Introduces the supporting Auth configuration for  MFA (Phone). This include:

1. The respective enable / disable flows for each toggle `MFA.Phone.EnrollEnabled` and `MFA.Phone.VerifyEnabled`
2. `MFA.MaxEnrolledFactors`  which is already exposed on the platform
4. Phone also comes with a configurable OTP Length as well as Template. We don't include the provider specific configuration as they are derived from the primary provider. We also expose MaxFrequency

### DIFF
--- a/internal/start/start.go
+++ b/internal/start/start.go
@@ -616,8 +616,6 @@ EOF
 				env,
 				"GOTRUE_MFA_PHONE_ENROLL_ENABLED=true",
 				"GOTRUE_MFA_PHONE_VERIFY_ENABLED=true",
-				// TODO: Revisit once we have connected the "Enabled" field with the Phone fields
-
 			)
 		}
 		if utils.Config.Auth.MFA.TOTP.EnrollEnabled && utils.Config.Auth.MFA.TOTP.VerifyEnabled {
@@ -625,8 +623,6 @@ EOF
 				env,
 				"GOTRUE_MFA_TOTP_ENROLL_ENABLED=true",
 				"GOTRUE_MFA_TOTP_VERIFY_ENABLED=true",
-				// TODO: Revisit once we have connected the "Enabled" field with the Phone fields
-
 			)
 		}
 
--- a/internal/start/start.go
+++ b/internal/start/start.go
@@ -611,6 +611,24 @@ EOF
 				"GOTRUE_HOOK_SEND_EMAIL_SECRETS="+utils.Config.Auth.Hook.SendEmail.Secrets,
 			)
 		}
+		if utils.Config.Auth.MFA.Phone.Enroll.Enabled && utils.Config.Auth.MFA.Phone.Verify.Enabled {
+			env = append(
+				env,
+				"GOTRUE_MFA_PHONE_ENROLL_ENABLED=true",
+				"GOTRUE_MFA_PHONE_VERIFY_ENABLED=true"
+				// TODO: Revisit once we have connected the "Enabled" field with the Phone fields
+
+			)
+		}
+		if utils.Config.Auth.MFA.TOTP.Enroll.Enabled && utils.Config.Auth.MFA.TOTP.Verify.Enabled {
+			env = append(
+				env,
+				"GOTRUE_MFA_TOTP_ENROLL_ENABLED=true",
+				"GOTRUE_MFA_TOTP_VERIFY_ENABLED=true"
+				// TODO: Revisit once we have connected the "Enabled" field with the Phone fields
+
+			)
+		}
 
 		for name, config := range utils.Config.Auth.External {
 			env = append(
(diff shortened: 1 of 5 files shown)


internal/start/start.go
```diff
@@ -609,6 +609,20 @@ EOF
 				"GOTRUE_HOOK_SEND_EMAIL_SECRETS="+utils.Config.Auth.Hook.SendEmail.Secrets,
 			)
 		}
+		if utils.Config.Auth.MFA.Phone.EnrollEnabled && utils.Config.Auth.MFA.Phone.VerifyEnabled {
+			env = append(
+				env,
+				"GOTRUE_MFA_PHONE_ENROLL_ENABLED=true",
```
J0, line 615:
```suggestion
				"GOTRUE_MFA_PHONE_ENROLL_ENABLED=false",
```

hf: Yup this should be false

hf (commented):
Seems fine.

J0:
How this was tested:

Place this set of entries in the `config.toml`

```
[auth.mfa.totp]
enroll_enabled = true
verify_enabled = true

[auth.mfa]
max_enrolled_factors = 20

[auth.mfa.phone]
enroll_enabled = true
verify_enabled = true
otp_length = 6
template = "Your Code is {{ .Code }}"
max_frequency = "10s"
```

Run `docker inspect` on the Auth container and check that the Auth config is appropriately set

pkg/config/config.go
```diff
@@ -293,6 +294,27 @@ type (
...
+		VerifyEnabled bool `toml:"verify_enabled"`
+	}
+	totpFactorTypeConfiguration struct {
+		factorTypeConfiguration
+	}
+
+	phoneFactorTypeConfiguration struct {
+		factorTypeConfiguration
+		OtpLength    int           `toml:"otp_length"`
+		Template     string        `toml:"template"`
```
sweatybridge, line 308: Is this referring to a template string or path to a template file?

pkg/config/config.go
```diff
@@ -293,6 +294,27 @@ type (
...
+		EnrollEnabled bool `toml:"enroll_enabled"`
+		VerifyEnabled bool `toml:"verify_enabled"`
+	}
+	totpFactorTypeConfiguration struct {
+		factorTypeConfiguration
+	}
+
+	phoneFactorTypeConfiguration struct {
+		factorTypeConfiguration
+		OtpLength    int           `toml:"otp_length"`
```
sweatybridge, line 307:
```suggestion
		OtpLength    uint          `toml:"otp_length"`
```

pkg/config/config.go
```diff
@@ -293,6 +294,27 @@ type (
...
+		Template     string        `toml:"template"`
+		MaxFrequency time.Duration `toml:"max_frequency"`
+	}
+
+	mfa struct {
+		TOTP  totpFactorTypeConfiguration  `toml:"totp"`
+		Phone phoneFactorTypeConfiguration `toml:"phone"`
+		// set to float64 for backward compatibility
+		MaxEnrolledFactors float64 `toml:"max_enrolled_factors"`
```
sweatybridge, line 316: Do we need to consider backwards compatibility if it's a new config? probably use `uint` or the correct type.

J0: Yeah good point - the CLI config doesn't need to be compatible against the underlying env config. Let's change

pkg/config/config.go
J0: This should refer to a Template string. Perhaps can add a comment?

Can change the name of course but the config name in Auth is `Template` so hoping not to deviate.

Open to other suggestions as well

pkg/config/config.go
```diff
@@ -294,6 +295,27 @@ type (
...
+		OtpLength    uint          `toml:"otp_length"`
+		Template     string        `toml:"template"`
+		MaxFrequency time.Duration `toml:"max_frequency"`
+	}
+
+	mfa struct {
+		TOTP  totpFactorTypeConfiguration  `toml:"totp"`
+		Phone phoneFactorTypeConfiguration `toml:"phone"`
+		// set to float64 for backward compatibility
+		MaxEnrolledFactors float64 `toml:"max_enrolled_factors"`
```
J0, line 317:
```suggestion
		MaxEnrolledFactors uint `toml:"max_enrolled_factors"`
```

J0:
Deploy has completed

pkg/config/config.go
```diff
@@ -294,6 +295,27 @@ type (
 		SendSMS                     hookConfig `toml:"send_sms"`
 		SendEmail                   hookConfig `toml:"send_email"`
 	}
+	factorTypeConfiguration struct {
+		EnrollEnabled bool `toml:"enroll_enabled"`
+		VerifyEnabled bool `toml:"verify_enabled"`
+	}
+	totpFactorTypeConfiguration struct {
+		factorTypeConfiguration
+	}
```
sweatybridge, line 304:
```suggestion
```

Let's defer adding this type until we actually need to extend it?

pkg/config/config.go
```diff
@@ -294,6 +295,27 @@ type (
...
+
+	phoneFactorTypeConfiguration struct {
+		factorTypeConfiguration
+		OtpLength    uint          `toml:"otp_length"`
+		Template     string        `toml:"template"`
+		MaxFrequency time.Duration `toml:"max_frequency"`
+	}
+
+	mfa struct {
+		TOTP  totpFactorTypeConfiguration  `toml:"totp"`
```
sweatybridge, line 314:
```suggestion
		TOTP  factorTypeConfiguration       `toml:"totp"`
```

pkg/config/config.go
sweatybridge: Always good to add an example entry to [config.toml](https://github.com/supabase/cli/blob/develop/pkg/config/templates/config.toml#L116), for eg.

```
[auth.mfa.phone]
enroll_enabled = true
...
```

sweatybridge (approved):

PR #2595, opened 2024-08-07
feat: add third-party auth support for local dev

hf:
Adds support for configuring a Third-Party Auth provider to be used alongside Supabase Auth.

Internally both Storage, Realtime and PostgREST support this if they're configured with a JWKS.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1114,7 +1114,7 @@ func (tpa *thirdParty) IssuerURL() string {
 // ResolveJWKS creates the JWKS from the JWT secret and Third-Party Auth
 // configs by resolving the JWKS via the OIDC discovery URL.
 // It always returns a JWKS string, except when there's an error fetching.
-func (a *auth) ResolveJWKS() (string, error) {
+func (a *auth) ResolveJWKS(ctx context.Context) (string, error) {
 	var jwks struct {
 		Keys []json.RawMessage `json:"keys"`
 	}
@@ -1123,12 +1123,14 @@ func (a *auth) ResolveJWKS() (string, error) {
 	if issuerURL != "" {
 		discoveryURL := issuerURL + "/.well-known/openid-configuration"
 
+		t := &http.Client{Timeout: 10 * time.Second}
 		client := fetcher.NewFetcher(
-			discoveryURL,
+			issuerURL,
+			fetcher.WithHTTPClient(t),
 			fetcher.WithExpectedStatus(http.StatusOK),
 		)
 
-		resp, err := client.Send(context.Background(), http.MethodGet, "", nil)
+		resp, err := client.Send(ctx, http.MethodGet, "", nil)
 		if err != nil {
 			return "", err
 		}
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -837,6 +853,45 @@ func (c *config) Validate() error {
...
 		}
 	}
+	// Validate Third-Party Auth config
+	if c.ThirdPartyAuth.Enabled {
+		allDisabled := true
+		allDisabled = allDisabled && c.ThirdPartyAuth.FirebaseProjectID == ""
+		allDisabled = allDisabled && c.ThirdPartyAuth.Auth0Tenant == ""
+		allDisabled = allDisabled && c.ThirdPartyAuth.Auth0TenantRegion == ""
+		allDisabled = allDisabled && c.ThirdPartyAuth.AWSCognitoUserPoolID == ""
+		allDisabled = allDisabled && c.ThirdPartyAuth.AWSCognitoUserPoolRegion == ""
```
sweatybridge, line 863: This is unfortunately verbose... I really want to start using https://github.com/go-playground/validator for our config validation soon.

pkg/config/templates/config.toml
```diff
@@ -176,6 +176,18 @@ url = ""
 # If enabled, the nonce check will be skipped. Required for local sign in with Google auth.
 skip_nonce_check = false
 
+# Use a Third-Party Auth provider alongside Supabas Auth. Configure only one of: Firebase Auth, Auth0 or AWS Congito User Pool.
+[third_party_auth]
```
sweatybridge, line 180:
```suggestion
[auth.third_party.firebase]
enabled = true
project_id = "my-firebase-project"

[auth.third_party.auth0]
enabled = true
tenant = "my-auth0-tenant"
tenant_region = "us"

[auth.third_party.cognito]
enabled = true
user_pool_id = "my-user-pool-id"
user_pool_region = "us-east-1"
```

May be something like this would be nice.

hf: This should be better now?

pkg/config/config.go
hf: Tried to simplify it by having separate validate functions on the structs.

pkg/config/templates/config.toml
sweatybridge: Yes, it's very clear now with the new validation logic 🚀

sweatybridge (approved):

pkg/config/config.go
```diff
@@ -978,3 +1016,184 @@ func ValidateBucketName(name string) error {
...
+		Keys []json.RawMessage `json:"keys"`
+	}
+
+	issuerURL := a.ThirdParty.IssuerURL()
+	if issuerURL != "" {
+		discoveryURL := issuerURL + "/.well-known/openid-configuration"
+
+		t := &http.Client{Timeout: 10 * time.Second}
+		client := fetcher.NewFetcher(
+			issuerURL,
```
dagingaa, line 1128: Should this have been discoveryURL?

dagingaa: Fixed in https://github.com/supabase/cli/pull/2655

PR #2610, opened 2024-08-13
fix: add comment for mfa settings

J0:

Add comments to clarify mfa config


pkg/config/templates/config.toml
```diff
@@ -157,13 +157,15 @@ max_frequency = "5s"
 enabled = false
 account_sid = ""
 message_service_sid = ""
+
```
sweatybridge, line 160:
```suggestion
```

sweatybridge (approved):

Issue #2632, opened 2024-08-23
URGENT: Supabase DB Dump Corrupting Data Due to UNIQ


tmountain:
**Describe the bug**

UPDATE: `supabase db dump` is running `uniq` on its output which removed any contingous lines that are identical. This is corrupting output data and needs an urgent fix. (see my comments below).

When I run `supabase db dump` I get back a different function definition than is present on our production host. When I run `pg_dump` on the same host, I get the correct definition. If I log into the host via the Supabase UI, I see the correct function def. If I query the function def via SQL, I get the correct result:

```
SELECT proname, prosrc  FROM pg_proc
WHERE proname = 'handle_lesson_completion';
```
**To Reproduce**

1. Dump the schema via `pg dump dump`

```
supabase db dump --db-url 'postgresql://postgres.<instance_id>:[Password Removed]@aws-0-us-east-1.pooler.supabase.com:6543/postgres' > prod.sql
```

2. Dump the same schema via `pg_dump`

```
pg_dump 'postgresql://postgres.<instance_id>:[Password Removed]@aws-0-us-east-1.pooler.supabase.com:6543/postgres' --schema-only -f
```

3. Review the `handle_lesson_completion` function. It is different between the two dumps.

**Expected behavior**

The functions should be the same.

**Screenshots**

Excerpt of correct function def:

<img width="514" alt="Screenshot 2024-08-23 at 3 49 09 PM" src="https://github.com/user-attachments/assets/fa2c238e-eed1-4799-a886-a4cbb5b0f443">

Excerpt of incorrect function def:

<img width="495" alt="Screenshot 2024-08-23 at 3 49 49 PM" src="https://github.com/user-attachments/assets/d6e65231-3ae6-4b7b-87e6-34b5884baae5">

**System information**
Rerun the failing command with `--create-ticket` flag.
No additional output is provided.

**Additional context**
I can provide my supabase instance id if needed.

tmountain:
I figured out the cause myself. This is a serious bug. If you do a `supabase db dump --dry-run`, you will see that it runs a series of sed commands to post-process the data that it generates from `pg_dump`.

```
pg_dump \
    --schema-only \
    --quote-all-identifier \
[...]

PR #2633, opened 2024-08-23
fix: remove uniq to avoid data corruption (#2632)

tmountain:

This resolves:

 https://github.com/supabase/cli/issues/2632

The current behavior is that contiguous unique data in sql dumps is removed (corrupting data).

https://github.com/supabase/cli/issues/2632

## What is the new behavior?

The new behavior is that contiguous data is preserved.

## Additional context


sweatybridge (approved):
Thanks you for the thorough investigation and fix!

Issue #2527, opened 2024-07-16
Allow the API_EXTERNAL_URL used by the gotrue service to be set


toBeOfUse:
**Is your feature request related to a problem? Please describe.**
All of the email verification links generated when using a Supabase instance created by the CLI start with http://127.0.0.1. I believe that this is because this CLI always sets the `API_EXTERNAL_URL` environment variable to http://127.0.0.1 when creating the supabase/gotrue container. This is a problem when creating a Supabase instance using the CLI, and then trying to use it from outside of the current host, where 127.0.0.1 no longer refers to the machine that the gotrue container is running on.

**Describe the solution you'd like**
Ideally, the hostname in the links that the gotrue service creates would not always be 127.0.0.1. I believe that if a configuration option to set the `API_EXTERNAL_URL` option existed, it would allow us to change the generated URLs to use a different address or domain.

**Describe alternatives you've considered**
If there is any alternative way to set the hostnames that are used in email verification URLs, that would also be great.

PR #2634, opened 2024-08-23
fix(api): add external_url while a replacement is found for auth/studio

StanGirard:
Fixes https://github.com/supabase/cli/issues/2527

Currently, the external_api_url is not read from the config.toml. 

I know there is an ongoing subject about mutualizing auth/studio url but it requires too much changes. 

This is a quick fix that improves the features of the CLI without removing any


sweatybridge (approved):
Yup, agree this is a good interim solution. Thanks for contributing!

PR #2643, opened 2024-08-30
feat: add secure password change config

silentworks:

This feature can be found in the Supabase Hosted Dashboard but doesn't have a way to configure it in the `config.toml`.

No way to set **Secure password change**

## What is the new behavior?

Can now set **Secure password change** in the `config.toml` file.

## Additional context

### DIFF
--- a/pkg/config/templates/config.toml
+++ b/pkg/config/templates/config.toml
@@ -109,7 +109,7 @@ enable_signup = true
 double_confirm_changes = true
 # If enabled, users need to confirm their email address before signing in.
 enable_confirmations = false
-# if enabled, users will need to be recently logged in to change their password without requiring reauthentication.
+# If enabled, users will need to reauthenticate or have logged in recently to change their password.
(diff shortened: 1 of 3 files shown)


pkg/config/templates/config.toml
```diff
@@ -109,6 +109,8 @@ enable_signup = true
 double_confirm_changes = true
 # If enabled, users need to confirm their email address before signing in.
 enable_confirmations = false
+# if enabled, users will need to be recently logged in to change their password without requiring reauthentication.
```
J0, line 112: /nit
```suggestion
# if enabled, users will need to reauthenticate or have logged in recently to change their password.
```

J0 (approved):

J0:
Probably don't need a test in `config_test.go` for this one since it should load similarly to the other auth config variables

sweatybridge:
Thanks for patching and reviewing!

Issue #2658, opened 2024-09-04
`supabase db reset` fails if storage is disabled


wyozi:
**Describe the bug**
If storage is disabled via supabase config.toml, resetting the database via `supabase db reset` fails

**To Reproduce**
1. Disable storage in config.toml (`[storage] enabled = false`)
2. Try to reset db with `supabase db reset`
3. "failed to inspect container health: Error response from daemon: No such container: supabase_storage_raxalle"

**Expected behavior**
Reset succeeds

**System information**
 - Version of CLI: 1.191.3

avallete:
Thanks for reporting this issue! I've reproduced the bug and I'm working on a fix. I'll update you once it's resolved.

PR #2663, opened 2024-09-09
fix: db reset with disabled services

avallete:

When performing the "reset db" check against the config to find "enabled" services and only restart the ones that are enabled. Also only "seed storage" service if the storage service is actually enabled.

Fixes https://github.com/supabase/cli/issues/2658

## What is the new behavior?

- Make "restartServices" only restart enabled services
- Skip the bucket seeding phase if "storage" service isn't enabled.

In terms of output there should be no differences.

### DIFF
--- a/internal/db/reset/reset.go
+++ b/internal/db/reset/reset.go
@@ -208,7 +208,6 @@ func RestartDatabase(ctx context.Context, w io.Writer) error {
 func restartServices(ctx context.Context) error {
 	// No need to restart PostgREST because it automatically reconnects and listens for schema changes
 	services := listServicesToRestart()
-
 	result := utils.WaitAll(services, func(id string) error {
 		if err := utils.Docker.ContainerRestart(ctx, id, container.StopOptions{}); err != nil && !errdefs.IsNotFound(err) {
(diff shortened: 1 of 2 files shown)


internal/db/reset/reset.go
```diff
@@ -217,7 +220,22 @@ func restartServices(ctx context.Context) error {
...
+		services = append(services, utils.RealtimeId)
+	}
+	if utils.Config.Db.Pooler.Enabled {
+		services = append(services, utils.PoolerId)
+	}
+	if utils.Config.Auth.Enabled {
+		services = append(services, utils.GotrueId)
+	}
+
+	return services
```
sweatybridge, line 238: I think this is unnecessary because if a service isn't already started, restarting it will throw 404 which is then ignored by cli https://github.com/supabase/cli/pull/2663/files#diff-e7e370d31621951333d60503924eb420ab55bf0bc8e903e7b86e1766baef60edR213

Additionally if user runs supabase start, then disable a service in config.toml, running db reset should probably stop the service instead of restarting. But to avoid complicating the logic, I think it's ok to let users manually stop and restart.

sweatybridge (approved):

Issue #2534, opened 2024-07-19
Allow to skip seed even if file exists


T04435:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I have to update/rename the seed.sql to avoid seeding data.

**Describe the solution you'd like**
Add CLI db reset option --skip-seed which will ignore the seed.sql

**Describe alternatives you've considered**
- rename seed.sql => _seed.sql every time I want to seed from a different source
- ...?

**Additional context**

sweatybridge:
@avallete I wonder if this issue can be better solved by a config option, and if so what would it be?

Perhaps something that allows users to specify different sources or skip completely if omitted.

```
[db.seed]
scripts_path = "./seed/*.sql"
```

avallete:
@sweatybridge,

I like the idea of adding a config option. My only worry is that it might be harder for users to find compared to a CLI flag, and it could be less convenient for one-time use. It depends on how often people would need to change this setting.

For compatibility, can we add this new config automatically when the CLI runs for the first time? Or should we use some logic like:

- **No `[db.seed]` in config**: Use the current behavior.
- **`[db.seed]` in config**: Seed from the given path or skip if empty/disabled.

Also, for different branches, should we still use `"supabase/seed/*.sql"` by default, or follow the path set in `config.toml`?

Another solution could be adding a `--seed-path` flag. If this flag is given an empty value, it would skip seeding. This way, users can control seeding behavior directly from the command line.

What do you think?

sweatybridge:
> Or should we use some logic like:
> - No [db.seed] in config: Use the current behavior.
> - [db.seed] in config: Seed from the given path or skip if empty/disabled.

This sounds good to me. We can start with a cli flag for adhoc usage, followed by the config approach for branching use cases.

The config approach would eventually support different sources so that users can control what to seed for each branch.

PR #2665, opened 2024-09-09
feat: db reset skip seed flag

avallete:

feature add `--skip-seed` flag to `db reset` command

Currently, when resetting the local database using supabase db reset, the command always applies the seed data from seed.sql. There's no option to skip the seeding process, which can be inconvenient when users want to reset the database structure without populating it with seed data.

See: #2534 

## What is the new behavior?

This PR introduces a new `--skip-seed` flag for the supabase db reset command. When this flag is used, the reset process will recreate the database and apply migrations, but it will skip the seeding step. This allows users to reset their database structure without automatically populating it with seed data.

### DIFF
--- a/cmd/db.go
+++ b/cmd/db.go
@@ -189,7 +189,9 @@ var (
 		Use:   "reset",
 		Short: "Resets the local database to current migrations",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			utils.Config.Db.Seed.Enabled = !noSeed
+			if noSeed {
+				utils.Config.Db.Seed.Enabled = false
+			}
 			return reset.Run(cmd.Context(), migrationVersion, flags.DbConfig, afero.NewOsFs())
 		},
 	}
--- a/cmd/db.go
+++ b/cmd/db.go
@@ -187,7 +187,8 @@ var (
 		Use:   "reset",
 		Short: "Resets the local database to current migrations",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return reset.Run(cmd.Context(), migrationVersion, flags.DbConfig, afero.NewOsFs())
+			skipSeed, _ := cmd.Flags().GetBool("skip-seed")
+			return reset.Run(cmd.Context(), migrationVersion, flags.DbConfig, afero.NewOsFs(), skipSeed)
 		},
 	}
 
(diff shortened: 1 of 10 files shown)


docs/supabase/db/reset.md
```diff
@@ -7,3 +7,7 @@ Requires the local development stack to be started by running `supabase start`.
 Recreates the local Postgres container and applies all local migrations found in `supabase/migrations` directory. If test data is defined in `supabase/seed.sql`, it will be seeded after the migrations are run. Any other data or schema changes made during local development will be discarded.
 
 Note that since Postgres roles are cluster level entities, those changes will persist between resets. In order to reset custom roles, you need to restart the local development stack.
+
+### Options
+
+- `--skip-seed`: Skips seeding the database with test data.
```
sweatybridge, line 13:
```suggestion
```

No need to repeat the flag as it will be merged with the rest of the generated docs.

cmd/db.go
```diff
@@ -299,6 +300,7 @@ func init() {
 	resetFlags.String("db-url", "", "Resets the database specified by the connection string (must be percent-encoded).")
 	resetFlags.Bool("linked", false, "Resets the linked project with local migrations.")
 	resetFlags.Bool("local", true, "Resets the local database with local migrations.")
+	resetFlags.Bool("skip-seed", false, "Skip running the seed script after reset")
```
sweatybridge, line 303:
```suggestion
	resetFlags.Bool("no-seed", false, "Skip running the seed script after reset")
```

Let's call this `--no-seed` for consistency with other flags.

cmd/db.go
```diff
@@ -187,7 +187,8 @@ var (
 		Use:   "reset",
 		Short: "Resets the local database to current migrations",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return reset.Run(cmd.Context(), migrationVersion, flags.DbConfig, afero.NewOsFs())
+			skipSeed, _ := cmd.Flags().GetBool("skip-seed")
```
sweatybridge, line 190: Bearing in mind of future work to add config option, we can add a new option now `utils.Config.Db.Seed.Enabled = false` instead of passing down this bool through all the layers.

cmd/db.go
```diff
@@ -183,10 +183,13 @@ var (
 		},
 	}
 
+	noSeed bool
+
 	dbResetCmd = &cobra.Command{
 		Use:   "reset",
 		Short: "Resets the local database to current migrations",
 		RunE: func(cmd *cobra.Command, args []string) error {
+			utils.Config.Db.Seed.Enabled = !noSeed
```
sweatybridge:
```suggestion
			if noSeed {
				utils.Config.Db.Seed.Enabled = false
			}
```

If config is disabled, but flag is left at default, we still want to respect the config value.

Issue #2104, opened 2024-03-29
Can we get a `supabase stop --all` to stop all running local projects [Feature request]


jonniedarko:
**Is your feature request related to a problem? Please describe.**
As a Developer who uses Supabase for almost all projects I can I want to be able to stop any and all local Supabase projects so that switch between many projects with as little friction as possible.

I use Supabase locally on most projects I start now and I often switch between them on any given day. Its easy to forget I have Supabase running in another project until I realise I need to free up local resources (I run llms and other docker images locally ). it slows me down and breaks my focus when I need to find the project thats eating my resources and stop the ones I no longer need.

**Describe the solution you'd like**
It would be great if we could have a `supabase stop --all` to stop all local instances. 

**Describe alternatives you've considered**
Another option could be a `supabase ls` command to list all **local** running projects (possibly with a `--all` flag for all local projects) that could be piped into `supabase stop --project-id ...`. 

I am not sure what the best approach is for this but one potential idea is supporting a global config file that persists all local projects ids: 
  `~/.supabase/config.yml` ?

And while out of scope for this request, It could also potentially offer an avenue for extended features later such as overrides for default settings, support multiple project folders for single Supabase project (multi repo vs mono repos) and local templates for quick set up (e.g. maybe a common Supabase function starting point 🤷)  or reduced need for multiple instances of same docker images.

**Additional context**
I would ~be happy~ Love to contribute to this in anyway I can with some guidance 😃

sweatybridge:
You can list all locally running supabase containers using a docker filter, for eg.


```bash
$ docker ps -f 'name=supabase_db' --format '{{ .Names }}'
supabase_db_cli
```

After that you can cut the id and pipe to `supabase stop --project-id ...` via bash.

I'm not sure how useful it would be to have a global config because the services for different local projects all need to expose separate ports, and sometime separate service versions.

PR #2672, opened 2024-09-11
feat: add --all flag to stop command

avallete:

This PR introduces a new feature to the Supabase CLI that allows users to stop all local Supabase instances across different projects with a single command via a `--all` flag.

Currently, users can only stop Supabase instances for individual projects. When working with multiple projects, developers need to manually navigate to each project directory and stop the instances one by one. This can be time-consuming and may lead to forgotten running instances that consume system resources.

Closes https://github.com/supabase/cli/issues/2104

## What is the new behavior?

This PR adds a new `--all` flag to the `supabase stop` command. When used, it stops all local Supabase instances across all projects on the machine. 

## Additional context


cmd/stop.go
```diff
@@ -30,5 +31,6 @@ func init() {
 	flags.StringVar(&projectId, "project-id", "", "Local project ID to stop.")
 	cobra.CheckErr(flags.MarkHidden("backup"))
 	flags.BoolVar(&noBackup, "no-backup", false, "Deletes all data volumes after stopping.")
+	flags.BoolVar(&all, "all", false, "Stop all local Supabase instances from all projects across the machine.")
```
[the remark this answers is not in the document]
avallete: Done here https://github.com/supabase/cli/pull/2672/commits/da51b6649ae9972a4e94a8f87cab93d73d3e19ae

internal/stop/stop.go
```diff
@@ -6,12 +6,21 @@ import (
...
 	"github.com/spf13/afero"
 	"github.com/supabase/cli/internal/utils"
 )
 
-func Run(ctx context.Context, backup bool, projectId string, fsys afero.Fs) error {
+func Run(ctx context.Context, backup bool, projectId string, all bool, fsys afero.Fs) error {
+	if all {
+		return stopAllProjects(ctx, backup, fsys)
+	}
+	return stopProject(ctx, backup, projectId, fsys)
```
sweatybridge, line 20: Looking at `utils.DockerRemoveAll` https://github.com/supabase/cli/blob/develop/internal/utils/docker.go#L97, it might be better to extend it with a `projectId` arg. If passed down an empty string, the `CliProjectFilter` would simply match all running projects for deletion.

avallete: Yup, only thing is that it's not exactly the same filter. In one case, you want something like `label=supa-project=projectid` in the other you want `label=supa-project`. So I had to push the `projectId` info down to the `CliProjectFilter` where the actual docker filter is constructed.

internal/stop/stop.go
```diff
@@ -11,33 +11,45 @@ import (
...
+		// Sanity checks.
+		if len(projectId) > 0 {
+			utils.Config.ProjectId = projectId
+		} else if err := utils.LoadConfigFS(fsys); err != nil {
+			return err
+		}
+		searchProjectIdFilter = utils.Config.ProjectId
+	} else {
+		// Search for all project ids
+		searchProjectIdFilter = ""
```
sweatybridge, line 26:
```suggestion
```

redundant as new string defaults to empty

sweatybridge (approved):

PR #2673, opened 2024-09-11
Add support for json output

### LINKED ISSUE #2669: Machine readable output
With https://github.com/marmelab/atomic-crm, we want to offer a good local development experience as well as making it easy to deploy the project. To do so, we included the supabase CLI as a development dependency and created this script to deploy a new hosted instance when ready: https://github.com/marmelab/atomic-crm/blob/main/scripts/supabase-remote-init.mjs.

However, as the CLI only display the commands output for humans, it's hard to integrate it in another tool.

A `--json` flag or similar that would make the CLI display the commands output in a machine readable format.

We currently parse the output as best as we can and cross our fingers that it doesn't change.

djhi:

It add a global `json` flag that instruct some commands to output their result in JSON format allowing those results to be consumed by other tools.

I only updated the commands we need in #2669:
- projects list, create and api-keys
- orgs list and create (not needed but I wanted to ensure it worked in more commands)

All commands print their result in a human readable format

## What is the new behavior?

Some commands output JSON when the flag is set

## Additional context

Fixes #2669 

Please let me know whether this goes in the right direction :pray: If it is, I'd be happy to apply it in more commands

### DIFF
--- a/cmd/orgs.go
+++ b/cmd/orgs.go
@@ -29,17 +29,19 @@ var (
 			if err != nil {
 				return err
 			}
-			table := `|ID|NAME|
-			|-|-|
-			`
-			for _, org := range *orgs {
-				table += fmt.Sprintf("|`%s`|`%s`|\n", org.Id, strings.ReplaceAll(org.Name, "|", "\\|"))
-			}
+			if orgs != nil {
+				table := `|ID|NAME|
+|-|-|
+`
+				for _, org := range *orgs {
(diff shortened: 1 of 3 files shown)


sweatybridge:
Thank you for contributing this PR.

We've actually started using the [output flag](https://github.com/supabase/cli/blob/develop/cmd/status.go#L17) in some commands to output structured format. It would be great if you can reuse this flag in other commands as well.

sweatybridge:
Yup, I think it's general enough to be worth lifting to a global flag. I will review your changes more closely when I get some time tmr.

djhi:
I'll close this one and open another PR then

djhi:
Closing in favor of #2674

PR #2674, opened 2024-09-11
feat: add support for multiple output formats

### LINKED ISSUE #2669: Machine readable output

djhi:

It add a global `output` flag that instruct some commands to output their result in the provided format allowing those results to be consumed by other tools.

I only updated the commands we need in #2669:
- projects list
- projects create
- projects api-keys

All commands print their result in a human readable format

## What is the new behavior?

Some commands output the provided format when the flag is set

## Additional context

Supersedes #2673 
Fixes #2669

Please let me know whether this goes in the right direction :pray: If it is, I'd be happy to apply it in more commands

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -243,6 +243,11 @@ func init() {
 	flags.String("workdir", "", "path to a Supabase project directory")
 	flags.Bool("experimental", false, "enable experimental features")
 	flags.String("network-id", "", "use the specified docker network instead of a generated one")
+	var outputFlag = utils.EnumFlag{
+		Allowed: append([]string{utils.OutputEnv}, utils.OutputDefaultAllowed...),
+		Value:   utils.OutputPretty,
+	}
+	flags.Var(&outputFlag, "output", "output format of status variables")
 	flags.Var(&utils.DNSResolver, "dns-resolver", "lookup domain names using the specified resolver")
 	flags.BoolVar(&createTicket, "create-ticket", false, "create a support ticket for any CLI error")
 	cobra.CheckErr(viper.BindPFlags(flags))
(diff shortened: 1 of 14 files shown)


cmd/root.go
```diff
@@ -243,6 +243,11 @@ func init() {
 	flags.String("workdir", "", "path to a Supabase project directory")
 	flags.Bool("experimental", false, "enable experimental features")
 	flags.String("network-id", "", "use the specified docker network instead of a generated one")
+	var outputFlag = utils.EnumFlag{
+		Allowed: append([]string{utils.OutputEnv}, utils.OutputDefaultAllowed...),
+		Value:   utils.OutputPretty,
+	}
+	flags.Var(&outputFlag, "output", "output format of status variables")
 	flags.Var(&utils.DNSResolver, "dns-resolver", "lookup domain names using the specified resolver")
```
djhi, line 251: Should I introduce another util like you did with `DNSResolver`?

sweatybridge: I added `utils.DNSResolver` to primarily avoid cyclic imports from utils => cmd package. And yes, I think it makes changes simpler if we do the same for `utils.OutputFormat`.

For eg. you won't need to change every function signature to accept a format arg, because you can compare `utils.OutputFormat == utils.OutputPretty` directly in internal packages.

djhi: Done

sweatybridge (approved):
Added some minor edits. LGTM

sweatybridge (approved):

PR #2683, opened 2024-09-14
fix: db test on SELinux enabled systems

### LINKED ISSUE #2659: Cannot run db tests (Permission denied on /tmp)
**Describe the bug**
Cannot run tests using `supabase db test` or `supabase test db`

**To Reproduce**
Run `supabase db test` via the CLI.  The following error is observed:
```
Connecting to local database...
Can't opendir(/tmp): Permission denied
 at /usr/share/perl5/core_perl/App/Prove/State.pm line 422.
Files=0, Tests=0,  0 wallclock secs ( 0.00 usr +  0.00 sys =  0.00 CPU)
Result: NOTESTS
```

**Expected behavior**
Tests begin.

**System information**
- Supabase 1.192.2
- Fedora 39
- Docker 24.0.5

**Additional context**
Have tried to unset the sticky bit on my local /tmp directory using `chmod -t /tmp`, but this does not solve the problem.  There's no selinux alerts raised when I run this command.

I assume it's some sort of permissions setting for the local directory.

avallete:

This PR adds the `:z` flag to the volume binding for SELinux systems when using the `supabase db test` command.

Fixes #2659

When running `supabase db test` on a system with SELinux enabled, `pg_tap` fails to execute because the `/tmp` volume does not have the appropriate permissions, causing issues with the bind mount.

## What is the new behavior?

With this change, the `:z` flag ensures the `/tmp` volume is properly relabeled to allow `pg_tap` to run successfully on SELinux-enabled systems. There are no expected impacts on systems where SELinux is not enabled.

## Additional context

I'm hesitant about applying this change across all volume bindings in the CLI codebase. While this should fix the issue for SELinux users without affecting other systems, there might be security concerns, especially for "dynamic" bindings where the exact directory being mounted is unpredictable. Since this change is intended for local development, the security impact should be minimal, but I’m open to feedback on whether this is the best approach.


internal/db/test/test.go
```diff
@@ -43,7 +43,7 @@ func Run(ctx context.Context, testFiles []string, config pgconn.Config, fsys afe
 		return errors.Errorf("failed to resolve absolute path: %w", err)
 	}
 	dstPath := "/tmp"
-	binds := []string{fmt.Sprintf("%s:%s:ro", srcPath, dstPath)}
+	binds := []string{fmt.Sprintf("%s:%s:ro,z", srcPath, dstPath)}
```
sweatybridge, line 46: There's another problem with `z` label for podman on macOS. https://github.com/supabase/cli/pull/2392

I think we have to use a platform specific implementation for this, ie. docker_linux.go

avallete: Ah ! Nice catch, thanks for the context. I'll create an utils function checking if SELinux is present and "wrapping" the volumes mount with or without the `z` if necessary.

Something like: `GetVolumeBindMode('ro' | 'rw')`

avallete:
Made the changes to scope the `z` flags to Linux with SELinux installed environments only.

Changed the volume binding option across the codebase instead of just for the `db test` command.

sweatybridge:
I'm not fully convinced about the scope of this change. We use different docker volume mount types, including named, anonymous, and bind mounts. Perhaps it's only bind mount that's breaking in selinux rather than every mount type.

Therefore, I think it's worth setting up a [selinux test](https://github.com/supabase/supabase-action-example/pull/39) to reproduce this error, and to find out if other commands like `functions deploy` also need to be fixed.

There's also the alternative to disable selinux for a specific container which should be fine for running db tests https://jaosorior.dev/2018/selinux-and-docker-notes/

avallete:
Setting up testing before making changes indeed sounds like a better plan. Plus it'll add coverage for one more platform, I'll do that, thank's for pointing in the right direction.

avallete:
I was able to reproduce the issue locally on a virtual machine. The `SecurityOpt` option mentioned in the article you linked, @sweatybridge, successfully resolves the problem with the `db test` run.

However, I wasn't able to find a way to test this in a CI environment. Interestingly, the tests passed in my reproduction environment as well, likely due to our use of mocked filesystems and Docker.

I encountered a separate issue related to an invalid `RLIMIT_NOFILE` when running both the `storage` and `realtime` containers. This seems like a distinct problem that should be addressed independently.

sweatybridge:
Amazing, thank you for going the extra mile to reproduce this.

Is the rlimit issue a warning or straight error starting storage container? Either way, I'm happy to deal with that separately.

sweatybridge (approved):

PR #2688, opened 2024-09-16
feat: add enabled flag for functions under config.toml

### LINKED ISSUE #1507: Ignore function when deploy all
I just switched to the new method to deploy all my functions:
`supabase functions deploy`

I would like a way to set in `config.toml` some function to be ignored during deploy all.

Furthermore, I have some demo function to help open-source maintainers create one, and this shouldn't be deployed.

Something like:
```toml
[functions.hello-world]
no_deploy = true
```
Would fix my issue

avallete:

This PR add the capability of having a `no_deploy` flag under the config.toml [fuctions.<function-name>] allowing to exclude some functions from the deployment.

Fixes #1507

## Additional context

Documentation added here: https://github.com/supabase/supabase/pull/29340


internal/functions/serve/serve.go
```diff
@@ -217,6 +217,7 @@ func populatePerFunctionConfigs(cwd, importMapPath string, noVerifyJWT *bool, fs
 	if err != nil {
 		return nil, "", err
 	}
+	functionsConfig = deploy.FilterFunctionsToDeploy(functionsConfig)
```
avallete, line 220: Here I wonder, should the flag also skip the serve of those functions locally ?

sweatybridge: Yes, I would think so. We want to have functions served locally to match those deployed.

pkg/config/config.go
```diff
@@ -412,6 +412,7 @@ type (
 		VerifyJWT  *bool  `toml:"verify_jwt" json:"verifyJWT"`
 		ImportMap  string `toml:"import_map" json:"importMapPath,omitempty"`
 		Entrypoint string `json:"-"`
+		NoDeploy   bool   `toml:"no_deploy"`
```
sweatybridge, line 415:
```suggestion
		Enabled   bool   `toml:"enabled"`
```

We can use `enabled` option for consistency with rest of the config.

internal/functions/deploy/deploy.go
```diff
@@ -93,3 +97,17 @@ func GetFunctionConfig(slugs []string, importMapPath string, noVerifyJWT *bool,
 	}
 	return functionConfig, nil
 }
+
+func FilterFunctionsToDeploy(functionsConfig config.FunctionConfig) (functionsToDeploy config.FunctionConfig, skippedFunctions []string) {
```
sweatybridge, line 101: This can be private method called by `GetFunctionSlugs`

internal/functions/deploy/deploy.go
```diff
@@ -93,3 +100,12 @@ func GetFunctionConfig(slugs []string, importMapPath string, noVerifyJWT *bool,
...
 	return functionConfig, nil
 }
+
+func isFunctionEnabled(slug string) bool {
+	functionConfig := utils.Config.Functions[slug]
+	// If the function config Enabled is not defined, or defined and set to true
+	if functionConfig.Enabled == nil || (functionConfig.Enabled != nil && *functionConfig.Enabled) {
+		return true
+	}
+	return false
```
sweatybridge, line 110:
```suggestion
	return functionConfig.Enabled == nil || functionConfig.Enabled
```

internal/functions/serve/serve.go
```diff
@@ -209,10 +209,13 @@ func parseEnvFile(envFilePath string, fsys afero.Fs) ([]string, error) {
 }
 
 func populatePerFunctionConfigs(cwd, importMapPath string, noVerifyJWT *bool, fsys afero.Fs) ([]string, string, error) {
-	slugs, err := deploy.GetFunctionSlugs(fsys)
+	slugs, skippedFunctions, err := deploy.GetFunctionSlugs(fsys)
 	if err != nil {
 		return nil, "", err
 	}
+	if len(skippedFunctions) > 0 {
+		fmt.Fprintf(utils.GetDebugLogger(), "Skipped serve the following functions: %s\n", strings.Join(skippedFunctions, ", "))
```
sweatybridge, line 217:
```suggestion
		fmt.Fprintf(utils.GetDebugLogger(), "Skipped serving the following functions: %s\n", strings.Join(skippedFunctions, ", "))
```

sweatybridge (approved):

PR #2697, opened 2024-09-19
chore: start command validate exclude arguments

avallete:

When providing --exclude flags, check the arguments match actuals images names show a warning otherwise.
Also fix a bug where stack couldnt start if storage was enabled in config but excluded in the arguments

Related: https://www.notion.so/supabase/exclude-doesnt-seem-to-work-f38b8444b62f4308aef657a52d51692b

## What is the new behavior?

Screenshot 2024-09-19 at 15 33 59

## Additional context

Also might be worth considering allowing to alias between config services names and "containers exclusions" names. As it's not clear for instance that "gotrue" container relate to the "auth" service.

### DIFF
--- a/internal/start/start.go
+++ b/internal/start/start.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 	"text/template"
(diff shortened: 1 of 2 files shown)


internal/start/start.go
```diff
@@ -827,7 +859,7 @@ EOF
 	}
 
 	// Start Storage.
-	if utils.Config.Storage.Enabled && !isContainerExcluded(utils.Config.Storage.Image, excluded) {
```
sweatybridge, line 830: Why does moving the same check earlier make a difference?

internal/start/start.go
```diff
@@ -43,9 +44,39 @@ func suggestUpdateCmd(serviceImages map[string]string) string {
...
+	if len(invalidContainers) > 0 {
+		// Sort the names list so it's easier to visually spot the one you looking for
+		sort.Strings(validContainers)
+		warning := fmt.Sprintf("%s The following container names are not valid to exclude: %s\nValid containers to exclude are: %s\n",
+			utils.Yellow("Warning:"),
+			utils.Aqua(strings.Join(invalidContainers, ", ")),
+			utils.Aqua(strings.Join(validContainers, ", ")))
+		fmt.Fprint(os.Stderr, warning)
+	}
+}
```
sweatybridge, line 74: Let's move this check to `cmd/start.go` since it's closely related to validating command line flag.

internal/start/start.go
avallete: Good catch, I think the second part of the condition is unnecessary I can just check over the `isStorageEnabled`.

As for the why I moved it to a dedicated value it's to ensure that we always check the same values when we check "isStorageEnabled" aka: "is the config true, and storage not excluded from the start".

I had this error where, if you had storage `enabled` in the config but excluded with the flags, then, the storage container was not started (as expected) but the "imageTransform" container was attempting to start (because it was only checking over the config value which was true) and not against the excluded flags. But since the Image container depends on the storage to work properly, the stack was failing to start.

sweatybridge (approved):

PR #2702, opened 2024-09-23
feat: add custom seed path to config

avallete:

Feature enhancement: Improved seed data handling and configuration

Currently, the system uses a single, fixed seed file path (`utils.SeedDataPath`) for database seeding. This limits flexibility in organizing and applying seed data.

## What is the new behavior?

1. Introduces a new configuration option in `config.toml` for specifying multiple seed file paths using glob patterns:

```toml
[db.seed]
enabled = true
path = ['./seed.sql', './seeds/*.sql']
```

2. Implements a new function `GetSeedFiles` that:
   - Matches files based on the configured glob patterns
   - Deduplicates matched files to avoid duplicate seeding
   - Sorts the files to ensure consistent application order

3. Updates various parts of the codebase to use the new seeding mechanism:
   - Modifies the `SeedDatabase` function to use the new `GetSeedFiles` function
   - Updates error handling and logging related to seed file operations
   - Adjusts test cases to accommodate the new seeding behavior

4. Change the logging to display the seed file paths being used

## Additional context

This change allows for more flexible organization of seed data, enabling users to split seed data across multiple files and use patterns to include them. It maintains backwards compatibility by defaulting to the previous behavior if no custom configuration is provided.

I'm just wondering about how this should work with supabase "branch deployments" and so on. Locally we'll always find the seeds files at the right location. I wonder how the branching approach work with that. From my understanding it does "clone" the user repository so it should work the same as locally but I might need some help to get that behavior tested.

### DIFF
--- a/internal/db/push/push_test.go
+++ b/internal/db/push/push_test.go
@@ -162,7 +162,7 @@ func TestPushAll(t *testing.T) {
 
 	t.Run("throws error on seed failure", func(t *testing.T) {
 		// Setup in-memory fs
-		fsys := &fstest.OpenErrorFs{DenyPath: utils.SeedDataPath}
+		fsys := &fstest.OpenErrorFs{DenyPath: utils.DefaultSeedDataPath}
 		path := filepath.Join(utils.MigrationsDir, "0_test.sql")
 		require.NoError(t, afero.WriteFile(fsys, path, []byte{}, 0644))
 		// Setup mock postgres
--- a/internal/db/push/push_test.go
+++ b/internal/db/push/push_test.go
@@ -163,6 +163,7 @@ func TestPushAll(t *testing.T) {
 	t.Run("throws error on seed failure", func(t *testing.T) {
 		// Setup in-memory fs
 		fsys := &fstest.OpenErrorFs{DenyPath: utils.DefaultSeedDataPath}
+		fsys.Create(utils.DefaultSeedDataPath)
 		path := filepath.Join(utils.MigrationsDir, "0_test.sql")
 		require.NoError(t, afero.WriteFile(fsys, path, []byte{}, 0644))
 		// Setup mock postgres
--- a/internal/db/push/push_test.go
+++ b/internal/db/push/push_test.go
@@ -162,8 +162,9 @@ func TestPushAll(t *testing.T) {
 
 	t.Run("throws error on seed failure", func(t *testing.T) {
 		// Setup in-memory fs
-		fsys := &fstest.OpenErrorFs{DenyPath: filepath.Join(utils.SupabaseDirPath, "seed.sql")}
-		_, _ = fsys.Create(filepath.Join(utils.SupabaseDirPath, "seed.sql"))
+		seedPath := filepath.Join(utils.SupabaseDirPath, "seed.sql")
+		fsys := &fstest.OpenErrorFs{DenyPath: seedPath}
+		_, _ = fsys.Create(seedPath)
 		path := filepath.Join(utils.MigrationsDir, "0_test.sql")
 		require.NoError(t, afero.WriteFile(fsys, path, []byte{}, 0644))
 		// Setup mock postgres
--- a/internal/db/push/push_test.go
+++ b/internal/db/push/push_test.go
@@ -163,7 +163,7 @@ func TestPushAll(t *testing.T) {
 	t.Run("throws error on seed failure", func(t *testing.T) {
 		// Setup in-memory fs
 		fsys := &fstest.OpenErrorFs{DenyPath: utils.DefaultSeedDataPath}
-		fsys.Create(utils.DefaultSeedDataPath)
+		_, _ = fsys.Create(utils.DefaultSeedDataPath)
 		path := filepath.Join(utils.MigrationsDir, "0_test.sql")
 		require.NoError(t, afero.WriteFile(fsys, path, []byte{}, 0644))
 		// Setup mock postgres
--- a/internal/db/push/push_test.go
+++ b/internal/db/push/push_test.go
@@ -162,8 +162,8 @@ func TestPushAll(t *testing.T) {
 
 	t.Run("throws error on seed failure", func(t *testing.T) {
 		// Setup in-memory fs
-		fsys := &fstest.OpenErrorFs{DenyPath: utils.DefaultSeedDataPath}
-		_, _ = fsys.Create(utils.DefaultSeedDataPath)
+		fsys := &fstest.OpenErrorFs{DenyPath: filepath.Join(utils.SupabaseDirPath, "seed.sql")}
+		_, _ = fsys.Create(filepath.Join(utils.SupabaseDirPath, "seed.sql"))
 		path := filepath.Join(utils.MigrationsDir, "0_test.sql")
 		require.NoError(t, afero.WriteFile(fsys, path, []byte{}, 0644))
 		// Setup mock postgres
--- a/internal/db/reset/reset_test.go
+++ b/internal/db/reset/reset_test.go
@@ -362,7 +362,7 @@ func TestResetRemote(t *testing.T) {
 		fsys := afero.NewMemMapFs()
 		path := filepath.Join(utils.MigrationsDir, "0_schema.sql")
 		require.NoError(t, afero.WriteFile(fsys, path, nil, 0644))
-		seedPath := filepath.Join(utils.SeedDataPath)
+		seedPath := filepath.Join(utils.DefaultSeedDataPath)
 		// Will raise an error when seeding
 		require.NoError(t, afero.WriteFile(fsys, seedPath, []byte("INSERT INTO test_table;"), 0644))
 		// Setup mock postgres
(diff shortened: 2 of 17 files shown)


internal/db/push/push_test.go
```diff
@@ -162,7 +162,8 @@ func TestPushAll(t *testing.T) {
 
 	t.Run("throws error on seed failure", func(t *testing.T) {
 		// Setup in-memory fs
-		fsys := &fstest.OpenErrorFs{DenyPath: utils.SeedDataPath}
+		fsys := &fstest.OpenErrorFs{DenyPath: utils.DefaultSeedDataPath}
+		_, _ = fsys.Create(utils.DefaultSeedDataPath)
```
avallete, line 166: With the new files searching behavior we need to ensure the file actually exist on the filesystem. Otherwise, the "Glob" filter will just not return anything, the seeds will be skipped and the file will never be attempted to open.

pkg/migration/seed.go
```diff
@@ -12,8 +12,7 @@ import (
 
 func SeedData(ctx context.Context, pending []string, conn *pgx.Conn, fsys fs.FS) error {
 	for _, path := range pending {
-		filename := filepath.Base(path)
-		fmt.Fprintf(os.Stderr, "Seeding data from %s...\n", filename)
+		fmt.Fprintf(os.Stderr, "Seeding data from %s...\n", path)
```
avallete, line 15: Since the seed can now be in multiples locations, I changed to log the path instead of only the filename.

The result will be something like:

```
supabase/seeds/seed1.sql
supabase/another/seed1.sql
```

pkg/config/config.go
```diff
@@ -465,6 +466,7 @@ func NewConfig(editors ...ConfigEditor) config {
 			},
 			Seed: seed{
 				Enabled: true,
+				Path:    []string{"./seed.sql"},
```
sweatybridge, line 469:
```suggestion
```

not necessary as `templates/config.toml` will fill it up if left empty

pkg/config/config.go
```diff
@@ -164,7 +164,8 @@ type (
 	}
 
 	seed struct {
-		Enabled bool `toml:"enabled"`
+		Enabled bool     `toml:"enabled"`
+		Path    []string `toml:"path"`
```
sweatybridge, line 168:
```suggestion
		SqlPaths    []string `toml:"sql_paths"`
```

May be call this sql path to be more specific.

internal/utils/misc.go
```diff
@@ -157,6 +158,43 @@ var (
...
+		for _, match := range matches {
+			if _, exists := fileSet[match]; !exists {
+				fileSet[match] = struct{}{}
+				files = append(files, match)
+			} else {
+				fmt.Fprintf(GetDebugLogger(), "Duplicate seed file found skipping: %s\n", match)
+			}
+		}
+	}
+	return files, nil
```
sweatybridge, line 195:
```suggestion
	return utils.RemoveDuplicates(files), nil
```

there's a utils for this to reduce some code duplication.

internal/utils/misc.go
```diff
@@ -148,7 +149,7 @@ var (
 	FallbackImportMapPath = filepath.Join(FunctionsDir, "import_map.json")
 	FallbackEnvFilePath   = filepath.Join(FunctionsDir, ".env")
 	DbTestsDir            = filepath.Join(SupabaseDirPath, "tests")
-	SeedDataPath          = filepath.Join(SupabaseDirPath, "seed.sql")
+	DefaultSeedDataPath   = filepath.Join(SupabaseDirPath, "seed.sql")
```
sweatybridge, line 152: Q: is it possible to remove this derived field altogether? so we can have `Db.Seed.SqlPaths` as the single source of truth

internal/utils/misc.go
```diff
@@ -157,6 +158,43 @@ var (
...
+func GetSeedFiles(fsys afero.Fs) ([]string, error) {
+	seedPaths := Config.Db.Seed.Path
+	fileSet := make(map[string]struct{})
+	var files []string
+	for _, pattern := range seedPaths {
+		fullPattern := filepath.Join(SupabaseDirPath, pattern)
+		matches, err := afero.Glob(fsys, fullPattern)
+		if err != nil {
+			return nil, errors.Errorf("failed to apply glob pattern for %w", err)
+		}
```
sweatybridge, line 182: How about throwing an error when `len(matches) == 0`? This helps users catch typos in their seed paths.

internal/utils/misc.go
avallete: I think we can, only issue is at init time, we create a seed file by writing the "seed.sql" file onto `supabase/` folder.

But maybe we don't need the "write a default" file logic if we just skip the seed if the default `seed.sql` isn't present.

internal/utils/misc.go
avallete: Yeah tough about that as well.
I was thinking about logging a warning in that case something like `Warning: Your pattern XXXX matched 0 seeds files.` but I can see the benefit of an early fail as well.

sweatybridge: true, happy to go with a warning too.

pkg/config/config.go
avallete: There is cases in our tests where we don't call the `Load` config but just use `NewConfig()`, so this values isn't set by the template and it end up causing tests failure. I could try to refactor the tests to make them call `Load` just didn't looked like it was the best option.

internal/db/reset/reset_test.go
```diff
@@ -362,7 +362,7 @@ func TestResetRemote(t *testing.T) {
 		fsys := afero.NewMemMapFs()
 		path := filepath.Join(utils.MigrationsDir, "0_schema.sql")
 		require.NoError(t, afero.WriteFile(fsys, path, nil, 0644))
-		seedPath := filepath.Join(utils.SeedDataPath)
+		seedPath := filepath.Join(filepath.Join(utils.SupabaseDirPath, "seed.sql"))
```
sweatybridge, line 365:
```suggestion
		seedPath := filepath.Join(utils.SupabaseDirPath, "seed.sql")
```

internal/db/push/push_test.go
```diff
@@ -162,7 +162,8 @@ func TestPushAll(t *testing.T) {
 
 	t.Run("throws error on seed failure", func(t *testing.T) {
 		// Setup in-memory fs
-		fsys := &fstest.OpenErrorFs{DenyPath: utils.SeedDataPath}
+		fsys := &fstest.OpenErrorFs{DenyPath: filepath.Join(utils.SupabaseDirPath, "seed.sql")}
+		_, _ = fsys.Create(filepath.Join(utils.SupabaseDirPath, "seed.sql"))
```
sweatybridge, line 166:
```suggestion
		seedPath := filepath.Join(utils.SupabaseDirPath, "seed.sql")
		fsys := &fstest.OpenErrorFs{DenyPath: seedPath}
		f, _ := fsys.Create(seedPath)
		require.NoError(t, f.Close())
```

internal/utils/misc.go
```diff
@@ -157,6 +157,36 @@ var (
 	ErrNotRunning  = errors.Errorf("%s is not running.", Aqua("supabase start"))
 )
 
+func GetSeedsFilepaths() string {
+	if seedPaths, _ := GetSeedFiles(afero.NewOsFs()); seedPaths != nil {
+		return fmt.Sprintf("%v", seedPaths)
+	}
+	return ""
+}
```
sweatybridge, line 165: We can inline this function since it is only called once.

internal/migration/apply/apply_test.go
```diff
@@ -99,10 +99,15 @@ func TestSeedDatabase(t *testing.T) {
...
 
 	t.Run("throws error on read failure", func(t *testing.T) {
-		// Setup in-memory fs
-		fsys := &fstest.OpenErrorFs{DenyPath: utils.SeedDataPath}
+		// Wrap the fs with OpenErrorFs
+		errorFs := &fstest.OpenErrorFs{
+			DenyPath: filepath.Join(utils.SupabaseDirPath, "seed.sql"),
+		}
+		_, _ = errorFs.Create(filepath.Join(utils.SupabaseDirPath, "seed.sql"))
+
```
sweatybridge, line 107:
```suggestion
		seedPath := filepath.Join(utils.SupabaseDirPath, "seed.sql")
		errorFs := &fstest.OpenErrorFs{DenyPath: seedPath}
		f, _ := fsys.Create(seedPath)
		require.NoError(t, f.Close())
```

sweatybridge (approved):

jacob-8:
This broke my seeding ability on Windows. No matter if I use the default `supabase/seed.sql` location or use something like `sql_paths = ['./seeds/*.sql']` I always get the response `No seed files matched pattern: supabase\seed.sql` (or whatever my pattern happens to be). Is anyone else having this problem?

avallete:
> This broke my seeding ability on Windows. No matter if I use the default `supabase/seed.sql` location or use something like `sql_paths = ['./seeds/*.sql']` I always get the response `No seed files matched pattern: supabase\seed.sql` (or whatever my pattern happens to be). Is anyone else having this problem?

Hey thanks for reporting, this is already handled and should be fixed by https://github.com/supabase/cli/pull/2735 in the next upcoming release sorry about that.

Please use github issues to report such cases as it has a much better discoverability for the team than comments on closed PR 👍

PR #2707, opened 2024-09-24
chore: use _supabase database for internals

avallete:
This is to avoid overloading our user postgres database with every new addition to _analytics or _realtime

## What is the new behavior?

## Additional context


internal/db/dump/dump.go
```diff
@@ -100,9 +100,6 @@ func dumpData(ctx context.Context, config pgconn.Config, schema, excludeTable []
 		// "storage",
 		// "supabase_functions",
 		"supabase_migrations",
-		"_analytics",
-		"_realtime",
-		"_supavisor",
```
sweatybridge, line 105: We have to keep this for a few versions because users could be using new cli with and old backup db.

internal/db/start/start.go
```diff
@@ -122,7 +124,12 @@ func StartDatabase(ctx context.Context, fsys afero.Fs, w io.Writer, options ...f
 		},
 	}
 	if utils.Config.Db.MajorVersion <= 14 {
-		config.Entrypoint = nil
+		config.Entrypoint = []string{"sh", "-c", `
+			cat <<'EOF' > /docker-entrypoint-initdb.d/supabase_schema.sql
+` + _supabaseSchema + `
+EOF
+			docker-entrypoint.sh postgres -D /etc/postgresql
+		`}
```
avallete, line 132: I had to do that to make the start also work for postgres v14 and v13. I'm not sure it's the best option tough, maybe I'm missing why it was nil before 🤔 

I also wonder if in those cases, we should migrate the part where there is "_realtime" schema seeding inside `13.sql` and `14.sql` to it or not.

internal/db/start/templates/_supabase.sql
```diff
@@ -0,0 +1,18 @@
...
+
+-- Switch to the newly created _supabase database
+\c _supabase
+-- Create schemas in _supabase database for
+-- internals tools and reports to not overload user database
+-- with non-user activity
+CREATE SCHEMA IF NOT EXISTS _realtime;
+ALTER SCHEMA _realtime OWNER TO postgres;
+CREATE SCHEMA IF NOT EXISTS realtime;
+ALTER SCHEMA realtime OWNER TO postgres;
```
avallete, line 11: For some reason, if I don't include the `realtime` schema as well and only the `_realtime` the `initRealtime` function will fail requesting this other schema.

internal/start/start.go
```diff
@@ -750,7 +750,7 @@ EOF
 					fmt.Sprintf("DB_PORT=%d", dbConfig.Port),
 					"DB_USER=supabase_admin",
 					"DB_PASSWORD=" + dbConfig.Password,
-					"DB_NAME=" + dbConfig.Database,
```
sweatybridge, line 753: `realtime` schema should be initialised inside the postgres database. We probably need to check with @filipecabaco if there's additional config for separating the control plane database.

```
psql (16.3, server 15.6)
Type "help" for help.

_supabase=> \dn
        List of schemas
    Name    |       Owner
------------+-------------------
 _analytics | postgres
 _realtime  | postgres
 _supavisor | postgres
 public     | pg_database_owner
 realtime   | postgres
(5 rows)

_supabase=> \dt realtime.*
                       List of relations
  Schema  |       Name        | Type  |          Owner
----------+-------------------+-------+-------------------------
 realtime | messages          | table | supabase_realtime_admin
 realtime | schema_migrations | table | supabase_admin
 realtime | subscription      | table | supabase_admin
(3 rows)
```

internal/db/start/start.go
sweatybridge: Yup, this is a good fix. Previously we set entrypoint to nil to use the default value that came with 13 and 14 images.

internal/db/start/start.go
```diff
@@ -82,7 +84,7 @@ func NewContainerConfig() container.Config {
 			Retries:  3,
 		},
 		Entrypoint: []string{"sh", "-c", `cat <<'EOF' > /etc/postgresql.schema.sql && cat <<'EOF' > /etc/postgresql-custom/pgsodium_root.key && docker-entrypoint.sh postgres -D /etc/postgresql
-` + initialSchema + `
+` + initialSchema + "\n" + _supabaseSchema + "\n" + `
```
sweatybridge, line 87:
```suggestion
` + initialSchema + `
` + _supabaseSchema + `
```

slightly more readable

PR #2726, opened 2024-10-04
chore: glob seed paths when loading config

sweatybridge:

refactor

## What is the new behavior?

Globs seed files on config load. This helps detect path errors early and loading paths consistently in branching.

## Additional context


pkg/config/config.go
```diff
@@ -172,8 +173,9 @@ type (
 	}
 
 	seed struct {
-		Enabled  bool     `toml:"enabled"`
-		SqlPaths []string `toml:"sql_paths"`
+		Enabled      bool     `toml:"enabled"`
+		GlobPatterns []string `toml:"sql_paths"`
+		SqlPaths     []string `toml:"-"`
```
avallete, line 178: **praise**

I really need to remember this trick to add stuff in the config structure that are not in the toml.

avallete (approved):

PR #2760, opened 2024-10-11
feat: read env in auth.additional_redirect_urls values

jgoux:

feature

The values in `auth.additional_redirect_urls` can't read from the env.

## What is the new behavior?

The values in `auth.additional_redirect_urls` can read from the env. 😄 

## Additional context

I need that in order to have an environment agnostic configuration for the example I'm building showcasing social login in a Next.js app using `@supabase-labs/nextjs`.

In principle every time we have a config value being URL related, it should be able to read from the env.


avallete (commented):
LGTM waiting leaving the approval to qiao

sweatybridge (approved):
Looks good. We probably want to start doing this for every string field by adding a custom toml decode function.

PR #2761, opened 2024-10-12
feat(config): add api config for branching override

avallete:

Feature add tooling around api config to be used inside branching for config override.

## What is the new behavior?

## Additional context


internal/utils/diff/diff.go
```diff
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-package link
+package diff
```
avallete, line 5: Couldn't put this under `utils` because import in `config` would cause cyclic dependencies.

pkg/config/api.go
```diff
@@ -0,0 +1,96 @@
...
+	v1API "github.com/supabase/cli/pkg/api"
+)
+
+type (
+	RemoteApi struct {
+		Enabled         bool     `toml:"enabled"`
+		Schemas         []string `toml:"schemas"`
+		ExtraSearchPath []string `toml:"extra_search_path"`
+		MaxRows         uint     `toml:"max_rows"`
+	}
```
avallete, line 16: **question:**

This is what I ended up with, this is public facing so it can be imported inside `branching` and we can use it's methods. And basically contain the fields that will be actually shared between remote and local config.

pkg/config/api.go
```diff
@@ -0,0 +1,96 @@
...
+	api struct {
+		RemoteApi
+		Image     string  `toml:"-"`
+		KongImage string  `toml:"-"`
+		Port      uint16  `toml:"port"`
+		Tls       tlsKong `toml:"tls"`
+		// TODO: replace [auth|studio].api_url
+		ExternalUrl string `toml:"external_url"`
+	}
```
avallete, line 25: The cli specifics config remains in a private dedicated structure.

pkg/config/config.go
```diff
@@ -732,7 +716,19 @@ func (c *config) Load(path string, fsys fs.FS) error {
...
+			duplicatedRemotesProjectsIds[base.ProjectId] = append(duplicatedRemotesProjectsIds[base.ProjectId], name)
+		}
+		// Check for duplicate project IDs across remotes
+		for projectId, remotes := range duplicatedRemotesProjectsIds {
+			if len(remotes) > 1 {
+				fmt.Fprintf(os.Stderr, "WARN: Multiple remotes (%s) have the same project_id: %s. This may lead to unexpected config override.\n", strings.Join(remotes, ", "), projectId)
+			}
+		}
 		if err := base.Validate(fsys); err != nil {
+			fmt.Fprintf(os.Stderr, "Error with remote config %s\n", name)
```
avallete, line 731: **remark**

While testing, when I set up a remote with `project_id = ""` I had the error raising about the "whole config". Not mentioning the actual issue was inside one of the remote definition. This should help provide better guidance in such cases.

pkg/config/api.go
```diff
@@ -0,0 +1,98 @@
...
+	if a.MaxRows > 0 {
+		intValue := cast.UintToInt(a.MaxRows)
+		body.MaxRows = &intValue
+	}
+
+	// Note: DbPool is not present in the Api struct, so it's not set here
+	return body
+}
+
+func (a *RemoteApi) fromRemoteApiConfig(remoteConfig v1API.PostgrestConfigWithJWTSecretResponse) RemoteApi {
```
sweatybridge, line 65:
```suggestion
func NewRemoteApiConfig(remoteConfig v1API.PostgrestConfigWithJWTSecretResponse) remoteApi {
```

Do we need to override existing config? If not, it's more idiomatic to declare the constructor on its own.

avallete:
@sweatybridge

> I don't quite follow the intended used case on branching. I'd imagine the following code snippet is close to how it's used?
> ```
> cliConfig, _ := config.NewConfig()
> body := cliConfig.Api.ToUpdatePostgrestConfigBody()
> ```
> It's allowed to call public methods on private struct.

Yes, but the actual use case is for splitting each of the "service config update" in their dedicated scope like so: https://github.com/supabase/branching/blob/41c0391c65126fc9f0da978043a2182cac75c2b7/internal/task/configure.go#L13-L35

The idea is to expose what's "shared" but keep what's only cli specific private.

pkg/config/api.go
avallete: I do actually "override" the existing config. We don't replace it since we clone it but we keep it's fields and "replace" the ones coming from the "remoteConfig". That allow for having a cleaner diff in some cases (for instance, when Enabled = false), since the rest of the "RemoteConfig" will be kept the same. Rather than having a diff with empty db schemas.

pkg/config/config.go
```diff
@@ -1327,3 +1322,28 @@ func (a *auth) ResolveJWKS(ctx context.Context) (string, error) {
...
+	// If no matching remote config is found, return the base config and false
+	return c.baseConfig, false
+}
+
+func ToTomlBytes(config any) []byte {
+	var buf bytes.Buffer
+	enc := toml.NewEncoder(&buf)
+	enc.Indent = ""
+	if err := enc.Encode(config); err != nil {
+		fmt.Fprintln(os.Stderr, "failed to marshal toml config:", err)
```
sweatybridge, line 1346: I think we want to return this error and let caller decide what to do with it.

For link, it's safe to log / ignore because it shouldn't affect the exit code. But for `DiffWithRemote`, we most likely want to bubble up the error since it indicates an irrecoverable failure.

pkg/config/updater.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (u *ConfigUpdater) UpdateRemoteConfig(ctx context.Context, remote baseConfig) error {
+	if err := u.UpdateApiConfig(ctx, remote.ProjectId, remote.Api); err != nil {
+		return err
+	}
+	// TODO: implement other service configs, ie. auth
+	return nil
+}
+
+func (u *ConfigUpdater) UpdateApiConfig(ctx context.Context, projectRef string, c api) error {
```
sweatybridge, line 28: I've moved most of your [implementation](https://github.com/supabase/branching/pull/163/files#diff-2f6ae6efa668b8891389d9d65b8b18cefc13b580c4a4c16138c75b26d9df8cbcR13) here because it can be reused for `config push` in the future.

pkg/config/config.go
```diff
@@ -1327,3 +1322,28 @@ func (a *auth) ResolveJWKS(ctx context.Context) (string, error) {
 
 	return string(jwksEncoded), nil
 }
+
+// Retrieve the final base config to use taking into account the remotes override
+func (c *config) GetRemoteOverride(project_ref string) (overrideConfig baseConfig, overrideExist bool) {
```
sweatybridge, line 1327: IMO this should throw an error if projectRef is not found.

```suggestion
func (c *config) GetRemoteOverride(projectRef string) (baseConfig, error) {
```

pkg/config/config.go
```diff
@@ -719,6 +702,7 @@ func (c *config) Load(path string, fsys fs.FS) error {
 		return err
 	}
 	c.Remotes = make(map[string]baseConfig, len(c.Overrides))
+	duplicatedRemotesProjectsIds := make(map[string][]string)
```
sweatybridge, line 705: I'd suggest moving this check to `GetRemoteOverride` because the logic is closely related.

pkg/config/updater.go
avallete: Right ! Plus it allow to get rid of the public `RemoteApi` I needed in branching.

I wonder if the "functions upsert" we have on branching already should fit there as well, since it's kindda in-between a config and something else.

pkg/config/config.go
```diff
@@ -732,8 +716,19 @@ func (c *config) Load(path string, fsys fs.FS) error {
 		} else if undecoded := metadata.Undecoded(); len(undecoded) > 0 {
 			fmt.Fprintf(os.Stderr, "Unknown config fields: %+v\n", undecoded)
 		}
+		if base.ProjectId == c.baseConfig.ProjectId {
+			fmt.Fprintf(os.Stderr, "WARN: project_id is missing for remote %s this config won't apply to any branch\n", name)
```
sweatybridge, line 720:
```suggestion
			fmt.Fprintf(os.Stderr, "WARN: project_id is missing for [remotes.%s]\n", name)
```

There's an edge case where user may declare `project_id = "<project_ref>` in the base config, which means the id is inherited over. Let's just not mention that it won't apply to any branch.

pkg/config/updater.go
sweatybridge: I see what you mean. The function bundling part is very tricky because we have to convert the config to docker paths when bundling from cli, but leave them as host paths when ran from branching executor.

May be we can unify the approaches if edge runtime starts releasing cross platform binaries which allow would allow us to only use the OS native path.

sweatybridge (approved):

pkg/config/config.go
avallete: I wonder, it felt weird to split the validation logic in the getter since this is where we check for the other fields. And project_id in my mind is just one more field with custom cross fields (across remotes) validation logic.

I can move it, but I think might be best to see thoses issues when your config is parsed and validated, not when you actually try to access one of the declaration. After the Load/Validate, your config is correct, the access logic shouldn't worry about global config consistency / validation.

sweatybridge: hmm the way I see it, when calling `GetRemoteByProjectRef`, a reasonable expectation is to throw an error if there are multiple remotes with the same matching id.

What you said about validating the config once at load time also makes sense. But perhaps we need something stronger than a warning? Because the iteration order of golang map is randomized. So it would be very surprising for the user if `GetRemoteByProjectRef` returns different config on separate calls.

avallete: Maybe we can throw an error if duplicates are seen at validation time in that case, keep the warning for "missing" ones that are just undeclared.

And have the `GetRemoteProjectRef` only worry about throwing an error if the requested remote isn't present (and is the same as the base project id maybe ?).

That way user get early feedback about messing up their config, we avoid inconsistent behavior at runtime (the map iteration) and we keep validation in a single place. WDYT ?

PR #2763, opened 2024-10-14
feat: Option to create Edge Functions with JavaScript syntax

laktek:

This PR adds the option to create, serve, and deploy pure JS functions instead of using TypeScript.

Users can set `EdgeRuntime.default_language` to either JavaScript or TypeScript to set the language template they want to use. This can be overridden with `lang` flag when using `functions new`.

`functions serve` and `functions deploy` will now treat for both `index.js` and `index.ts` as valid entrypoints.

### DIFF
--- a/internal/functions/new/new.go
+++ b/internal/functions/new/new.go
@@ -7,6 +7,8 @@ import (
 	"html/template"
 	"os"
 	"path/filepath"
+	"slices"
+	"strings"
 
 	"github.com/go-errors/errors"
 	"github.com/spf13/afero"
@@ -17,6 +19,10 @@ var (
 	//go:embed templates/index.ts
 	indexEmbed    string
 	indexTemplate = template.Must(template.New("indexl").Parse(indexEmbed))
+
+	//go:embed templates/index.js
+	indexJsEmbed    string
+	indexJsTemplate = template.Must(template.New("indexjs").Parse(indexJsEmbed))
 )
 
 type indexConfig struct {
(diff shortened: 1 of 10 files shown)


pkg/config/config.go
```diff
@@ -412,10 +412,11 @@ type (
...
 	edgeRuntime struct {
-		Enabled       bool          `toml:"enabled"`
-		Image         string        `toml:"-"`
-		Policy        RequestPolicy `toml:"policy"`
-		InspectorPort uint16        `toml:"inspector_port"`
+		Enabled         bool          `toml:"enabled"`
+		Image           string        `toml:"-"`
+		Policy          RequestPolicy `toml:"policy"`
+		InspectorPort   uint16        `toml:"inspector_port"`
+		DefaultLanguage string        `toml:"default_language"`
```
sweatybridge, line 419: Does this mean we can't support js and ts at the same time? How does platform handle one function in ts and another in js?

laktek: hmm..this is probably a bad place to place this setting, given it may convey the wrong message. `DefaultLanguage` is only used when creating a new function to select the right template. Edge Runtime itself can handle both ts/js functions without any modifications (need to set via Entrypoint when bundling code).

Alternatively, we can add a top-level config `FunctionsDefaultLanguage`? (can't add it under `Functions` due to possible naming conflicts)

sweatybridge: discussed offline, will include a comment to clarify

internal/functions/new/new.go
```diff
@@ -17,43 +19,92 @@ var (
...
+			items := []utils.PromptItem{
+				{Summary: "TypeScript"},
+				{Summary: "JavaScript"},
+			}
+			choice, err := utils.PromptChoice(ctx, title, items)
+			if err != nil {
+				return err
+			}
+			// update config
+			if err := utils.InitConfig(utils.InitParams{EdgeRuntimeDefaultLanguage: choice.Summary, Overwrite: true}, fsys); err != nil {
```
sweatybridge, line 63: reminder to self: support in-place config update

laktek:
Will revisit this after we have support to update `config.toml`

PR #2771, opened 2024-10-16
feat: load fallback import map with config

sweatybridge:

feature

## What is the new behavior?

Not sure if this is the desired approach because `functions/import_map.json` is a local fallback that's primarily useful for development. When deploying to platform, each function should only declare the import map it needs.

## Additional context


sweatybridge:
closing to encourage declaring import map in config.toml for better isolation https://supabase.slack.com/archives/C02KMRX22NR/p1729522457208809?thread_ts=1729522237.335929&cid=C02KMRX22NR

PR #2776, opened 2024-10-17
chore(docs): correct CLI linking to /usr/bin/supabase

n0tank3sh:

Fixes the documentation.


sweatybridge (approved):
Good catch!

Issue #2348, opened 2024-05-30
podman service health check fails


acsezen:
**Describe the bug**
Hello,

I am trying to run Quivr project. After running supabase start i received this error;
`service not healthy: [supabase_realtime_secondbrain supabase_pg_meta_secondbrain supabase_studio_secondbrain]`

**To Reproduce**
1. git clone https://github.com/quivrhq/quivr.git && cd quivr
2. cp .env.example .env (Make necessary changes)
3. cd backend && supabase start

**Expected behavior**

**Screenshots**

**System information**
Rerun the failing command with `--create-ticket` flag.
 - Ticket ID: fa5f99540ba34304b26a2324fce58cde
 - Version of OS: RHEL 9
 - Version of CLI: v1.169.8
 - Version of Docker: podman version 5.0.2
 - Versions of services: 
```
% supabase services


        SERVICE IMAGE      │        LOCAL         │ LINKED
  ─────────────────────────┼──────────────────────┼─────────
    supabase/postgres      │ 15.1.1.41            │ -
    supabase/gotrue        │ v2.151.0             │ -
    postgrest/postgrest    │ v12.0.1              │ -
    supabase/realtime      │ v2.28.32             │ -
    supabase/storage-api   │ v1.0.6               │ -
    supabase/edge-runtime  │ v1.53.3              │ -
    supabase/studio        │ 20240514-6f5cabd     │ -
    supabase/postgres-meta │ v0.80.0              │ -
    supabase/logflare      │ 1.4.0                │ -
    bitnami/pgbouncer      │ 1.20.1-debian-11-r39 │ -
    darthsim/imgproxy      │ v3.8.0               │ -
```

Best Regards,

sweatybridge:
Could you share the logs from unhealthy containers? I can't reproduce locally or on gitpod.

acsezen:
Hello,

Today i tried with a clean install of a supabase db with supabase init and have the same result. Here is the full log.

[formatted_supabase.log](https://github.com/user-attachments/files/15519097/formatted_supabase.log)


Seems this is the crucial part;
```bash
supabase_realtime_tmp container logs:
21:47:23.156 [info] Migrations already up
21:47:23.513 [debug] QUERY OK db=1.0ms queue=60.8ms idle=0.0ms
begin []
21:47:23.529 [debug] QUERY OK source="tenants" db=3.8ms
[...]

PR #2781, opened 2024-10-20
fix: prune volumes without all flag for pre-Docker v1.42

aloisklink:

**Bug fix** for `supabase stop --no-backup` for users on old versions of Docker or on Podman.

Before Docker API version 1.42, the `all` flag in `Docker.VolumePrune` is not required.

Podman prints an error like the following when the `all` flag is passed ([Podman's only compatible with Docker API v1.41][1]), e.g. when running `supabase stop --no-backup`:

```console
alois@pc:~/app (develop)$ npx supabase@1.207.4 stop --no-backup
failed to prune volumes: Error response from daemon: failed to parse filters for all=true&label=com.supabase.cli.project%3Dapp: "all" is an invalid volume filter
```

[1]: https://github.com/containers/podman/blob/290d94d3c00857dd582ffbee6bd0677a0904c783/version/version.go#L46

See: https://github.com/supabase/cli/issues/2348#issuecomment-2149936472, who had this same issue.
See: https://github.com/supabase/cli/pull/1989 (which added the `all` flag)

## What is the new behavior?

I've added an `if versions.GreaterThanOrEqualTo(Docker.ClientVersion(), "1.42")` check before adding the `all` flag.

A similar check is done by the official Docker CLI: https://github.com/docker/cli/blob/062eecf14af34d7295da16c23c2578fcf4aa0196/cli/command/volume/prune.go#L67-L78

This works on Podman. I suspect it would work on older versions of Docker too, but I don't have one of those environments to test with!

## Additional context

**I'm a complete beginner in golang** (this is only my second time writing a golang PR), **so please review my code carefully**. It's very likely I made a dumb mistake since I didn't know any better.

### DIFF
--- a/internal/stop/stop_test.go
+++ b/internal/stop/stop_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/network"
 	"github.com/docker/docker/api/types/volume"
+	"github.com/docker/docker/client"
 	"github.com/h2non/gock"
 	"github.com/spf13/afero"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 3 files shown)


sweatybridge (approved):
Thanks for contributing this fix with unit tests!

Issue #2611, opened 2024-08-13
How to update the PostgreSQL config in the local environment?


vinch:
Hello,

I tried to setup a local environment and copying my hosted database to my local environment but I encountered some kind of memory issues that you can see here when doing `supabase db reset`:

```
Resetting local database...
Recreating database...
Setting up initial schema...
Applying migration 20240813070343_remote_schema.sql...
ERROR: out of shared memory (SQLSTATE 53200)
At statement 843: SECURITY LABEL FOR "pgsodium" ON COLUMN "public"."abc"."cdf" IS 'ENCRYPT WITH KEY ID xyz'
Try rerunning the command with --debug to troubleshoot the error.
```

When I troubleshoot the issue with the `--debug` flag, it tells me to update the `max_locks_per_transaction` value in the PostgreSQL config. By default it's 64 and I would like to upgrade to 512 like I did on my hosted database.

On my hosted database, I was able to do it with the `supabase postgres-config update` command (documented here https://supabase.com/docs/guides/platform/custom-postgres-config) but it does not seem to work locally as it requires a `project-ref` that I don't have on my local environment.

So, I decided to go inside the docker instance for my database, find the postgresql.conf file located in /etc/postgresql but if I update the `max_locks_per_transaction` there and then restart the instance (by doing `supabase stop` then `supabase start`), my changes in the postgresql.conf file are erased.

So my question is: what's the best way to update the PostgreSQL config when using a local environment?

Thank you.

omaiboroda:
@vinch , we have overcome this with

```sh
docker exec -it <container_name> bash -c "echo \"maintenance_work_mem = '128MB'\" >> /etc/postgresql/postgresql.conf && su postgres -c '/usr/lib/postgresql/15/bin/pg_ctl reload'"
```

PR #2787, opened 2024-10-21
feat: add db postgres settings to config

avallete:

Add "postgres" settings to the config so they can be synced on branching (and maybe pushed all at once from cli later)

Will work along with:

- https://github.com/supabase/infrastructure/pull/20325

## What is the new behavior?

## Additional context


pkg/config/db.go
```diff
@@ -0,0 +1,134 @@
...
+	v1API "github.com/supabase/cli/pkg/api"
+	"github.com/supabase/cli/pkg/cast"
+	"github.com/supabase/cli/pkg/diff"
+)
+
+type (
+	PoolMode string
+
+	// All of thoses are remote only settings that'll apply to supabase hosted database
+	remoteDb struct {
```
avallete, line 13: **remark**

I've also considered moving this under `[experimental.db]` config instead, maybe that will be more suitable ? Since we'll need to update the docs for it anyways 🤔

pkg/config/db.go
```diff
@@ -0,0 +1,158 @@
...
+
+// Compare two db config, if changes requires restart return true, return false otherwise
+func requireDbRestart(a *db, b *db) bool {
+	return !isPointerValueEquals(a.MaxConnections, b.MaxConnections) ||
+		!isPointerValueEquals(a.MaxWorkerProcesses, b.MaxWorkerProcesses) ||
+		!isPointerValueEquals(a.MaxParallelWorkers, b.MaxParallelWorkers) ||
+		!isPointerValueEquals(a.MaxWalSenders, b.MaxWalSenders) ||
+		!isPointerValueEquals(a.MaxReplicationSlots, b.MaxReplicationSlots) ||
+		!isPointerValueEquals(a.SharedBuffers, b.SharedBuffers)
+}
```
avallete, line 88: **remark**

I see that the update endpoint allow to pass a `restart` parameter so it's externally controlled. I guess we want to gives users full control about wether to restart or not when they change each field one by one. But when we update all the parameters at once (like branching) we wanna automatically restart if needed.

pkg/config/db.go
sweatybridge: We can probably nest it under `[db.settings]`. It would nicely solve https://github.com/supabase/cli/issues/2611 once we support it locally.

avallete: Yup remembered I saw something like this somewhere. Gonna put that in my todo list thank's for pointing it out.

pkg/cast/cast.go
```diff
@@ -20,6 +20,31 @@ func IntToUint(value int) uint {
...
+func UintToIntPtr(value *uint) *int {
+	if value == nil {
+		return nil
+	}
+	if *value <= math.MaxInt {
+		result := UintToInt(*value)
+		return &result
+	}
+	maxInt := math.MaxInt
+	return &maxInt
```
sweatybridge, line 32:
```suggestion
	return Ptr(UintToInt(*value))
```

pkg/cast/cast.go
```diff
@@ -20,6 +20,31 @@ func IntToUint(value int) uint {
...
+	var result uint
+	result = 0
+	if value == nil {
+		return nil
+	}
+	if *value < 0 {
+		return &result
+	}
+	result = IntToUint(*value)
+	return &result
```
sweatybridge, line 45:
```suggestion
	if value == nil {
		return nil
	}
	return Ptr(IntToUint(*value))
```

sweatybridge (commented):
nitpicks

pkg/config/db.go
```diff
@@ -0,0 +1,158 @@
...
+		return true
+	}
+	if a == nil || b == nil {
+		return false
+	}
+	return *a == *b
+}
+
+// Compare two db config, if changes requires restart return true, return false otherwise
+func requireDbRestart(a *db, b *db) bool {
```
sweatybridge, line 79:
```suggestion
func (a settings) requireDbRestart(b settings) bool {
```

Seems nicer to make this a struct method.

Also you can use `cmp.Equal(a.MaxConnections, b.MaxConnections)` instead of implementing it ourselves.

pkg/config/db.go
```diff
@@ -0,0 +1,156 @@
+package config
+
+import (
+	v1API "github.com/supabase/cli/pkg/api"
+	"github.com/supabase/cli/pkg/cast"
+	"github.com/supabase/cli/pkg/diff"
+)
+
+type (
+	PoolMode string
```
sweatybridge, line 10: Let's move the entire enum declaration here https://github.com/supabase/cli/blob/0377d4730f5f9ba3a8c7c64cfef37584a9d82150/pkg/config/config.go#L60-L65

pkg/config/db.go
```diff
@@ -0,0 +1,156 @@
...
+		MaxParallelWorkers            *uint   `toml:"max_parallel_workers"`
+		MaxParallelWorkersPerGather   *uint   `toml:"max_parallel_workers_per_gather"`
+		MaxReplicationSlots           *uint   `toml:"max_replication_slots"`
+		MaxSlotWalKeepSize            *string `toml:"max_slot_wal_keep_size"`
+		MaxStandbyArchiveDelay        *string `toml:"max_standby_archive_delay"`
+		MaxStandbyStreamingDelay      *string `toml:"max_standby_streaming_delay"`
+		MaxWalSize                    *string `toml:"max_wal_size"`
+		MaxWalSenders                 *uint   `toml:"max_wal_senders"`
+		MaxWorkerProcesses            *uint   `toml:"max_worker_processes"`
+		SessionReplicationRole        *string `toml:"session_replication_role"`
```
sweatybridge, line 27: And also declare an enum for `SessionReplicationRole` similar to PoolMode

pkg/config/db.go
```diff
@@ -0,0 +1,156 @@
...
+	result.Settings.SessionReplicationRole = (*string)(remoteConfig.SessionReplicationRole)
+	result.Settings.SharedBuffers = remoteConfig.SharedBuffers
+	result.Settings.StatementTimeout = remoteConfig.StatementTimeout
+	result.Settings.WalKeepSize = remoteConfig.WalKeepSize
+	result.Settings.WalSenderTimeout = remoteConfig.WalSenderTimeout
+	result.Settings.WorkMem = remoteConfig.WorkMem
+	return result
+}
+
+func (a *db) DiffWithRemote(remoteConfig v1API.PostgresConfigResponse) ([]byte, error) {
```
sweatybridge, line 145: Perhaps one more suggestion is to make these methods of `settings` struct so that it's cleaner to support syncing `pooler` config in the future.

```suggestion
func (a *settings) DiffWithRemote(remoteConfig v1API.PostgresConfigResponse) ([]byte, error) {
```

pkg/config/config.go
```diff
@@ -775,6 +738,12 @@ func (c *baseConfig) Validate(fsys fs.FS) error {
 		}
 	}
 	// Validate db config
+	if c.Db.Settings.SessionReplicationRole != nil {
+		allowedRoles := []SessionReplicationRole{"origin", "replica", "local"}
```
sweatybridge, line 742:
```suggestion
		allowedRoles := []SessionReplicationRole{SessionReplicationRoleOrigin, SessionReplicationRoleReplica, SessionReplicationRoleLocal}
```

pkg/config/updater.go
```diff
@@ -40,10 +42,48 @@ func (u *ConfigUpdater) UpdateApiConfig(ctx context.Context, projectRef string,
...
+		return err
+	} else if len(dbDiff) == 0 {
+		fmt.Fprintln(os.Stderr, "Remote DB config is up to date.")
+		return nil
+	}
+	fmt.Fprintln(os.Stderr, "Updating DB service with config:", string(dbDiff))
+	remoteConfig := s.fromRemoteConfig(*dbConfig.JSON200)
+	restartRequired := s.requireDbRestart(remoteConfig)
+	if restartRequired {
+		fmt.Fprintln(os.Stderr, "DB service updates will require database restart...")
```
sweatybridge, line 72:
```suggestion
		fmt.Fprintln(os.Stderr, "Database will be restarted to apply config updates...")
```

sweatybridge (approved):

PR #2788, opened 2024-10-21
fix: empty array is valid api config

sweatybridge:

Bug fix

## What is the new behavior?

Empty array is a valid config if user doesn't want to expose any schemas or search paths.

```
[api]
schemas = []
extra_search_path = []
```

## Additional context

### DIFF
--- a/pkg/config/api.go
+++ b/pkg/config/api.go
@@ -38,16 +38,10 @@ func (a *api) ToUpdatePostgrestConfigBody() v1API.UpdatePostgrestConfigBody {
 	}
 
 	// Convert Schemas to a comma-separated string
-	if len(a.Schemas) > 0 {
-		schemas := strings.Join(a.Schemas, ",")
-		body.DbSchema = &schemas
-	}
+	body.DbSchema = cast.Ptr(strings.Join(a.Schemas, ","))
 
 	// Convert ExtraSearchPath to a comma-separated string
-	if len(a.ExtraSearchPath) > 0 {
-		extraSearchPath := strings.Join(a.ExtraSearchPath, ",")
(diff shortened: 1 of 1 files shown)


pkg/config/api.go
```diff
@@ -38,16 +38,10 @@ func (a *api) ToUpdatePostgrestConfigBody() v1API.UpdatePostgrestConfigBody {
...
-		body.DbSchema = &schemas
-	}
+	body.DbSchema = cast.Ptr(strings.Join(a.Schemas, ","))
 
 	// Convert ExtraSearchPath to a comma-separated string
-	if len(a.ExtraSearchPath) > 0 {
-		extraSearchPath := strings.Join(a.ExtraSearchPath, ",")
-		body.DbExtraSearchPath = &extraSearchPath
-	}
+	body.DbExtraSearchPath = cast.Ptr(strings.Join(a.ExtraSearchPath, ","))
```
avallete, line 44: **question**

Doesn't that end up being the same as if the user did:

```
[api]
enabled = False
```

I guess both can work anyways but their results will be the same runtime wise 🤔

avallete (approved):

pkg/config/api.go
sweatybridge: hmm currently it's a bit confusing that setting empty arrays is the same as not updating this field.

I will try to make them pointers like you said to distinguish default and non-default values.

sweatybridge: Actually this whole toggle to disable api is pretty weird. I'd expect us to remove the `/rest/v1` path from api gateway or something. Currently you can still hit `https://ref.supabase.co/rest/v1/` for an empty openapi spec 🤷‍♂️

PR #2794, opened 2024-10-23
feat(config): experimental config webhooks

avallete:

- Add a [experimental.webhooks] section to the config
- Add necessary changes for use with the updater

Related:

https://github.com/supabase/infrastructure/pull/20348

## What is the new behavior?

## Additional context

### DIFF
--- a/pkg/config/updater.go
+++ b/pkg/config/updater.go
@@ -24,7 +24,7 @@ func (u *ConfigUpdater) UpdateRemoteConfig(ctx context.Context, remote baseConfi
 	if err := u.UpdateDbConfig(ctx, remote.ProjectId, remote.Db); err != nil {
 		return err
 	}
-	if err := u.UpdateExperimentalConfig(ctx, remote.ProjectId, &remote.Experimental); err != nil {
+	if err := u.UpdateExperimentalConfig(ctx, remote.ProjectId, remote.Experimental); err != nil {
 		return err
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


pkg/config/updater.go
```diff
@@ -24,6 +24,9 @@ func (u *ConfigUpdater) UpdateRemoteConfig(ctx context.Context, remote baseConfi
 	if err := u.UpdateDbConfig(ctx, remote.ProjectId, remote.Db); err != nil {
 		return err
 	}
+	if err := u.UpdateExperimentalConfig(ctx, remote.ProjectId, &remote.Experimental); err != nil {
```
sweatybridge, line 27:
```suggestion
	if err := u.UpdateExperimentalConfig(ctx, remote.ProjectId, remote.Experimental); err != nil {
```

`experimental` block should always exist, so let's pass by value.

pkg/config/updater.go
```diff
@@ -87,3 +90,20 @@ func (u *ConfigUpdater) UpdateDbConfig(ctx context.Context, projectRef string, c
...
+	if exp == nil {
+		return nil
+	}
+
+	if exp.Webhooks != nil && exp.Webhooks.Enabled {
+		fmt.Fprintln(os.Stderr, "Enabling webhooks for the project...")
+
+		if _, err := u.client.V1EnableDatabaseWebhookWithResponse(ctx, projectRef); err != nil {
+			return errors.Errorf("failed to enable webhooks: %w", err)
+		}
```
sweatybridge, line 104: Remember to check response status code in case it's not 200.

avallete: Oh right, I assumed if it's not we'll get an err. Nice catch.

sweatybridge (approved):

PR #2796, opened 2024-10-23
feat(config): allow local postgres configuration

### LINKED ISSUE #2611: How to update the PostgreSQL config in the local environment?

avallete:

- Apply settings set under db.settings to the local running db instance

## What is the new behavior?

## Additional context

Closes #2611

### DIFF
--- a/internal/db/start/start.go
+++ b/internal/db/start/start.go
@@ -56,7 +56,6 @@ func NewContainerConfig() container.Config {
 	env := []string{
 		"POSTGRES_PASSWORD=" + utils.Config.Db.Password,
 		"POSTGRES_HOST=/var/run/postgresql",
-		"POSTGRES_INITDB_ARGS=--lc-ctype=C.UTF-8",
 		"JWT_SECRET=" + utils.Config.Auth.JwtSecret,
 		fmt.Sprintf("JWT_EXP=%d", utils.Config.Auth.JwtExpiry),
 	}
@@ -81,13 +80,18 @@ func NewContainerConfig() container.Config {
 			Timeout:  2 * time.Second,
 			Retries:  3,
 		},
-		Entrypoint: []string{"sh", "-c", `cat <<'EOF' > /etc/postgresql.schema.sql && cat <<'EOF' > /etc/postgresql-custom/pgsodium_root.key && docker-entrypoint.sh postgres -D /etc/postgresql
+		Entrypoint: []string{"sh", "-c", `
+cat <<'EOF' > /etc/postgresql.schema.sql && \
+cat <<'EOF' > /etc/postgresql-custom/pgsodium_root.key && \
+cat <<'EOF' >> /etc/postgresql/postgresql.conf && \
+docker-entrypoint.sh postgres -D /etc/postgresql
 ` + initialSchema + `
 ` + _supabaseSchema + `
 EOF
 ` + utils.Config.Db.RootKey + `
 EOF
-`},
+` + utils.Config.Db.Settings.ToPostgresConfig() + `
+EOF`},
 	}
 	if utils.Config.Db.MajorVersion >= 14 {
 		config.Cmd = []string{"postgres",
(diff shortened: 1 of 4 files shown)


internal/db/start/start.go
```diff
@@ -56,7 +56,6 @@ func NewContainerConfig() container.Config {
 	env := []string{
 		"POSTGRES_PASSWORD=" + utils.Config.Db.Password,
 		"POSTGRES_HOST=/var/run/postgresql",
-		"POSTGRES_INITDB_ARGS=--lc-ctype=C.UTF-8",
```
sweatybridge, line 59: is this intentional? we need this to set the same locale as platform db

internal/db/start/start.go
```diff
@@ -81,13 +80,18 @@ func NewContainerConfig() container.Config {
 			Timeout:  2 * time.Second,
 			Retries:  3,
 		},
-		Entrypoint: []string{"sh", "-c", `cat <<'EOF' > /etc/postgresql.schema.sql && cat <<'EOF' > /etc/postgresql-custom/pgsodium_root.key && docker-entrypoint.sh postgres -D /etc/postgresql
+		Entrypoint: []string{"sh", "-c", `
+cat <<'EOF' > /etc/postgresql.schema.sql && \
+cat <<'EOF' > /etc/postgresql-custom/pgsodium_root.key && \
+cat <<'EOF' >> /etc/postgresql/postgresql.conf && \
```
sweatybridge, line 86: Isn't it problematic to append to `postgresql.conf` on every db restart? the file will keep growing in size. I'm not sure what would be a better solution though

avallete: I don't think `/etc/postgresql` is mounted in a volume, so restart should always start with the same file. I'll have a look to be sure.

internal/db/start/start.go
avallete: Yeah this is override down the line by an if/else statement: https://github.com/supabase/cli/pull/2796/files/de420f3044ea89994162aea89953559299968ba5#diff-f635ef2693ae7184092f49b3ed3c642a7d3bf0e727de4ff03ea207591c1ef0c2R62-R73

So seemed to me that this doesn't apply.

internal/db/start/start.go
sweatybridge: thanks for verifying

pkg/config/db.go
```diff
@@ -146,6 +150,62 @@ func (a *settings) fromRemoteConfig(remoteConfig v1API.PostgresConfigResponse) s
 	return result
 }
 
+// create a valid string to append to /etc/postgresql/postgresql.conf
+func (a *settings) ToPostgresConfig() string {
```
sweatybridge, line 154: I think we can implement a shortcut here 

```
const pgConfHeader = "# supabase [db.settings] configuration\n"

// create a valid string to append to /etc/postgresql/postgresql.conf
func (a *settings) ToPostgresConfig() string {
	data, _ := ToTomlBytes(*a)
	body := bytes.ReplaceAll(data, []byte{'"'}, []byte{'\''})
	return pgConfHeader + string(body)
}
```

sweatybridge (approved):

PR #2813, opened 2024-10-29
feat(functions): allow custom functions entrypoint path

sweatybridge:

feature

## What is the new behavior?

Allow functions entrypoint to be customized, ie. `functions/my-slug/index.js`

TODO:
- [x] support arbitrary entrypoint path, ie. `../slug/sample.js`

## Additional context

### DIFF
--- a/internal/functions/serve/templates/main.ts
+++ b/internal/functions/serve/templates/main.ts
@@ -2,6 +2,7 @@ import {
   STATUS_CODE,
   STATUS_TEXT,
 } from "https://deno.land/std/http/status.ts";
+import * as posix from "https://deno.land/std/path/posix/mod.ts";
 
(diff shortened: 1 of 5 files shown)


internal/functions/serve/templates/main.ts
```diff
@@ -2,6 +2,7 @@ import {
   STATUS_CODE,
   STATUS_TEXT,
 } from "https://deno.land/std/http/status.ts";
+import * as posix from "https://deno.land/std/path/posix/mod.ts";
```
laktek, line 5: can use `path/mod.ts` right? doesn't need `posix`

sweatybridge: `path/posix` imports fewer dependencies, ie. excluding path/windows which is never needed in our containerized runtime.

Issue #2552, opened 2024-07-23
Show auth config diff after supabase link


mezoistvan:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

After `supabase init`, if the linked and local config is different, there is a message which lists the differences in the config:

```
$ supabase link
Selected project: .................................
Enter your database password (or leave blank to skip): 
Finished supabase link.
WARNING: Local config differs from linked project. Try updating supabase/config.toml
[api]
enabled = true
port = 54321
schemas = ["public", "graphql_public"]
extra_search_path = ["public", "extensions"]
max_rows = 1000
[api.tls]
enabled = false
```

This message does not include all differences. For example if `[auth.email] enable_confirmations` differs, it's not shown. This is misleading and confusing.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Go to a remote project on supabase.com, enable email auth and **disable** `enable_confirmations`
2. Create a local project and set up `config.toml` with `[email.auth] enable_confirmations` **enabled**.
3. Link your remote project to your local one and see the difference.

## Expected behavior

All diffs should be visible in the message

## System information

- OS: macOS
- Version of supabase cli: 1.187.3
- Version of Node.js: 20

## Additional context

Is there a way to see when a linked config like this was changed? Is there a way to log these config changes on a linked project?

kiarafbickers:
I'm having this same issue.

kiarafbickers:
This might be similar to issue https://github.com/supabase/cli/issues/1483

ThingEngineer:
This is happening again in v1.226.3

No issues, just the warning message after liking and a diff between the local and remote supabase/config.toml.

sweatybridge:
As requested by the original ticket, we have released enhancement to diff the entire auth config between local and linked project. 

You can update your local config.toml according to the suggested diff to resolve the warnings.

ThingEngineer:
Thank you @sweatybridge, since this does not cause an actual problem with anything I don't feel it warrants it's own issue. However, just as an fyi there may be one side effect to consider. 

[...]

PR #2814, opened 2024-10-29
feat(config): add diff with remotes for auth

avallete:

feature
relates to https://github.com/supabase/cli/issues/2552

## What is the new behavior?

## Additional context


pkg/config/auth.go
```diff
@@ -0,0 +1,1124 @@
...
+	}
+	// Sensitives fields
+	if remoteConfig.SmtpAdminEmail != nil {
+		result.Email.Smtp.AdminEmail = *remoteConfig.SmtpAdminEmail
+	}
+	if remoteConfig.SmtpHost != nil {
+		result.Email.Smtp.Host = *remoteConfig.SmtpHost
+	}
+	if remoteConfig.SmtpPass != nil {
+		result.Email.Smtp.Pass = *remoteConfig.SmtpPass
```
sweatybridge, line 471: I think this value is masked in api response.

pkg/config/auth.go
```diff
@@ -0,0 +1,1124 @@
...
+
+func (original *auth) compareAndHideSensitiveFields(remote *auth) {
+	// This function compares the original auth struct with a remote auth struct
+	// and hides sensitive fields in both structs for secure comparison
+	// SMTP sensitive fields
+	diff.CompareSensitiveField(&original.Email.Smtp.AdminEmail, &remote.Email.Smtp.AdminEmail)
+	diff.CompareSensitiveField(&original.Email.Smtp.Host, &remote.Email.Smtp.Host)
+	diff.CompareSensitiveField(&original.Email.Smtp.User, &remote.Email.Smtp.User)
+	diff.CompareSensitiveField(&original.Email.Smtp.SenderName, &remote.Email.Smtp.SenderName)
+	diff.CompareSensitiveField(&original.Email.Smtp.Pass, &remote.Email.Smtp.Pass)
```
sweatybridge, line 847: We probably only need to treat secrets, api keys, and passwords as sensitive. Other strings are ok

pkg/config/auth.go
```diff
@@ -0,0 +1,1162 @@
...
+
+func compareHashedField(secret string, local, remote *string) {
+	if remote == nil {
+		return
+	}
+	if *local == "" && *remote == "" {
+		return
+	}
+	localHashed := sha256Hmac(secret, *local)
+	diff.CompareSensitiveField(&localHashed, remote)
```
avallete, line 207: **remark**

Here I still use `CompareSentitiveField` to keep a human readable result message. We could just use the standard comparaison over the hashed values but I think that might confuse the users wondering why the value they set in their config get a different value in their config diff.

pkg/config/auth.go
```diff
@@ -0,0 +1,1156 @@
...
+
+func (a *auth) mapRemoteEmailTemplates(remoteConfig v1API.AuthConfigResponse) {
+	for name, template := range a.Email.Template {
+		switch name {
+		case "invite":
+			if remoteConfig.MailerSubjectsInvite != nil {
+				template.Subject = *remoteConfig.MailerSubjectsInvite
+			}
+			if remoteConfig.MailerTemplatesInviteContent != nil {
+				template.ContentPath = *remoteConfig.MailerTemplatesInviteContent
```
sweatybridge, line 555: We need to implement a new solution for this. The api returns file content but here we are assigning to a file path.

Perhaps writing to a predefined path would be sufficient.

sweatybridge: I'm also thinking to deprecate this `ContentPath` for simply `Content` field that accepts multi line string as input.

One inconsistency with the current `ContentPath` field is that it's relative to workdir and not the supabase dir like all other paths in config.toml.

avallete: So the "template" would be directly embedded within the `config.toml`, I guess that would match what we have for the `mfa.phone` template message  🤔 

I do like to split my things up so having the templates in dedicated files in dedicated path if the content is more than a few lines long, but I can see the benefits of "having it all in a single place" as well. So this one is really a matter of preference.

If we wanna go that way basically, when we load the config, we look the content in the config, if not found, fallback on the path, and use this content sounds good to me.

sweatybridge:
closing as all parts have been split to smaller PRs

PR #2818, opened 2024-10-30
feat(inspect): add role-configs command to list db roles with custom settings

encima:

Feature

## What is the new behavior?

New `role-configs` command under `inspect` to show all users that have a configuration that is different from the standard

## Additional context

Useful for debugging and auditing

### DIFF
--- a/internal/inspect/role_configs/role_configs.go
+++ b/internal/inspect/role_configs/role_configs.go
@@ -1,0 +1,50 @@
+package role_configs
+
+import (
+	"context"
+	_ "embed"
+	"fmt"
+
+	"github.com/go-errors/errors"
+	"github.com/jackc/pgconn"
+	"github.com/jackc/pgx/v4"
+	"github.com/spf13/afero"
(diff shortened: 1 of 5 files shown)


internal/inspect/role_configs/role_configs.go
```diff
@@ -0,0 +1,50 @@
...
+	table := "|Role Name|Custom config|\n|-|-|\n"
+	for _, r := range result {
+		table += fmt.Sprintf("|`%s`|`%s`|\n", r.Role_name, r.Custom_config)
+	}
+
+	if err := list.RenderTable(table); err != nil {
+		return err
+	}
+
+	return nil
```
sweatybridge, line 49:
```suggestion
	table := "|Role name|Custom config|\n|-|-|\n"
	for _, r := range result {
		table += fmt.Sprintf("|`%s`|`%s`|\n", r.Role_name, r.Custom_config)
	}

	return list.RenderTable(table)
```

sweatybridge (approved):
just a nitpick

PR #2824, opened 2024-11-01
feat(config): load local deno.json file as import map

jgoux:

feature

Currently each edge function shares a global `import_map.json` by default.
To use one particular import map per function the user has to define it manually in `config.toml`

## What is the new behavior?

We now automatically load `deno.json` and `deno.jsonc` under the `supabase/functions/<slug>` directory.
Each function should be treated as a different project with its own set of dependencies.

## Additional context

`import_map.json` disappeared from Deno's doc since as of v1.19.2: https://deno.land/x/manual@v1.19.1/npm_nodejs/import_maps.md?source

Imports should now be declared as part of deno's config file `deno.json`.

We might want to deprecate `import_map.json`, especially when we will ship deno v2 support.

### DIFF
--- a/internal/functions/deploy/deploy.go
+++ b/internal/functions/deploy/deploy.go
@@ -81,20 +81,12 @@ func GetFunctionConfig(slugs []string, importMapPath string, noVerifyJWT *bool, 
 	functionConfig := make(config.FunctionConfig, len(slugs))
 	for _, name := range slugs {
 		function := utils.Config.Functions[name]
-		functionDir := filepath.Join(utils.FunctionsDir, name)
 		// Precedence order: flag > config > fallback
 		if len(function.Entrypoint) == 0 {
-			function.Entrypoint = filepath.Join(functionDir, "index.ts")
+			function.Entrypoint = filepath.Join(utils.FunctionsDir, name, "index.ts")
 		}
-		// Check for deno.json or deno.jsonc in functionDir
-		denoJsonPath := filepath.Join(functionDir, "deno.json")
-		denoJsoncPath := filepath.Join(functionDir, "deno.jsonc")
 		if len(importMapPath) > 0 {
 			function.ImportMap = importMapPath
-		} else if _, err := fsys.Stat(denoJsonPath); err == nil {
-			function.ImportMap = denoJsonPath
-		} else if _, err := fsys.Stat(denoJsoncPath); err == nil {
-			function.ImportMap = denoJsoncPath
 		} else if len(function.ImportMap) == 0 && fallbackExists {
 			function.ImportMap = utils.FallbackImportMapPath
 		}
--- a/internal/functions/deploy/deploy.go
+++ b/internal/functions/deploy/deploy.go
@@ -81,12 +81,20 @@ func GetFunctionConfig(slugs []string, importMapPath string, noVerifyJWT *bool, 
 	functionConfig := make(config.FunctionConfig, len(slugs))
 	for _, name := range slugs {
 		function := utils.Config.Functions[name]
+		functionDir := filepath.Join(utils.FunctionsDir, name)
 		// Precedence order: flag > config > fallback
 		if len(function.Entrypoint) == 0 {
-			function.Entrypoint = filepath.Join(utils.FunctionsDir, name, "index.ts")
+			function.Entrypoint = filepath.Join(functionDir, "index.ts")
 		}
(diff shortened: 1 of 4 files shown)


internal/functions/deploy/deploy.go
```diff
@@ -81,12 +81,20 @@ func GetFunctionConfig(slugs []string, importMapPath string, noVerifyJWT *bool,
...
 		}
+		// Check for deno.json or deno.jsonc in functionDir
+		denoJsonPath := filepath.Join(functionDir, "deno.json")
+		denoJsoncPath := filepath.Join(functionDir, "deno.jsonc")
 		if len(importMapPath) > 0 {
 			function.ImportMap = importMapPath
+		} else if _, err := fsys.Stat(denoJsonPath); err == nil {
+			function.ImportMap = denoJsonPath
+		} else if _, err := fsys.Stat(denoJsoncPath); err == nil {
+			function.ImportMap = denoJsoncPath
```
sweatybridge, line 97: Is this also the default path that deno also looks for?

Either way, we'd want to move this code to [config](https://github.com/supabase/cli/blob/develop/pkg/config/config.go#L666) so it works for serve and branching deployments too.

jgoux: Yes it works exactly like `package.json`:

> Deno will automatically detect a deno.json or deno.jsonc configuration file if it's in your current working directory or parent directories. The --config flag can be used to specify a different configuration file.

I'll move the code to config then. 👍

internal/functions/deploy/deploy_test.go
jgoux, line 1: I'm keeping this change because previously we used the fallback path as the cli flag which isn't showing that the CLI really overrides the default fallback as they have the same path. 😅

pkg/config/config.go
```diff
@@ -654,19 +653,46 @@ func (c *config) Load(path string, fsys fs.FS) error {
...
+		c.Functions = make(FunctionConfig)
+	}
+	for _, entry := range entries {
+		if !entry.IsDir() {
+			continue
+		}
+		slug := entry.Name()
+		fn, exists := c.Functions[slug]
+		if !exists {
+			fn = function{}
```
jgoux, line 675: I had to rename `function` to `fn` to avoid conflicts with the type `function`.

pkg/config/config.go
```diff
@@ -654,19 +653,46 @@ func (c *config) Load(path string, fsys fs.FS) error {
 		}
 		c.Storage.Buckets[name] = bucket
 	}
-	for slug, function := range c.Functions {
+
+	// Load functions config
+	entries, err := fs.ReadDir(fsys, builder.FunctionsDir)
```
jgoux, line 658: We now have to scan the functionsDir in order to detect `deno.json(c)` files and populate the `ImportMap` accordingly.

internal/functions/deploy/deploy.go
jgoux: Logic moved to config. ✅

pkg/config/config.go
```diff
@@ -654,19 +653,50 @@ func (c *config) Load(path string, fsys fs.FS) error {
...
+	// Load functions config
+	pattern := filepath.Join(builder.FunctionsDir, "*", "index.ts")
+	paths, err := fs.Glob(fsys, pattern)
+	if err != nil {
+		return errors.Errorf("failed to glob function slugs: %w", err)
+	}
+
+	if c.Functions == nil {
+		c.Functions = make(FunctionConfig)
+	}
```
sweatybridge, line 666: Globbing functions this way is very fragile because users often have shared components that they don't want to deploy as functions, for eg. `functions/shared/index.ts`.

sweatybridge (changes requested):

sweatybridge (approved):

pkg/config/config.go
jgoux: I took the code from https://github.com/supabase/cli/blob/4cf189afa6dc8b9738b1de31e4c06fcb6d5e1aaf/internal/functions/deploy/deploy.go#L49-L66
Should we also change this part then?

sweatybridge: No need to update deploy for the time being. We left it for backwards compatibility with those who use setup-cli GitHub action.

Users only need to declare if they want to deploy to branches using our new integration.

tazmaniax:
As per the additional context in the description can a global deno.json now be used instead of an import_map.json with this change?

PR #2835, opened 2024-11-05
chore: refactor let api decide for restart

avallete:

Following: https://github.com/supabase/infrastructure/pull/20462

Let the api decide if the service restart is necessary.

**caution**
Deduplicating the logic remove our ability to log a message to the user but since we didn't allowed to act on it (agree / refute) I think it's alright to let the API be the source of truth.

## What is the new behavior?

## Additional context

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,6 @@ require (
 	github.com/go-xmlfmt/xmlfmt v1.1.2
 	github.com/golang-jwt/jwt/v5 v5.2.1
 	github.com/golangci/golangci-lint v1.61.0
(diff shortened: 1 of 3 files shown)


sweatybridge (approved):

sweatybridge:
The other change that we want to do on the worker side is to trigger SIGUP on the database to apply the new config when not restarting https://supabase.slack.com/archives/C02FHG9QQAF/p1729925142614479?thread_ts=1728485737.099679&cid=C02FHG9QQAF

PR #2839, opened 2024-11-06
feat(config): sync external provider config to remote

sweatybridge:

part 3 of https://github.com/supabase/cli/pull/2814

## What is the new behavior?

Also adds `figma` and `kakao` auth providers to local.

## Additional context

### DIFF
--- a/pkg/config/auth.go
+++ b/pkg/config/auth.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"maps"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/config/auth.go
```diff
@@ -285,6 +291,265 @@ func (s *sms) fromAuthConfig(remoteConfig v1API.AuthConfigResponse) {
 	}
 }
 
+func (e external) toAuthConfigBody(body *v1API.UpdateAuthConfigBody) {
+	if len(e) == 0 {
+		return
+	}
+	var p *provider
+	// Ignore configs of disabled providers because their envs are not loaded
```
avallete, line 299: **remark**

I think that might be a marker that we wanna refactor the way we load the env/config 🤔 

One approach would be to always load it everywhere it's mentioned, whether the actual service is enabled or not. And if it raise an error, then the user can just "comment" or remove the errored line if he still want to keep the thing around.

avallete (approved):

pkg/config/auth.go
sweatybridge: Yea possibly. We would have to consider this change carefully though because it could break CI/CD pipelines that rely on setting `enabled = false` but not commenting out the usage of env.

Issue #2846, opened 2024-11-07
Custom path mapping for dependencies with deno.json seems to be broken


SanjoSolutions:
**Describe the bug**
The custom path mapping for dependencies with deno.json as described in the [documentation](https://supabase.com/docs/guides/functions/import-maps#using-denojson-recommended) seems to be broken.

It works with the [legacy import maps method](https://supabase.com/docs/guides/functions/import-maps#using-import-maps-legacy).

**System information**
 - Version of OS: Windows 11
 - Version of CLI: 1.215.0

I hope the information is sufficient to reproduce it.

sweatybridge:
Hi @SanjoSolutions, thanks for reporting this issue. Are you perhaps using relative path imports in deno.json? Could you give us a minimal reproducible example to illustrate what's broken?

SanjoSolutions:
> Are you perhaps using relative path imports in deno.json?

Yes.

In my case the content is:

```json
{
  "imports": {
    "basejump/": "../../../basejump-deno-packages/"
  }
}
```

With the legacy import maps this works.
With the deno.json method with an import like `import {
  billingFunctionsWrapper,
  stripeFunctionHandler,
} from "basejump/billing-functions/mod.ts";` I receive the following error:

```sh
$ npx supabase functions serve
Setting up Edge Functions runtime...
Serving functions on http://127.0.0.1:54321/functions/v1/<function-name>
Using supabase-edge-runtime-1.60.1 (compatible with Deno v1.45.2)
serving the request with supabase/functions/billing-functions
worker boot error: failed to create the graph: Relative import path "basejump/billing-functions/mod.ts" not prefixed with / or ./ or ../
    at file:///Users/jonas/saas-template/supabase/functions/billing-functions/index.ts:5:8
worker boot error: failed to create the graph: Relative import path "basejump/billing-functions/mod.ts" not prefixed with / or ./ or ../
    at file:///Users/jonas/saas-template/supabase/functions/billing-functions/index.ts:5:8
InvalidWorkerCreation: worker boot error: failed to create the graph: Relative import path "basejump/billing-functions/mod.ts" not prefixed with / or ./ or ../
    at file:///Users/jonas/saas-template/supabase/functions/billing-functions/index.ts:5:8
    at async UserWorker.create (ext:sb_user_workers/user_workers.js:145:15)
    at async Object.handler (file:///root/index.ts:156:22)
    at async respond (ext:sb_core_main_js/js/http.js:163:14) {
[...]

PR #2852, opened 2024-11-07
fix: auto detect deno.json for globbed slugs

sweatybridge:

Bug fix https://github.com/supabase/cli/issues/2846

`import_map` is only set if the function is declared in config.toml

## What is the new behavior?

Also set `import_map` for function slugs globbed from local directories.

## Additional context


jgoux (approved):
LGTM!

PR #2861, opened 2024-11-08
feat(config): sync hook and mfa config to remote

sweatybridge:

part 4 of https://github.com/supabase/cli/pull/2814

## What is the new behavior?

Sync `hook`, `mfa`, and `sessions` config to remote.

## Additional context

### DIFF
--- a/pkg/config/auth_test.go
+++ b/pkg/config/auth_test.go
@@ -87,6 +87,124 @@ func TestHookDiff(t *testing.T) {
 	})
 }
 
+func TestMfaDiff(t *testing.T) {
+	t.Run("local and remote enabled", func(t *testing.T) {
+		c := auth{EnableSignup: true, MFA: mfa{
+			TOTP: factorTypeConfiguration{
+				EnrollEnabled: true,
+				VerifyEnabled: true,
+			},
+			Phone: phoneFactorTypeConfiguration{
+				factorTypeConfiguration: factorTypeConfiguration{
+					EnrollEnabled: true,
+					VerifyEnabled: true,
+				},
+				OtpLength:    6,
+				Template:     "Your code is {{ .Code }}",
+				MaxFrequency: 5 * time.Second,
+			},
+			WebAuthn: factorTypeConfiguration{
+				EnrollEnabled: true,
+				VerifyEnabled: true,
+			},
+			MaxEnrolledFactors: 10,
+		}}
+		// Run test
+		diff, err := c.DiffWithRemote("", v1API.AuthConfigResponse{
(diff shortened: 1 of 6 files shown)


pkg/config/auth_test.go
```diff
@@ -9,6 +9,202 @@ import (
...
+		assert.Contains(t, string(diff), `[hook.mfa_verification_attempt]`)
+		assert.Contains(t, string(diff), `-enabled = true`)
+		assert.Contains(t, string(diff), `+enabled = false`)
+		assert.Contains(t, string(diff), `uri = ""`)
+		assert.Contains(t, string(diff), `secrets = ""`)
+
+		assert.Contains(t, string(diff), `[hook.custom_access_token]`)
+		assert.Contains(t, string(diff), `-enabled = false`)
+		assert.Contains(t, string(diff), `+enabled = true`)
+		assert.Contains(t, string(diff), `uri = ""`)
```
avallete, line 70: **question**

Is there an equivalent to `toMatchInlineSnapshot ` in Go ? That could maybe be useful instead of repeating the contains for every line as it would also test for the lines order.

pkg/config/templates/config.toml
```diff
@@ -184,12 +184,12 @@ enroll_enabled = true
...
-# verify_enabled = true
-# otp_length = 6
-# template = "Your code is {{ `{{ .Code }}` }} ."
-# max_frequency = "10s"
+[auth.mfa.phone]
+enroll_enabled = false
+verify_enabled = false
+otp_length = 6
+template = "Your code is {{ `{{ .Code }}` }}"
+max_frequency = "5s"
```
avallete, line 192: **question**

Will it cause any issue if it stay commented instead of being explicitly set to false ? Wondering about retro-compatibility with existing config.

avallete (approved):

pkg/config/auth_test.go
sweatybridge: Not that I'm aware of. But I'm interested as well.

A low hanging fruit is to write the expected diff output to a file under `testdata` for comparison.

pkg/config/templates/config.toml
sweatybridge: One problem is if user sets `verify_enabled = true` without uncommenting the other fields, the update code would sync zero values to the api. I think that's rarely desired because it would override platform defaults like `otp_length`.

We can of course handle zero values in code, by treating it as null and not sending to api. But 0 is actually a valid config value according to auth team.

So I think it's cleaner to set non-zero default values for these fields to match our platform. If we want to support nullable fields in the future, we can also change them into pointers.

PR #2874, opened 2024-11-12
fix(postgres-config): always restart if --no-restart is omitted

avallete:

With the new api changes: https://github.com/supabase/infrastructure/pull/20462

We must always the `restart_database` setting to the endpoint. Otherwise we go from the logic of "always restart except if --no-restart is specified" to a "restart only if the changed settings require a restart and --no-restart is not specified".

Which in the case of `maintenance_work_mem` might sometimes don't work as expected

Related: https://app.hubspot.com/live-messages/19953346/inbox/8348874005#comment

PS: An alternative would be to mark `maintenance_work_mem` as a parameter requiring restart on api side, but according to this documentation: https://postgresqlco.nf/doc/en/param/maintenance_work_mem/ it isn't the case.

## What is the new behavior?

## Additional context

### DIFF
--- a/internal/postgresConfig/update/update.go
+++ b/internal/postgresConfig/update/update.go
@@ -50,9 +50,8 @@ func Run(ctx context.Context, projectRef string, values []string, replaceOverrid
 	}
 	// 4. update config overrides and print out final result
 	{
-		if noRestart {
-			finalOverrides["restart_database"] = false
-		}
+		// Should always restart database except if --no-restart flag is provided
+		finalOverrides["restart_database"] = !noRestart
 		bts, err := json.Marshal(finalOverrides)
 		if err != nil {
 			return errors.Errorf("failed to serialize config overrides: %w", err)


sweatybridge:
> Which in the case of maintenance_work_mem might sometimes don't work as expected

I think a better fix for this is on infra worker, by sending a SIGUP to admin api running on both primary and read replicas. More context https://supabase.slack.com/archives/C02FHG9QQAF/p1729922433212149?thread_ts=1728485737.099679&cid=C02FHG9QQAF

avallete:
> > Which in the case of maintenance_work_mem might sometimes don't work as expected
> 
> I think a better fix for this is on infra worker, by sending a post request to reload config to admin api running on both primary and read replicas. This is needed whenever we update config without restarting database. More context https://supabase.slack.com/archives/C02FHG9QQAF/p1729922433212149?thread_ts=1728485737.099679&cid=C02FHG9QQAF

Closing in favor of: https://github.com/supabase/infrastructure/pull/20575

PR #2885, opened 2024-11-15
feat: add minimum password length and password requirements config

silentworks:

This feature can be found in the Supabase Hosted Dashboard but doesn't have a way to configure it in the `config.toml`.

No way to set **Password minimum length** and **Password requirements**

## What is the new behavior?

Can now set **Password minimum length** and **Password requirements** in the `config.toml` file.

## Additional context


pkg/config/auth.go
```diff
@@ -18,6 +18,8 @@ type (
 		SiteUrl                    string   `toml:"site_url"`
 		AdditionalRedirectUrls     []string `toml:"additional_redirect_urls"`
 		JwtExpiry                  uint     `toml:"jwt_expiry"`
+		MinimumPasswordLength      uint     `toml:"minimum_password_length"`
+		PasswordRequirements       string   `toml:"password_requirements"`
```
sweatybridge, line 22: Since we started supporting `config push` command, adding new auth config also requires mapping to the api request / response fields https://github.com/supabase/cli/blob/develop/pkg/config/auth.go#L186.

Once that's mapped, remember to update the diff snapshots in `pkg/config/testdata` to resolve the failing tests.

silentworks: @sweatybridge I think this is going to be an issue for me as I don't know what these keys are on the hosted platform.

sweatybridge: No worries, I can take over from here next week.

Fyi, all fields are more or less documented in the [openapi spec](https://github.com/supabase/cli/blob/develop/pkg/api/types.gen.go#L465).

silentworks: I figured it out. The only part is the `testdata` I'm having trouble with now.

silentworks: @sweatybridge Are the diff files manually created or is there a command to run to generate them?

pkg/config/templates/config.toml
```diff
@@ -98,6 +98,13 @@ additional_redirect_urls = ["https://127.0.0.1:3000"]
 # How long tokens are valid for, in seconds. Defaults to 3600 (1 hour), maximum 604,800 (1 week).
 jwt_expiry = 3600
 # If disabled, the refresh token will never expire.
+minimum_password_length = 6
+# Passwords shorter than this value will be rejected as weak. Minimum 6, recommended 8 or more.
+password_requirements = ""
+# Passwords that do not have at least one of each will be rejected as weak.
+# Letters and digits: abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ:0123456789
+# Lowercase, uppercase letters and digits: abcdefghijklmnopqrstuvwxyz:ABCDEFGHIJKLMNOPQRSTUVWXYZ:0123456789
+# Lowercase, uppercase letters, digits and symbols: abcdefghijklmnopqrstuvwxyz:ABCDEFGHIJKLMNOPQRSTUVWXYZ:0123456789:!@#$%^&*()_+-=[]{};'\\\\:\"|<>?,./`~
```
sweatybridge, line 107: We should make this an enum and validate it, similar to frontend.

Screenshot 2024-11-18 at 3 11 03 PM

pkg/config/auth.go
```diff
@@ -192,6 +194,8 @@ func (a *auth) ToUpdateAuthConfigBody() v1API.UpdateAuthConfigBody {
 		SecurityManualLinkingEnabled:      &a.EnableManualLinking,
 		DisableSignup:                     cast.Ptr(!a.EnableSignup),
 		ExternalAnonymousUsersEnabled:     &a.EnableAnonymousSignIns,
+		PasswordMinLength:                 cast.UintToIntPtr(&a.MinimumPasswordLength),
+		PasswordRequiredCharacters:        (*v1API.UpdateAuthConfigBodyPasswordRequiredCharacters)(&a.PasswordRequirements),
```
sweatybridge, line 198: Also need to update `fromRemoteAuthConfig` for the reverse mapping.

pkg/config/auth.go
silentworks: Some tests don't work in WSL due to the reliance on keyring which is not available in WSL. This means I can't run the test suite from my end.

image

sweatybridge: To run a single test case:

```
go test -v ./pkg/config -run "TestHookDiff/local_and_remote_enabled"
```

pkg/config/templates/config.toml
sweatybridge: Sounds good to me

silentworks:
Add screen recording to show this working. 

Turn down your audio as I might be a bit loud in the recording.

https://github.com/user-attachments/assets/19184c99-6428-4976-a2e4-371a0ba00d3a

pkg/config/auth.go
silentworks: @sweatybridge I've tried rerunning the corresponding test but it isn't generating any new diff file.

pkg/config/auth.go
silentworks: @sweatybridge how can I find out what value `fromRemoteAuthConfig` will return? because I then need to convert these from `regex` or a `string` to match the enum map I'm using in the CLI.

pkg/config/auth.go
sweatybridge: And you've deleted the diff file before running the single test? Could be a windows issue then 😅

pkg/config/auth.go
sweatybridge: It's ok to cast the response field to enum.

pkg/config/auth.go
silentworks: OK got this working now. I used the run test functionality of VSCode instead because the go test command didn't seem like it was creating new diff files.

pkg/config/auth.go
silentworks: @sweatybridge How can I test the fromRemoteAuthConfig functionality?

sweatybridge (approved):

pkg/config/auth.go
sweatybridge: I've added the tests as an example.

PR #2895, opened 2024-11-16
feat(link): diff more services config

avallete:

- Re-use updater to diff remote and local config in link command
- split updater between two files `remote_updater` (local->remote) and `local_updater`(remote -> local)


avallete:
Closing in favor of: https://github.com/supabase/cli/pull/2894

PR #2912, opened 2024-11-25
feat: allow setting compute multiplier on functions

laktek:

Introduce compute multiplier setting on Edge Functions.

### DIFF
--- a/internal/functions/serve/templates/main.ts
+++ b/internal/functions/serve/templates/main.ts
@@ -156,7 +156,8 @@ Deno.serve({
     console.error(`serving the request with ${servicePath}`);
 
     // Ref: https://supabase.com/docs/guides/functions/limits
-    const memoryLimitMb = 256;
+    const computeMultiplier = Math.max(Math.min(parseFloat(functionsConfig[functionName].computeMultiplier ?? '1'), 4), 1)
+    const memoryLimitMb = 256 * computeMultiplier;
     const workerTimeoutMs = isFinite(WALLCLOCK_LIMIT_SEC) ? WALLCLOCK_LIMIT_SEC * 1000 : 400 * 1000;
     const noModuleCache = false;
     const envVarsObj = Deno.env.toObject();
@@ -168,7 +169,7 @@ Deno.serve({
     const forceCreate = false;
     const customModuleRoot = ""; // empty string to allow any local path
     const cpuTimeSoftLimitMs = 1000;
-    const cpuTimeHardLimitMs = 2000;
+    const cpuTimeHardLimitMs = 2000 * computeMultiplier;
 
     // NOTE(Nyannyacha): Decorator type has been set to tc39 by Lakshan's request,
     // but in my opinion, we should probably expose this to customers at some
@@ -197,7 +198,8 @@ Deno.serve({
         },
       });
 
-      return await worker.fetch(req);
+      const res = await worker.fetch(req);
+      res.headers.set('x-sb-compute-multiplier', computeMultiplier)
     } catch (e) {
       console.error(e);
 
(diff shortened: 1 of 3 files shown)


internal/functions/serve/templates/main.ts
```diff
@@ -156,7 +156,8 @@ Deno.serve({
     console.error(`serving the request with ${servicePath}`);
 
     // Ref: https://supabase.com/docs/guides/functions/limits
-    const memoryLimitMb = 256;
+    const computeMultiplier = Math.max(Math.min(parseFloat(functionsConfig[functionName].computeMultiplier ?? '1'), 4), 1)
```
sweatybridge, line 159:
```suggestion
    const computeMultiplier = Math.max(Math.min(functionsConfig[functionName].computeMultiplier ?? 1, 4), 1)
```

No need to parse from string because serialised json already sets this field as number.

Should also update the type definition for `functionConfig` https://github.com/supabase/cli/pull/2912/files#diff-44113b100b5a3840b42b5d08768b70a8f82055f9a7d11b37f31e2ea72d51e50dR52

```
computeMultiplier?: number
```

Optionally, consider emitting a log message when this value is clamped so users are aware of this limit when serving locally, ie. 1 <= x <=4. Otherwise the only way to know is via the api error when deploying.

pkg/function/batch.go
```diff
@@ -52,21 +53,23 @@ func (s *EdgeRuntimeAPI) UpsertFunctions(ctx context.Context, functionConfig con
 		upsert := func() error {
 			if _, ok := exists[slug]; ok {
 				if resp, err := s.client.V1UpdateAFunctionWithBodyWithResponse(ctx, s.project, slug, &api.V1UpdateAFunctionParams{
-					VerifyJwt:      function.VerifyJWT,
-					ImportMapPath:  toFileURL(function.ImportMap),
-					EntrypointPath: toFileURL(function.Entrypoint),
+					VerifyJwt:         function.VerifyJWT,
+					ImportMapPath:     toFileURL(function.ImportMap),
+					EntrypointPath:    toFileURL(function.Entrypoint),
+					ComputeMultiplier: computeMultiplier,
```
sweatybridge, line 59:
```suggestion
					ComputeMultiplier: function.ComputeMultiplier,
```

pkg/function/batch.go
```diff
@@ -52,21 +53,23 @@ func (s *EdgeRuntimeAPI) UpsertFunctions(ctx context.Context, functionConfig con
...
-					Name:           &slug,
-					VerifyJwt:      function.VerifyJWT,
-					ImportMapPath:  toFileURL(function.ImportMap),
-					EntrypointPath: toFileURL(function.Entrypoint),
+					Slug:              &slug,
+					Name:              &slug,
+					VerifyJwt:         function.VerifyJWT,
+					ImportMapPath:     toFileURL(function.ImportMap),
+					EntrypointPath:    toFileURL(function.Entrypoint),
+					ComputeMultiplier: computeMultiplier,
```
sweatybridge, line 72:
```suggestion
					ComputeMultiplier: function.ComputeMultiplier,
```

pkg/function/batch.go
```diff
@@ -44,6 +44,7 @@ func (s *EdgeRuntimeAPI) UpsertFunctions(ctx context.Context, functionConfig con
 				continue
 			}
 		}
+		computeMultiplier := function.ComputeMultiplier
```
sweatybridge, line 47:
```suggestion
```

nitpick for consistency with how other configs are accessed

laktek:
I'm closing this for now until we figure out some internal decisions.

PR #2931, opened 2024-11-29
fix(config): syncing for postgres hook without secrets

avallete:

- Postgres custom auth hook doesn't have any secret value, this is only required for http endpoint. We turn the secret into an optional value in the config structure so it can be omitted in such case rather than being filled with an empty string.

## What is the new behavior?

## Additional context

### DIFF
--- a/pkg/config/auth.go
+++ b/pkg/config/auth.go
@@ -171,9 +171,9 @@ type (
 	}
 
 	hookConfig struct {
-		Enabled bool    `toml:"enabled"`
-		URI     string  `toml:"uri"`
-		Secrets *string `toml:"secrets"`
+		Enabled bool   `toml:"enabled"`
+		URI     string `toml:"uri"`
+		Secrets string `toml:"secrets"`
 	}
 
 	sessions struct {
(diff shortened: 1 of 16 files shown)


pkg/config/testdata/TestEmailDiff/local_disabled_remote_enabled.diff
```diff
@@ -22,7 +22,7 @@ diff remote[auth] local[auth]
  [email.template]
  [email.template.confirmation]
  content_path = ""
-@@ -71,13 +71,6 @@
+@@ -66,13 +66,6 @@
```
avallete, line 25: **remark**

I'm not totally sure why those snapshot have differences.

pkg/config/config.go
```diff
@@ -997,10 +997,16 @@ func (h *hookConfig) validate(hookType string) (err error) {
 	} else if parsed, err := url.Parse(h.URI); err != nil {
 		return errors.Errorf("failed to parse template url: %w", err)
 	} else if !(parsed.Scheme == "http" || parsed.Scheme == "https" || parsed.Scheme == "pg-functions") {
-		return errors.Errorf("Invalid HTTP hook config: auth.hook.%v should be a Postgres function URI, or a HTTP or HTTPS URL", hookType)
+		return errors.Errorf("Invalid auth hook config: auth.hook.%v should be a Postgres function URI, or a HTTP or HTTPS URL", hookType)
+	} else if (parsed.Scheme == "http" || parsed.Scheme == "https") && h.Secrets == nil {
+		return errors.Errorf("Invalid HTTP config: auth.hook.%v missing required secret value for the http hook endpoint.", hookType)
```
avallete, line 1002: **remark**

Splitted up the URI and Secret checking in two different error message since secrets is only required for http endpoints.

sweatybridge:
Could you elaborate again why empty string was causing an issue?

pkg/config/auth_test.go
```diff
@@ -121,29 +136,44 @@ func TestHookDiff(t *testing.T) {
...
-			HookPasswordVerificationAttemptUri:     cast.Ptr(""),
-			HookPasswordVerificationAttemptSecrets: cast.Ptr("b613679a0814d9ec772f95d778c35fc5ff1697c493715653c6c712144292c5ad"),
+			HookPasswordVerificationAttemptUri:     cast.Ptr("pg-functions://functionName"),
+			HookPasswordVerificationAttemptSecrets: nil,
+		})
+		// Check error
+		assert.NoError(t, err)
+		assert.Empty(t, string(diff))
+	})
+	t.Run("local and remote enabled pg-function custom hook", func(t *testing.T) {
```
sweatybridge, line 167: no need a new test for this. let's merge it to `local and remote enabled` for simplicity

pkg/config/auth_test.go
```diff
@@ -121,29 +136,44 @@ func TestHookDiff(t *testing.T) {
...
-			CustomAccessToken:           hookConfig{Enabled: true},
-			SendSMS:                     hookConfig{Enabled: true},
-			SendEmail:                   hookConfig{Enabled: true},
-			MFAVerificationAttempt:      hookConfig{Enabled: true},
-			PasswordVerificationAttempt: hookConfig{Enabled: true},
+			CustomAccessToken:           defaultHttpHookConfig(),
+			SendSMS:                     defaultHttpHookConfig(),
+			SendEmail:                   defaultHttpHookConfig(),
+			MFAVerificationAttempt:      defaultHttpHookConfig(),
+			PasswordVerificationAttempt: defaultPostgresHookConfig(),
```
sweatybridge, line 143: Let's not refactor the test helpers here. explicitly defining each hook is more easier to read. Also means we can set pg-function and http hook in one test

pkg/config/config.go
```diff
@@ -997,10 +997,16 @@ func (h *hookConfig) validate(hookType string) (err error) {
 	} else if parsed, err := url.Parse(h.URI); err != nil {
 		return errors.Errorf("failed to parse template url: %w", err)
 	} else if !(parsed.Scheme == "http" || parsed.Scheme == "https" || parsed.Scheme == "pg-functions") {
-		return errors.Errorf("Invalid HTTP hook config: auth.hook.%v should be a Postgres function URI, or a HTTP or HTTPS URL", hookType)
+		return errors.Errorf("Invalid auth hook config: auth.hook.%v should be a Postgres function URI, or a HTTP or HTTPS URL", hookType)
+	} else if (parsed.Scheme == "http" || parsed.Scheme == "https") && h.Secrets == nil {
```
sweatybridge, line 1001: I think it's cleaner to use a switch case here

```
switch parsed.Scheme {
...
}
```

pkg/config/testdata/TestEmailDiff/local_disabled_remote_enabled.diff
sweatybridge: coz we added uri and secrets fields in tests, so there are more lines in the config dump.

pkg/config/auth.go
```diff
@@ -851,20 +864,20 @@ func (a *auth) HashSecrets(key string) {
...
 	}
-	if a.Hook.SendSMS.Enabled {
-		a.Hook.SendSMS.Secrets = hash(a.Hook.SendSMS.Secrets)
+	if a.Hook.SendSMS.Enabled && a.Hook.SendSMS.Secrets != nil {
+		a.Hook.SendSMS.Secrets = cast.Ptr(hash(*a.Hook.SendSMS.Secrets))
 	}
-	if a.Hook.SendEmail.Enabled {
-		a.Hook.SendEmail.Secrets = hash(a.Hook.SendEmail.Secrets)
+	if a.Hook.SendEmail.Enabled && a.Hook.SendEmail.Secrets != nil {
+		a.Hook.SendEmail.Secrets = cast.Ptr(hash(*a.Hook.SendEmail.Secrets))
```
sweatybridge, line 880: We should handle all nullable secret fields the same way for consistency. We currently check `len(smtp.pass) == 0` so probably continue using that instead of changing to pointer https://github.com/supabase/cli/pull/2931/files#diff-ceb99eefb6843a3aa3db9de47b504518a30b320f117166511f0fa05156d2b767R851

sweatybridge (commented):
Sorry for nitpicking. We want to be extremely strict about the coding standards for these configs. The impact of large frequent changes is very bad on cognitive load as well as maintainability, especially since we've spent a lot of time crafting it carefully.

avallete:
> Sorry for nitpicking. We want to be extremely strict about the coding standards for these configs. The impact of large frequent changes is very bad on cognitive load as well as maintainability, especially since we've spent a lot of time crafting it carefully.

No worries thank's for the thorough review, I've changed the code accordingly to your comments.

sweatybridge (approved):

PR #2932, opened 2024-11-29
feat(updater): allow to skip secrets sync for branching

avallete:

On branching side we'll be able to do:

```
updater.skipSecrets = true
```

## What is the new behavior?

## Additional context

### DIFF
--- a/pkg/config/auth.go
+++ b/pkg/config/auth.go
@@ -217,6 +217,60 @@ type (
 	}
 )
 
+func (a *auth) StripUpdateBodySecrets(body v1API.UpdateAuthConfigBody) v1API.UpdateAuthConfigBody {
+	body.SmtpPass = nil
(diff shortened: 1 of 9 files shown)


pkg/config/auth.go
```diff
@@ -217,6 +217,60 @@ type (
 	}
 )
 
+func (a *auth) StripUpdateBodySecrets(body v1API.UpdateAuthConfigBody) v1API.UpdateAuthConfigBody {
+	body.SmtpPass = nil
```
sweatybridge, line 221: Unfortunately most of these secrets are required fields that we validate on api side when something like smtp is enabled. Setting them to nil would essentially block everyone from updating the whole config block.

An alternative is to let users opt out of those configs by turning the struct into pointers. Similar to https://github.com/supabase/cli/pull/2954

PR #2937, opened 2024-12-02
fix(studio): Bump studio version to 20241202-71e5240

ivasilov:


sweatybridge (approved):
I will release this together with cli v2.

PR #2945, opened 2024-12-04
Add `supabase studio` command

jsj:
Feature - Adds a new CLI command `supabase studio` to open Supabase Studio directly from the terminal

Currently, users need to manually open their browser and navigate to the Supabase Studio dashboard.

## What is the new behavior?
This PR adds a new `studio` command that automatically opens the Supabase Studio dashboard in the user's default browser. The command:
- Works cross-platform (macOS, Windows, and Linux)
- Uses the appropriate system command to open the browser (`open` for macOS, `start` for Windows, `xdg-open` for Linux)
- Provides feedback to the user by displaying the URL being opened
- Handles errors gracefully with descriptive error messages

Example usage:
```bash
supabase studio

### DIFF
--- a/cmd/studio.go
+++ b/cmd/studio.go
@@ -4,23 +4,41 @@ import (
 	"fmt"
 	"os/exec"
 	"runtime"
+	"os"
 
 	"github.com/spf13/cobra"
+	"github.com/spf13/afero"
 	"github.com/supabase/cli/internal/utils"
+	"github.com/supabase/cli/internal/utils/flags"
+	"github.com/supabase/cli/internal/login"
 )
 
 var studioCmd = &cobra.Command{
 	Use:   "studio",
 	Short: "Opens Supabase Studio in your browser",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		url := utils.GetSupabaseDashboardURL()
+
+		fsys := afero.NewOsFs()
+		if err := flags.LoadConfig(fsys); err != nil {
+			return err
+		}
+		
+		var url string
+		if utils.Config.Studio.Enabled {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: cmd/studio.go [hunk 1]
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/spf13/cobra"
+	"github.com/supabase/cli/internal/utils"
+)
+
+var studioCmd = &cobra.Command{
+	Use:   "studio",
+	Short: "Opens Supabase Studio in your browser",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		url := utils.GetSupabaseDashboardURL()
```
sweatybridge, line 16: This opens the hosted dashboard and not local studio. Is that your intention?

It might be a good idea to support `--local` and `--linked` flag to distinguish these use cases.

cmd/studio.go
```diff
@@ -0,0 +1,39 @@
...
+		var err error
+
+		switch runtime.GOOS {
```
sweatybridge, line 19: Let's use the `RunOpenCmd` function to handle this more reliably on Linux and Windows.

- https://github.com/supabase/cli/blob/develop/internal/login/login_linux.go
- https://github.com/supabase/cli/blob/develop/internal/login/login_windows.go

sweatybridge (changes requested):

cmd/studio.go
jsj: Yeah this was wrong the goal is to open the local studio url, this comes from using Windsurf, Cursor it will hallucinate this command to go adjust settings only available in the Studio

PR #2949, opened 2024-12-06
fix: skip pushing site url config for localhost

sweatybridge:

relates to https://github.com/supabase/cli/issues/2552#issuecomment-2515058735

## What is the new behavior?

Skip site_url if it's pointing to localhost.

## Additional context

### DIFF
--- a/pkg/config/testdata/TestAuthDiff/local_enabled_and_disabled.diff
+++ b/pkg/config/testdata/TestAuthDiff/local_enabled_and_disabled.diff
@@ -3,7 +3,7 @@ diff remote[auth] local[auth]
 +++ local[auth]
 @@ -1,14 +1,14 @@
  enabled = false
--site_url = ""
+ site_url = "http://127.0.0.1:3000"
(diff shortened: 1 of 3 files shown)


pkg/config/testdata/TestAuthDiff/local_enabled_and_disabled.diff
```diff
@@ -3,7 +3,7 @@ diff remote[auth] local[auth]
 +++ local[auth]
 @@ -1,14 +1,14 @@
  enabled = false
--site_url = ""
+ site_url = "http://127.0.0.1:3000"
 -additional_redirect_urls = ["https://127.0.0.1:3000", "https://ref.supabase.co"]
```
avallete, line 7: **question**
Shouldn't we still have the `https://ref.supabase.co` in the diff ?

sweatybridge: I updated the default additional redirect to empty array, which is why its diff is changed.

sweatybridge:
Closing as probably not needed after we start pushing down envs from vercel.

PR #2966, opened 2024-12-11
fix: disable image transformation by default

sweatybridge:

## What is the new behavior?

## Additional context


avallete (approved):

avallete:
Sorry didn't noticed. But shouldn't we also update the default value here @sweatybridge : https://github.com/supabase/cli/blob/fix/update-free-tier-projects/pkg/config/config.go#L321-L324

sweatybridge:
@avallete good catch. I've removed the default value.

Issue #2983, opened 2024-12-19
Supabase functions prioritize 'import_map.json' over 'deno.json' for import maps.


avallete:
I'm noting here that the Supabase functions appear to search for "import_map.json" within the "functions" directory. The newly recommended method of defining import maps in the "deno.json" file does not seem to be working.

_Originally posted by @AndryHTC in https://github.com/supabase/cli/issues/1338#issuecomment-2288580336_

Issue #2979, opened 2024-12-17
`supabase functions serve` and private packages


elyobo:
**Is your feature request related to a problem? Please describe.**

Support for `.npmrc` authentication for private packages was added in https://github.com/supabase/edge-runtime/pull/427 but in order for private packages to work a `.npmrc` needs to be placed in the directory for [each and every function](https://github.com/supabase/edge-runtime/pull/427#issuecomment-2547398875).

**Describe the solution you'd like**

For local development, it would be great to respect the general `.npmrc` rules (find the cloest up the directory structure).
For deployment to production, a `.npmrc` in `supabase/functions` that's used for all would probably make sense (and this would fit with the suggestion above).

**Describe alternatives you've considered**

Adding `.npmrc` to the `.gitignore` and copying them to each directory as required is viable, but the requirement should be spelled out in docs I guess (although I don't know where I would have expected to find them).

**Additional context**
N/A

JohnBra:
@elyobo I looked into this and made it work locally with the current cli version as follows:
1. create a new edge function (i just created the hello-world example)
2. delete `.npmrc` from the newly created function directory
3. create `.npmrc` up the directory tree, but not higher than `supabase/functions`
4. add a private package ref like so:
```ini
#.npmrc
@johnbra:registry="https://npm.pkg.github.com"
//npm.pkg.github.com/:_authToken=<TOKEN>
registry=https://registry.npmjs.org/
```
5. reference your package in your function like so:
```typescript
import { appendExclamationMark } from "npm:@johnbra/node-package@0.1.0";
// ...
```

Currently, the local directory `supabase/functions` is bound to the container, i.e. everything above isn't. This means the `supabase` root directory including `config.toml` etc. aren't visible to the `edge-runtime`. As long as you place your `.npmrc` in the `supabase/functions` directory or in children of it, it should work.

I'm not sure if I'm missing something, but I think this should solve your problem since the `supabase/functions` dir should be the root for all functions _(please correct me if I'm wrong)_ and @nyannyacha's implementation of [find_up](https://github.com/nyannyacha/edge-runtime/blob/fb36a4ab72939cbbf748f65f41e5922800ed801c/crates/base/src/utils/path.rs#L3) should [look for the closest .npmrc up the tree](https://github.com/nyannyacha/edge-runtime/blob/fb36a4ab72939cbbf748f65f41e5922800ed801c/crates/base/src/deno_runtime.rs#L337).

I haven't tested if this works with the `deploy` command.

[...]

PR #2984, opened 2024-12-19
fix(edge-functions): add warning for edge-functions configuration

avallete:

Closes: https://github.com/supabase/cli/issues/2983

## What is the new behavior?

Add a warning preventing about deprecated global deps use, and `import_map` usage:
Here 3 functions, `hello-world`, `hello-word2`, `hello-word3` only one of them declare a valid `deno.jsonc` deps, here the result:

Screenshot 2025-01-06 at 17 30 05


avallete:
This would fix the issue, but I wonder if we should rather move toward a more "isolation" approach with something like this in the docs for the edge functions:

```
**Note:** Each Edge Function should be treated as an independent project with its own `deno.json` or `import_map.json` file. Avoid placing shared configurations at the root of the `functions` directory, as this can lead to conflicts. This ensures functions remain self-contained and portable.
```

Could also mention the `.npmrc` as mentioned here: https://github.com/supabase/cli/issues/2979

I'm happy to close the PR and go toward this approach if we choose it instead.

jgoux (dismissed):

sweatybridge:
Two suggestions. 

1. We should handle both .json and .jsonc like in per function config parsing.
2. I agree we want to encourage isolation for deployed functions. The default import map was only added for convenience for local dev and serving.

I'd prefer updating the docs to mention the suggested path for deno.json for deployment so it's less confusing for new users.

avallete:
> Two suggestions.
> 
> 1. We should handle both .json and .jsonc like in per function config parsing.
> 2. I agree we want to encourage isolation for deployed functions. The default import map was only added for convenience for local dev and serving.
> 
> I'd prefer updating the docs to mention the suggested path for deno.json for deployment so it's less confusing for new users.

Good spot for the `jsonc`.

I think the issue with the "convenience shortcut" is that it create the expectation that things will work the same way between all the ways you can define your deps. And even if we mention things in the docs, it's not the best for discovery.

Here is what I would like to propose:

1. Support the "global deno.json/deno.jsonc" for consistency with the `import_map.json` shortcut behavior.
2. Mention in the docs that each function should be isolated, etc...
3. Add a warning for every function that use the fallback on the cli side. So it stand out for the user that this is not a suitable ways to do things and should only be used for development, redirect to the docs (prepare to remove this completely for the next major as a breaking change thing ? We could provide a shellscript'ish way to easily do the migration by copying the global into the subfolders)
4. Maybe a good way to nudge the users in the right direction is that when we do `functions new`, on top of the `index.ts` also create a `deno.json` and `.npmrc` files in the created folders ?

WDYT ? @sweatybridge @jgoux

internal/functions/deploy/deploy.go
```diff
@@ -92,20 +84,37 @@ func GetFunctionConfig(slugs []string, importMapPath string, noVerifyJWT *bool,
...
 			}
 		}
 		if noVerifyJWT != nil {
 			function.VerifyJWT = cast.Ptr(!*noVerifyJWT)
 		}
 		functionConfig[name] = function
 	}
+	// Check validity of ImportMap paths
+	functionsWithFallback := []string{}
+	if fallbacksPath, err := utils.GetImportsFilePath(utils.FunctionsDir, fsys); err == nil {
```
sweatybridge, line 100: If we really want to support fallback import map for deno.json, then this should be moved to pkg/config.

So each function config's import_map field will point to the fully resolved path after considering all possible fallbacks.

avallete: **question**

It seems like we didn't supported this fallback before ? I don't know how that was handled with branching 🤔 

I have two question about this:

1. Won't it have side effects with the diffing of the config in such case ?
2. I think we can't import utils within pkg/config without causing a cycling deps issue. What would be the best approach here ?

jgoux:
+1 with @sweatybridge on only updating the docs and explaining that one `deno.json(c)` per function is the best practice, and maybe add a deprecated mention on the import_map.json file in `functions`.

avallete (commented):
Changed the PR to remove the "global fallback handling" and keep focus over:

1. Providing warning and feedbacks if the function isn't configured the recommended way.
2. Add empty `deno.jsonc` and `.npmrc` template when running `function new` to setup user for success.

internal/functions/new/new.go
```diff
@@ -38,25 +43,41 @@ func Run(ctx context.Context, slug string, fsys afero.Fs) error {
...
+		if err := afero.WriteFile(fsys, filepath.Join(funcDir, ".npmrc"), []byte(npmrcEmbed), 0644); err != nil {
+			return errors.Errorf("failed to create .npmrc config: %w", err)
 		}
 	}
 
 	fmt.Println("Created new Function at " + utils.Bold(funcDir))
 	return nil
 }
+
+func createTemplateFile(fsys afero.Fs, path string, tmpl *template.Template, data interface{}) error {
```
sweatybridge, line 72: Where are we calling this method with `data == nil`?

internal/functions/deploy/deploy.go
```diff
@@ -94,18 +96,30 @@ func GetFunctionConfig(slugs []string, importMapPath string, noVerifyJWT *bool,
...
 			denoJsonPath := filepath.Join(functionDir, "deno.json")
 			denoJsoncPath := filepath.Join(functionDir, "deno.jsonc")
+			importMapPath := filepath.Join(functionDir, "import_map.json")
 			if _, err := fsys.Stat(denoJsonPath); err == nil {
 				function.ImportMap = denoJsonPath
 			} else if _, err := fsys.Stat(denoJsoncPath); err == nil {
 				function.ImportMap = denoJsoncPath
+			} else if _, err := fsys.Stat(importMapPath); err == nil {
+				function.ImportMap = importMapPath
+				functionsUsingDeprecatedImportMap = append(functionsUsingDeprecatedImportMap, name)
```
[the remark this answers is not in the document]
sweatybridge: It was actually intentional because I didn't want to support `functions/import_map.json` for those using branching. When deploying from cli, we kept [this code](https://github.com/supabase/cli/blob/develop/internal/functions/deploy/deploy.go#L74) for backwards compatibility and local development convenience.

Perhaps we can move the `functionsUsingDeprecatedImportMap` check to only cli as well.

internal/functions/new/new.go
avallete: We don't good catch.

internal/functions/deploy/deploy.go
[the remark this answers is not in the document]
avallete: Sorry I think I don't get it.

The fallback allow to have a single `functions/import_map.json` and will use it for deploy in both the cli and the branching to deploy the function (or I'm I wrong) ?

So what we don't want is to have a "local" override for a function such as `functions/test/import_map.json` to not be taken into account for branching ? But isn't that inconsistent if `deno.json(c)` is used ?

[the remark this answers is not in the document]
sweatybridge: Ohh you are right. I thought you were updating the code in pkg/config. But it's all good now that I looked at it again.

I agree that `functions/slug/import_map.json` should take precedence over `functions/import_map.json`. Good catch!

sweatybridge (approved):

PR #2989, opened 2024-12-19
feat(cli): add shadow database host configuration

nzlz:
- Add ShadowHost config option to control shadow database connection
- Default to Hostname if ShadowHost not set
---
The scenario is the following:
- local supabase deployment, self hosted.
- running migrations inside our own container which has supa running via docker compose.
- attempt supabase `db diff --db-url` ...

This does not work since the docker running the shadow database is created on host, and 127.0.0.1 for host network is not accessible from inside our container.

In my case, i set the following in the config.toml >> `shadow_host = "172.18.0.1"` and now db diff works.

(open to alternative solutions!)


sweatybridge:
> This does not work since the docker running the shadow database is created on host, and 127.0.0.1 for host network is not accessible from inside our container.

The shadow db container is always created using docker [host network](https://docs.docker.com/engine/network/drivers/host/) mode, so 127.0.0.1 should resolve to your host network from inside the container. I suspect the root cause is something else.

Could you try the following

```bash
DOCKER_HOST=172.18.0.1 supabase db diff --db-url ...
```

PR #2995, opened 2024-12-24
feat: support passing `--env-file` flag to functions deploy command.

laktek:

Add the option to pass an env file via the `--env-flag` option to the `functions deploy` command. It will fallback to use the default env file set in `supabase/functions/.env` if no explicit flag is provided.

Currently, the only useful env variable is `NPM_CONFIG_REGISTRY`, which can override the default NPM registry URL used for downloading dependencies. There may be other configuration options via env vars in future.


laktek:
Superseded by #3020

PR #2996, opened 2024-12-25
fix: obtain Postgres URL from branches get command

### LINKED ISSUE #2964: Obtaining DB_URL from the CLI
I am attempting to set up branch previews (similar to Vercel) in [railway](railway.app). To do this, I need to update some environment variables on the railway side, including SUPABASE_DB_URL. 

I'd like to be able to run some supabase CLI command to obtain the DB_URl automatically, so I don't have to add it manually as a variable whenever I make a new branch. However, I cannot find a CLI command that will do this. `gen keys` gives me the password but not the user, the port, the database, or the host. `branches get` has the user, port, and host, but these conflict with the information shown in my dashboard. (I assume the database is always `postgres`?)

I've looked in the docs for the CLI and the javascript library, and neither of them seems to provide the database url. Any help is appreciated!

yaten2302:

This PR introduces a new feature to enable users to obtain the Postgres db url from the .env using the `branches get` command.

Fixes #2964 

## What is the new behavior?

This PR adds a new flag to the `branches get` command to help users to get the postgres db url from the env.

## Additional context


yaten2302:
@sweatybridge , I've a doubt, that as you mentioned in [this](https://github.com/supabase/cli/issues/2964#issuecomment-2561252928) comment, that a new `--output` flag should be added to the `branches get` command. But, while implementing that, it's showing an error that the `output` flag is being redeclared. So, is it possible that instead of using the output flag, we can use a Boolean var `postgres-url`?

Also, as you mentioned, that the postgres_url and prostgres_url_non_pooling has to be added if this flag is mentioned with the command and the code can be reused from the `bootstrap` command. But, while I was checking the `writeDotEnv` func, it writes the complete thing to the .env file. But, I'm confused that how can I get the postgres_db_url from the `writeDotEnv` func? Because, I was having some issues in adding the postgres_db_url to the [`resp`](https://github.com/supabase/cli/blob/develop/internal/branches/get/get.go#L13), could you guide that how can I add the same to the resp, so that it can be used to display it on the output table for the user.

yaten2302:
Hey @sweatybridge , could you please have a look at this PR, I've made the required changes, kindly review and LMK if any more are required!!
Thanks

sweatybridge:
Let's do this in smaller steps so I can provide more detailed guidance.

> But, while implementing that, it's showing an error that the output flag is being redeclared.

This is how golang works: any variable declared in the same package is shared by all code in that package. Since the output variable is already declared in `cmd` package, you don't need to redeclare for this command.

For your first commit, simply passing it down to branches get command will do. No need to implement the actual behaviour for env output.

After you've done that, tag me again so I can check before providing more details for the actual implementation.

sweatybridge (changes requested):
Feel free to open a new PR if it's easier for you.

sweatybridge:
Also, sorry for the late reply as I was on holiday for the last 2 weeks and there were a lot to catch up on. I will review more promptly this week and next if you plan to make the changes.

yaten2302:
Heyy @sweatybridge , sorry, actually, I was pushing some changes right now only and that's why didn't see your comment on this PR.

> Let's do this in smaller steps so I can provide more detailed guidance.
> 
> > But, while implementing that, it's showing an error that the output flag is being redeclared.
> 
> This is how golang works: any variable declared in the same package is shared by all code in that package. Since the output variable is already declared in `cmd` package, you don't need to redeclare for this command.
> 
> For your first commit, simply passing it down to branches get command will do. No need to implement the actual behaviour for env output.
> 
> After you've done that, tag me again so I can check before providing more details for the actual implementation.

Also, yes, for this, I was also researching about the same and now I understood that, vars are shared across the same package files in Go.
I've made some changes now, I think that these might fix this `--output-env` flag. Could you have a look at it, whenever you're free!!
Thanks :)

cmd/branches.go
```diff
@@ -165,6 +171,8 @@ func init() {
 	createFlags.Var(&branchRegion, "region", "Select a region to deploy the branch database.")
 	createFlags.Var(&size, "size", "Select a desired instance size for the branch database.")
 	createFlags.BoolVar(&persistent, "persistent", false, "Whether to create a persistent branch.")
+	getFlags := branchGetCmd.Flags()
+	getFlags.BoolVar(&postgres_url, "output-env", false, "Output the Postgres URL in the table.")
```
sweatybridge, line 175: We want to use an enum flag named output, not a new flag. For eg. https://github.com/supabase/cli/blob/e303c411d0722056beab59edb0b6c1117a9d156c/cmd/status.go#L44

sweatybridge: @yaten2302 let me know if everything is clear. The rest of this PR depends on passing down this flag correctly so I won't review further until the first part is done.

yaten2302: @sweatybridge , I've pushed a commit, what I've understood is that, the `output` flag is already present and that has to be used and checked if `output == "env"`, then the postgres url should be displayed on the table, am I correct?

sweatybridge: Yes, but you want to check `output == utils.OutputEnv` since it's an enum. Let's make the flag change first and iterate from there.

yaten2302: Got it now, made the changes 👍

cmd/branches.go
```diff
@@ -75,14 +76,18 @@ var (
...
+			cmdFlags := cmd.Flags()
 			if len(args) == 0 {
 				if err := promptBranchId(ctx, flags.ProjectRef); err != nil {
 					return err
 				}
 			} else {
 				branchId = args[0]
+				if cmdFlags.Changed("output-env") {
+					output.Value = "env"
+				}
```
sweatybridge, line 88:
```suggestion
```

Not needed anymore

cmd/branches.go
```diff
@@ -75,14 +76,18 @@ var (
 		Args:  cobra.MaximumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := cmd.Context()
+			cmdFlags := cmd.Flags()
```
sweatybridge, line 79:
```suggestion
```

Ditto

cmd/branches.go
```diff
@@ -75,14 +76,18 @@ var (
...
 					return err
 				}
 			} else {
 				branchId = args[0]
+				if cmdFlags.Changed("output-env") {
+					output.Value = "env"
+				}
 			}
-			return get.Run(ctx, branchId)
+			return get.Run(ctx, branchId, output.Value, pgconn.Config{}, afero.NewOsFs())
```
sweatybridge, line 90:
```suggestion
			return get.Run(ctx, branchId, output.Value, afero.NewOsFs())
```

Also no need to pass in pgconn, as we will be constructing the connection config from branch details returned from api.

yaten2302: @sweatybridge , but if we don't pass pgconn here, then it will give an error because in get.Run(), 5 arguments are expected, so are we gonna change the Run() func in `internal/branches/get/get.go`?

sweatybridge: Yes, let's change the Run method to remove this arg since it's not needed.

yaten2302:
Now, how should the Run() func be changed in `internal/branches/get/get.go` in order to get the POSTGRES_URL_NON_POOLING from the `internal/bootstrap/bootstrap.go`?

cmd/branches.go
```diff
@@ -6,6 +6,7 @@ import (
 	"os"
 
 	"github.com/go-errors/errors"
+	"github.com/jackc/pgconn"
```
sweatybridge, line 9:
```suggestion
```

Removed unused import

internal/branches/get/get.go
```diff
@@ -29,10 +32,20 @@ func Run(ctx context.Context, branchId string) error {
 		resp.JSON200.JwtSecret = &masked
 	}
```
sweatybridge, line 34: Here you want to instantiate pgconn.Config with params from api response. For eg

```go
config := pgconn.Config{Host: utils.GetSupabaseDbHost(resp.JSON200.DbHost)}
```

Please fill in the other fields from `resp` as well.

internal/branches/get/get.go
```diff
@@ -29,10 +32,20 @@ func Run(ctx context.Context, branchId string) error {
...
+
+	table += "\n|-|-|-|-|-|-|-|"
+	if env == utils.OutputEnv {
+		table += "-|"
+	}
+
+	table += "\n"
+
+	row := fmt.Sprintf(
+		"|`%s`|`%d`|`%s`|`%s`|`%s`|`%s`|`%s`|",
```
sweatybridge, line 48: The output fields should not change based on output format.

Since the default output format is pretty print, let's keep the old format string and simply append the newly constructed url with header `POSTGRES URL`.

internal/branches/get/get.go
yaten2302: @sweatybridge , I've a doubt in here, that, if we pass the DbHost in `utils.GetSupabaseDbHost(resp...)`, then our result will be - `db.projectRef.supabase.co`, am I correct?
If yes, then will it not return the wrong URL, because we want [this non_pooling url](https://github.com/supabase/cli/blob/develop/internal/bootstrap/bootstrap.go#L207), to be returned?

yaten2302: Also, I've to pass this `config := pgconn.Config{Host: utils.GetSupabaseDbHost(resp.JSON200.DbHost)}`, in the `get.go`, similar to [this](https://github.com/supabase/cli/blob/develop/internal/branches/get/get.go#L21-L30), right?

sweatybridge: `db.*.supabase.co` is the non pooling url. The url for connection pooler must be fetched separately from api.

internal/branches/get/get.go
yaten2302: So, like, I've don't have to change [this](https://github.com/supabase/cli/blob/develop/internal/branches/get/get.go#L32-L43), but, simply append the `POSTGRES_URL` column if `env == utils.OutputEnv`, am I right?

internal/branches/get/get.go
sweatybridge: > Also, I've to pass this config := pgconn.Config{Host: utils.GetSupabaseDbHost(resp.JSON200.DbHost)}, in the get.go, similar to this, right?

There's no need to mask again if that's your question. As long as you instantiate at this specific line.

internal/branches/get/get.go
sweatybridge: Don't handle OutputEnv for now. It will be too involved. Just focus on making the pretty print work as expected.

After that I will show you the code to support all output formats, including env, json, etc.

internal/branches/get/get.go
yaten2302: Yes, got it now 👍

internal/branches/get/get.go
yaten2302: I haven't added this change in the latest commit, doing it now 👍

yaten2302: Previously, the pretty print was like [this](https://github.com/supabase/cli/blob/develop/internal/branches/get/get.go#L32-L43), so should I first revert back to the same as given here?

sweatybridge: > Previously, the pretty print was like this, so should I first revert back to the same as given here?

Yup, we are only appending a new column to the previous table.

yaten2302: I've reverted back to the previous table structure, now should I use the `env` var to check the condition and add new column accordingly?

internal/branches/get/get.go
sweatybridge: Please add in other fields like Port, Pass, and User.

internal/branches/get/get.go
sweatybridge: No, just add in the column unconditionally. Like this

```
	table := `|HOST|PORT|USER|PASSWORD|JWT SECRET|POSTGRES VERSION|STATUS|POSTGRES URL|
|-|-|-|-|-|-|-|-|
` + fmt.Sprintf(
```

internal/branches/get/get.go
yaten2302: For the other fields like port, password, user etc, can it be done by using `resp.JSON200.Port`, is this valid?

yaten2302: image

sweatybridge: Yes, you are on the right track. There are convenience methods in cast.go to help you convert from between different integer and pointer types safely.

Remember to format each field on a new line for readability.

yaten2302: image
Is this looking good?

sweatybridge: Let's move Host: to a new line as well.

I think casting to uint is unnecessary. Your previous solution `uint16(port)` should be fine.

The rest looks ok from screenshot.

yaten2302: Made the changes 👍 
image

yaten2302: May I push this change if it looks good?

internal/branches/get/get.go
yaten2302: But, if we add this column and the user hasn't specified the `-o` flag, then this will still be shown (but with nil value), will this not look somewhat odd?

sweatybridge: You are misunderstanding what the `-o` flag does. It should format all fields using one of the supported encoding, ie. json or .env. It is not meant to be appending or removing fields from the tabular output.

internal/branches/get/get.go
sweatybridge: Feel free to push it and I will take another look.

internal/branches/get/get.go
sweatybridge: For eg. this is how branches get command should eventually be implemented https://github.com/supabase/cli/blob/7a513e346ab55ad2e3a90a8675445121b66c3761/internal/projects/list/list.go#L44

yaten2302: Ohh, yes, okay, I think now I got it, as mentioned [here](https://supabase.com/docs/reference/cli/supabase-status), the `-o` flag gives the format in which the output should be shown to the user, am I right? And the table here mentioned is used only to render the output which has to be shown to the user. <br>
image

yaten2302: But, if this is the case, then how will we get to know that the user wants the DB_URL or not? Are we gonna add that irrespective of the user's preference or should a new flag be created to take into account that if the user wants the DB_URL, then it'll be shown otherwise it won't be?

sweatybridge: > Are we gonna add that irrespective of the user's preference

Yup we want to always show the db url. I think most users want the connection string for convenience. If not, there are other ways to filter the output like piping to grep or sed.

internal/branches/get/get.go
yaten2302: Pushed these changes 👍

internal/branches/get/get.go
yaten2302: Okay, got it now 👍 
So, in the `POSTGRES URL` column in the table, we've to write - https://supabase.com/docs/guides/database/connecting-to-postgres this connection string?<br>
image

I was also checking about this on stackoverflow, I found [this](https://stackoverflow.com/questions/78455486/whats-the-correct-postgres-url-when-hosting-supabase-with-coolify).

sweatybridge: Yes, and you can call ToPostgresURL to convert from pgconn.Config to the full connection string.

yaten2302: Hi, sorry for the delay, I've added the whole connection string, could you please have a look 👍

sweatybridge: Looks good! Once you resolve the lint errors, we can merge and work on part 2.

internal/bootstrap/bootstrap.go
```diff
@@ -272,6 +272,37 @@ func writeDotEnv(keys []api.ApiKeyResponse, config pgconn.Config, fsys afero.Fs)
...
+
+	// Check if POSTGRES_URL_NON_POOLING is present
+	if value, exists := initial[POSTGRES_URL_NON_POOLING]; exists {
+		return value
+	}
+
+	errors.New("error fetching POSTGRES_URL_NON_POOLING")
+	return ""
+}
+
```
sweatybridge, line 305: Let's remove these changes for part 1.

sweatybridge: I mean we don't need the entire `GetPostgresURLNonPooling` method.

In branches/get, we can serialise a map to env file like this.

```
	if utils.OutputFormat.Value != utils.OutputPretty {
		envs := map[string]string{
			"POSTGRES_URL": postgresConnectionString,
		}
		return utils.EncodeOutput(utils.OutputFormat.Value, os.Stdout, envs)
	}
```

internal/branches/get/get.go
```diff
@@ -5,11 +5,13 @@ import (
...
 
 	"github.com/go-errors/errors"
+	"github.com/jackc/pgconn"
+	"github.com/spf13/afero"
 	"github.com/supabase/cli/internal/migration/list"
 	"github.com/supabase/cli/internal/utils"
 )
 
-func Run(ctx context.Context, branchId string) error {
+func Run(ctx context.Context, branchId string, env string, fsys afero.Fs) error {
```
sweatybridge, line 14:
```suggestion
func Run(ctx context.Context, branchId string, fsys afero.Fs) error {
```

cmd/branches.go
```diff
@@ -82,7 +82,7 @@ var (
 			} else {
 				branchId = args[0]
 			}
-			return get.Run(ctx, branchId)
+			return get.Run(ctx, branchId, output.Value, afero.NewOsFs())
```
sweatybridge, line 85:
```suggestion
			return get.Run(ctx, branchId, afero.NewOsFs())
```

internal/branches/get/get.go
```diff
@@ -29,17 +31,27 @@ func Run(ctx context.Context, branchId string) error {
...
-	table := `|HOST|PORT|USER|PASSWORD|JWT SECRET|POSTGRES VERSION|STATUS|
-|-|-|-|-|-|-|-|
+	config := pgconn.Config{
+		Host: utils.GetSupabaseDbHost(resp.JSON200.DbHost),
+		Port: uint16(resp.JSON200.DbPort),
+		User: *resp.JSON200.DbUser,
+		Password: *resp.JSON200.DbPass,
+	}
+
+	postgresConnectionString := utils.ToPostgresURL(config)
```
sweatybridge, line 41:
```suggestion
	postgresConnectionString := utils.ToPostgresURL(config)

	if utils.OutputFormat.Value != utils.OutputPretty {
		envs := map[string]string{
			"POSTGRES_URL": postgresConnectionString,
		}
		return utils.EncodeOutput(utils.OutputFormat.Value, os.Stdout, envs)
	}

```

yaten2302: Already made this change 👍

cmd/branches.go
```diff
@@ -165,6 +165,8 @@ func init() {
 	createFlags.Var(&branchRegion, "region", "Select a region to deploy the branch database.")
 	createFlags.Var(&size, "size", "Select a desired instance size for the branch database.")
 	createFlags.BoolVar(&persistent, "persistent", false, "Whether to create a persistent branch.")
+	getFlags := branchGetCmd.Flags()
+	getFlags.VarP(&output, "output", "o", "Output format of branch details.")
```
sweatybridge, line 169:
```suggestion
	getFlags.VarP(&utils.OutputFormat, "output", "o", "Output format of branch details.")
```

internal/branches/get/get.go
```diff
@@ -3,14 +3,18 @@ package get
...
+	"github.com/jackc/pgconn"
+	"github.com/spf13/afero"
 	"github.com/supabase/cli/internal/migration/list"
 	"github.com/supabase/cli/internal/utils"
 )
 
-func Run(ctx context.Context, branchId string) error {
+func Run(ctx context.Context, branchId string, fsys afero.Fs) error {
 	resp, err := utils.GetSupabase().V1GetABranchConfigWithResponse(ctx, branchId)
+	envs := map[string]string{}
```
sweatybridge, line 17:
```suggestion
```

internal/branches/get/get.go
```diff
@@ -29,17 +32,34 @@ func Run(ctx context.Context, branchId string) error {
 		resp.JSON200.JwtSecret = &masked
 	}
 
-	table := `|HOST|PORT|USER|PASSWORD|JWT SECRET|POSTGRES VERSION|STATUS|
-|-|-|-|-|-|-|-|
+	config := pgconn.Config{
+		Host: utils.GetSupabaseDbHost(resp.JSON200.DbHost),
+		Port: uint16(resp.JSON200.DbPort),
```
sweatybridge, line 37:
```suggestion
		Port: uint16(cast.IntToUint(resp.JSON200.DbPort)),
```

To fix lint check

sweatybridge: @yaten2302 in case you missed it, there are more linter errors to fix https://github.com/supabase/cli/actions/runs/12608039666

yaten2302: @sweatybridge , yes, ig I forgot about the linter errors, I've a doubt, that in [here](https://github.com/supabase/cli/pull/2996/files#annotation_30429263353), these 2 errors are being show <br>
image

In the 2nd one, should I remove `cast.IntToUInt(...)`, I think the error is arising because of that only?
For the first one, I've pushed a commit, I think it should be fixed now (formatted the file)

sweatybridge: > In the 2nd one, should I remove cast.IntToUInt(...), I think the error is arising because of that only?

We need to implement a new method for safely casting UInt to UInt16 and call it. You can check the existing examples in cast.go on how to check for overflows. But if you feel stuck, I can also take over from here.

yaten2302: I'll research upon it a and then ask if I encounter any doubts! Also, I wanted to ask that, is it possible to check the linter errors are failing or passing without pushing the changes?

yaten2302: Hey, I just now pushed a commit to try to fix the linter error, but unfortunately it's failing, could you please guide that is there some other file, in which some issue similar to this is being handled? I would like to understand from that and then try to implement that here.

sweatybridge:
```
// UIntToUInt16 converts a uint to an uint16, handling potential overflow
func UIntToUInt16(value uint) int {
	if value <= math.MaxUInt16 {
		return uint16(value)
	}
	return math.MaxUInt16
}
```

You want to add the above to cast.go and call it like this


```
Port: cast.UIntToUInt16(cast.IntToUInt(reps.JSON200.Port))
```

yaten2302: Oh, okay, got it now, thanks for your help 👍 
But, in the first line, shouldn't the return value of the `UIntToUInt16` func be `uint16`?

sweatybridge (approved):

yaten2302:
Thank you so much @sweatybridge  for helping me out with this PR, this was my very first contribution with supabase, I'll definitely check out some other issues and work on them 👍

PR #3007, opened 2025-01-02
fix: bump gotrue to 2.167.0

monicakh:
## Additional context

PR to update the version of gotrue. The change aligns with the existing version outlined in [line 27 of bump-gotrue-generic.yml](https://github.com/supabase/helper-scripts/blob/master/ansible/bump-gotrue-generic.yml#L27).


monicakh:
@J0 should I also update it to match the latest version (v2.167.0)?

J0:
Yes, the AMI is now on v2.167.0 so it should be safe to bump: https://github.com/supabase/postgres/pull/1389

J0 (approved):

Issue #3003, opened 2025-01-01
config.toml do not accepts env variables for `auth.hook.send_email.uri`


KajSzy:
**Describe the bug**
I am unable to use env variables in `config.toml` file to define custom email hooks url

**To Reproduce**

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Define urls inside `.env` file
2. Reference them inside config.toml file
```toml
[auth.hook.send_email]
enabled = true
uri = "env(SUPABASE_SEND_MAIL_HOOK_URL)"
secrets = "env(SUPABASE_MAIL_SECRET)"
```
3. Run `supabase db reset` script

**Expected behavior**
I expect it to allow using env variables for those config.toml properties

**System information**
Rerun the failing command with `--create-ticket` flag.
 - Ticket ID: [e.g. ab1ac733e31e4f928a4d7c8402543712]
 - Version of OS: MacOS
 - Version of CLI: 2.2.1
 - Version of Docker: 27.4.0
 - Versions of services: 
![Zrzut ekranu 2025-01-2 o 00 16 04](https://github.com/user-attachments/assets/e30a163f-4dd3-45b7-98da-99f8a9fac9cd)

sreejithnnair:
Faced a similar issue in CLI 2.2.1 , even when the email webhook secret is defined explicitly in the terminal, Supabase start throws an error Invalid hook config: auth.hook.send_email.secrets must be formatted as "v1,whsec_<base64_encoded_secret>". Even if you prefix or not prefix - v1,whsec_ with the secret, it seems to throw the same error. Only if the webhook is disabled, does supabase start

Issue #2981, opened 2024-12-18
env variables not working in config.toml


ryanmeisters:
**Describe the bug**
.env variables don't appear to be working. Using supabase cli 2.1.1.

**To Reproduce**
1. `mkdir envtest && cd envtest`
2. `supabase init`
3. `echo "PROJECT_ID=TESTPROJ" > .env`
4. set `project_id = "env(PROJECT_ID)"` in config.toml
5. `supabase start`
6. observe that project id is not set from .env

`WARN: project_id field in config is invalid. Auto-fixing to env_PROJECT_ID_`

**Expected behavior**
.env file should be read and should function according to the [Managing config and secrets](https://supabase.com/docs/guides/local-development/managing-config) doc

**Screenshots**

**System information**
Rerun the failing command with `--create-ticket` flag.
 - Ticket ID:
 - Version of OS: Macos 15.1
 - Version of CLI: 2.1.1
 - Version of Docker: 27.3.1
 - Versions of services: [output from `supabase services` command]
 ```
 SERVICE IMAGE      │      LOCAL       │ LINKED
  ─────────────────────────┼──────────────────┼─────────
    supabase/postgres      │ 15.6.1.143       │ -
    supabase/gotrue        │ v2.164.0         │ -
    postgrest/postgrest    │ v12.2.0          │ -
    supabase/realtime      │ v2.33.58         │ -
    supabase/storage-api   │ v1.14.5          │ -
    supabase/edge-runtime  │ v1.65.4          │ -
    supabase/studio        │ 20241202-71e5240 │ -
    supabase/postgres-meta │ v0.84.2          │ -
    supabase/logflare      │ 1.4.0            │ -
    supabase/supavisor     │ 1.1.56           │ -
```

Issue #2774, opened 2024-10-16
env() does not work in config.toml arrays


schwartzadev:
**Describe the bug**
The `env()` function to load environment variables into the config.toml does not work for arrays in the `config.toml`.

**To Reproduce**
1. Add this to the project `config.toml`

```
[db.seed]
enabled = true
sql_paths = [
    "env(SEEDS_PATH)"
]
```

3. Run `SEEDS_PATH=abc123 supabase db reset`
4. Observe `No seed files matched pattern: supabase/env(SEEDS_PATH)` in the console output. This is the first line in the output.

**Expected behavior**
Supabase uses the path in the environment variable to source database seeds.

**System information**
Rerun the failing command with `--create-ticket` flag.
 - Ticket ID: This does not output from running `SEEDS_PATH=abc123 supabase db reset --create-ticket`
 - Version of OS: Mac 14.4 (23E214)
 - Version of CLI: 1.204.3
 - Version of Docker: Docker version 27.2.0, build 3ab4256
 - Versions of services: 
```

  
        SERVICE IMAGE      │      LOCAL       │   LINKED
  ─────────────────────────┼──────────────────┼─────────────
    supabase/postgres      │ 15.1.0.122       │ 15.1.0.122
    supabase/gotrue        │ v2.162.1         │ v2.162.1
    postgrest/postgrest    │ v11.2.0          │ v11.2.0
    supabase/realtime      │ v2.30.34         │ -
    supabase/storage-api   │ v1.11.7          │ v1.11.14
    supabase/edge-runtime  │ v1.58.12         │ -
    supabase/studio        │ 20240930-16f2b8e │ -
    supabase/postgres-meta │ v0.83.2          │ -
    supabase/logflare      │ 1.4.0            │ -
    supabase/supavisor     │ 1.1.56           │ -
```

**Additional context**
I have tested this with a `.env` file as well, with no luck.

Related issues include:

* https://github.com/supabase/cli/issues/1551
* https://github.com/supabase/cli/issues/1325
* https://github.com/supabase/cli/issues/278
* https://github.com/supabase/cli/issues/210

These issues are related but **distinct** -- as far as I can tell there is no issue for this array-specific bug.

sweatybridge:
May I clarify if your use case is to change seed files based on different environments?

Instead of customising each field by env var, we probably want to offer a more reliable alternative like `SUPABASE_PROJECT_ID=some-ref supabase db reset` which would customise all configs for a specific project.

[...]

Issue #2647, opened 2024-08-30
env replacement in [auth.sms.twilio] / `account_sid`


mfazekas:
**Describe the bug**
It seems that `env(...)` is not evaluated in `account_sid`. Twilio works if it put value directly in `account_sid` but it doesn't if i try to use and env. `env(...)` works in `auth_token`

**To Reproduce**

Edit your config.toml
```toml
[auth.sms.twilio]
enabled = true
account_sid = "env(FOO12)"
message_service_sid = "env(FOO23)"
auth_token = "env(FOO34)"
```

Start supabase with the referenced variables set:
```sh
FOO12=foo12 FOO23=foo23 FOO34=foo34 npx supabase@1.191.3 start
```

Inspect variables in docker:
```
% docker inspect supabase_auth_db |  grep TWILIO
                "GOTRUE_SMS_TWILIO_ACCOUNT_SID=env(FOO12)",
                "GOTRUE_SMS_TWILIO_AUTH_TOKEN=foo34",
                "GOTRUE_SMS_TWILIO_MESSAGE_SERVICE_SID=env(FOO23)",
```

**Expected behavior**
```
% docker inspect supabase_auth_db |  grep TWILIO
                "GOTRUE_SMS_TWILIO_ACCOUNT_SID=foo12",
                "GOTRUE_SMS_TWILIO_AUTH_TOKEN=foo34",
                "GOTRUE_SMS_TWILIO_MESSAGE_SERVICE_SID=foo23",
```

**Screenshots**

**System information**
Rerun the failing command with `--create-ticket` flag.
 - Ticket ID: n/a
 - Version of OS: `Darwin mbpm2.t.hu 23.5.0 Darwin Kernel Version 23.5.0: Wed May  1 20:14:38 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6020 arm64`
 - Version of CLI: `1.191.3` 
 - Version of Docker: `Docker version 27.1.1, build 6312585`
 - Versions of services: 
```
        SERVICE IMAGE      │      LOCAL       │ LINKED
  ─────────────────────────┼──────────────────┼─────────
    supabase/postgres      │ 15.1.1.78        │ -
    supabase/gotrue        │ v2.158.1         │ -
    postgrest/postgrest    │ v12.2.0          │ -
    supabase/realtime      │ v2.30.23         │ -
    supabase/storage-api   │ v1.10.1          │ -
    supabase/edge-runtime  │ v1.56.1          │ -
[...]

PR #3019, opened 2025-01-06
feat: support env loading for all string fields

sweatybridge:

closes https://github.com/supabase/cli/issues/3003
closes https://github.com/supabase/cli/issues/2981
closes https://github.com/supabase/cli/issues/2774
closes https://github.com/supabase/cli/issues/2647

## What is the new behavior?

Ensures all string fields can accept `env()` templates.

## Additional context


pkg/config/config_test.go
```diff
@@ -388,6 +388,7 @@ func TestLoadFunctionImportMap(t *testing.T) {
 			"supabase/config.toml": &fs.MapFile{Data: []byte(`
 			project_id = "test"
 			[functions.hello]
+			enabled = true
```
avallete, line 391: **question**

Why does this had to be added ? Is it related to the changes or just an unrelated fix ?

avallete (approved):

pkg/config/config_test.go
sweatybridge: For some reason viper doesn't handle maps where keys are strings but values are empty structs. It thinks that the key doesn't exist and completely skips it when unmarshaling.

I'm still looking into ways to handle this programmatically.

PR #3020, opened 2025-01-06
fix: pass NPM_CONFIG_REGISTRY env variable when bundling functions

laktek:

This PR provides the option to set a custom NPM registry using the `NPM_CONFIG_REGISTRY` environment variable. This is used when deploying functions.

### DIFF
--- a/internal/functions/deploy/bundle.go
+++ b/internal/functions/deploy/bundle.go
@@ -57,12 +57,16 @@ func (b *dockerBundler) Bundle(ctx context.Context, entrypoint string, importMap
 	if viper.GetBool("DEBUG") {
 		cmd = append(cmd, "--verbose")
 	}
+	env := []string{}
+	if custom_registry := os.Getenv("NPM_CONFIG_REGISTRY"); custom_registry != "" {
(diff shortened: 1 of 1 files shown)


internal/functions/deploy/bundle.go
```diff
@@ -57,12 +57,16 @@ func (b *dockerBundler) Bundle(ctx context.Context, entrypoint string, importMap
 	if viper.GetBool("DEBUG") {
 		cmd = append(cmd, "--verbose")
 	}
+	env := []string{}
+	if custom_registry := os.Getenv("NPM_CONFIG_REGISTRY"); custom_registry != "" {
+		env = append(env, "NPM_CONFIG_REGISTRY="+custom_registry)
```
sweatybridge, line 62:
```suggestion
	if customRegistry := os.Getenv("NPM_CONFIG_REGISTRY"); customRegistry != "" {
		env = append(env, "NPM_CONFIG_REGISTRY="+customRegistry)
```

Just a nitpick on casing

sweatybridge (approved):

PR #3042, opened 2025-01-14
fix(config): dotenvx loading

avallete:

- Also search for `.env` file within `supabase` folder so you can have dedicated .env for supabase deployments.
- Seach for `.env.keys` for local dotenvx development
- Try all keys matching the prefix `DOTENV_PRIVATE_KEY` so you can have `DOTENV_PRIVATE_KEY_PRODUCTION/LOCAL/PREVIEW`

## What is the new behavior?

## Additional context

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -806,6 +806,7 @@ func sanitizeProjectId(src string) string {
 
 func loadDefaultEnv() error {
 	env := viper.GetString("ENV")
+
 	if env == "" {
 		env = "development"
 	}
@@ -813,7 +814,7 @@ func loadDefaultEnv() error {
 	if env != "test" {
 		filenames = append(filenames, ".env.local")
 	}
-	filenames = append(filenames, ".env."+env, ".env")
+	filenames = append(filenames, ".env."+env, ".env", "supabase/.env."+env, "supabase/.env", ".env.keys", "supabase/.env.keys")
 	for _, path := range filenames {
 		if err := loadEnvIfExists(path); err != nil {
 			return err
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -814,7 +814,7 @@ func loadDefaultEnv() error {
 	if env != "test" {
 		filenames = append(filenames, ".env.local")
 	}
-	filenames = append(filenames, ".env."+env, ".env", "supabase/.env."+env, "supabase/.env")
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -806,14 +806,15 @@ func sanitizeProjectId(src string) string {
...
+
 	if env == "" {
 		env = "development"
 	}
 	filenames := []string{".env." + env + ".local"}
 	if env != "test" {
 		filenames = append(filenames, ".env.local")
 	}
-	filenames = append(filenames, ".env."+env, ".env")
+	filenames = append(filenames, ".env."+env, ".env", "supabase/.env."+env, "supabase/.env", ".env.keys", "supabase/.env.keys")
```
sweatybridge, line 817: Shall we avoid loading `.env.keys` in local? I think it's fine to use plaintext values in `.env.local` since it's gitignored.

If they really want to use encrypted values locally, there's an alternative to use

```
npx @dotenvx/dotenvx run -- supabase config push
```

avallete: Yeah, there is always workaround, the friction I had in my case here: https://github.com/supabase/supabase/pull/31213

Was that since I declared the `.env` in `supabase/.env`  and that there was different environment key, but that they're depending over `SUPABASE_ENV` where `dotenvx` make it depend to `ENVIRONMENT` by default. It was either, telling me that the key was wrong to decrypt the secrets, or if I was located in the parent folder, that the keys weren't present (because they were inside `supabase/xxx` instead of being in the project root).

My guess was that if we find a `.env.keys` it probably means that the users want them loaded, and if the user have multiples of them, we want to try them until we find the one able to decrypt the secrets. But can revert tough.

pkg/config/secret.go
```diff
@@ -64,8 +64,20 @@ func DecryptSecretHookFunc(hashKey string) mapstructure.DecodeHookFunc {
 		if t != reflect.TypeOf(result) {
 			return data, nil
 		}
+		// Get all env vars and filter for DOTENV_PRIVATE_KEY
+		var privateKeys []string
+		for _, env := range os.Environ() {
+			key := strings.Split(env, "=")[0]
+			if strings.HasPrefix(key, "DOTENV_PRIVATE_KEY") {
```
sweatybridge, line 71: I think this would also match `DOTENV_PRIVATE_KEYabc`. May be a regex match here would be nice.

avallete: Used a direct match + prefix instead.

sweatybridge (approved):

Issue #2998, opened 2024-12-29
Segmentation fault on db pull


svickers:
**Describe the bug**
I've had no luck getting supabase set up for local development with an existing project.  All the containers start fine but I end up with a blank file `migrations/20241229224617_remote_schema.sql` file and a segmentation fault error.  I've tried db reset multiple times as well but with no luck.

**To Reproduce**
1. npx supabase init
2. npx supabase login
3. npx supabase link
4. npx supabase start
5. npx supabase db pull --linked

**Expected behavior**
New file in /migrations that contains the init script for the database.

**Screenshots**
![Screenshot_624](https://github.com/user-attachments/assets/e9d87fc8-df68-4747-b879-443cab83f36d)

**System information**
 - Ticket ID: 170026373437488bb89e4a8ba115ea6b
 - Version of OS: Windows 11
 - Version of CLI: 2.1.1
 - Version of Docker: 27.4.0
 - Versions of services: 

| SERVICE IMAGE | LOCAL | LINKED |
| --- | --- | --- |
| supabase/postgres  | 15.8.1.020 | 15.8.1.020
|    supabase/gotrue |  v2.167.0         |  v2.167.0
|    postgrest/postgrest   | v12.2.3          | v12.2.3
|    supabase/realtime      | v2.33.58         | -
|    supabase/storage-api   | v1.14.4         | v1.14.4
|    supabase/edge-runtime | v1.65.4         | -
|    supabase/studio       | 20241202-71e5240 | -
|    supabase/postgres-meta | v0.84.2         | -
|    supabase/logflare     | 1.4.0           | -
|    supabase/supavisor     | 1.1.56           | -

sweatybridge:
This is a bit strange. Could you try the following command to help rule out potential issues with the postgres image

```bash
echo '15.6.1.143' > supabase/.temp/postgres-version
npx supabase db pull --linked
```

svickers:
@sweatybridge Same issue still unfortunately.

auskim:
Seeing this same segfault message, on CLI version 2.2.1, macOS ARM binary. I also rolled back to 2.0.0 and manually ran the `pg_dump` commands against our Supabase remote DB, all with the same result. I tried the `pg_dump` command with every `--schema` and `--table` we have, and those do actually all work. 

I didn't see this problem earlier today. `supabase db pull` worked, but I reverted the migration history on the remote and cleared all local migration files to try and do a 'clean pull' and I believe this issue started immediately afterward.

[...]

PR #3043, opened 2025-01-14
fix: exclude pgmq schema from db dump and pull

sweatybridge:

Bug fix https://github.com/supabase/cli/issues/2998

pgmq causes seg fault when running pg_dump. Steps to reproduce

1. Create `supabase/migrations/0_schema.sql`

```sql
CREATE EXTENSION pgmq;
SELECT pgmq.create('my_queue');
```

2. Run `supabase db dump --local`

```bash
Dumping schemas from local database...
--: line 47:     7 Segmentation fault      pg_dump --schema-only --quote-all-identifier --exclude-schema "${EXCLUDED_SCHEMAS:-}" ${EXTRA_FLAGS:-}
...
error running container: exit 139
```

## What is the new behavior?

Exclude `pgmq` schema to unblock db dump.

Schema diffing for pgmq queues wouldn't work anyway because it inserts into pgmq.meta table.

The workaround is to manage those queues manually with migration files like (1).

## Additional context

### DIFF
--- a/internal/db/dump/dump.go
+++ b/internal/db/dump/dump.go
@@ -82,6 +82,7 @@ func dumpData(ctx context.Context, config pgconn.Config, schema, excludeTable []
 		"graphql",
 		"graphql_public",
 		// "net",
+		// "pgmq",
 		// "pgsodium",
 		// "pgsodium_masks",
 		"pgtle",
--- a/internal/utils/misc.go
+++ b/internal/utils/misc.go
@@ -74,6 +74,7 @@ var (
 		"graphql",
 		"graphql_public",
 		"net",
+		"pgmq",
 		"pgsodium",
 		"pgsodium_masks",
 		"pgtle",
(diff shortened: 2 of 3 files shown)


internal/db/dump/dump.go
```diff
@@ -82,6 +82,7 @@ func dumpData(ctx context.Context, config pgconn.Config, schema, excludeTable []
 		"graphql",
 		"graphql_public",
 		// "net",
+		// "pgmq",
```
avallete, line 85: **question**

If we don't want pgmq shouldn't this be uncommented ?

avallete (approved):

internal/db/dump/dump.go
sweatybridge: This is specific to `--data-only` dump. Since I'm not sure what kind of data pgmq stores in its own tables, I'm including pgmq schema to avoid unintentionally missing some data.

PR #3045, opened 2025-01-15
Support adding static files to Edge Functions

laktek:

This PR introduces a new config `static_files`, which can be defined per function in config.toml. This can be used to add static files (eg: HTML, WASM) to Edge Functions.

Eg:
```toml
[functions.serve-html]
static_files = [ "functions/serve-html/*.html" ]
```

Which can then be read in Edge Function, like:

```js
const file = await Deno.readTextFile(new URL("foo.html", import.meta.url))
```


pkg/config/config.go
```diff
@@ -592,6 +593,13 @@ func (c *baseConfig) resolve(builder pathBuilder, fsys fs.FS) error {
 		} else if !filepath.IsAbs(function.ImportMap) {
 			function.ImportMap = filepath.Join(builder.SupabaseDirPath, function.ImportMap)
 		}
+		if len(function.StaticFiles) != 0 {
+			s := []string{}
+			for _, val := range function.StaticFiles {
+				s = append(s, filepath.Join(builder.SupabaseDirPath, val))
+			}
+			function.StaticFiles = s
+		}
```
sweatybridge, line 602: Might need to refer to loadSeedFiles for handling glob patterns.

Also need to handle nested directory when mounting to docker, ie. user may add both html/dir and html/dir/file.html, which docker will reject.

sweatybridge (commented):
Generally ok, commented on some edge cases

pkg/config/config.go
laktek: > Might need to refer to loadSeedFiles for handling glob patterns.

Edge Runtime already handles the glob patterns.

>Also need to handle nested directory when mounting to docker, ie. user may add both html/dir and html/dir/file.html, which docker will reject.

Config only passes glob patterns of paths to look for static files. The functions directory (and its nested directories) is mounted as before. Edge Runtime would look for matching files based on the patterns in the mounted functions directory. 

Also, users can't add a directory (`html/dir`) as a static file. Edge Runtime will throw an error in both `serve` and `deploy`. It has to be either `html/dir/*.html` or `html/dir/file.html`. Will document them in config docs.

sweatybridge: Ic, can we also support relative path `../../index.html` like import map? I think it's only a matter of time before user asks for it

internal/functions/serve/serve.go
```diff
@@ -230,6 +230,13 @@ func populatePerFunctionConfigs(cwd, importMapPath string, noVerifyJWT *bool, fs
 		fc.ImportMap = utils.ToDockerPath(fc.ImportMap)
 		fc.Entrypoint = utils.ToDockerPath(fc.Entrypoint)
 		functionsConfig[slug] = fc
+		if len(fc.StaticFiles) != 0 {
+			s := []string{}
+			for _, val := range fc.StaticFiles {
+				s = append(s, utils.ToDockerPath(val))
+			}
+			fc.StaticFiles = s
+		}
```
sweatybridge, line 239:
```suggestion
		for i, val := range fc.StaticFiles {
			fc.StaticFiles[i] = utils.ToDockerPath(val)
		}
```

Nitpick

pkg/config/config.go
laktek: Only supports paths inside the functions directory. I think it's fine for now, since we want to encourage to keep dependencies of a function inside the same directory.

sweatybridge: Yes, it's fine as long as we document the intended use case and limitations.

Shall we update the config.toml example too https://github.com/supabase/cli/blob/d27db81d36e480a61cb3738981718740b77f6534/pkg/config/templates/config.toml#L259

laktek: added in https://github.com/supabase/cli/pull/3045/commits/29a24a2657880ef8299383c03efac59d4a2cd39f

sweatybridge, line 602:
```suggestion
		for i, val := range function.StaticFiles {
			function.StaticFiles[i] = append(s, filepath.Join(builder.SupabaseDirPath, val))
		}
```

pkg/config/templates/config.toml
```diff
@@ -257,6 +257,11 @@ inspector_port = 8083
 # Uncomment to specify a custom file path to the entrypoint.
 # Supported file extensions are: .ts, .js, .mjs, .jsx, .tsx
 # entrypoint = "./functions/MY_FUNCTION_NAME/index.ts"
+# Use static_files to specify static files to be bundled with the function.
+# You can use explicit paths or glob patterns.
+# For example, if you want to serve static HTML pages in your function:
+# static_files = [ "./functions/MY_FUNCTION_NAME/*.html" ]
+
```
sweatybridge, line 264:
```suggestion
# Specifies static files to be bundled with the function. Supports glob patterns.
# For example, if you want to serve static HTML pages in your function:
# static_files = [ "./functions/MY_FUNCTION_NAME/*.html" ]
```

pkg/function/bundle_test.go
```diff
@@ -36,7 +36,7 @@ func TestBundleFunction(t *testing.T) {
 		// Setup mock bundler
 		bundler := nativeBundler{fsys: fsys}
 		// Run test
-		err := bundler.Bundle(context.Background(), "hello/index.ts", "", &body)
+		err := bundler.Bundle(context.Background(), "hello/index.ts", "", []string{}, &body)
```
sweatybridge, line 39:
```suggestion
		err := bundler.Bundle(context.Background(), "hello/index.ts", "", nil, &body)
```

PR #3046, opened 2025-01-15
feat(cli): add repeatable migrations

### LINKED ISSUE #2784: Support for repeatable migrations
When updating definitions for objects such as views or functions, a new migration is required, which means we have to duplicate the definition to add the necessary modification.

Have some repeatable migrations that are applied with each `migrations up` command, something similar to [Flyway's Repeatable Migrations](https://documentation.red-gate.com/fd/tutorial-repeatable-migrations-184127625.html)

Let's say we have these files in the `migrations` folder

```
[timestamp1]_migration_1.sql
r_my_view_or_function.sql
```

Running `migrations up` would apply `migration_1` and then `my_view_or_function` (it can use `current_timestamp` as id).
We now add a new migration:

```
[timestamp1]_migration_1.sql
[timestamp2]_migration_2.sql
r_my_view_or_function.sql
```

Running `migrations up` would skip `migration_1`, apply `migration_2`, and then re-apply `my_view_or_function` (with a different `current_timestamp` as id).

Manually executing `r_my_view_or_function.sql` after applying migrations by connecting to the database directly

**Additional context**
N/A

JohnBra:

Similar to [flyways repeatable migrations](https://documentation.red-gate.com/fd/tutorial-repeatable-migrations-184127625.html) this adds rudimentary functionality for repeatable database migrations to the supabase cli.

- Adds a flag to `supabase migration new` to create a repeatable migration
- Implementation to differentiate between repeatable and common migrations in `supabase migration up`
- Tests

Instead of versioning repeatable migration files they are applied after all pending migrations locally and can then be pushed to remote.

Currently database objects like views, functions etc. must be duplicated in the migration files when changed. E.g. to change a `view` we create a new versioned migration in which we duplicate the definition of the `view`.

## What is the new behavior?

- `supabase migration new` can now be used to create repeatable migrations
- `supabase migration up` now applies repeatable migrations after all pending migrations

## Additional context

Should close #2784

### DIFF
--- a/cmd/migration.go
+++ b/cmd/migration.go
@@ -152,7 +152,7 @@ func init() {
 	migrationCmd.AddCommand(migrationFetchCmd)
 	// Build new command
 	newFlags := migrationNewCmd.Flags()
-	newFlags.BoolVarP(&repeatable, "repeatable", "r", false, "Creates a repeatable migration instead of a common migration.")
+	newFlags.BoolVarP(&repeatable, "repeatable", "r", false, "Creates a repeatable migration instead of a versioned migration.")
 	migrationCmd.AddCommand(migrationNewCmd)
 	rootCmd.AddCommand(migrationCmd)
 }
--- a/cmd/migration.go
+++ b/cmd/migration.go
@@ -39,12 +39,14 @@ var (
 		},
 	}
 
+	repeatable bool
+
 	migrationNewCmd = &cobra.Command{
 		Use:   "new <migration name>",
 		Short: "Create an empty migration script",
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return new.Run(args[0], os.Stdin, afero.NewOsFs())
+			return new.Run(repeatable, args[0], os.Stdin, afero.NewOsFs())
 		},
 	}
 
@@ -149,6 +151,8 @@ func init() {
 	migrationFetchCmd.MarkFlagsMutuallyExclusive("db-url", "linked", "local")
 	migrationCmd.AddCommand(migrationFetchCmd)
 	// Build new command
+	newFlags := migrationNewCmd.Flags()
+	newFlags.BoolVarP(&repeatable, "repeatable", "r", false, "Creates a repeatable migration instead of a common migration.")
 	migrationCmd.AddCommand(migrationNewCmd)
 	rootCmd.AddCommand(migrationCmd)
 }
--- a/internal/migration/new/new.go
+++ b/internal/migration/new/new.go
@@ -11,8 +11,16 @@ import (
 	"github.com/supabase/cli/internal/utils"
 )
 
-func Run(migrationName string, stdin afero.File, fsys afero.Fs) error {
-	path := GetMigrationPath(utils.GetCurrentTimestamp(), migrationName)
+func Run(repeatable bool, migrationName string, stdin afero.File, fsys afero.Fs) error {
+	path := ""
+
+	if repeatable {
+		// if migration name already exists, repeatable migration will be overwritten
+		path = GetRepeatableMigrationPath(migrationName)
+	} else {
+		path = GetMigrationPath(utils.GetCurrentTimestamp(), migrationName)
+	}
+
 	if err := utils.MkdirIfNotExistFS(fsys, filepath.Dir(path)); err != nil {
 		return err
 	}
@@ -30,6 +38,11 @@ func Run(migrationName string, stdin afero.File, fsys afero.Fs) error {
 
 func GetMigrationPath(timestamp, name string) string {
 	fullName := fmt.Sprintf("%s_%s.sql", timestamp, name)
+	return filepath.Join(utils.MigrationsDir, fullName)
+}
+
+func GetRepeatableMigrationPath(name string) string {
+	fullName := fmt.Sprintf("r_%s.sql", name)
 	return filepath.Join(utils.MigrationsDir, fullName)
 }
 
--- a/internal/migration/new/new.go
+++ b/internal/migration/new/new.go
@@ -12,7 +12,7 @@ import (
 )
 
 func Run(repeatable bool, migrationName string, stdin afero.File, fsys afero.Fs) error {
-	path := ""
+	var path string
 
 	if repeatable {
 		// if migration name already exists, repeatable migration will be overwritten
--- a/internal/migration/up/up.go
+++ b/internal/migration/up/up.go
@@ -46,14 +46,6 @@ func GetPendingMigrations(ctx context.Context, includeAll bool, conn *pgx.Conn, 
 		}
 		utils.CmdSuggestion = suggestIgnoreFlag(diff)
 	}
-	if err != nil {
-		return diff, err
-	}
-	repeatableMigrations, err := migration.ListRepeatableMigrations(utils.MigrationsDir, afero.NewIOFS(fsys))
-	if err != nil {
-		return nil, err
-	}
-	diff = append(diff, repeatableMigrations...)
 	return diff, err
 }
 
--- a/internal/migration/up/up.go
+++ b/internal/migration/up/up.go
@@ -24,17 +24,7 @@ func Run(ctx context.Context, includeAll bool, config pgconn.Config, fsys afero.
 		return err
 	}
 
-	err = migration.ApplyMigrations(ctx, pending, conn, afero.NewIOFS(fsys))
-	if err != nil {
-		return err
-	}
-
-	repeatableMigrations, err := migration.ListRepeatableMigrations(utils.MigrationsDir, afero.NewIOFS(fsys))
-	if err != nil {
-		return err
-	}
-
-	return migration.ApplyMigrations(ctx, repeatableMigrations, conn, afero.NewIOFS(fsys))
+	return migration.ApplyMigrations(ctx, pending, conn, afero.NewIOFS(fsys))
 }
 
 func GetPendingMigrations(ctx context.Context, includeAll bool, conn *pgx.Conn, fsys afero.Fs) ([]string, error) {
@@ -56,6 +46,11 @@ func GetPendingMigrations(ctx context.Context, includeAll bool, conn *pgx.Conn, 
 		}
 		utils.CmdSuggestion = suggestIgnoreFlag(diff)
 	}
+	repeatableMigrations, err := migration.ListRepeatableMigrations(utils.MigrationsDir, afero.NewIOFS(fsys))
+	if err != nil {
+		return nil, err
+	}
+	diff = append(diff, repeatableMigrations...)
 	return diff, err
 }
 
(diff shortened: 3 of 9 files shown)


JohnBra:
@sweatybridge I already see the mistake, will push a fixed version soon

JohnBra:
@sweatybridge as mentioned in the issue #2784 there are a few different options to implement this.

I have opted for a simple solution where repeatable migrations are treated like a seed file.
I.e. all repeatable migrations are applied after the pending migrations have been applied in each `supabase migration up` run.

Currently they won't be pushed to remote or show up in the studio, as this would create discrepancies when versioning with timestamps and subsequent comparisons between remote/local. They are treated the exact same as the seed file.

I experimented locally with creating repeatable migrations and then using the current timestamp as new version for each repeatable migration when applied. (See [opening comment of issue](https://github.com/supabase/cli/issues/2784#issue-2600837627)).
This will create a mismatch between local and remote every time the `migration up` command is run, because the version on remote will always be different to local.

A way around this would be to remove (?!) the remote migration and apply the local one, since the local migration will always supersede the remote migration. Not really sure what to think of this.

internal/migration/up/up.go
```diff
@@ -23,7 +23,18 @@ func Run(ctx context.Context, includeAll bool, config pgconn.Config, fsys afero.
...
 		return err
 	}
-	return migration.ApplyMigrations(ctx, pending, conn, afero.NewIOFS(fsys))
+
+	err = migration.ApplyMigrations(ctx, pending, conn, afero.NewIOFS(fsys))
+	if err != nil {
+		return err
+	}
+
+	repeatableMigrations, err := migration.ListRepeatableMigrations(utils.MigrationsDir, afero.NewIOFS(fsys))
```
sweatybridge, line 32: I think it's worth moving this to inside GetPendingMigrations so other commands which depend on migrations can have these to. For eg. db push and db diff.

cmd/migration.go
```diff
@@ -149,6 +151,8 @@ func init() {
 	migrationFetchCmd.MarkFlagsMutuallyExclusive("db-url", "linked", "local")
 	migrationCmd.AddCommand(migrationFetchCmd)
 	// Build new command
+	newFlags := migrationNewCmd.Flags()
+	newFlags.BoolVarP(&repeatable, "repeatable", "r", false, "Creates a repeatable migration instead of a common migration.")
```
sweatybridge, line 155:
```suggestion
	newFlags.BoolVarP(&repeatable, "repeatable", "r", false, "Creates a repeatable migration instead of a versioned migration.")
```

Use consistent terminology as flyway

internal/migration/new/new.go
```diff
@@ -11,8 +11,16 @@ import (
 	"github.com/supabase/cli/internal/utils"
 )
 
-func Run(migrationName string, stdin afero.File, fsys afero.Fs) error {
-	path := GetMigrationPath(utils.GetCurrentTimestamp(), migrationName)
+func Run(repeatable bool, migrationName string, stdin afero.File, fsys afero.Fs) error {
+	path := ""
+
```
sweatybridge, line 16:
```suggestion
	var path string
```

Nitpick: since the empty string is not used, it's more idiomatic to declare the variable for assignment later.

sweatybridge:
> all repeatable migrations are applied after the pending migrations have been applied in each supabase migration up run

I think this is the right approach for as far as I understand the behaviour of flyway.

> They are treated the exact same as the seed file.

Currently seed files are also versioned and tracked in seed_files table, similar to versioned migration. But it used to be untracked so I can understand where the misconception came from.

pkg/migration/list.go
```diff
@@ -45,6 +46,10 @@ func ListLocalMigrations(migrationsDir string, fsys fs.FS, filter ...func(string
 			fmt.Fprintf(os.Stderr, "Skipping migration %s... (replace \"init\" with a different file name to apply this migration)\n", filename)
 			continue
 		}
+		if strings.HasPrefix(filename, "r_") {
+			// silently skip repeatable migrations
+			continue
+		}
```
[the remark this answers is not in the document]
JohnBra: Added to not confuse users with the "bla file pattern" console output when the filename doesn't start with a timestamp. I can still remove it if you want

internal/migration/up/up.go
JohnBra: Yep, will move it into `GetPendingMigrations`.

IMO, squashing would have to ignore repeatable migrations. Otherwise we are mixing versioned migrations (usually not repeatable) with repeatable migrations. Another option would be squashing repeatable migrations separately as in squash all versioned migrations together and squash all repeatable migrations together.

pkg/migration/list.go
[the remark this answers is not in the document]
sweatybridge: Ic, it's fine to leave it in then.

JohnBra:
I guess I should also add the repeatable migrations to the `supabase db reset` and add tests in `supabase migration squash`.

JohnBra:
@sweatybridge that should be everything now.

I wasn't aware the seed files are versioned as well. I think it would be good to version repeatable migrations, too if possible.

Does the cli manage the seed file versioning, or is that done somewhere else? I couldn't find it in `internal` or `cmd`.

sweatybridge:
> I think it would be good to version repeatable migrations, too if possible.

What's the use case for tracking repeatable migrations? My understanding is that repeatable migrations should always after all versioned migrations, even if there are no changes.

> Does the cli manage the seed file versioning, or is that done somewhere else?

The versioning of seed file is done in https://github.com/supabase/cli/blob/develop/pkg/migration/seed.go#L33

pkg/migration/list.go
```diff
@@ -60,6 +65,24 @@ func ListLocalMigrations(migrationsDir string, fsys fs.FS, filter ...func(string
...
+func ListRepeatableMigrations(migrationsDir string, fsys fs.FS) ([]string, error) {
+	localMigrations, err := fs.ReadDir(fsys, migrationsDir)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return nil, errors.Errorf("failed to read directory: %w", err)
+	}
+	var repeatable []string
+
+	for _, migration := range localMigrations {
+		filename := migration.Name()
+		if strings.HasPrefix(filename, "r_") && strings.HasSuffix(filename, ".sql") {
```
sweatybridge, line 77: How about moving this check into ListLocalMigrations so it lists both versioned and repeatable? I think that will cover more scenarios than calling from utils.GetPendingMigrations.

sweatybridge:
A few more complications that need answers
1. What should happen when user runs `supabase db reset --version <timestamp>`? Some repeatable migrations might depend on new schemas defined in later versions.
2. Can it handle `supabase db reset --linked` which triggers the code path here https://github.com/supabase/cli/blob/develop/internal/db/reset/reset.go#L235

JohnBra:
> > I think it would be good to version repeatable migrations, too if possible.
> 
> What's the use case for tracking repeatable migrations? My understanding is that repeatable migrations should always after all versioned migrations, even if there are no changes.

I suppose tracking them within the db isn't necessary considering they are tracked in version control and always applied after the versioned migrations. The only use case I see is that you could inspect them in supabase studio, which currently isn't possible. Leaning towards not tracking them then.

> > Does the cli manage the seed file versioning, or is that done somewhere else?
> 
> The versioning of seed file is done in https://github.com/supabase/cli/blob/develop/pkg/migration/seed.go#L33

Thanks, will have a look at it.

pkg/migration/list.go
JohnBra: If I understand correctly, your suggestion is to include all repeatable migrations in the `ListLocalMigrations` result.

I separated `ListRepeatableMigrations` from `ListLocalMigrations` to prevent running into errors when diffing local and remote migrations [in `migration up`](https://github.com/supabase/cli/blob/cbec946e595950f5f1e779c6a1b9b9086b9c8e7c/internal/migration/up/up.go#L39). Since the repeatable migrations aren't kept in the versioned migrations table, we would always run into a err `ErrMissingRemote`. That's why the repeatable migrations are added to the diff slice after the check.

If I include repeatable migrations in the `ListLocalMigrations` result, I will have to tweak the [`FindPendingMigrations`](https://github.com/supabase/cli/blob/develop/pkg/migration/apply.go#L21) func to not return an error for repeatable migrations. LMK if that's your preferred solution.

NOTE: There could be further side effects of including repeatable migrations in `ListLocalMigrations`.

JohnBra:
> A few more complications that need answers
> 
> 1. What should happen when user runs `supabase db reset --version <timestamp>`? Some repeatable migrations might depend on new schemas defined in later versions.

I think it would be ok to break if the reset can be reverted, or alternatively not allow the command with a warning in this case. Maybe there we could do a dry run, see if it succeeds and react accordingly.

> 2. We also need to handle `supabase db reset --linked` which triggers the code path here https://github.com/supabase/cli/blob/develop/internal/db/reset/reset.go#L235
> 3. Should repeatable migrations appear in `migration list` and be downloadable from `migration fetch`? If so, we probably need to record the contents of repeatable migrations in the `schema_migrations` table as well, together with versioned ones.

True, 2 needs to be considered.

I guess at this point it might make sense to treat repeatable migrations as versioned migrations with only one or an empty/`""` version field.

We could then load them through `LoadPartialMigrations`/`LoadLocalMigrations`, which should solve point 2. above as well as make them listable via `migration list` and inspectable in the studio.

What are your thoughts on that? @sweatybridge

sweatybridge:
> I guess at this point it might make sense to treat repeatable migrations as versioned migrations with only one or an empty/"" version field.
>
> We could then load them through LoadPartialMigrations/LoadLocalMigrations, which should solve point 2. above as well as make them listable via migration list and inspectable in the studio.

Many birds in one stone. I think this is the most ideal solution. Thank you for coming up with the suggestion.

JohnBra:
@sweatybridge the current implementation should satisfy most requirements.

I changed it so repeatable migrations are now treated as versioned migrations, with their name as version. Leaving it as an empty string wouldn't work because the `version` is the PK in the `schema_migrations` table.

The [regex to check the migration files](https://github.com/supabase/cli/blob/322044b462784586f5ffb688788986fd01a6f41d/pkg/migration/file.go#L26) for correct formatting and `version`/`name` group matching now includes an adjustment to match repeatable migrations as well.

I also modified the `makeTable` func to [append the repeatable migrations to the versioned migrations](https://github.com/supabase/cli/blob/322044b462784586f5ffb688788986fd01a6f41d/internal/migration/list/list.go#L73). They always append to the end (chronologically newest) and don't include a timestamp, as they currently don't have one. If there are discrepancies between local and remote, this should also be visible:
Screenshot 2025-01-24 at 3 44 09 PM

The repeatable migrations now also show up in the studio:
Screenshot 2025-01-24 at 3 46 06 PM

Before I implement the `supabase db reset --version <version>`, does this seem better than the solution before?

sweatybridge:
Sorry again for the delay. Given the complexity of this change, I took a step back and suggested an alternative in the original issue https://github.com/supabase/cli/issues/2784#issuecomment-2623582604

Let's see if users are comfortable with declarative schema which imo better addresses the duplication issue.

JohnBra:
Hey @sweatybridge, just checking in if this still relevant. Happy to close if not.

IAmRiteshKoushik:
Is this PR still relevant / being worked upon ?

PR #3052, opened 2025-01-17
feat(settings): add track_activity_query_size setting configuration

avallete:
To merge after: https://github.com/supabase/infrastructure/pull/21241

- Add track_activity_query_size setting customization
- Update types and beta.yaml to latest api schema


cmd/snippets.go
```diff
@@ -30,6 +32,9 @@ var (
 		Long:  "Download contents of the specified SQL snippet.",
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if !utils.UUIDPattern.MatchString(args[0]) {
+				return errors.Errorf("snippet ID %q is not a UUID", args[0])
+			}
```
sweatybridge, line 37:
```suggestion
```

We can just rely on the uuid.Parse called inside Run

api/beta.yaml
```diff
@@ -13,13 +13,13 @@ paths:
           schema:
             type: string
       responses:
-        '200':
-          description: ''
```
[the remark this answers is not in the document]
avallete: We can, but the issue is that this is the default everytime you generate things from `v1-yaml` endpoint. Changing it, mean changing it everytime we sync things up, so I'll be more in favor to either edit the yaml generation or just change it here.

pkg/config/db.go
```diff
@@ -43,6 +43,7 @@ type (
 		SessionReplicationRole        *SessionReplicationRole `toml:"session_replication_role"`
 		SharedBuffers                 *string                 `toml:"shared_buffers"`
 		StatementTimeout              *string                 `toml:"statement_timeout"`
+		TrackActivityQuerySize        *uint                   `toml:"track_activity_query_size"`
```
sweatybridge, line 46: This field needs to be a string because it can support "1MB" https://postgresqlco.nf/doc/en/param/track_activity_query_size/

avallete: It doesn't it's an integer, the "1MB" is for when you set it's value to: 1048576

See:
https://www.postgresql.org/docs/17/runtime-config-statistics.html#GUC-TRACK-ACTIVITY-QUERY-SIZE

cmd/snippets.go
avallete: Yeah, added it here so it's 1-1 with the other places where we validate UUID format like here: https://github.com/supabase/cli/blob/feat/add-track-activity-query-size-setting/cmd/sso.go#L63

api/beta.yaml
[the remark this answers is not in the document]
sweatybridge: > default everytime you generate things from v1-yaml endpoint

Actually our https://api.supabase.com/api/v1-yaml endpoint always serves single quote. It's probably your vscode prettier plugin that's formatting it to double quotes.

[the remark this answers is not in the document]
avallete: Ohh right, I thought it was the other way around, my bad will change it back to single quotes.

pkg/config/db.go
sweatybridge: Well.. that's the confusing part of their docs. I realised this when addin other `*_size` fields like https://postgresqlco.nf/doc/en/param/effective_cache_size/

You can run the following query to check that 1kB is indeed valid.

```sql
show track_activity_query_size;
```

avallete: Good catch, updated accordingly both the cli and the api PR: https://github.com/supabase/infrastructure/pull/21241/files

sweatybridge (approved):
lgtm. pending next api deploy

avallete:
Ready to go, tested with the latest api deploy on a project, seems all good.

PR #3063, opened 2025-01-22
fix: Exclude realtime messages publication from publication dropping

filipecabaco:

Exclude realtime messages publication from publication dropping

Requires https://github.com/supabase/realtime/pull/1274 to be merged


sweatybridge (approved):
we can release to beta channel since this change is fully backwards compatible

Issue #1190, opened 2023-06-07
Set up Webhooks in the config file


NatoNathan:
At the moment you have to manually setup database webhooks. You also need to redo the setup when you reset the DB. This becomes a problem when collaborating with others and sharing projects. It make it harder to provide a single command to setup and configure a project locally or remote.

It would be nice to be able to use the `config.toml` file to define webhooks that are automatically  setup when you start or reset a project. Maybe something like this?

```toml
[[db.webhooks]]
name = "my_webhook"
table= "auth.users"
operation = ["INSERT", "UPDATE", "DELETE"]
url = "http://localhost:3000/webhook"
params = { key = "value" }
headers = { key = "value" }
```

You can use SQL to make custom webhooks with triggers. but then you can't easily use environment variables for things like the url, params or headers.

sweatybridge:
An alternative is to create web hooks from studio UI and save them as as a new migration. 

```bash
supabase db diff --schema supabase_functions -f my_webhook
```

Would the above work for you?

NatoNathan:
The important thing is to be able to set a webhook like this

```toml
[[db.webhooks]]
name = "my_webhook"
table= "auth.users"
operation = ["INSERT", "UPDATE", "DELETE"]
url = "env(API_URL)"
params = { Authorization = "env(JWT)" }
headers = { key = "value" }
```

where `API_URL` and `JWT` change between local, staging and prod. otherwise I would be happy to use SQL to define the trigger.

duyleekun:
Would `localhost` works in your case @NatoNathan

NatoNathan:
> Would `localhost` works in your case @NatoNathan

Not really, I need to be able to target the docker host when running supabase locally and need to target supabase functions when we deploy it.

bombillazo:
Yup, supabase hardcodes the webhook URL into the trigger definition, so any migration breaks the environment this was not build for...

marcfrankel:
Jumping in to say this is also something we could use. It's hard to use db webhooks right now because the URL can't be dynamic based on environment variables

rafaelmagalhaes:
has anyone come up with a temp solutions for this, facing some issues between development and production environments

jmwind:
@sweatybridge I'm open to contributing a PR for to allow webhook config in the config.toml file. It's such a pain for local dev now that we depend on a few webhooks. 

[...]

PR #3080, opened 2025-01-28
feat: update vault secrets from config

sweatybridge:

feature

## What is the new behavior?

Updates vault secrets before migration.

- migration up
- db diff
- db push
- db reset
- db start

## Additional context

relates to https://github.com/supabase/cli/issues/1190

### DIFF
--- a/internal/db/start/start.go
+++ b/internal/db/start/start.go
@@ -368,9 +368,24 @@ func SetupLocalDatabase(ctx context.Context, version string, fsys afero.Fs, w io
 	return apply.MigrateAndSeed(ctx, version, conn, fsys)
 }
 
+const CREATE_VAULT_SECRET = "SELECT vault.create_secret($1, $2)"
+
 func SetupDatabase(ctx context.Context, conn *pgx.Conn, host string, w io.Writer, fsys afero.Fs) error {
 	if err := initSchema(ctx, conn, host, w); err != nil {
 		return err
+	}
+	// Create vault secrets first so roles.sql can reference them
+	batch := pgx.Batch{}
+	for k, v := range utils.Config.Db.Vault {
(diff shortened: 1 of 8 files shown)


avallete (approved):
LGTM

pkg/config/templates/config.toml
```diff
@@ -42,6 +42,9 @@ default_pool_size = 20
 # Maximum number of client connections allowed.
 max_client_conn = 100
 
+# [db.vault]
+# secret_key = "env(SECRET_VALUE)"
+
 [db.seed]
 # If enabled, seeds the database after migrations during a db reset.
```
avallete, line 49: **suggestion**

Should we put the vault under:

```
[db.seed]
[db.seed.vault]
```

So it's clear when those values are imported ?

sweatybridge: For vault to be useful, this has to run before migrations so users can reference it when creating triggers. Seed on the other hand has to run after migrations.

avallete: Hummm I see. Yeah even if the inserting data feels like more seeding the time it execute isn't the same.

internal/db/start/start.go
```diff
@@ -368,10 +368,25 @@ func SetupLocalDatabase(ctx context.Context, version string, fsys afero.Fs, w io
 	return apply.MigrateAndSeed(ctx, version, conn, fsys)
 }
 
+const CREATE_VAULT_SECRET = "SELECT vault.create_secret($1, $2)"
```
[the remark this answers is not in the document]
sweatybridge: I think we need to rerun this on migration up and db push. So yes, I will have to pull and merge with the current secrets before updating.

PR #3091, opened 2025-01-31
fix: print notices from migration statements

sweatybridge:

Bug fix

## What is the new behavior?

Improve migration debugging experience by logging notice messages and error details.

```
Applying migration 1_schema.sql...
ERROR: syntax error at or near "exist" (SQLSTATE 42601)
At statement 0:                                        
create table if not exist my_table (name text)         
                    ^                           
```

## Additional context

### DIFF
--- a/pkg/migration/file.go
+++ b/pkg/migration/file.go
@@ -90,24 +90,28 @@ func (m *MigrationFile) ExecBatch(ctx context.Context, conn *pgx.Conn) error {
 		}
 		var pgErr *pgconn.PgError
 		if errors.As(err, &pgErr) {
-			lines := strings.Split(stat, "\n")
-			pos := int(pgErr.Position)
-			for j, r := range lines {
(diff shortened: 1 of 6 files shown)


pkg/migration/file.go
```diff
@@ -86,7 +88,24 @@ func (m *MigrationFile) ExecBatch(ctx context.Context, conn *pgx.Conn) error {
...
+					continue
+				}
+				if pos > 0 {
+					caret := append(bytes.Repeat([]byte{' '}, pos-1), '^')
+					lines = append(lines[:j+1], string(caret))
+				}
+				break
+			}
+			stat = strings.Join(lines, "\n")
+		}
```
avallete, line 107: **suggestion**

Add a comment on top of this block explaining what it does, as it's not trivial to figure out from the code and the string/chars manipulations in it 😅

avallete (approved):

Issue #2139, opened 2024-04-10
Allow customization of the supabase root directory name


Bowbee:
**Is your feature request related to a problem? Please describe.**
I have a monorepo that has a very specific naming structure with `/apps` and `/libs` folders. The option for `--workdir` is great when initializing a new project, however I would love to specify the name of the root directory alongside this.

**Describe the solution you'd like**
Having a cli flag for `--rootname <folder name>` would be great.

**Describe alternatives you've considered**
The alternative is to place a parent folder with the desired name before initializing a project. Eg: `apps/testproject/supabase`.
Something like this is required for multiple projects to exist within a monorepo.

**Additional context**
This appears to currently be hardcoded in [internal/utils/misc.go](https://github.com/supabase/cli/blob/develop/internal/utils/misc.go#L169).
I am unsure whether other considerations are needed such as an indicator for what is a supabase directory or not. Maybe the CLI can automatically detect this based off `--workdir` or the contents of the `cwd`.

sweatybridge:
Do you mean to customize like `apps/testproject/custom-supabase`? That wouldn't be possible because a lot of our platform's GitHub integrations depend on a directory named `supabase`.

If you mean `apps/testproject/supabase`, I think it's already supported with `supabase init --workdir apps/testproject` flag.

Bowbee:
Yes I mean `apps/testpropject/custom-supabase` - this would be great. Maybe this can be accomplished by having the directory name specified in the `config.toml` instead of assumed to be `supabase` always?

sweatybridge:
`config.toml` lives inside `supabase` directory though. How would you suggest we find the right config.toml file if we support `custom-supabase` directory?

tomx-sh:
@sweatybridge I'm also interested in the possibility to place Supabase directories anywhere I want in my project.
Next.JS docs encourage a structure where everything is in the `src` or `app` folder, and the root directory is only used for config files.
I can imagine a `supabase.config.toml` file here, or a `.supabase` directory ' (personally prefer the first solution) where you would provide the paths of supabase stuff!

sweatybridge:
I see, so perhaps a `--config <path-to-toml>` flag that can be passed to CLI to override the default location of `supabase/config.toml`.

Within `config.toml`, users can declare path overrides for other directories relative to the location of `config.toml`. For eg.

```toml
workdir = '..'

[migrations]
path = './migrations/*.sql'

[seed]
path = './seed.sql'

[functions.slug]
[...]

PR #3092, opened 2025-01-31
Add Custom Config File Support

StoneTapeStudios:
# Add Custom Config File Support

Introduces a new `--config-file` flag to the Supabase CLI, enabling users to specify alternative configuration files. This enhancement allows running multiple parallel local development environments with different configurations (https://github.com/supabase/cli/issues/2139).

## Example Usage

```bash
supabase start --config-file=supabase/config.test.toml
```

## Benefits
- Supports multiple local database instances with different configurations
- Simplifies management of different environment setups (e.g., testing, staging)
- Maintains backwards compatibility with default config path


sweatybridge (commented):
Generally agree to add this functionality. However, the current implementation doesn't handle [changing workdir](https://github.com/supabase/cli/blob/develop/cmd/root.go#L94) based on the given config path.

StoneTapeStudios:
> Generally agree to add this functionality. However, the current implementation doesn't handle [changing workdir](https://github.com/supabase/cli/blob/develop/cmd/root.go#L94) based on the given config path.

Thanks for the feedback! I think I have addressed the issues with my latest changes.

PR #3101, opened 2025-02-04
fix: improve function config parse error message

avallete:

- Improve the functions parse config error to give more actionable error message for functions in common case of `Unknown field` or `functions[0] ... expected a map` error.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -426,7 +426,7 @@ func (c *config) loadFromReader(v *viper.Viper, r io.Reader) error {
 		dc.TagName = "toml"
 		dc.Squash = true
 		dc.ZeroFields = true
-		dc.DecodeHook = c.newDecodeHook(LoadEnvHook, ValidateFunctionsHookFunc)
+		dc.DecodeHook = c.newDecodeHook(LoadEnvHook, ValidateFunctionsHook)
 	}); err != nil {
 		return errors.Errorf("failed to parse config: %w", err)
 	}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -16,7 +16,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"reflect"
 	"regexp"
 	"sort"
 	"strconv"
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -396,6 +396,28 @@ func (c *config) loadFromFile(filename string, fsys fs.FS) error {
 	return c.loadFromReader(v, f)
 }
 
+// Used to improve config parsing errors into a more contextual and actionable
+// error message
+func wrapTomlError(err error) error {
```
sweatybridge, line 401: Can we implement this as a decode hook? By checking if the source type is a non-map value and the target type is `FunctionConfig`.

avallete: I extracted the custom decode hooks into their own file so it's easier to find them and refactored to use a custom decode hook as you suggested here: https://github.com/supabase/cli/pull/3101/commits/c256ee3697efb3d38e67668059e12322a8f49c6e

pkg/config/decode_hooks.go
```diff
@@ -0,0 +1,75 @@
...
+	if f.Kind() != reflect.Map {
+		return nil, errors.New(`Invalid functions config format. Functions should be configured as:
+
+[functions.<function-name>]
+field = value
+
+Example:
+[functions.hello]
+verify_jwt = true
+`)
```
sweatybridge, line 45: Let's refactor this string to a constant so it can be reused on line 61

pkg/config/decode_hooks.go
```diff
@@ -0,0 +1,75 @@
...
+	if matches := envPattern.FindStringSubmatch(value); len(matches) > 1 {
+		if env := os.Getenv(matches[1]); len(env) > 0 {
+			value = env
+		}
+	}
+	return value, nil
+}
+
+// ValidateFunctionsHookFunc is a mapstructure decode hook that validates the functions config format.
+func ValidateFunctionsHookFunc(f reflect.Type, t reflect.Type, data interface{}) (interface{}, error) {
```
sweatybridge, line 29:
```suggestion
// ValidateFunctionsHook is a mapstructure decode hook that validates the functions config format.
func ValidateFunctionsHook(f reflect.Type, t reflect.Type, data interface{}) (interface{}, error) {
```

nitpick: a `HookFunc` is a function that returns a hook. Subtle difference but I think it's worth having consistent names.

sweatybridge (commented):
looks good!

sweatybridge (approved):

PR #3102, opened 2025-02-04
fix: improve config parsing error message for env usage

avallete:

- An attempt to provide better error messages for misuses of `env.VARIABLE` syntax by directly parsing the TOML before viper.


pkg/config/config.go
```diff
@@ -415,12 +416,45 @@ Example:
 verify_jwt = true
 `)
 	}
+
 	return err
 }
 
+func wrapTomlUnmarshalError(err toml.ParseError, content *string) error {
```
sweatybridge, line 423: I think it's better to check this in [LoadEnvHook](https://github.com/supabase/cli/blob/develop/pkg/config/config.go#L779), possibly using a regex.

avallete: I think this won't be possible as the decode happen after the parsing doesn't it ?

In this case we're not checking if a valid syntax match some pattern and will be valid config, but we're catching and giving a better error about a parsing issue that producing an invalid toml:

```toml
some_value = env.SOMETHING
```

Isn't a valid toml value, which raising issue at parsing time.

sweatybridge: hmm this exact error is produced by the following config.

```
project_id = "test"

env.OPENAI_API_KEY = ""
```

It feels to me doing anything more than hooks would be a little overkill for the purpose of error reporting. The toml syntax can be easily validated by a vscode plugin so we should just focus on semantic errors.

avallete: Hum, right I though the issue was about users doing:

```
project_id = "test"

open_api_key = env.OPEN_API_KEY
```

Which I could understand. I'll drop this PR for now we'll see if we wanna revisit.

PR #3103, opened 2025-02-04
feat: make config file optional

sweatybridge:

feature

## What is the new behavior?

Use sane defaults when config file is not present.

## Additional context

### DIFF
--- a/internal/utils/flags/config_path.go
+++ b/internal/utils/flags/config_path.go
@@ -1,10 +1,6 @@
 package flags
 
 import (
-	"fmt"
-	"os"
-
-	"github.com/go-errors/errors"
 	"github.com/spf13/afero"
 	"github.com/supabase/cli/internal/utils"
 )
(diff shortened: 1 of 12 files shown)


internal/utils/flags/config_path.go
```diff
@@ -1,20 +1,13 @@
...
 	"github.com/supabase/cli/internal/utils"
 )
 
 func LoadConfig(fsys afero.Fs) error {
 	utils.Config.ProjectId = ProjectRef
 	if err := utils.Config.Load("", utils.NewRootFS(fsys)); err != nil {
-		if errors.Is(err, os.ErrNotExist) {
-			utils.CmdSuggestion = fmt.Sprintf("Have you set up the project with %s?", utils.Aqua("supabase init"))
-		}
 		return err
```
avallete, line 11: **suggestion**

Maybe we might still want to print some kind of warning in that case rather than error out ?

avallete (approved):

internal/utils/flags/config_path.go
sweatybridge: oh this code path is dead because [config.Load](https://github.com/supabase/cli/pull/3103/files#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5R392) now catches the missing file error.

Issue #3118, opened 2025-02-07
`api.external_url` config field is always overridden


toBeOfUse:
**Describe the bug**
In theory, the `api.external_url` config.toml field allows the user to tell a Supabase instance what URL its API will be accessible at when it's created. This URL should be used as the base by Supabase Auth endpoints like ["POST /admin/generate_link"](https://github.com/supabase/auth?tab=readme-ov-file#post-admingenerate_link), which needs to use the base URL in the URLs it generates.

However, in practice, the value supplied for `external_url` in config.toml is always overwritten by the following code:

https://github.com/supabase/cli/blob/9142412562468ac0adfdbb9e7ebf43560b1387f6/pkg/config/config.go#L547-L556

**To Reproduce**
1. Supply an `external_url` string under the `[api]` section in config.toml.
2. Start a new Supabase instance with `supabase start`.
3. Generate a link through the Supabase Auth API.
4. The link will start with 127.0.0.1, in spite of whatever you supplied as the `external_url`.

**Expected behavior**
I expect the generated link to start with the `external_url` I supplied.

**System information**
Rerun the failing command with `--create-ticket` flag.

There is no specific failing command, only a failure to use the config.

PR #3119, opened 2025-02-07
Fix: Don't overwrite `api.external_url` from config.toml

toBeOfUse:

This fixes a bug that prevents users from setting a config value containing the external URL that the Supabase API will be deployed at.

As stated in #3118, although the `api.external_url` config field exists, its value is always overwritten in pkg/config/config.go. This prevents the user from setting the external URL that is used for things like Supabase Auth generating links.

## What is the new behavior?

Now, if there is a value supplied for `api.external_url`, it will be used as the API's external URL. If there is not a value supplied for `api.external_url`, it will still be set to a value derived from the hardcoded hostname, "127.0.0.1".


sweatybridge (approved):
Thanks for the quick fix!

PR #3121, opened 2025-02-09
feat: add mailpit image

silentworks:

Replace inbucket with mailpit. For now I'm opening this to start a discussion about this change as I believe mailpit provides more verbose features and works quite well after testing it for a week now. I've kept this PR at a minimal in order to keep backward compatibility and not have a breaking change. I did at first changed all references to `inbucket` to be `mailpit` but realised this would cause a break in everyone's `config.toml` since they would still be referencing `[inbucket]`. What could be done instead is to create a new config for mailpit but this time it should be given a universal name rather than the service we are using name. So instead of `[inbucket]` or `[mailpit]` it should be more generic like `[mail]` this way if this is to change in the future it would all be internal changes and none breaking to it's external use.

The benefits I see in mailpit is that it has features like HTML check and an API for e2e testing too. The UI looked more refined and having the various views like HTML, HTML Source, Text, Headers, Raw amongst others was also good.

To test this you would have to set the registry to `docker.io` as Supabase doesn't have a mirror image of mailpit.

Note `supalocal` is just an alias in my terminal to the built executable of the `cli`.

```sh
SUPABASE_INTERNAL_IMAGE_REGISTRY=docker.io supalocal start
```

## What is the new behavior?

## Additional context


sweatybridge (commented):
Very nice! I will start adding the image registry mirrors for ECR and ghcr.

sweatybridge:
I've mirrored the image to ghcr to unblock CI tests. https://github.com/orgs/supabase/packages/container/package/mailpit

Once this PR is approved, mirror to ECR will also happen.

sweatybridge (approved):

silentworks:
@chekdotdev I honestly did think about this when making the PR and you are right that instructions and a heads up should have been provided around this as I too do use inbucket in my e2e tests with playwright. My main thought though was that mailpit provides better API for e2e testing than inbucket as with inbucket you don't get an API endpoint at all. The only suggestion I can now give after the fact is that you can pin your CLI verison to an older version whilst making the transition of your e2e tests to mailpit. You can ping me on the Supabase discord with my same username here if you have any specific questions or need any further help. Sorry again for not putting some information out there before this PR got merged in.

Issue #3117, opened 2025-02-07
Support remote function deployments without docker


Thorsten-Kd:
**Is your feature request related to a problem? Please describe.**
We're running supabase project creation inside our aws ecs container where we can't run a local docker container to deploy functions to our freshly created remote supabase instance.

**Describe the solution you'd like**
Ideally, the CLI wouldn't need to depend on docker for remote function deployment and generates the resulting zip file without docker. As far as I can tell, the docker engine is only used to create a zip file that is uploaded to supabase using the CLI.
Alternatively, could the  CLI  expose a command to deploy the functions using a custom generated zip file that we generate outside of docker? We could probably replicate some of the zip file creation logic on our end to achieve that.

**Describe alternatives you've considered**
Looked into Podman, but it's not feasible on ECS as ECS relies on their docker engine and configuration.

**Additional context**

sweatybridge:
Hi @Thorsten-Kd, we have released an experimental flag to deploy functions without docker. You can try it out using the beta release channel.

```bash
npx supabase@beta functions deploy --use-api
```

Let me know if it works for your use case.

Thorsten-Kd:
Thank you @sweatybridge, that works great for our use case! I found one issue that is not a blocker for us as we want to move away from the index file export approach, but here is an example in case you want to address it before releasing out of beta state:
Let's assume we have this structure in our functions folder:
```
shared
--- index.ts 
--- file1.ts
--- file2.ts
my-function
--- index.ts
```
**shared/index.ts:**
```
export * from ./file1.ts
export * from ./file2.ts
```

**shared/file1.ts**
`export const abc = 'abc';`

**shared/file2.ts**
`export const dce = 'dce';`

**my-function/index.ts:**
`import { abc } from ../shared`

When I deploy `my-function` with your command it starts uploading all files including `shared/index.ts` and `shared/file1.ts`, but not `shared/file2.ts` which is also not used in `my-function`, so it's expected. 
But than I get an error message that looks like this:
```
[...]

PR #3123, opened 2025-02-10
feat: support deploying functions without docker

sweatybridge:

feature
closes https://github.com/supabase/cli/issues/3117

## What is the new behavior?

```bash
supabase functions deploy [--use-docker|--use-api]
```

## Additional context


internal/functions/deploy/upload.go
```diff
@@ -0,0 +1,183 @@
...
+// Ref: https://regex101.com/r/DfBdJA/1
+var importPathPattern = regexp.MustCompile(`(?i)import\s+(?:{[^{}]+}|.*?)\s*(?:from)?\s*['"](.*?)['"]|import\(\s*['"](.*?)['"]\)`)
+
+func walkImportPaths(srcPath string, importMap utils.ImportMap, fsys afero.Fs, callback func(curr string, data []byte) error) error {
+	seen := map[string]struct{}{}
+	// BFS so we can list paths in increasing depth
+	q := make([]string, 1)
+	q[0] = srcPath
+	for len(q) > 0 {
+		curr := q[len(q)-1]
```
avallete, line 141: **question**

Comment says `BFS` but seems like we're popping from the end (`curr := q[len(q)-1]`) wouldn't that make it a DFS ?

internal/functions/deploy/upload.go
```diff
@@ -0,0 +1,183 @@
...
+			}
+			mod = strings.TrimSpace(mod)
+			// Substitute kv from import map
+			for k, v := range importMap.Imports {
+				if strings.HasPrefix(mod, k) {
+					mod = v + mod[len(k):]
+				}
+			}
+			if strings.HasPrefix(mod, "./") || strings.HasPrefix(mod, "../") {
+				mod = path.Join(path.Dir(curr), mod)
```
avallete, line 175: **question**

I know there is a OS relative reason to use `filepath.Join` or `path.Join` I'm not sure in this case, but would that yield the proper result on windows ?

internal/functions/deploy/upload.go
```diff
@@ -0,0 +1,183 @@
...
+			} else if !strings.HasPrefix(mod, "/") {
+				continue
+			}
+			q = append(q, mod)
```
avallete, line 179: **question**

Would it be worth to use `filepath.Clean` to always get the shortest consistant path ? Not sure about if that could break the bundling tough 🤔

avallete (approved):

internal/functions/deploy/upload.go
sweatybridge: Yes, good catch. DFS is slightly more efficient here to avoid array allocation.

internal/functions/deploy/upload.go
sweatybridge: I'm treating all file paths as unix paths for compatibility with our api. The path is converted to OS specific separator when actually reading from the file system.

Also interesting to note that glob only works with unix path patterns. So I think going forward, we want to use unix path in all the middle layers and only convert just before passing to FS.

internal/functions/deploy/upload.go
sweatybridge: Hmm good catch. I think cleaning will reduce duplicates since the seen map is keyed by full path.

internal/functions/deploy/testdata/modules/imports.ts
```diff
@@ -0,0 +1,90 @@
+import {
```
laktek, line 1: @sweatybridge another scenario we might have to consider is dynamic imports referencing files in the same directory as entrypoint.

Currently this code works when deployed, because we include all source files in a function directory. Parsing dynamic imports would be hard, if the paths are not staticly analyzable (eg: `await import(getFilename())`).

```
await import("./foo.ts")
Deno.serve(() => new Response("ok"))
```

I think in the release we can call it out as a breaking change and suggest:
1. Change local dynamic imports to hoisted imports.
2. Add those files to `staticPatterns` in config.yaml

internal/functions/deploy/testdata/modules/imports.ts
```diff
@@ -0,0 +1,90 @@
...
+    PlaneBufferGeometry,
+    OctahedronGeometry,
+    TorusBufferGeometry
+} from '../geometries/Geometries.js';
+
+import {
+    PlaneBufferGeometry,
+    OctahedronGeometry,
+    TorusBufferGeometry
+} from '../geometries/Geometries.js'
```
laktek, line 44: are we also allowing importing files outside `supabase/functions` directory? eg: `../../../geometries/Geometries.js`? If so, would be good to add a test scenario for it too.

internal/functions/deploy/testdata/modules/imports.ts
sweatybridge: Yup, good shout. But is this case supported by `edge-runtime bundle` command? `getFilename()` could come from a user request for example.

In either case, I plan to keep `--use-docker` as the default flag for the time being.

laktek: Currently, it's implicitly supported because we bundle all source files within a function's directory.

internal/functions/deploy/testdata/modules/imports.ts
sweatybridge: Not possible at the moment because the workdir is always set to parent of supabase directory. So that's the top level it can reach. But once we fix the workdir issue, triple `..` should also work.

laktek: Can users continue to add files outside of supabase directory to import map and have those be included?

Issue #3112, opened 2025-02-06
Supabase local client-side error viewing postgres logs


sproctor:
**Describe the bug**
When clicking on a log entry in the postgres log viewer, an error screen appears

**To Reproduce**
1. Go to Logs > Postgres
2. Click on any entry that is not a warning or error
3. See error: "Error: c.metadata is null"

2.11.0 gave me a 404 when trying to view the logs, so I couldn't test it.

**Expected behavior**
The details of the log entry should show on the side panel

**System information**
 - Version of OS: Ubuntu 24.04
 - Version of CLI: 2.6.8, 2.9.6, and 2.9.7
 - Version of Docker: 27.5.1, build 9f9e405
 - Versions of services:

```
        SERVICE IMAGE      │      LOCAL       │   LINKED
  ─────────────────────────┼──────────────────┼─────────────
    supabase/postgres      │ 15.8.1.030       │ 15.8.1.030
    supabase/gotrue        │ v2.168.0         │ v2.168.0
    postgrest/postgrest    │ v12.2.3          │ v12.2.3
    supabase/realtime      │ v2.34.7          │ -
    supabase/storage-api   │ v1.17.1          │ v1.17.1
    supabase/edge-runtime  │ v1.66.5          │ -
    supabase/studio        │ 20250113-83c9420 │ -
    supabase/postgres-meta │ v0.84.2          │ -
    supabase/logflare      │ 1.4.0            │ -
    supabase/supavisor     │ 1.1.56           │ -
```

**Additional context**
If applicable, add any other context about the problem here.
 - Browser Chrome and Firefox

sproctor:
Local logs on 2.12.0 are now completely broken for me.

I get this error message when trying to view any of them:
```
{
  "code": 404,
[...]

Issue #3086, opened 2025-01-30
Studio: 502 error when trying to view cron or edge functions logs on localhost


stefan-girlich:
**Describe the bug**
Error 502 is shown when trying to view logs for cron or edge functions. All other types of logs work as expected.

**To Reproduce**
1. Follow steps from thread on [similar issue](https://github.com/supabase/supabase/issues/26785#issuecomment-2255463012)
2. Open Supabase Studio in browser
3. Go to log section "Cron" or "Edge Functions"

```json
{
  "code": 502,
  "errors": [],
  "message": "Something went wrong! Unknown error. If this continues please contact support.",
  "status": "UNKNOWN"
}
```

**Expected behavior**
Logs are shown.

**System information**
Rerun the failing command with `--create-ticket` flag.
 - Ticket ID: -
 - Version of OS: MacOS 14.5
 - Version of CLI:  2.9.6 or 2.9.7 (beta)
 - Version of Docker: 27.4.0, build bde2b89
 - Versions of services: (for CLI v2.9.6) 

**Additional context**
If applicable, add any other context about the problem here.
 - Browser: Version 1.74.50 Chromium: 132.0.6834.111

PR #3130, opened 2025-02-12
Revert "fix(studio): Bump studio version to 20250130-b048539"

sweatybridge:
Reverts supabase/cli#3085

### DIFF
--- a/internal/start/start.go
+++ b/internal/start/start.go
@@ -992,7 +992,7 @@ EOF
 					"HOSTNAME=0.0.0.0",
 				},
 				Healthcheck: &container.HealthConfig{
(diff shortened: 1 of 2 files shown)


silentworks:
@sweatybridge what was the issue that caused this to be reverted? there was a fix for not being able to delete users from the Dashboard in the image that you reverted. It would be useful to know when this fix will be re-instated as there is a huge thread about it in the issues on the supabase/supabase repo.

I don't want to link it here so this PR doesn't get flooded with folks complaining about this.

sweatybridge:
Quite a few issues came up since the last studio update in cli.

- https://github.com/supabase/cli/issues/3129
- https://github.com/supabase/cli/issues/3128
- https://github.com/supabase/cli/issues/3112
- https://github.com/supabase/cli/issues/3112
- https://github.com/supabase/cli/issues/3086

I think most of those are fixed now so it should be ok to update the beta channel again.

PR #3152, opened 2025-02-17
fix: verify jwt by default when deploying without config

sweatybridge:

Bug fix

## What is the new behavior?

## Additional context


avallete (approved):
**question**

Should that be released in a major with a breaking change warning ? Wondering if that might break some user deployment scripts by making some public functions unaccessible.

sweatybridge:
According to our docs https://supabase.com/docs/guides/functions/deploy#deploy-your-edge-functions

> By default, Edge Functions require a valid JWT in the authorization header. If you want to use Edge Functions without Authorization checks (commonly used for Stripe webhooks), you can pass the --no-verify-jwt flag when deploying your Edge Functions.

So this is more of a bug than breaking change.

May be it will break a few. But I'd rather let users know they should declare `verify_jwt = false` in either config or cli flag than finding out afterwards that their function was accidentally exposed.

PR #3156, opened 2025-02-17
Fix/cross platform tests (compatible with windows and unix)

theRealThomasDavidson:

Bug fix (cross-platform test compatibility)

### Current Behavior
Tests fail on Windows due to:
1. Path comparisons using forward slashes fail against Windows backslashes
   ```go
   // Fails on Windows:
   assert.Equal(t, cwd+"/common", resolved.Imports["root"])
   ```

2. Line ending comparisons fail due to CRLF vs LF differences
   ```go
   // Fails when comparing files with different line endings:
   assert.Equal(t, string(expected), string(actual))
   ```

### New Behavior
1. Path comparisons work cross-platform using filepath.ToSlash:
   ```go
   assert.Equal(t, filepath.ToSlash(cwd+"/common"), 
                  filepath.ToSlash(resolved.Imports["root"]))
   ```

2. Line endings are normalized before comparison:
   ```go
   func normalizeLineEndings(s string) string {
       s = strings.ReplaceAll(s, "\r\n", "\n")
       s = strings.ReplaceAll(s, "\r", "\n")
       return s
   }
   ```

## Additional context

###Goal/Purpose

- Tests should run consistently regardless of platform
- No functional changes to production code except URL path handling
- Changes focused on test reliability and maintainability


theRealThomasDavidson:
shit i didn't realize this had the change to temp folder i had tried earlier. please reject this one i will have one with just the right files and will fix the one lint error.

PR #3201, opened 2025-02-24
fix: support main.ts as an entrypoint

laktek:

Adds support for `main.ts` to be treated as a valid entrypoint along with `index.ts`. This allows functions generated with `deno init` to work without having to change the filename.

### DIFF
--- a/internal/functions/deploy/deploy.go
+++ b/internal/functions/deploy/deploy.go
@@ -54,7 +54,7 @@ func Run(ctx context.Context, slugs []string, useDocker bool, noVerifyJWT *bool,
 }
 
 func GetFunctionSlugs(fsys afero.Fs) (slugs []string, err error) {
-	pattern := filepath.Join(utils.FunctionsDir, "*", "index.ts")
+	pattern := filepath.Join(utils.FunctionsDir, "*", "*.ts")
 	paths, err := afero.Glob(fsys, pattern)
 	if err != nil {
 		return nil, errors.Errorf("failed to glob function slugs: %w", err)
@@ -97,7 +97,15 @@ func GetFunctionConfig(slugs []string, importMapPath string, noVerifyJWT *bool, 
 		// Precedence order: flag > config > fallback
 		functionDir := filepath.Join(utils.FunctionsDir, name)
 		if len(function.Entrypoint) == 0 {
-			function.Entrypoint = filepath.Join(functionDir, "index.ts")
+			indexEntrypoint := filepath.Join(functionDir, "index.ts")
+			mainEntrypoint := filepath.Join(functionDir, "main.ts")
+			if _, err := fsys.Stat(indexEntrypoint); err == nil {
+				function.Entrypoint = indexEntrypoint
+			} else if _, err := fsys.Stat(mainEntrypoint); err == nil {
+				function.Entrypoint = mainEntrypoint
+			} else {
+				return nil, errors.Errorf("Cannot find a valid entrypoint file (index.ts or main.ts) for the function. Set the custom entrypoint path in config.toml")
+			}
 		}
 		if len(importMapPath) > 0 {
 			function.ImportMap = importMapPath
--- a/internal/functions/deploy/deploy.go
+++ b/internal/functions/deploy/deploy.go
@@ -104,7 +104,7 @@ func GetFunctionConfig(slugs []string, importMapPath string, noVerifyJWT *bool, 
 			} else if _, err := fsys.Stat(mainEntrypoint); err == nil {
 				function.Entrypoint = mainEntrypoint
 			} else {
-				return nil, errors.Errorf("Cannot find a valid entrypoint file (index.ts or main.ts) for the function. Set the custom entrypoint path in config.toml")
+				return nil, errors.Errorf("Cannot find a valid entrypoint file (index.ts or main.ts) for the '%s' function. Set the custom entrypoint path in config.toml", name)
 			}
 		}
 		if len(importMapPath) > 0 {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -654,7 +654,13 @@ func (c *baseConfig) resolve(builder pathBuilder, fsys fs.FS) error {
 	// Resolve functions config
 	for slug, function := range c.Functions {
 		if len(function.Entrypoint) == 0 {
-			function.Entrypoint = filepath.Join(builder.FunctionsDir, slug, "index.ts")
+			indexEntrypoint := filepath.Join(builder.FunctionsDir, slug, "index.ts")
+			mainEntrypoint := filepath.Join(builder.FunctionsDir, slug, "main.ts")
+			if _, err := fs.Stat(fsys, indexEntrypoint); err == nil {
+				function.Entrypoint = indexEntrypoint
+			} else if _, err := fs.Stat(fsys, mainEntrypoint); err == nil {
+				function.Entrypoint = mainEntrypoint
+			}
 		} else if !filepath.IsAbs(function.Entrypoint) {
 			// Append supabase/ because paths in configs are specified relative to config.toml
 			function.Entrypoint = filepath.Join(builder.SupabaseDirPath, function.Entrypoint)
--- a/internal/functions/deploy/deploy_test.go
+++ b/internal/functions/deploy/deploy_test.go
@@ -31,6 +31,10 @@ func TestDeployCommand(t *testing.T) {
 		// Setup in-memory fs
 		fsys := afero.NewMemMapFs()
 		require.NoError(t, utils.WriteConfig(fsys, false))
+
+		require.NoError(t, afero.WriteFile(fsys, filepath.Join(utils.FunctionsDir, slug, "index.ts"), []byte{}, 0644))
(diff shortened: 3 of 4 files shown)


internal/functions/deploy/deploy.go
```diff
@@ -97,7 +97,15 @@ func GetFunctionConfig(slugs []string, importMapPath string, noVerifyJWT *bool,
...
-			function.Entrypoint = filepath.Join(functionDir, "index.ts")
+			indexEntrypoint := filepath.Join(functionDir, "index.ts")
+			mainEntrypoint := filepath.Join(functionDir, "main.ts")
+			if _, err := fsys.Stat(indexEntrypoint); err == nil {
+				function.Entrypoint = indexEntrypoint
+			} else if _, err := fsys.Stat(mainEntrypoint); err == nil {
+				function.Entrypoint = mainEntrypoint
+			} else {
+				return nil, errors.Errorf("Cannot find a valid entrypoint file (index.ts or main.ts) for the function. Set the custom entrypoint path in config.toml")
+			}
```
avallete, line 108: **nitpick**

I think we might want to extract that in a dedicated function as it seems to repeat here:

https://github.com/supabase/cli/blob/0b8c968b470aa0aaf13874f01b618e792bfeecac/pkg/config/config.go#L656-L658

Also we probably need to change it there as well.

laktek: ah good catch!

laktek: Technically, it doesn't matter though. If the `index.ts` file doesn't exist, `main.ts` will be tried in `GetFunctionConfig`. We probably don't need to set a default in config. The same applies to import map paths, which we seem to repeat in config.go and `GetFunctionConfig`. 

Is there another place where resolved config is used without calling `GetFunctionConfig`? If that's the case, `GetFunctionConfig` should either be in config or in `pkg/utility` instead of the `deploy` package.

cc: @sweatybridge

sweatybridge: `GetFunctionConfig` is mostly here for cli flag overrides, default import map, and auto detecting function slugs. They are not moved to `pkg/config` because the logic is either unnecessary or fragile for config based deployment.

To support `main.ts` as entrypoint, we should be updating `pkg/config` instead so it works for both branching and cli.

internal/functions/deploy/deploy.go
```diff
@@ -54,7 +54,7 @@ func Run(ctx context.Context, slugs []string, useDocker bool, noVerifyJWT *bool,
 }
 
 func GetFunctionSlugs(fsys afero.Fs) (slugs []string, err error) {
-	pattern := filepath.Join(utils.FunctionsDir, "*", "index.ts")
+	pattern := filepath.Join(utils.FunctionsDir, "*", "*.ts")
```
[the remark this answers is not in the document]
laktek: hmm..don't we already have a check to exclude shared modules? https://github.com/supabase/cli/blob/cf4d8032f86d1df7f213762d53278d7f12f099e3/internal/functions/deploy/deploy.go#L64 (we already do a wildcard match on the directory, I don't think this makes it worse than that)

If there's a Glob to explicitly match both `index.ts` and `main.ts`, I'm happy to change that (I couldn't get it to work). 

The problem with having `[functions.some-slug]` in config.toml is users have to add it manually. And if they do `deno init --serve supabase/functions/my-func` and then try to access it wouldn't work. If we can auto-populate the valid function slugs, it would improve the DX a lot.

internal/functions/deploy/deploy.go
laktek: For now, I followed the same as import map paths and copied over the same code to `pkg/config`. https://github.com/supabase/cli/pull/3201/commits/5af8b8a2e2846eb7f4edb834c3d593a584075587

I couldn't really figure out a clean way to abstract this logic. We should do a refactor for both entrypoint and import maps in a separate PR. Currently, config only handles explicitly defined function slugs, we should make it implicit (ie. go through function directories and then config.toml). Then we can resolve the function config once and reuse everywhere.

internal/functions/deploy/deploy.go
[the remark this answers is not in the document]
sweatybridge: > (we already do a wildcard match on the directory, I don't think this makes it worse than that)

The existing wildcard match does NOT eliminate shared modules like `functions/_shared/index.ts`. It will deploy a function named `_shared`.

Unfortunately this isn't something we can easily change now without risking breaking user's CI/CD pipeline. Because users may very well have a function named `_something` and expects it to be deployed.

IMO, the root of this evil is due to the implicit behaviour of finding entrypoints by glob patterns. It was ok for small scale local development, but not acceptable for CI/CD when you have 100s of functions to manage.

> And if they do deno init --serve supabase/functions/my-func and then try to access the function it wouldn't work.

I suspect you've misunderstood my suggestion. I'm saying to parse `main.ts` as fallback entrypoint in `pkg/config`, not to completely ignore it. That way, both `deno init --serve` and `functions deploy` work.

internal/functions/deploy/deploy.go
sweatybridge: I'd say keep only your changes to `pkg/config.go` and remove all changes to `deploy.go`.

If users run `functions deploy` without declaring anything in config.toml, only `index.ts` is going to be used, exactly the same as today. But if they declare `[functions.some-slug]`, then `main.ts` can be used as fallback.

This way, we avoid introducing more implicit behaviour which makes maintenance difficult. And we nudge users towards best practices for CI/CD by declaring their functions in config.toml.

laktek: > But if they declare [functions.some-slug], then main.ts can be used as a fallback.

Sadly, this would mean users must do another step after `deno init --serve` before they can test their functions. If users are going to do another step, it doesn't even matter to add support to `main.ts`. We can simply ask them to rename the file to `index.ts` instead.

The ideal DX I'm aiming for:

* deno init --serve supabase/functions/hello-world (or copy an existing project to `supabase/functions`)
* npx supabase function serve
* visit http://localhost:54321/functions/v1/hello-world and see the response

NOT having to manually edit `config.toml` or rename files/directories is important so users can get functions to work on the first attempt.

sweatybridge: Another option is allow specifying the full entrypoint path when deploying, exactly like deno.

pkg/config/config.go
```diff
@@ -654,7 +654,13 @@ func (c *baseConfig) resolve(builder pathBuilder, fsys fs.FS) error {
...
 	for slug, function := range c.Functions {
 		if len(function.Entrypoint) == 0 {
-			function.Entrypoint = filepath.Join(builder.FunctionsDir, slug, "index.ts")
+			indexEntrypoint := filepath.Join(builder.FunctionsDir, slug, "index.ts")
+			mainEntrypoint := filepath.Join(builder.FunctionsDir, slug, "main.ts")
+			if _, err := fs.Stat(fsys, indexEntrypoint); err == nil {
+				function.Entrypoint = indexEntrypoint
+			} else if _, err := fs.Stat(fsys, mainEntrypoint); err == nil {
+				function.Entrypoint = mainEntrypoint
+			}
```
sweatybridge, line 663:
```suggestion
			function.Entrypoint = filepath.Join(builder.FunctionsDir, slug, "index.ts")
			if _, err := fs.Stat(fsys, function.Entrypoint); errors.Is(err, os.ErrNotExist) {
				mainEntrypoint := filepath.Join(builder.FunctionsDir, slug, "main.ts")
				if _, err := fs.Stat(fsys, mainEntrypoint); err == nil {
					function.Entrypoint = mainEntrypoint
				}
			}
```

We should still pass down default entrypoint path.

internal/functions/deploy/deploy.go
```diff
@@ -97,7 +97,15 @@ func GetFunctionConfig(slugs []string, importMapPath string, noVerifyJWT *bool,
...
-			function.Entrypoint = filepath.Join(functionDir, "index.ts")
+			indexEntrypoint := filepath.Join(functionDir, "index.ts")
+			mainEntrypoint := filepath.Join(functionDir, "main.ts")
+			if _, err := fsys.Stat(indexEntrypoint); err == nil {
+				function.Entrypoint = indexEntrypoint
+			} else if _, err := fsys.Stat(mainEntrypoint); err == nil {
+				function.Entrypoint = mainEntrypoint
+			} else {
+				return nil, errors.Errorf("Cannot find a valid entrypoint file (index.ts or main.ts) for the '%s' function. Set the custom entrypoint path in config.toml", name)
+			}
```
sweatybridge, line 108:
```suggestion
			function.Entrypoint = filepath.Join(functionDir, "index.ts")
			if _, err := fs.Stat(fsys, function.Entrypoint); errors.Is(err, os.ErrNotExist) {
				mainEntrypoint := filepath.Join(functionDir, "main.ts")
				if _, err := fs.Stat(fsys, mainEntrypoint); err == nil {
					function.Entrypoint = mainEntrypoint
				}
			}
```

ditto

One way to reduce duplication is to add a `function.ResolvePaths` method. I can work on that separately.

laktek:
Sync'd offline to come up with an alternate workflow. Closing this PR.

PR #3206, opened 2025-02-25
feat(config): add rate limit configuration for emails in toml

avallete:
Allow to configure rate limits via the `config.toml`

Comes with the docs PR here: https://github.com/supabase/supabase/pull/33830


pkg/config/templates/config.toml
```diff
@@ -127,6 +127,15 @@ minimum_password_length = 6
...
 
+# Rate limit for unauthenticated requests per minute
+rate_limit_anonymous_users = 30
+# Rate limit for token refresh requests per minute
+rate_limit_token_refresh = 150
+# Rate limit for OTP generation requests per minute
+rate_limit_otp = 30
+# Rate limit for verification requests per minute
+rate_limit_verify = 30
+
```
sweatybridge, line 138: I think it's worth moving them to a new struct so it aligns with our [dashboard](https://supabase.com/dashboard/project/_/auth/rate-limits). For eg.

```toml
[auth.rate_limits]
# Number of emails that can be sent per hour. Only available with custom SMTP config.
emails_sent = 2
# Number of SMS messages that can be sent per hour.
sms_sent = 30
# Number of sessions that can be refreshed in a 5 minute interval per IP address.
token_refreshes = 150
# Number of OTP / Magic link verifications that can be made in a 5 minute interval per IP address.
token_verifications = 30
# Number of anonymous sign-ins that can be made per hour per IP address.
anonymous_users = 30
# Number of sign up and sign-in requests that can be made in a 5 minute interval per IP address (excludes anonymous users).
sign_in_sign_ups = 30
```

Btw it's really confusing to me why `rate_limit_otp` maps to number of sign-in and sign-ups https://github.com/supabase/supabase/blob/master/apps/studio/components/interfaces/Auth/RateLimits/RateLimits.tsx#L417

avallete: Regrouped it into the same stuct, had to update a bunch of test snapshot because of it tough 😅

pkg/config/templates/config.toml
```diff
@@ -127,6 +127,21 @@ minimum_password_length = 6
 # are: `letters_digits`, `lower_upper_letters_digits`, `lower_upper_letters_digits_symbols`
 password_requirements = ""
 
+[auth.rate_limit]
+# Number of emails that can be sent per hour. Requires auth.email.smtp to be enabled.
+# email_sent = 2
+# Number of SMS messages that can be sent per hour. Requires auth.sms to be enabled.
+# sms_sent = 30
```
sweatybridge, line 134:
```suggestion
email_sent = 2
# Number of SMS messages that can be sent per hour. Requires auth.sms to be enabled.
sms_sent = 30
```

It's a bit annoying that we can't leave these uncommented for consistency. Can update api to support configuring values < shared quota?

avallete: I'm not sure how much of a side effect this could have looking at those lines on API side:

https://github.com/supabase/infrastructure/blob/43c6d1a97c73eda0c171990ecd4e4ede5bafd185/api/src/common/projects/project-gotrue-config.service.ts#L359-L366

I think it might break some other behaviors.

pkg/config/auth.go
```diff
@@ -107,6 +108,15 @@ type (
 		Cognito  tpaCognito  `toml:"aws_cognito"`
 	}
 
+	rateLimit struct {
+		AnonymousUsers     *uint `toml:"anonymous_users"`
+		TokenRefresh       *uint `toml:"token_refresh"`
+		SignInSignUps      *uint `toml:"sign_in_sign_ups"`
+		TokenVerifications *uint `toml:"token_verifications"`
+		EmailSent          *uint `toml:"email_sent"`
+		SmsSent            *uint `toml:"sms_sent"`
```
sweatybridge, line 117: Any objection to make these non-pointers?

sweatybridge (approved):

PR #3212, opened 2025-02-26
fix: append config.toml when creating new function

sweatybridge:

Bug fix

## What is the new behavior?

Make it easier to initialise new functions by appending to config.toml.

## Additional context


avallete (approved):
LGTM

PR #3221, opened 2025-02-27
feat: output standard env vars from branches get

sweatybridge:

feature

## What is the new behavior?

```
$ supabase --experimental branches get --output json
Selected branch ID: 709a7cc3-2cfd-417d-8828-76c8bbc58a1e
{
  "POSTGRES_URL": "postgresql://...@aws-0-ap-southeast-1.pooler.supabase.com:6543/postgres?connect_timeout=10",
  "POSTGRES_URL_NON_POOLING": "postgresql://...@db.kpojrjxbvujjhamjpcux.supabase.co:5432/?connect_timeout=10",
  "SUPABASE_ANON_KEY": "...",
  "SUPABASE_SERVICE_ROLE_KEY": "...",
  "SUPABASE_URL": "https://kpojrjxbvujjhamjpcux.supabase.co"
}
```

## Additional context

### DIFF
--- a/internal/branches/get/get.go
+++ b/internal/branches/get/get.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"strings"
 
 	"github.com/go-errors/errors"
 	"github.com/jackc/pgconn"
@@ -102,15 +101,9 @@ func toStandardEnvs(detail api.BranchDetailResponse, pooler api.SupavisorConfigR
 	} else {
 		config.Password = direct.Password
 	}
-	envs := map[string]string{
-		"POSTGRES_URL":             utils.ToPostgresURL(*config),
-		"POSTGRES_URL_NON_POOLING": utils.ToPostgresURL(direct),
-		"SUPABASE_URL":             "https://" + utils.GetSupabaseHost(detail.Ref),
-	}
-	for _, entry := range keys {
-		name := strings.ToUpper(entry.Name)
-		key := fmt.Sprintf("SUPABASE_%s_KEY", name)
-		envs[key] = entry.ApiKey
-	}
+	envs := apiKeys.ToEnv(keys)
+	envs["POSTGRES_URL"] = utils.ToPostgresURL(*config)
+	envs["POSTGRES_URL_NON_POOLING"] = utils.ToPostgresURL(direct)
+	envs["SUPABASE_URL"] = "https://" + utils.GetSupabaseHost(detail.Ref)
 	return envs
 }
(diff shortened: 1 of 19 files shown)


internal/branches/get/get.go
```diff
@@ -32,35 +69,48 @@ func Run(ctx context.Context, branchId string, fsys afero.Fs) error {
...
+	config, err := utils.ParsePoolerURL(pooler.ConnectionString)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, utils.Yellow("WARNING:"), err)
+		config = &direct
+	} else {
+		config.Password = direct.Password
+	}
+	envs := map[string]string{
+		"POSTGRES_URL":             utils.ToPostgresURL(*config),
+		"POSTGRES_URL_NON_POOLING": utils.ToPostgresURL(direct),
```
jgoux, line 107: Should we follow the naming we use publicly in the docs? `DIRECT_DATABASE_URL` and `POOLER_DATABASE_URL`? And if we want to default to be the pooler we can also alias it with `DATABASE_URL`?

pkg/function/batch.go
```diff
@@ -34,14 +34,15 @@ func (s *EdgeRuntimeAPI) UpsertFunctions(ctx context.Context, functionConfig con
...
+OUTER:
 	for slug, function := range functionConfig {
 		if !function.Enabled {
 			fmt.Fprintln(os.Stderr, "Skipped deploying Function:", slug)
 			continue
 		}
 		for _, keep := range filter {
 			if !keep(slug) {
-				continue
+				continue OUTER
```
jgoux, line 45: Labeled statement, I don't see that everyday, nice. 😄

jgoux (approved):
LGTM, just a tiny comment on the env variable names. 👍

internal/branches/get/get.go
jgoux: In fact if we want to keep it namespaced we can also invert the two first words to have `DATABASE` first like `DATABASE_DIRECT_URL` and `DATABASE_POOLER_URL`. 😄

sweatybridge: Our dashboard is bit funny because both direct and database url point to the pooler.

```
# Connect to Supabase via connection pooling with Supavisor.
DATABASE_URL="postgresql://...@aws-0-ap-southeast-1.pooler.supabase.green:6543/postgres?pgbouncer=true"
# Direct connection to the database. Used for migrations.
DIRECT_URL="postgresql://...@aws-0-ap-southeast-1.pooler.supabase.green:5432/postgres"
```

The current convention is following [bootstrap command](https://github.com/supabase/cli/blob/develop/internal/bootstrap/bootstrap.go#L201) which seems to be more backwards compatible.

PR #3229, opened 2025-02-28
fix: remove dependency on global viper

sweatybridge:

Bug fix https://github.com/orgs/supabase/discussions/18937#discussioncomment-12327922

## What is the new behavior?

- Do not rely on global viper state in shared package.
- Fixes plaintext secrets in `.env.local` not loading.

## Additional context


jgoux (approved):
LGTM

Issue #3146, opened 2025-02-14
type generation not including my specified schema for helper types


GeekyMonkey:
**Describe the bug**
When I generate types, it does the basic processing of the schema I specified, but the helper types are only generated for the "public" schema, which is blank in my case.

**To Reproduce**
- Have a database with no tables in the public schema
- Create a "robo" schema with one table
- Run the gentypes command ```supabase gen types typescript --schema robo --local > supabase/dbTypes.ts```

**Expected behavior**
in my dbTypes.ts I should have helper types based on the specified "robo" schema.

**Screenshots**

![Image](https://github.com/user-attachments/assets/6da76b2a-ff4a-4a2a-a7c5-ece402ffd387)

The only schema in the database is "robo", but the only helpers are created from the "public" schema

**System information**
 - Version of OS: Windows 11
 - Version of CLI: 2.12.1
 - Version of Docker: [e.g. v25.0.3]
 - Versions of services: 
       SERVICE IMAGE      │      LOCAL       │   LINKED
  ─────────────────────────┼──────────────────┼─────────────
    supabase/postgres      │ 15.6.1.121       │ 15.6.1.121
    supabase/gotrue        │ v2.168.0         │ v2.168.0
    postgrest/postgrest    │ v12.2.3          │ v12.2.3
    supabase/realtime      │ v2.34.7          │ -
    supabase/storage-api   │ v1.17.1          │ v1.17.1
    supabase/edge-runtime  │ v1.67.0          │ -
    supabase/studio        │ 20250113-83c9420 │ -
    supabase/postgres-meta │ v0.84.2          │ -
    supabase/logflare      │ 1.4.0            │ -
    supabase/supavisor     │ 1.1.56           │ -

**Additional context**
If applicable, add any other context about the problem here.
 - Version of supabase-js: 2.48.1
 - Version of Node.js: 20.18.2

GeekyMonkey:
See also https://github.com/supabase/cli/issues/3061

Issue #3061, opened 2025-01-21
Support for multiple schemas in type generating


jljl1337:
**Is your feature request related to a problem? Please describe.**
I am writing a web application using Next.js and typescript. I have a `private` schema for my backend with a table `note`, I ran the `supabase gen types` command with the `-s` flag, but when i want to use the table type in my logic, it was resolved as `unknown` type:

<img width="248" alt="Image" src="https://github.com/user-attachments/assets/c013da01-6acb-468e-8051-54ba0259dc98" />

When I try to debug, I found this line in the generated file:

```
type PublicSchema = Database[Extract<keyof Database, "public">];
```

When I change `"public"` to `"private"`, the table type can be referred.

It seems that only the tables from the `public` schema can be referred, it would be much better if tables from all schemas can be referred, especially when the schema(s) is/are specified when generating the types.

**Describe the solution you'd like**
The correct table (and enum) type of all the generated schemas should be able to be referred, especially when the schema(s) is/are specified when generating the types.

**Describe alternatives you've considered**
Mentioned the expected outcome and/or any alternate solution in the document of the CLI

**Additional context**
N/a

GeekyMonkey:
I also reported this on https://github.com/supabase/cli/issues/3146

PR #3243, opened 2025-03-03
fix: automatically set default schema env var for type generation

bug-author:

This PR lets the CLI automatically detect when exactly one non‑public schema is provided and sets the environment variable `PG_META_GENERATE_TYPES_DEFAULT_SCHEMA` accordingly.

Currently, when running `supabase gen types`, the CLI passes the schema flags directly to `postgres‑meta` without automatically setting a default schema. This results in the generated TypeScript file always defaulting to the "public" alias:

```ts
export type PublicSchema = Database[Extract<keyof Database, "public">]
```

Even if a non‑public schema is specified (for example, `--schema robo`), helper types (like `Tables<"test_table">`) incorrectly reference the public schema. 

Relevant issues:
#3061 
#3146 

## What is the new behavior?

The CLI now automatically checks if exactly one schema is provided (and if that schema is not "public"). If so, it sets the environment variable `PG_META_GENERATE_TYPES_DEFAULT_SCHEMA` to that schema's name. This allows `postgres‑meta` to generate a default alias based on the provided schema (for example, for "robo", the generated file will include):

```ts
export type DefaultSchema = Database[Extract<keyof Database, "robo">]
export type DefaultSchemaOrPublic = DefaultSchema
```

In single‑schema mode, shorthand type references (like `Tables<"test_table">`) correctly resolve to the non‑public schema. In multi‑schema scenarios, the variable is not set, and users must explicitly specify the schema for non‑public tables.

## Additional context

Corresponding [postgres-meta PR](https://github.com/supabase/postgres-meta/pull/904)


internal/gen/types/types.go
```diff
@@ -35,6 +35,11 @@ func Run(ctx context.Context, projectId string, dbConfig pgconn.Config, lang str
 	}
 	included := strings.Join(schemas, ",")
 
+	var defaultSchemaEnv string
+	if len(schemas) == 1 && schemas[0] != "public" {
+		defaultSchemaEnv = schemas[0]
+	}
```
sweatybridge, line 41: We should add a new flag so it doesn't break existing users that rely on the generated default schema.

bug-author: @sweatybridge Thanks. I've implemented a `set-default` flag for backwards compatibility that makes the behavior now explicit and opt-in. The new behavior is now:

```sh
supabase gen types --schema private --set-default
```

bug-author:
I've added 4 tests as well to test the core feature, backward compatibility and edge cases.

PR #3246, opened 2025-03-04
feat: support setting function secrets from config

sweatybridge:

feature https://github.com/orgs/supabase/discussions/18937#discussioncomment-12342719

## What is the new behavior?

```toml
[edge_runtime.secrets]
MY_SECRET = "env(some_key)"
```

## Additional context

TODO:
- [x] find an alternative to preserve case sensitivity

### DIFF
--- a/pkg/config/testdata/config.toml
+++ b/pkg/config/testdata/config.toml
@@ -260,6 +260,9 @@ enabled = true
 policy = "per_worker"
 inspector_port = 8083
 
+[edge_runtime.secrets]
+test_key = "test_value"
+
 [analytics]
 enabled = true
 port = 54327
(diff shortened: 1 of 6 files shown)


pkg/config/testdata/config.toml
```diff
@@ -260,6 +260,9 @@ enabled = true
 policy = "per_worker"
 inspector_port = 8083
 
+[edge_runtime.secrets]
+test_key = "test_value"
+
```
avallete, line 265: **question**

Just to make sure I get it, in production this would be something like:

```
test_key = "encrypted:xxxxxXXXXX"
```

And decrypted by the branching executor to deploy the "clear secret" to the branch ?

Followup question, I guess when merging, this'll also deploy the secret to the production branch ?

avallete (approved):

pkg/config/testdata/config.toml
sweatybridge: I think the behaviour should be to set these secrets if they are declared, before functions are deployed.

So yes, it applies to both dev and production branches if declared in base config. But users can override it for specific remotes using either `.env.preview` (all dev) or `[remotes.production.edge_runtime.secrets]`.

PR #3273, opened 2025-03-09
fix: clerk tpa domain pattern works for non-production mode domains

hf:
Missing `+` wasn't caught.

### DIFF
--- a/pkg/config/templates/config.toml
+++ b/pkg/config/templates/config.toml
@@ -273,7 +273,7 @@ enabled = false
 # Use Clerk as a third-party provider alongside Supabase Auth.
 [auth.third_party.clerk]
 enabled = false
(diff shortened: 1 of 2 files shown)


sweatybridge (approved):

pkg/config/templates/config.toml
```diff
@@ -273,7 +273,7 @@ enabled = false
 # Use Clerk as a third-party provider alongside Supabase Auth.
 [auth.third_party.clerk]
 enabled = false
-# domain = "example.clerk.accounts.dev"
+# domain = "example.clerk.accounts.dev" obtain from https://clerk.com/setup/supabase
```
sweatybridge, line 276:
```suggestion
# Obtain from https://clerk.com/setup/supabase
# domain = "example.clerk.accounts.dev"
```

Issue #3353, opened 2025-03-16
over_email_send_rate_limit -running local dev - only 2 emails sent in past hour


magick93:
# Bug report
this is clearly a bug that has been introduced in a recent upgrade. 

I've been running this project locally for over a year now and have never had this issue. I decided to update the supabase cli, and now get this error. 

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

When performing a `supabase.auth.signUp()` invocation I now get:

```
supabase.error: {
      "__isAuthError": true,
      "name": "AuthApiError",
      "status": 429,
      "code": "over_email_send_rate_limit"
    }
```

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. run `supabase start` - this issue is occurring locally, not self hosted, not hosted by supabase.com
1. In Typescript / Javascript, peform a `signUp()` action, eg:

```ts
return await supabase.auth.signUp({
      email: form.data.email,
      password: form.data.password,
      options: {
        emailRedirectTo: `${url.origin}/auth/callback?next=/auth/onboarding`,
        data: userData(),
      }
    });
```
3. This is now fails with error

```js
supabase.error: {
      "__isAuthError": true,
      "name": "AuthApiError",
      "status": 429,
      "code": "over_email_send_rate_limit"
    }
```

## System information

- OS: Ubuntu 24.04.2 LTS
-  `"@supabase/ssr": "^0.5.2"`
-  `"@supabase/supabase-js": "^2.47.10"`

## Additional context

I notice that the dev email appliation has also changed. I discovered this bug after running cypress tests that usually pass without any issue. The first failed. I noticed that the dev email client had changed. I adjusted the test, it still failed. And at this point it had sent 2 emails. 

And this is running completely locally, so I dont see why any email limit would even make sense. 

# config.toml

Here are the relevant sections of my local config.toml:

```toml

[inbucket]
enabled = true
# Port to use for the email testing server web interface.
port = 54324
# Uncomment to expose additional ports for testing user applications that send emails.
smtp_port = 54325
# pop3_port = 54326

[auth]
enabled = true
# The base URL of your website. Used as an allow-list for redirects and for constructing URLs used
# in emails.
site_url = "http://localhost:5173"
[...]

PR #3356, opened 2025-03-26
fix(config): rate limit emails only if custom smtp enabled

silentworks:

This is to fix the email rate limit issue on local development.

Emails are being rate limited even without `auth.email.smtp` being enabled.

## What is the new behavior?

Emails are only rate limited if `auth.email.smtp` is enabled.

## Additional context

Related to https://github.com/supabase/cli/issues/3353


silentworks:
I could have added in the rate limit for anonymous users and sms sent in this PR too but I'm not sure if these are handled inside of Supabase Auth (aka GoTrue) itself.

sweatybridge (approved):
This is good enough. Thanks again @silentworks !

PR #3387, opened 2025-04-03
feat(cli): setup postgres17 for local development

avallete:

- Introduce pg17 as a possible image for the local dev experience.

### DIFF
--- a/internal/utils/config.go
+++ b/internal/utils/config.go
@@ -130,11 +130,7 @@ func InitConfig(params InitParams, fsys afero.Fs) error {
 	c := config.NewConfig()
 	c.ProjectId = params.ProjectId
 	if params.UseOrioleDB {
-		if c.Db.MajorVersion > 15 {
-			c.Experimental.OrioleDBVersion = "17.0.1.57"
-		} else {
-			c.Experimental.OrioleDBVersion = "15.1.0.150"
-		}
+		c.Experimental.OrioleDBVersion = "15.1.0.150"
 	}
 	// Create config file
 	if err := MkdirIfNotExistFS(fsys, SupabaseDirPath); err != nil {
--- a/internal/utils/config.go
+++ b/internal/utils/config.go
@@ -130,7 +130,7 @@ func InitConfig(params InitParams, fsys afero.Fs) error {
 	c := config.NewConfig()
 	c.ProjectId = params.ProjectId
 	if params.UseOrioleDB {
-		c.Experimental.OrioleDBVersion = "15.1.0.150"
+		c.Experimental.OrioleDBVersion = c.Db.MajorVersion > 15 ? "17.0.1.57" : "15.1.0.150"
 	}
 	// Create config file
 	if err := MkdirIfNotExistFS(fsys, SupabaseDirPath); err != nil {
--- a/internal/utils/config.go
+++ b/internal/utils/config.go
@@ -130,7 +130,11 @@ func InitConfig(params InitParams, fsys afero.Fs) error {
 	c := config.NewConfig()
 	c.ProjectId = params.ProjectId
 	if params.UseOrioleDB {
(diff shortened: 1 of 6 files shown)


pkg/config/config.go
```diff
@@ -600,6 +600,9 @@ func (c *config) Load(path string, fsys fs.FS) error {
 		if strings.HasPrefix(string(version), "15.") && semver.Compare(string(version[3:]), "1.0.55") >= 0 {
 			c.Db.Image = replaceImageTag(Images.Pg15, string(version))
 		}
+		if strings.HasPrefix(string(version), "17.") && semver.Compare(string(version[3:]), "1.0.55") >= 0 {
+			c.Db.Image = replaceImageTag(Images.Pg17, string(version))
+		}
```
avallete, line 605: **note**

This part I'm not sure about what it does exactly, duplicated what we do for the v15, wonder if it should replace it instead 🤔

internal/utils/config.go
```diff
@@ -130,7 +130,11 @@ func InitConfig(params InitParams, fsys afero.Fs) error {
 	c := config.NewConfig()
 	c.ProjectId = params.ProjectId
 	if params.UseOrioleDB {
-		c.Experimental.OrioleDBVersion = "15.1.0.150"
+		if c.Db.MajorVersion > 15 {
+			c.Experimental.OrioleDBVersion = "17.0.1.57"
+		} else {
+			c.Experimental.OrioleDBVersion = "15.1.0.150"
+		}
```
sweatybridge, line 137: There are some incompatibilities with orioledb 17. It's probably better to do this change separately https://github.com/supabase/cli/compare/develop...oriole-17

pkg/config/config.go
sweatybridge:
```suggestion
```

It was checking the minor versions of pg15 to make sure we don't override it to anything below 15.1.0.55. We don't need this for 17.

pkg/config/config.go
```diff
@@ -729,7 +732,7 @@ func (c *config) Validate(fsys fs.FS) error {
 		c.Db.Image = pg13
 	case 14:
 		c.Db.Image = pg14
-	case 15:
+	case 15, 17:
```
sweatybridge, line 735: Let's add a new case for pg17 instead. No need to touch orioledb for the time being.

```
	case 17:
		c.Db.Image = pg17
```

pkg/config/constants.go
```diff
@@ -16,6 +16,7 @@ const (
 
 type images struct {
 	Pg15 string `mapstructure:"pg15"`
+	Pg17 string `mapstructure:"pg17"`
```
sweatybridge, line 19:
```suggestion
	Pg string `mapstructure:"pg"`
```

Dependabot only updates one tag of the same image. So let's just rename it as pg.

pkg/config/templates/Dockerfile
```diff
@@ -1,5 +1,6 @@
 # Exposed for updates by .github/dependabot.yml
 FROM supabase/postgres:15.8.1.060 AS pg15
+FROM supabase/postgres:17.4.1.004 AS pg17
```
sweatybridge, line 3:
```suggestion
FROM supabase/postgres:17.4.1.004 AS pg
```

pkg/config/constants.go
```diff
@@ -11,11 +11,12 @@ import (
 const (
 	pg13  = "supabase/postgres:13.3.0"
 	pg14  = "supabase/postgres:14.1.0.89"
+	pg15  = "supabase/postgres:15.8.1.069"
```
avallete, line 14: **note**

To rename `Pg15` and `Pg17` under the single `Pg` I still needed a way to address the specific `pg15` version in some case. I went like for `pg13` and `pg14` and pointed to the latest available version at the time. Not sure if that's the way to go though.

sweatybridge (approved):

PR #3431, opened 2025-04-13
Optimize config, entrypoint, and Deno configurations

Soumojit-007:

Streamlined config.toml,index.ts and deno.json
No new features or bug fixing,purely optimization

Configurations were verbose and harder to maintain
Entrypoint (index.ts) had redundant logic
Deno dependencies weren't fully leveraging modern features

## What is the new behavior?
config.toml : Simplified structure,remove unused settings
index.ts: Reorganised logic for faster execution
deno.json: Updated imports and optimized task runners
(No visual changes-backend-only optimization)
## Additional context
deno test passes

Suggested labels: 
refractor,performance backend


internal/functions/migrations/templates/index.ts
sweatybridge, line 1: If I understand correctly, you want to make the default template created by `functions new` more useful? If so, we should introduce a new flag instead of modifying the default behaviour. For eg.

```
supabase functions new --with-template fetch-user
```

PR #3514, opened 2025-05-05
chore: add some utils tests 

egor-romanov:

- just add some utils tests


.github/workflows/ci.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - uses: t1m0thyj/unlock-keyring@v1
       - run: |
           go run gotest.tools/gotestsum -- -race -v -count=1 -coverprofile=coverage.out \
-          `go list ./... | grep -Ev 'cmd|docs|examples|pkg/api|tools'`
+          `go list ./... | grep -Ev 'cmd|docs|examples|pkg/api|tools|test|testing|mock'`
```
sweatybridge, line 25:
```suggestion
          `go list ./... | grep -Ev 'cmd|docs|examples|pkg/api|tools`
```

no need to include non-existent directories

sweatybridge (approved):

internal/utils/cloudflare/dns_test.go
sweatybridge, line 1: This is mostly covered by [internal/utils/api_test.go](https://github.com/supabase/cli/blob/develop/internal/utils/api_test.go) but was reported as 0% due to a bug in coverage reporting introduced by go 1.24 upgrade. I've fixed the coverage reporting issue so could you rebase and verify if these tests are improving coverage?

Historically a large part of this file was moved from [internal/utils/api.go](https://github.com/supabase/cli/blob/develop/internal/utils/api.go) and it wasn't worth the effort to rewrite or refractor the tests that already covered the code paths.

internal/utils/container_output_test.go
sweatybridge, line 1: This test [randomly fails](https://github.com/supabase/cli/actions/runs/14829242612/job/41627228076) due to a race condition in the underlying implementation. The code path is only affecting pgadmin schema diff tool which we plan to deprecate.

I'd suggest removing this test since there's no need to further maintain this code.

.github/workflows/ci.yml
egor-romanov: lets return `test|testing` as it also excludes nested directories like

```
internal/db/test
internal/testing
...
```

sweatybridge: `db test` is actually a command we expose to the end user so it should be included in coverage.

`internal/testing` may be excluded but it's also good to know if those utils are actually used by tests. I'm in favour of keeping.

sweatybridge (approved):

PR #3529, opened 2025-05-06
fix(config): sanitize dotenv errors avoiding data leaking

avallete:

- Print full error when in debug mode
- Otherwise, parse native godotenv error to sanitize values that might be leaking from the dotenv file.


pkg/config/config.go
```diff
@@ -929,7 +929,31 @@ func loadDefaultEnv(env string) error {
...
+		switch {
+		case strings.Contains(msg, "unexpected character"):
+			// Try to extract the character, fallback to generic
+			start := strings.Index(msg, "unexpected character \"")
+			if start != -1 {
+				start += len("unexpected character \"")
+				end := strings.Index(msg[start:], "\"")
+				if end != -1 {
+					char := msg[start : start+end]
+					return errors.Errorf("failed to parse environment file: .env (unexpected character '%s' in variable name)", char)
```
sweatybridge, line 946:
```suggestion
					return errors.Errorf("failed to parse environment file: %s (unexpected character '%s' in variable name)", path, char)
```

probably a good idea to include the full path to env file

pkg/config/config.go
```diff
@@ -929,7 +929,31 @@ func loadDefaultEnv(env string) error {
...
+				}
+			}
+			return errors.Errorf("failed to parse environment file: .env (unexpected character in variable name)")
+		case strings.Contains(msg, "unterminated quoted value"):
+			return errors.Errorf("failed to parse environment file: .env (unterminated quoted value)")
+		case strings.Contains(msg, "\n"):
```
sweatybridge, line 952: any explanation why we are checking for new lines?

avallete: The sanitizing should handle the current possible parsing error (from the code I read from the godotenv lib). But if some update introduce new errors, this might break.

When the error message contain newlines, chances are, the newlines are actually the dotenv values being leaked away. Will add a comment for context though.

pkg/config/config.go
```diff
@@ -929,7 +929,33 @@ func loadDefaultEnv(env string) error {
...
+			}
+			return errors.Errorf("failed to parse environment file: %s (unexpected character in variable name)", path)
+		case strings.HasPrefix(msg, "unterminated quoted value"):
+			return errors.Errorf("failed to parse environment file: %s (unterminated quoted value)", path)
+		// If the error message contains newlines, there is a high chance that the actual content of the
+		// dotenv file is being leaked. In such cases, we return a generic error to avoid unwanted leaks in the logs
+		case strings.Contains(msg, "\n"):
+			return errors.Errorf("failed to parse environment file: %s (syntax error)", path)
+		default:
+			return errors.Errorf("failed to load environment file: %s: %v", path, err)
```
sweatybridge, line 957:
```suggestion
			return errors.Errorf("failed to load %s: %w", path, err)
```

sweatybridge (commented):
just some nitpicks

sweatybridge (approved):

PR #3537, opened 2025-05-07
fix: apply the decorator flag only if uses deno 1

nyannyacha:

Bug fix

## Description

Edge runtime (deno 2) no longer accepts that flag, as it is possible to specify the decorator directly in deno.json (the default value is still tc39).

### DIFF
--- a/internal/functions/deploy/bundle.go
+++ b/internal/functions/deploy/bundle.go
@@ -60,7 +60,9 @@ func (b *dockerBundler) Bundle(ctx context.Context, slug, entrypoint, importMap 
 	if viper.GetBool("DEBUG") {
 		cmd = append(cmd, "--verbose")
 	}
-	cmd = append(cmd, function.BundleFlags...)
+	if utils.Config.EdgeRuntime.DenoVersion == 1 {
+		cmd = append(cmd, function.BundleFlags...)
+	}
 
 	env := []string{}
 	if custom_registry := os.Getenv("NPM_CONFIG_REGISTRY"); custom_registry != "" {


internal/functions/deploy/bundle.go
```diff
@@ -60,7 +60,9 @@ func (b *dockerBundler) Bundle(ctx context.Context, slug, entrypoint, importMap
 	if viper.GetBool("DEBUG") {
 		cmd = append(cmd, "--verbose")
 	}
-	cmd = append(cmd, function.BundleFlags...)
+	if utils.Config.EdgeRuntime.DenoVersion == 1 {
+		cmd = append(cmd, function.BundleFlags...)
+	}
```
sweatybridge, line 65: How would this work for the native bundler when we upgrade branching to edge runtime 2?

nyannyacha: @sweatybridge the decorator flag no longer exists in edge runtime 2. These properties will be available to set in deno.json.

deno.json
```jsonc
{
  "compilerOptions": {
    // These options enable the experimental TypeScript decorators feature.
    // If this is not enabled, the default is tc39. (It was also the default in edge runtime 1)
    "experimentalDecorators": true,
    "emitDecoratorMetadata": false
  }
}
```

nyannyacha:
Oh wait, the same thing needs to be applied to nativeBundler.

nyannyacha:
It seems `nativeBundler` is located in pkg/function, which causes an import cycle problem when trying to import utils.Config.

@sweatybridge Do you have any ideas on how to solve this?

nyannyacha:
btw, It appears that `NewNativeBundler` is not referenced anywhere. Am I seeing this correctly?

internal/functions/deploy/bundle.go
sweatybridge: Can we just silently ignore this flag on edge runtime 2 rather than erroring out?

Then we don't have to worry about the upgrade path potentially breaking anyone.

nyannyacha: Umm, I think that also will work. I'll do that. 😋

PR #3586, opened 2025-05-20
fix: escape underscore in sql like syntax

sweatybridge:

Bug fix

## What is the new behavior?

Underscores in SQL strings must be escaped.

## Additional context


avallete (approved):
TIL

Issue #2505, opened 2024-07-09
Vector fails to connect to docker socket on windows


seadynamic8:
**Describe the bug**

After update to v1.183.5, get this error when starting supabase cli: (local dev) ... now downgraded back to 1.178.2, get these errors:

with downgraded v1.178.2
```
supabase_realtime_chat_app container logs:
21:30:04.267 [info] Migrations already up
21:30:05.018 [debug] QUERY OK db=1.8ms queue=140.7ms idle=0.0ms
begin []
21:30:05.043 [debug] QUERY ERROR source="tenants" db=0.0ms
SELECT t0."id", t0."name", t0."external_id", t0."jwt_secret", t0."jwt_jwks", t0."postgres_cdc_default", t0."max_concurrent_users", t0."max_events_per_second", t0."max_bytes_per_second", t0."max_channels_per_client", t0."max_joins_per_second", t0."suspend", t0."enable_authorization", t0."inserted_at", t0."updated_at" FROM "tenants" AS t0 WHERE (t0."external_id" = $1) ["realtime-dev"]
21:30:05.046 [debug] QUERY OK db=0.3ms
rollback []
** (Postgrex.Error) ERROR 42703 (undefined_column) column t0.enable_authorization does not exist

    query: SELECT t0."id", t0."name", t0."external_id", t0."jwt_secret", t0."jwt_jwks", t0."postgres_cdc_default", t0."max_concurrent_users", t0."max_events_per_second", t0."max_bytes_per_second", t0."max_channels_per_client", t0."max_joins_per_second", t0."suspend", t0."enable_authorization", t0."inserted_at", t0."updated_at" FROM "tenants" AS t0 WHERE (t0."external_id" = $1)
    (ecto_sql 3.8.3) lib/ecto/adapters/sql.ex:932: Ecto.Adapters.SQL.raise_sql_call_error/1
    (ecto_sql 3.8.3) lib/ecto/adapters/sql.ex:847: Ecto.Adapters.SQL.execute/6
    (ecto 3.8.4) lib/ecto/repo/queryable.ex:221: Ecto.Repo.Queryable.execute/4
    (ecto 3.8.4) lib/ecto/repo/queryable.ex:19: Ecto.Repo.Queryable.all/3
    (ecto 3.8.4) lib/ecto/repo/queryable.ex:147: Ecto.Repo.Queryable.one/3
    /app/lib/realtime-2.28.32/priv/repo/seeds.exs:10: (file)
[...]

PR #3592, opened 2025-05-21
fix: early fail on non-exposed docker daemon

PhilipWee:

When trying to run supabase locally on windows, It checks if the docker daemon is properly exposed on `:2375` instead of printing a warning. 

If it is not running, it directs the user to the relevant documentation and gives a more verbose hint on how to solve the problem

If it is not exposed, supabase continues trying to start, and crashes with an obscure error message.

https://github.com/supabase/cli/issues/2505

## What is the new behavior?

Performs a get request on the docker daemon, and checks if it is available before running the analytics

## Additional context

image

### DIFF
--- a/internal/start/start.go
+++ b/internal/start/start.go
@@ -4,8 +4,10 @@ import (
 	"bytes"
 	"context"
 	_ "embed"
+	"encoding/json"
 	"fmt"
 	"net"
+	"net/http"
 	"net/url"
 	"os"
 	"path"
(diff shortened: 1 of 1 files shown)


internal/start/start.go
```diff
@@ -280,7 +282,25 @@ EOF
 			}
 			env = append(env, "DOCKER_HOST="+dindHost.String())
 		case "npipe":
-			fmt.Fprintln(os.Stderr, utils.Yellow("WARNING:"), "analytics requires docker daemon exposed on tcp://localhost:2375")
+			const dockerDaemonNeededErr = "Analytics requires docker daemon exposed on tcp://localhost:2375. Either disable analytics in supabase/config.toml, or expose the docker daemon. See https://supabase.com/docs/guides/local-development/cli/getting-started?queryGroups=platform&platform=windows#running-supabase-locally for more details"
+
+			resp, err := http.Get("http://localhost:2375/version")
```
sweatybridge, line 287: If docker host is already setup for `npipe` there's probably no need to call the tcp port.

IMO, it would be better to disable analytics here and let supabase start run to completion.

PhilipWee: Yes you are right my bad, the `tcp` port check will always fail because its already detected as `npipe` my bad

As for the suggestion regarding disabling analytics, I have a few ideas on how to implement it

Option A:
1. When installing supabase
2. Check if daemon running on `npipe`
3. If so, ask user if we want to disable analytics in the generated config
4. If yes, generate `config.toml` with `analytics=false`

Option B:
1. When running `supabase start`
2. Check if vector is failing
3. If vector is failing, then print the error, but continue allowing supabase to spin up fully

Let me know which one you think think is better and I can go ahead and do it. Thanks!

sweatybridge: Does it work on windows currently if you pass in `supabase start --ignore-health-check` flag? If so, then option B sounds like a good solution.

sweatybridge: Basically changing the following lines https://github.com/supabase/cli/blob/develop/internal/start/start.go#L331

```golang
		if parsed.Scheme != "npipe" {
			started = append(started, utils.VectorId)
		}
```

sweatybridge (approved):

PR #3600, opened 2025-05-22
fix: diff declarative schemas without supabase stop

sweatybridge:

feature

## What is the new behavior?

Running `supabase db diff` now uses declared schemas if they are defined. To diff the local database instead, users must comment out `[db.migrations] schema_paths` in config.toml.

This is made possible by creating a template database named `_shadow` that contains the initial schema without any user migrations. When diffing, a new database `contrib_regression` is created from template to apply all declarative schemas. A separate container is then created to apply all migrations on the `postgres` database. Finally, a diff is generated between `postgres` and `contrib_regression`.

When diffing linked or remote database, declarative schemas are not used.

## Additional context

Migrations which don't work

- [x] `create extension pg_cron`

```
ERROR: can only create extension in database postgres (SQLSTATE P0001)
Jobs must be scheduled from the database configured in cron.database_name, since the pg_cron background worker reads job descriptions from this database.
```

Possible solution is to remove pg_cron from declared schemas and diff output.

### DIFF
--- a/internal/db/diff/diff.go
+++ b/internal/db/diff/diff.go
@@ -19,6 +19,7 @@ import (
 	"github.com/docker/go-connections/nat"
 	"github.com/go-errors/errors"
 	"github.com/jackc/pgconn"
+	"github.com/jackc/pgerrcode"
 	"github.com/jackc/pgx/v4"
 	"github.com/spf13/afero"
 	"github.com/supabase/cli/internal/db/start"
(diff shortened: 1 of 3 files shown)


constantx:
this caught me off-guard, needs to be documented better in the documentation, but a very nice improvement indeed to make the declarative schemas more useful. 🙏

PR #3602, opened 2025-05-22
feat(typegen): add postgrest-version params

avallete:
Should be merged once: https://github.com/supabase/postgres-meta/pull/948/files is merged


internal/gen/types/types.go
```diff
@@ -60,6 +60,9 @@ func Run(ctx context.Context, projectId string, dbConfig pgconn.Config, lang str
 			return err
 		}
 
+		// Extract version from image tag and trim 'v' prefix
+		postgrestVersion = strings.TrimPrefix(utils.Config.Api.Image, "postgrest/postgrest:v")
+
```
sweatybridge, line 65: Can we move this to its own check that also applies to remote database?

```suggestion
	if len(postgrestVersion) == 0 {
		// Extract version from image tag and trim 'v' prefix
		postgrestVersion = strings.TrimPrefix(utils.Config.Api.Image, "postgrest/postgrest:v")
	}
```

cmd/gen.go
```diff
@@ -106,6 +110,7 @@ func init() {
 	typeFlags.StringSliceVarP(&schema, "schema", "s", []string{}, "Comma separated list of schema to include.")
 	typeFlags.Var(&swiftAccessControl, "swift-access-control", "Access control for Swift generated types.")
 	typeFlags.BoolVar(&postgrestV9Compat, "postgrest-v9-compat", false, "Generate types compatible with PostgREST v9 and below. Only use together with --db-url.")
+	typeFlags.StringVar(&postgrestVersion, "postgrest-version", "", "Generate types with __InternalSupabase schema using the right version of postgrest. Only use together with --db-url.")
```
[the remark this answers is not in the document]
avallete: I think we can infer v9 automatically from the postgrestVersion. Just wonder if that might break some user commands 🤔

internal/gen/types/types.go
sweatybridge: It's also possible to connect to remote db via direct connection string, via --db-url flag. Although it's more common to hit our hosted api like you mentioned.

Another use case is connecting to self hosted db that's not on localhost.

PR #3607, opened 2025-05-25
Update default Mailpit SMTP port to 1025 and pop3 port to 1110 in con…

raucheacho:
…fig.toml template

default config toml change:
Update default SMTP and pop3 configuration in config.toml template to use Mailpit

The cli and config.toml file still references Inbucket as the default local email testing server, with the SMTP and pop3 port set to 54324.

## What is the new behavior?

* The default **SMTP port** is updated from `54325` to `1025` and pop3 is updated from `54326` to `1110` to reflect the switch to **Mailpit**, which listens respectively on ports `1025` and `1110` by default.
* All references to **Inbucket** in descriptions or comments are replaced with **Mailpit**.

## Additional context

Supabase has recently migrated from **Inbucket** to **Mailpit** as the default local email testing service. Updating the `config.toml` file helps keep the developer documentation consistent and avoids confusion during local setup.

### DIFF
--- a/pkg/config/templates/config.toml
+++ b/pkg/config/templates/config.toml
@@ -75,13 +75,13 @@ openai_api_key = "env(OPENAI_API_KEY)"
 
 # Email testing server. Emails sent with the local dev setup are not actually sent - rather, they
 # are monitored, and you can view the emails that would have been sent from the web interface.
-[inbucket]
+[mailpit]
 enabled = true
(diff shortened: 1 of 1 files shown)


sweatybridge:
For backwards compatibility, you need to add a new config block for mailpit.

But even then, I don't think this change is necessary because the ports 1025 and 1110 are internal to mailpit container. They are mapped to different ports on localhost.

PR #3621, opened 2025-05-27
fix: use new api types

sweatybridge:

## What is the new behavior?

- parses string to uuid before calling api
- workaround for inline enums in query param
- many workarounds for using anonymous structs

## Additional context


hf (approved):
Looks to be OK.

PR #3638, opened 2025-05-31
chore: clarify project linking instructions and location of project ID

kenwuuu:

Doc update

### DIFF
--- a/docs/supabase/link.md
+++ b/docs/supabase/link.md
@@ -2,6 +2,8 @@
 
 Link your local development project to a hosted Supabase project.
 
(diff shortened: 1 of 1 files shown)


docs/supabase/link.md
```diff
@@ -2,6 +2,8 @@
 
 Link your local development project to a hosted Supabase project.
 
+Use your project ID for `--project-ref`, you can find it in your dashboard url: `https://app.supabase.com/project/<project-id>`
+
```
sweatybridge, line 6: The `--project-ref` flag is no longer required as you will be prompted to choose a project when running supabase link command without any flags. Could you elaborate what problems you ran into that motivates this change?

PR #3652, opened 2025-06-04
fix: update default postgres major version to 17

sweatybridge:

Bug fix

## What is the new behavior?

Bump default postgres version to 17.

## Additional context


jgoux (approved):
Let's gooo! 🚀

PR #3653, opened 2025-06-04
fix: adjust api key handling for logflare, studio, and vecor

Ziinc:
This PR fixes the API key handling for logflare, vector, and studio

- all services uses a private access token, since this is not meant for production purposes we don't bother generating another key to reduce complexity.
- vector config now passes logflare api key through headers instead of url, to avoid logging it

This PR requires a subsequent version bump after https://github.com/supabase/supabase/pull/36169 is merged

- [x] verify locally that all works


sweatybridge:
working as expected after updating studio

Screenshot 2025-06-05 at 4 26 15 PM

there are errors affecting edge function and cron logs. but those can be resolved separately

sweatybridge (approved):

PR #3675, opened 2025-06-07
feat: hot reload eszip bundle when function source changes

avallete:

- Feat: add hot reloading to functions serve command
- Functions url's in the serve command

https://github.com/user-attachments/assets/6f7cad96-8041-48c6-9918-1a11882049c9

Screenshot 2025-06-09 at 10 14 32


internal/utils/deno.go
```diff
@@ -209,20 +209,31 @@ func CopyDenoScripts(ctx context.Context, fsys afero.Fs) (*DenoScriptDir, error)
 	return &sd, nil
 }
 
-func BindHostModules(cwd, relEntrypointPath, relImportMapPath string, fsys afero.Fs) ([]string, error) {
-	var modules []string
+// ModulePaths contains both the raw paths and their Docker bind mount equivalents
+type ModulePaths struct {
+	Paths []string // Raw host paths
+	Binds []string // Docker bind mount strings
+}
```
avallete, line 216: **note**

Sightly changed this function to not only get the final bind but also the original host path that match it, so we can re-use it in the watcher.

internal/utils/logger.go
```diff
@@ -1,15 +1,45 @@
...
+func Debug(format string, args ...interface{}) {
+	message := fmt.Sprintf(format, args...)
+	fmt.Fprintf(GetDebugLogger(), "%s %s\n", Aqua("DEBUG:"), message)
+}
+
+// Warning logs a warning message to stderr with consistent formatting and styling
+func Warning(format string, args ...interface{}) {
+	message := fmt.Sprintf(format, args...)
+	fmt.Fprintf(os.Stderr, "%s %s\n", Yellow("WARNING:"), message)
+}
```
avallete, line 45: **note**

Attempt to setup some common logging to uniformize things on the whole cli. If that's accepted, will make follow-up refactors within another PR.

sweatybridge: I would probably use `slog` directly with a custom format so we don't need to create these methods ourselves.

internal/functions/serve/watcher.go
```diff
@@ -0,0 +1,316 @@
...
+		}
+	}
+}
+
+// FileWatcherSetup interface allows for dependency injection of file watching functionality
+type FileWatcherSetup interface {
+	SetupFileWatcher(fsys afero.Fs) (*fsnotify.Watcher, string, error)
+}
+
+type RealFileWatcherSetup struct{}
```
sweatybridge, line 302: I would move some of the args of `runFileWatcher` to its own internal state. For eg.

```suggestion
type RealFileWatcher struct {
	watcher     *fsnotify.Watcher
	watchedDirs map[string]bool
	RestartChan chan struct{}
}
```


Then `SetupFileWatcher` should also be renamed to an idiomatic constructor.

```go
func NewFileWatcher(watchedPath string, fsys afero.Fs) (FileWatcher, error)
```

internal/functions/serve/serve.go
```diff
@@ -59,179 +52,84 @@ func (i *RuntimeOption) toArgs() []string {
...
-const (
-	dockerRuntimeServerPort    = 8081
-	dockerRuntimeInspectorPort = 8083
-)
-
-//go:embed templates/main.ts
-var mainFuncEmbed string
-
 func Run(ctx context.Context, envFilePath string, noVerifyJWT *bool, importMapPath string, runtimeOption RuntimeOption, fsys afero.Fs) error {
+	return RunWithWatcher(ctx, envFilePath, noVerifyJWT, importMapPath, runtimeOption, fsys, &RealFileWatcherSetup{})
```
sweatybridge, line 56: How about we only run the file watcher if user sets the `policy = "per_worker"` in edge runtime config?

In oneshot policy, there's no need to watch for file changes.

internal/functions/serve/serve.go
```diff
@@ -59,179 +52,84 @@ func (i *RuntimeOption) toArgs() []string {
...
+			_ = utils.Docker.ContainerRemove(context.Background(), utils.EdgeRuntimeId, container.RemoveOptions{
+				RemoveVolumes: true,
+				Force:         true,
+			})
+			return ctx.Err()
+		default:
+			// Use network alias because Deno cannot resolve `_` in hostname
+			dbUrl := fmt.Sprintf("postgresql://postgres:postgres@%s:5432/postgres", utils.DbAliases[0])
+
+			serviceCancel, logsDone, err := manageFunctionServices(ctx, envFilePath, noVerifyJWT, importMapPath, dbUrl, runtimeOption, fsys, errChan)
```
sweatybridge, line 101: Instead of returning a service cancel, can we simply call `Docker.ContainerRestart` on receiving events from the restart channel?

I believe the docker log stream will be disconnected on container restart so we can catch and ignore the error to restream.

internal/functions/serve/serve.go
sweatybridge: Hmm in those cases we only need to update the `SUPABASE_INTERNAL_FUNCTIONS_CONFIG` env var. But yes file watching handles this use case too.

internal/functions/serve/serve.go
avallete: ContainerRestart won't be enough to update bindings and so on, we need a full container recreation.

internal/functions/serve/service.go
```diff
@@ -0,0 +1,230 @@
...
+func manageFunctionServices(
+	ctx context.Context,
+	envFilePath string,
+	noVerifyJWT *bool,
+	importMapPath string,
+	dbUrl string,
+	runtimeOption RuntimeOption,
+	fsys afero.Fs,
+	errChan chan<- error,
+) (context.CancelFunc, <-chan struct{}, error) {
```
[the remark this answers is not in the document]
avallete: Did that. Remove the `service.go` and merged the tests within `serve_test.go` also tried to reduce the number of linediffs.

internal/functions/serve/serve.go
```diff
@@ -75,23 +75,126 @@ func Run(ctx context.Context, envFilePath string, noVerifyJWT *bool, importMapPa
...
+func manageFunctionServices(
+	ctx context.Context,
+	envFilePath string,
+	noVerifyJWT *bool,
+	importMapPath string,
+	dbUrl string,
+	runtimeOption RuntimeOption,
+	fsys afero.Fs,
+	errChan chan<- error,
+) (context.CancelFunc, <-chan struct{}, error) {
```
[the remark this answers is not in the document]
avallete: Oh good catch. What might also happen is that the path for the impor_map or entrypoint get changed 🤔

internal/utils/deno.go
sweatybridge: I just read this post: https://unix.stackexchange.com/a/709937

We can rely on `strings.Split(bind, ":")[0]` to get the host path instead.

internal/functions/serve/serve.go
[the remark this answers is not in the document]
avallete: Back on this, I've tried the purely splitted approach once more here: e2f2a1fc8a8fb7b0a92f7bf89f9479de0aa6591c

And once more got sync issues for instance, when the container was restarted the logstream is lost. And I think it make sense since the container attached to it has actually been removed, and replaced by a new one.

```
INFO: Added directory to watcher: /Users/avallete/Programming/Supa/cli/supabase/functions
INFO: Added directory to watcher: /Users/avallete/Programming/Supa/cli/supabase
Serving functions on http://127.0.0.1:54321/functions/v1/<function-name>
 - http://127.0.0.1:54321/functions/v1/hello-world2
 - http://127.0.0.1:54321/functions/v1/hello-world88
Using supabase-edge-runtime-1.67.4 (compatible with Deno v1.45.2)
serving the request with supabase/functions/hello-world2
[Info] Hello from Functions!

serving the request with supabase/functions/hello-world2
// Changed the console.log in hello-world2 to print `Hello from Functions 2!`
INFO: File change detected: /Users/avallete/Programming/Supa/cli/supabase/functions/hello-world2/index.ts (WRITE)
INFO: Setting up Edge Functions runtime...
INFO: Edge Functions runtime is ready.
// Called the function with curl, no more logs from the docker container after the restart
```

To get back the proper behavior, I had to actually add a function in charge of syncing (creating the logstream) when the service restarted. Which end up very similar to the original approach (also had to introduce a singleton to check if the logstream was actually running to avoid race-conditions with starting containers) (see: 459a35f51916020519faeae5434e3846dfe8c96c)

Giving me the more expected:

```
INFO: Setting up Edge Functions runtime...

INFO: Edge Functions runtime is ready.

Serving functions on http://127.0.0.1:54321/functions/v1/<function-name>
 - http://127.0.0.1:54321/functions/v1/hello-world2
 - http://127.0.0.1:54321/functions/v1/hello-world88
Using supabase-edge-runtime-1.67.4 (compatible with Deno v1.45.2)
serving the request with supabase/functions/hello-world2
[Info] Hello from Functions!

INFO: File change detected: /Users/avallete/Programming/Supa/cli/supabase/functions/hello-world2/index.ts (WRITE)
INFO: Setting up Edge Functions runtime...

INFO: Edge Functions runtime is ready.

Serving functions on http://127.0.0.1:54321/functions/v1/<function-name>
 - http://127.0.0.1:54321/functions/v1/hello-world2
 - http://127.0.0.1:54321/functions/v1/hello-world88
Using supabase-edge-runtime-1.67.4 (compatible with Deno v1.45.2)
serving the request with supabase/functions/hello-world2
[Info] Hello from Functions 2!
```

If you prefer this approach or see another way to do it please let me know. Otherwise I can also revert this back to the original `manageServices` function with the logstream and restartChannel togethers.

internal/functions/serve/watcher.go
```diff
@@ -0,0 +1,368 @@
...
+			}
+			return
+		}
+	}
+}
+
+type FileWatcher interface {
+	Watch(ctx context.Context, fsys afero.Fs) (<-chan struct{}, <-chan error)
+	Close() error
+}
```
sweatybridge, line 324: This file watcher is tightly coupled with our internals of functions and config paths. I think we can go with a more general implementation

```suggestion
type FileWatcher interface {
	Watch(ctx context.Context) (<-chan struct{}, <-chan error)
	SetWatchPaths(watchPaths []string, fsys afero.Fs) error
	Close() error
}
```

The main responsibilities for file watcher are
1. Denounce fs events
2. Support recursive directories


After receiving an event from restart channel, it's the caller's responsibility to set the new watch paths.

sweatybridge: > if we automatically watch recursively all the directory from there we'll end up with a lot of watchers which might cause issues.

Good point. I wonder how nextjs solve it in their local dev?

Perhaps we can set a max recursion depth or other cleverer ways of pruning.

avallete:
Closing in favor of: https://github.com/supabase/cli/pull/3710

PR #3680, opened 2025-06-08
feat: add restart command

raideno:

Add a restart command


raideno:
The logout test doesn’t always work the same way, sometimes it passes and sometimes it fails, even tho i changed nothing related to it.

PR #3747, opened 2025-06-23
fix: diff declarative schemas without supabase stop

sweatybridge:

Bug fix

## What is the new behavior?

supersedes https://github.com/supabase/cli/pull/3600

## Additional context


dewinterjack:
Thank you!! This has been stopping me using declarative schemas

PR #3759, opened 2025-06-24
feat(config): add network_restrictions to db config

avallete:

## What is the new behavior?

## Additional context


pkg/config/db.go
```diff
@@ -67,18 +67,25 @@ type (
 		WorkMem                       *string                 `toml:"work_mem"`
 	}
 
+	NetworkRestrictions struct {
+		Enabled          bool     `toml:"enabled"`
+		DbAllowedCidrs   []string `toml:"db_allowed_cidrs"`
+		DbAllowedCidrsV6 []string `toml:"db_allowed_cidrs_v6"`
+	}
```
sweatybridge, line 74:
```suggestion
	networkRestrictions struct {
		Enabled          bool     `toml:"enabled"`
		AllowedCidrs   []string `toml:"allowed_cidrs"`
		AllowedCidrsV6 []string `toml:"allowed_cidrs_v6"`
	}
```

Since this block is under `[db.network_restrictions]`, we don't need to `db_` prefix in each config field.

Also the config struct should be priviate.

avallete: Was private at first, but then needed it there: https://github.com/supabase/cli/pull/3759/files#diff-b94d4322eb20f9cd90dc342d084155e9a360a54d58e2ee74170a73c80a967d42R213-R218

Could turn this into an `utils.Config.NewNetworkRestriction()` method WDYT ?

sweatybridge: I tried to simplify this with the following default config

```toml
[db.network_restrictions]
# Enable management of network restrictions.
enabled = false
# List of IPv4 CIDR blocks allowed to connect to the database.
# Defaults to allow all IPv4 connections. Set empty array to block all IPs.
allowed_cidrs = ["0.0.0.0/0"]
# List of IPv6 CIDR blocks allowed to connect to the database.
# Defaults to allow all IPv6 connections. Set empty array to block all IPs.
allowed_cidrs_v6 = ["::/0"]
```

I think it's ok since we don't support disabling network restrictions on platform. Will push a commit so you can take a look.

PR #3768, opened 2025-06-25
fix: skip unchanged functions when deploying

sweatybridge:

Bug fix

## What is the new behavior?

As title

## Additional context


avallete (approved):
Nice one 👍

PR #3841, opened 2025-07-10
feat: support jwt signing keys for local auth

cemalkilic:

command to generate jwt signing keys & spin up auth server with the configured JWK keys

### Steps
1. Generate the jwt signing keys by `supabase gen signing-key ES256`
3. Enable `signing_keys_path` line in the `config.toml`.
4. Once the auth server is up and running, the configured JWK should be available in `http://localhost:54321/auth/v1/.well-known/jwks.json`

```sh
$ ./bin/supabase gen signing-key --help
Securely generate a private JWT signing key for use in the CLI or to import in the dashboard.

Supported algorithms:
	ES256 - ECDSA with P-256 curve and SHA-256 (recommended)
	RS256 - RSA with SHA-256

Usage:
  supabase gen signing-key [flags]

Examples:
  supabase gen signing-key --algorithm RS256
  supabase gen signing-key --algorithm ES256
  supabase gen signing-key --algorithm RS256 --append

Flags:
      --algorithm [ RS256 | ES256 ]   Algorithm for signing key generation. (default ES256)
      --append                        Append new key to existing keys file instead of overwriting.
  -h, --help                          help for signing-key

...
```

## Command output
```sh
$ ./bin/supabase gen signing-key --algorithm RS256
/Users/cemalkilic/projects/cli/signing_keys.json
JWT signing keys saved to: /Users/cemalkilic/projects/cli/signing_keys.json
⚠️  IMPORTANT: Add this file to your .gitignore to prevent committing signing keys to version control

To enable JWT signing keys in your project:
1. Add the following to your config.toml file:
   signing_keys_path = "/Users/cemalkilic/projects/cli/signing_keys.json"
2. Restart your local development server:
   supabase start
```

## Run with `--append` flag
```sh
$ ./bin/supabase gen signing-key --algorithm RS256 --append
/Users/user/projects/cli/signing_keys.json
JWT signing key appended to: /Users/user/projects/cli/signing_keys.json (now contains 2 keys)
⚠️  IMPORTANT: Add this file to your .gitignore to prevent committing signing keys to version control

To enable JWT signing keys in your project:
1. Add the following to your config.toml file:
   signing_keys_path = "/Users/cemalkilic/projects/cli/signing_keys.json"
2. Restart your local development server:
   supabase start
```


hf (approved):
Looks OK to me, however you may want to explore a `jwk` library if you want.

hf:
Maybe the default output format should be the key, while passing another parameter to give you the config.toml setting?

Reason being this command can be used for people who want to import a private key they control.

cmd/gen.go
```diff
@@ -93,6 +94,32 @@ var (
...
   supabase gen types --db-url 'postgresql://...' --schema public --schema auth`,
 	}
+
+	keyOutputFormat = utils.EnumFlag{
+		Allowed: []string{"env", "json", "jwks"},
+		Value:   "json",
+	}
+
+	genGenerateKeyCmd = &cobra.Command{
+		Use:   "generate-key <algorithm>",
```
sweatybridge, line 104: Can we update the existing `gen keys` command please?

pkg/config/templates/config.toml
```diff
@@ -122,6 +122,9 @@ site_url = "http://127.0.0.1:3000"
 additional_redirect_urls = ["https://127.0.0.1:3000"]
 # How long tokens are valid for, in seconds. Defaults to 3600 (1 hour), maximum 604,800 (1 week).
 jwt_expiry = 3600
+# Use JWT_KEYS for enhanced security instead of JWT_SECRET. Generate keys using:
+# supabase gen generate-key RS256 --format=env
+# jwt_keys = "env(GOTRUE_JWT_KEYS)"
```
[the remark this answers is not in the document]
cemalkilic: Agree with the maintenance overhead but the JWT signing keys on the platform are also opt-in until September/October, so I think the CLI should also do so?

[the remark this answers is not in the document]
sweatybridge: IMO it's better to have local dev ahead of platform so users can experiment and discover issues before we roll out en masse. @hf Wdyt?

cmd/gen.go
cemalkilic: I see this is getting complicated already, keeping the JWT secret for anon & service_role keys but using asymmetric JWTs only for user tokens?

sweatybridge: Yea sorry for the lack of context. I should have explained that the existing `gen keys` command was added as a way to generate arbitrary JWT for anon and service role roles. However it was never working as intended because our api gateway only allows predefined keys to used as auth header.

So I'm also ok with removing the old command completely in favour of generating the new keys.

cemalkilic: for now, I added the new sub command `signing-key`. in a future PR, we can remove the `gen keys` command?

pkg/config/templates/config.toml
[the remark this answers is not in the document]
sweatybridge: Sorry if this is obvious but why is 1 necessary if we are only focusing on local?

cmd/gen.go
sweatybridge: Sure, that sounds better. Are we using `signing-key` consistently as a name in docs and dashboard? Just want to make sure we are using the same terminology in cli as well.

pkg/config/templates/config.toml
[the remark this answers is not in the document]
cemalkilic: > then it's not ideal to have a half baked solution where we gen keys in cli but upload via dashboard.

good point but also I want to highlight that JWT signing keys are also not enforced by deafult yet. once we have the full rollout/enforcement of JWT signing keys, I think we can make this flow entirely on the cli (no need to upload on the dashboard)

cmd/gen.go
```diff
@@ -93,6 +94,32 @@ var (
...
+	genSigningKeyCmd = &cobra.Command{
+		Use:   "signing-key <algorithm>",
+		Short: "Generate JWT signing keys",
+		Long: `Generate JWT signing keys for use with GOTRUE_JWT_KEYS.
+
+Supported algorithms:
+	ES256 - ECDSA with P-256 curve and SHA-256 (recommended)
+	RS256 - RSA with SHA-256
+
+Outputs the private key in JWK format by default. Use --format=env for GOTRUE_JWT_KEYS environment variable format.`,
```
hf, line 112:
```suggestion
		Long: `Securely generate a private JWT signing key for use in the CLI or to import in the dashboard.

Supported algorithms:
	ES256 - ECDSA with P-256 curve and SHA-256 (recommended)
	RS256 - RSA with SHA-256

Outputs the private key in JWK format by default. Use --format=env for GOTRUE_JWT_KEYS environment variable format.`,
```

cmd/gen.go
```diff
@@ -111,5 +138,8 @@ func init() {
 	keyFlags.StringVar(&flags.ProjectRef, "project-ref", "", "Project ref of the Supabase project.")
 	keyFlags.StringSliceVar(&override, "override-name", []string{}, "Override specific variable names.")
 	genCmd.AddCommand(genKeysCmd)
+	signingKeyFlags := genSigningKeyCmd.Flags()
+	signingKeyFlags.Var(&keyOutputFormat, "format", "Output format for the generated keys.")
+	genCmd.AddCommand(genSigningKeyCmd)
```
hf, line 143: Does cobra allow enums here, this can then automatically use <Tab> autocomplete and validation below might not be necessary?

hf (approved):
LGTM!

cmd/gen.go
cemalkilic: simplied the command by removing this (we dont need output format anymore as we're using file based approach and always store JWK format)

internal/gen/signingkeys/signingkeys.go
```diff
@@ -0,0 +1,227 @@
...
+			if err := dec.Decode(&jwkArray); err != nil {
+				return errors.Errorf("failed to decode signing key: %w", err)
+			}
+			if _, err = f.Seek(0, io.SeekStart); err != nil {
+				return errors.Errorf("failed to seek signing key: %w", err)
+			}
+		}
+		out = f
+	}
+	jwkArray = append(jwkArray, keyPair.PrivateKey)
```
sweatybridge, line 199: @cemalkilic do we need to show or save the public key to user?

cemalkilic: I think at this point, no.

pkg/config/config.go
```diff
@@ -707,6 +708,11 @@ func (c *baseConfig) resolve(builder pathBuilder, fsys fs.FS) error {
 		}
 		c.Storage.Buckets[name] = bucket
 	}
+	// Resolve signing keys path for cross-platform compatibility
+	if len(c.Auth.SigningKeysPath) > 0 && !filepath.IsAbs(c.Auth.SigningKeysPath) {
+		cwd := filepath.Dir(builder.SupabaseDirPath)
+		c.Auth.SigningKeysPath = filepath.Join(cwd, c.Auth.SigningKeysPath)
```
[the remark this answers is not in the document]
cemalkilic: Yeah, I was also confused with this because the existing template had `./supabase` for some values:

https://github.com/supabase/cli/blob/344fc1cc6cfdc40528b50ac941b3f4fa40bb6e29/pkg/config/templates/config.toml#L196

Thanks for all the fixes!

internal/start/start.go
```diff
@@ -508,6 +501,21 @@ EOF
...
+			"GOTRUE_JWT_AUD=authenticated",
+			"GOTRUE_JWT_DEFAULT_GROUP_NAME=authenticated",
+			fmt.Sprintf("GOTRUE_JWT_EXP=%v", utils.Config.Auth.JwtExpiry),
+			"GOTRUE_JWT_SECRET=" + utils.Config.Auth.JwtSecret.Value,
+			"GOTRUE_JWT_ISSUER=" + utils.GetApiUrl("/auth/v1"),
+		}...)
+
+		// Add JWT keys from file if configured
+		if keysData, err := utils.Config.Auth.GetSigningKeysData(fsys); err == nil && keysData != "" {
+			env = append(env, "GOTRUE_JWT_KEYS="+keysData)
```
sweatybridge, line 516: @cemalkilic can we add the jwks parsed from [line 148](https://github.com/supabase/cli/pull/3841/files#diff-1af1ea8d4dac7cf62987c89e0f1036ad192403d7135f445d719993f0f5303a96R148) directly here?

cemalkilic: No, that `jwks` is consumed by the other services (storage, realtime etc) to accept tokens from the third party auth ([docs](https://supabase.com/docs/guides/auth/third-party/overview#how-does-it-work)).
Supabase Auth server shouldn't serve the JWKS of the third party auth providers.

sweatybridge (approved):

logemann:
is there an easy way to track when this will be available as docker gotrue image? Happy to instantly use it ;-)

traviswimer:
@logemann It seems to be working for me after installing the `2.33.8` beta version of the cli.

I'm now also getting an error when trying to use the js client with a "service role key", but JWTs are working well with the JWKs.

marcfrankel:
@traviswimer Did you figure out a solution to the service role key not working? That's blocking me now. I keep getting "JWSError JWSInvalidSignature" errors when using it via the js-client

traviswimer:
@marcfrankel yep, you can generate your own JWT and use it as the service role key, but hopefully Supabase eventually provides a more streamlined solution.

For now anyway, you can use `supabase gen signing-key --algorithm ES256` to generate a JWK. Then you need to use a library to generate the JWT using that JWK.

In my case I used a node script with the [jose](https://www.npmjs.com/package/jose) library, so something like this:

```typescript
import { importJWK, SignJWT } from 'jose';
import jwks from './signing_keys.json';

const jwk = jwks[0];

delete jwk.key_ops;
const privateKey = await importJWK(jwk);

const jwt = await new SignJWT({ role: "service_role" })
	.setProtectedHeader({ alg: "ES256", kid: jwks[0].kid })
	.setIssuedAt()
	.setIssuer("supabase")
	.setAudience("authenticated")
	.setExpirationTime("30d")
	.sign(privateKey);
```

I'm not sure why, but the `key_ops` property of the JWK caused an error for me, so I had to delete it to get it to work.

When signing the JWT, you can set the expiration to whatever you like, but the strings used for the other fields are what supabase expects in a service role key.

---

*This is just the solution that worked for me. I have no idea if it is what the Supabase team would recommend, but since these are local JWTs that can't be used in production, I'm not too worried about it.*

cemalkilic:
Hey folks, sorry for the inconvenience. Working on the fix now to make it work with secret role keys.

BTW @traviswimer that's a nice approach and the root cause! With current implementation, we sign the anon/service role keys with JWT secret and auth server is failing to validate them as it doesn't know about the JWT secret anymore.

edit: solved in https://github.com/supabase/cli/pull/3969, will be available soon with a version release. thanks for bearing with us!

cwooldridge1:
On cli v2.39.2 I am getting Error status 400: {"statusCode":"403","error":"Unauthorized","message":"\"alg\" (Algorithm) Header Parameter value not allowed"}
Edit: Fixed by running `supabase stop` `supabase start` and then `supabase db reset --local`

PR #3969, opened 2025-08-01
feat: generate jwt tokens from signing key

sweatybridge:

feature

## What is the new behavior?

- Generates local anon and service role keys from signing key.
- Resolves only public keys when injecting to services.

## Additional context

TODO:
- [ ] handle multiple standby keys


hf:
Pushing a signing key is OK but it should come with a huge warning that the security of the key relies on the security of the device and person making the push.

Issue #3993, opened 2025-08-06
Shadowdb is unable to match production db during supabase db pull


ibash-corpusant:
**Describe the bug**
When running `supabase db pull` we always end up with a diff / new migration. This is because our supabase instance has some changes that were applied by superadmins on our behalf.

**To Reproduce**
1. Reset the db: `supabase db reset`
2. Run migrations: `supabase migration up`
3. Pull latest schema from prod: `supabase db pull`
4. Create any missing migrations
5. Repeat db reset, migration up and db pull

**Expected behavior**
After repeating the reset/migration steps the local db should be in sync with production.

**Actual behavior**

Each time we run `supabase db pull` we get a new migration with these contents:

```sql
create extension if not exists "pgstattuple" with schema "extensions";


alter extension "pg_repack" update to '1.5.2';
```

The pg_repack is because our instance was manually upgraded to 1.5.2 so we could run it.
I'm not sure why pgstattuple is in the migration.

If we try to run the new migration `supabase migration up` then we get this error:

```
ERROR: permission denied to create extension "pgstattuple" (SQLSTATE 42501)
At statement: 0                                                            
create extension if not exists "pgstattuple" with schema "extensions"      
Try rerunning the command with --debug to troubleshoot the error.
```

Because `supabase db pull` creates a new database each time there's not a way to
run these with superuser permissions on the shadow db:

```sql
CREATE EXTENSION IF NOT EXISTS "pgstattuple" WITH SCHEMA "extensions";

-- Instead of updating pg_repack to 1.5.2, we just pretend it's 1.5.2 to avoid the error
UPDATE pg_extension SET extversion = '1.5.2' WHERE extname = 'pg_repack';
```

A workaround is to grant postgres superuser permissions in `internal/db/start/templates/schema.sql` `ALTER ROLE postgres SUPERUSER;`. Which allows running a migration with the above during the shadowdb creation.

**Screenshots**

<img width="616" height="177" alt="Image" src="https://github.com/user-attachments/assets/4d4eb612-5d0b-479f-8ac4-ba3a291b1566" />

<img width="720" height="230" alt="Image" src="https://github.com/user-attachments/assets/faef802b-154a-4c78-a8bd-6fe201e72f1a" />

**System information**
Rerun the failing command with `--create-ticket` flag.
 - Ticket ID: [e.g. ab1ac733e31e4f928a4d7c8402543712]
 - Version of OS: [e.g. Ubuntu 22.04]
[...]

Issue #3919, opened 2025-07-25
Policies for tables are not added in migration on `db pull`


xidsyed:
**Describe the bug**
- perform `supabase db pull` on a fresh local instance from a fresh remote instane with one dummy `posts` table with few dummy rows . and rls policies. also one storage bucket `posts` with one rls policy on storage.objects table.

- migration file gets created on both sides.
- `supabase db reset` to apply migration and seed data
- everything shows up perfectly. EXCEPT THE POLICY FOR `STORAGE.OBJECTS`

Supabase CLI 2.31.8

<img width="800" height="541" alt="Image" src="https://github.com/user-attachments/assets/91cd99af-1038-414d-b4b8-46bfa13f89c4" />

should i not rely on the `supabase db diff` command to make migrations for me?

xidsyed:
```
$ supa db diff --local
Creating shadow database...
Initialising schema...
Seeding globals from roles.sql...
Applying migration 20250725101210_remote_schema.sql...
Connecting to local database...
Diffing schemas: extensions,public
Finished supabase db diff on branch main.

No schema changes found
```


I ended up adding the policy manually via the gui. and then i tried to diff it locally to see if it would pickup on the change and create a new migration. It didn't.

I guess it's a good thing that its consistently wrong both times.

longtimedeveloper:
@xidsyed I started using Supabase [Declarative database schemas](https://supabase.com/docs/guides/local-development/declarative-database-schemas).

Now, I don't need to pull from the database, instead, I only create migrations based on `supabase/schemas` folder `.sql` files.  This feature provides a single source of truth for your database.  I have been using this for over 3 weeks with zero issues.

I hope this helps.

xidsyed:
@longtimedeveloper 
appreciate the comment. is your current project complex enough. have u made several changes using the new declarative schema.

on the supabase website they claim that the declarative schema is still experimental. and they list out many usecases where the declarative schema fails as well.

 ill stick to diffs and  migrations since that's been out for a while now.

but apparently those aren't good enough either... I hope they stick to one tool and make it robust

longtimedeveloper:
@xidsyed Yes I have made migrations after the initial push.

I have not run into any of the limited edge cases they call out. 

If I need a policy change, I would drop the policy and then create it again in the `supabase/schemas/my_sql_file.sql`  This avoids the lack of alter policy support.

[...]

Issue #3379, opened 2025-04-01
Addition of `pgmq` extension not included in migration file made with `supabase db diff`


SalmonMode:
**Describe the bug**
After enabling the pgmq extension through the UI on my local environment, it does not include the extension in the migration I create right after using `supabase db diff -f some_name`

**To Reproduce**
1. enable the pgmq extension through the UI
2. Run `supabase db diff`
3. No diff will be found

**Expected behavior**
It would include the extension in the migration

**Screenshots**
n/a

**System information**
Rerun the failing command with `--create-ticket` flag.
 - Ticket ID: no ticket ID was produced
 - Version of OS: Ubuntu 22.04
 - Version of CLI: 2.20.5
 - Version of Docker: 27.5.1
 - Versions of services: 

        SERVICE IMAGE      │      LOCAL       │   LINKED
  ─────────────────────────┼──────────────────┼─────────────
    supabase/postgres      │ 15.8.1.054       │ 15.8.1.054
    supabase/gotrue        │ v2.170.0         │ v2.170.0
    postgrest/postgrest    │ v12.2.3          │ v12.2.3
    supabase/realtime      │ v2.34.43         │ -
    supabase/storage-api   │ v1.17.1          │ -
    supabase/edge-runtime  │ v1.67.4          │ -
    supabase/studio        │ 20250317-6955350 │ -
    supabase/postgres-meta │ v0.87.1          │ -
    supabase/logflare      │ 1.12.0           │ -
    supabase/supavisor     │ 2.4.14           │ -

**Additional context**
If applicable, add any other context about the problem here.
 - Version of supabase-js 2.47.2
 - Version of Node.js: 20.18.3

Issue #3342, opened 2025-03-23
`supabase db pull` ignores columns starting with `__xxx`


softmarshmallow:
supabase db pull ignores the column starting with double underscore. is this a feature or a bug?

for Example,

remote schema:
```sql
create table grida_canvas.canvas_document (
  created_at timestamp with time zone not null default now(),
  data jsonb not null,
  id uuid not null,
  __schema_version text not null default '0.0.1-beta.1+20250303'::text,
  constraint canvas_document_pkey primary key (id),
  constraint canvas_document_id_key unique (id),
  constraint canvas_document_id_fkey foreign KEY (id) references document (id) on delete CASCADE
) TABLESPACE pg_default;
```

local (pulled) schema:
```sql
create table grida_canvas.canvas_document (
  created_at timestamp with time zone not null default now(),
  data jsonb not null,
  id uuid not null,
  constraint canvas_document_id_key unique (id),
  constraint canvas_document_id_fkey foreign KEY (id) references document (id) on delete CASCADE
) TABLESPACE pg_default;
```

the `__schema_version` will be ignored

PR #3994, opened 2025-08-06
fix: diff schemas using pgkit migra

sweatybridge:

Bug fix

## What is the new behavior?

Using pgkit/migra solves a few issues:
1. Cross schema references now works properly when multiple schemas are excluded
2. Triggers and RLS policies in managed schemas are now captured automatically
3. Extensions in all schemas can be included (previously was just `extensions` schema)

Uses edge runtime to execute migra also reduces the number of container images needed.

## Additional context

closes https://github.com/supabase/cli/issues/3919
closes https://github.com/supabase/cli/issues/3993
closes https://github.com/supabase/cli/issues/3379
closes https://github.com/supabase/cli/issues/3342
closes https://github.com/supabase/cli/issues/3234
closes https://github.com/supabase/cli/issues/1113
closes https://github.com/supabase/cli/issues/1514
closes https://github.com/supabase/cli/issues/2620

### DIFF
--- a/internal/db/diff/templates/migra.ts
+++ b/internal/db/diff/templates/migra.ts
@@ -1,0 +1,99 @@
+import { createClient } from "npm:@pgkit/client";
+import { Migration } from "npm:@pgkit/migra";
+
+const clientBase = createClient(Deno.env.get("SOURCE"));
+const clientHead = createClient(Deno.env.get("TARGET"));
+const includedSchemas = Deno.env.get("INCLUDED_SCHEMAS");
+
+const managedSchemas = ["auth", "realtime", "storage"];
+const extensionSchemas = [
+  "pg_catalog",
+  "extensions",
+  "pgmq",
+  "tiger",
+  "topology",
+];
+const excludedSchemas = [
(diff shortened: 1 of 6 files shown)


internal/db/diff/templates/migra.ts
```diff
@@ -0,0 +1,99 @@
...
+  "_timescaledb_catalog",
+  "_timescaledb_config",
+  "_timescaledb_debug",
+  "_timescaledb_functions",
+  "_timescaledb_internal",
+  // Managed by Supabase
+  "pgbouncer",
+  "supabase_functions",
+  "supabase_migrations",
+];
```
avallete, line 45: **nitpick**

Since this is shared between go and ts code, I wonder if there's a way to have this unified in a single place somehow to avoid drift when this get updated 🤔 

Something similar to the `//go-embed` directive, maybe by setting a "pure data" file like `diffsettings.json`

avallete (approved):
LGTM ! 🚀

ibash-corpusant:
hi @avallete / @sweatybridge it seems this PR didn't fix #3993 unfortunately. Mind reopening the issue and investigating? Thanks!

PR #4010, opened 2025-08-10
fix: set healthcheck for Mailpit container

aloisklink:

Bug fix for users using podman (and I suspect older versions of Docker too).

The mailpit container's built-in healthcheck does not work with podman, which causes `supabase start` to fail without the `--ignore-health-check` flag.

I'm not 100% sure why this is, but I suspect it's because the mailpit [healthcheck has `--start-interval=1s`][1], and the `--start-interval` option was [only added to Docker Engine since v25.0][2] (or v1.44 of the Docker API), which podman does not yet suppprt (see https://github.com/containers/podman/issues/26505).

Looking at the logs in https://github.com/supabase/cli/issues/3566, this should fix the `supabase_inbucket_chat-monitor container is not ready:` issue reported there.

## What is the new behavior?

Setting an explicit healthcheck matches the healthchecks we set for the other containers, and fixes this issue.

`supabase start` no longer requires the `--ignore-health-check` flag when starting the mailpit container.

Fixes half of: https://github.com/supabase/cli/issues/3566 (only the `supabase_inbucket_chat-monitor container is not ready:` error).

## Additional context

This is the `HEALTHCHECK` command in the mailpit v1.22.3 `Dockerfile`:

https://github.com/axllent/mailpit/blob/c7d7810e687b943272c1c6b9aad07f9c59809a04/Dockerfile#L28

They specified the `internal` as 15 seconds, but I decided to keep it as 10, since every other healthcheck config in this supabase CLI uses 10 seconds.

[1]: https://github.com/axllent/mailpit/blob/d26e317d2507015d4d9fa247207ccd6a68fe645f/Dockerfile#L28
[2]: https://docs.docker.com/reference/dockerfile/#healthcheck


aloisklink:
The failing `CI / Codegen` check seems to be fixed by https://github.com/supabase/cli/pull/4008

sweatybridge (approved):

sweatybridge:
Thank you for the thorough investigation and PR!

PR #4036, opened 2025-08-13
fix: skip config diff to reduce output noise

sweatybridge:

Bug fix

## What is the new behavior?

Full diff is available when running `config push` if user is interested.

## Additional context


avallete (approved):
Might wanna have a specific "config diff" command the same way we have "db diff"

Issue #1325, opened 2023-07-21
Config.toml env substitution isn't working for booleans


goszczynskip:
**Describe the bug**
`env` isn't working for boolean config.toml keys for example `enabled = env(MY_VAR)`.

**To Reproduce**
1. Add following Google OAuth config in `config.toml` file
```
[auth.external.google]
enabled = env(GOOGLE_OAUTH_ENABLED)
client_id = "env(GOOGLE_CLIENT_ID)"
secret = "env(GOOGLE_SECRET)"
```
3. Add `.env` file in the project root with envs
```
GOOGLE_OAUTH_ENABLED=false
```  
3. Try to start supabase via `supabase start` command
5. See error
```
cannot read config in /[redacted]: toml: line 64 (last key "auth.external.google.enabled"): expected value but found "env" instead
```

**Expected behavior**
I can start the project with auth enabled or disabled based on the env value.

**Desktop (please complete the following information):**
 - OS: macOS
 - Version of CLI v1.77.9
 - Version of Node.js (if applicable) v18.14.0

G9000:
@sweatybridge any progress on this issue?
 Error evaluating "env(SUPABASE_AUTH_EXTERNAL_GOOGLE_CLIENT_ID)": environment variable SUPABASE_AUTH_EXTERNAL_GOOGLE_CLIENT_ID is unset.

sweatybridge:
@G9000 that looks like a config error. `env(SUPABASE_AUTH_EXTERNAL_GOOGLE_CLIENT_ID)` is present in your config.toml but the env var is undefined. Did you forget to define the env var?

G9000:
> @G9000 that looks like a config error. `env(SUPABASE_AUTH_EXTERNAL_GOOGLE_CLIENT_ID)` is present in your config.toml but the env var is undefined. Did you forget to define the env var?

I did, it seems not detecting at all. I saw this [thread ](https://github.com/supabase/cli/issues/278)but not working for me

sweatybridge:
> secret = "SUPABASE_AUTH_EXTERNAL_GOOGLE_SECRET"

That's just a static value and not using substitution.

You want to keep `secret = "env(SUPABASE_AUTH_EXTERNAL_GOOGLE_SECRET)"` in config.toml and try the following

```
SUPABASE_AUTH_EXTERNAL_GOOGLE_SECRET=abc supabase start
```

Or 

```
[...]

PR #4054, opened 2025-08-18
feat: allow env function for config.toml to have a default value

pathob:

Feature

Currently, the env function used in the config.toml file only supports reading environment variables without any fallback to a default value if the environment variable is not set:

```toml
secret_key = "env(SECRET_KEY)"
```

## What is the new behavior?

The env function used in the config.toml file now supports default values for environment variables using this syntax:

```toml
secret_key = "env(SECRET_KEY, my_default_secret)"
```

## Additional context

I need this functionality to implement dynamic environments while still being able to have hard-coded values in the config.toml file.

This feature was already wished by another user as I saw here: https://github.com/supabase/cli/issues/1325#issuecomment-2993714389

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -72,6 +72,51 @@ func TestConfigParsing(t *testing.T) {
 		// Run test
 		assert.Error(t, config.Load("", fsys))
 	})
+
+	t.Run("config file with env defaults uses defaults when vars not set", func(t *testing.T) {
+		// Setup in-memory fs
+		fsys := fs.MapFS{
+			"supabase/config.toml": &fs.MapFile{Data: []byte(`
+[auth]
+site_url = "env(SITE_URL, http://localhost:3000)"
+
+[auth.external.github]
+client_id = "env(GITHUB_CLIENT_ID, default_client_id)"
+`)},
+		}
+		config := NewConfig()
+		// Run test
+		assert.NoError(t, config.Load("", fsys))
+		// Check defaults are used
+		assert.Equal(t, "http://localhost:3000", config.Auth.SiteUrl)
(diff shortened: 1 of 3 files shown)


sweatybridge:
> I need this functionality to implement dynamic environments while still being able to have hard-coded values in the config.toml file.

Have you considered using .env files to control this? For eg. you can declare default values in `.env` and commit it to git.

```env
secret_key=my_default_secret
```

For environment specific overrides, add them either to `.env.local` or `.env.prod` file.

To use the prod override, you can set `SUPABASE_ENV=prod` before any CLI command, ie.

```bash
SUPABASE_ENV=prod supabase db push
```

I like your suggestion though, as it's more explicit than relying on the order of override like `.env.local > .env`.

pathob:
Hi @sweatybridge

> Have you considered using .env files to control this? For eg. you can declare default values in `.env` and commit it to git.

Not yet, but from this unmerged PR https://github.com/supabase/cli/pull/3741 I assume that environment variables take precedence over .env files so yes, that would be a workaround to achieve the default value.

> For environment specific overrides, add them either to `.env.local` or `.env.prod` file.
> 
> To use the prod override, you can set `SUPABASE_ENV=prod` before any CLI command, ie.
> 
> ```shell
> SUPABASE_ENV=prod supabase db push
> ```

Thank you for the hint, but with this approach I cannot really implement dynamic environments.

> I like your suggestion though, as it's more explicit than relying on the order of override like `.env.local > .env`.

Great, I hope you consider it for merging, because I also find it more clean in cases like mine. :)

sweatybridge:
> but with this approach I cannot really implement dynamic environments.

Could you elaborate on what you mean by dynamic environments?

PR #4076, opened 2025-08-25
fix: confirm cost before enabling paid mfa addons

sweatybridge:

## What is the new behavior?

## Additional context


sweatybridge:
> FMI how does it work for branches ? Enabling those for branches does also incurs extra cost on top of the project ? Is it included ?

Yup it will incur extra costs if enabled on branches. Probably would recommend disabling in base config and enable only in remotes block.

PR #4082, opened 2025-08-27
feat: support custom query timeout for type gen

sweatybridge:

## What is the new behavior?

## Additional context


avallete (approved):
**LGTM**

It really shouldn’t take that much time though 🤔

If you have a reproducible example (a database schema) that triggers this issue, I’d be happy to create an issue on the postgres-meta side. There’s probably some performance bottleneck we could investigate and optimize in such cases.

yilinjuang:
@avallete I have a supabase project that can reproduce the timeout. How can I share it with you privately?

avallete:
Hey ! Thank's for reaching out ! Yes I have. This allowed us to pinpoint that the cause of the issue was a bug in the introspection queries performed that postgres-meta, those only filter by schema/entity at the end of the query, constructing the CTE for all of the objects, before filtering out only the ones it needs.<br><br>I drafted a PR to adress this: supabase/postgres-meta#980<br><br>Which on my tests reduce typegen time from 22s+ to \~200ms. We'll try to roll that out ASAP.

yilinjuang:
Awesome! Yes this matches what we observed on our end. The more schemas we have the longer the generation takes even when we apply the schema filters. Look forward to the release.

avallete:
Hey @yilinjuang 

Just a quick update that the new version of `postgres-meta` has been released and should help with the typegen issue. It's not been released as the new default into the cli yet but you can try it early on by running:

```bash
echo "v0.91.6" > supabase/.temp/pgmeta-version
```

Before running your local stack, this should fix your local `postgres-meta` version to the one containing the fix. You should then be able to generate your types with:

```bash
npx supabase gen types --lang typescript --local  > database.generated.ts
```

Would love to hear if it solve your use case or any feedbacks.

yilinjuang:
@avallete thanks it works! It now takes 15 seconds to generate. It used to be 10 times. 

But I got another problem, using the latest cli [v2.40.4](https://github.com/supabase/cli/releases/tag/v2.40.4) or [v2.40.6](https://github.com/supabase/cli/releases/tag/v2.40.6), I cannot start supabase locally. Health check of postgrest failed. I filed another [issue](https://github.com/supabase/cli/issues/4126).

PR #4113, opened 2025-09-03
fix: no fallback from TLS to unsecure connection

sweatybridge:

Bug fix

## What is the new behavior?

1. No need to enable TLS when connecting to local postgres
2. Should not fallback to non-TLS when connecting to remote

## Additional context


avallete (approved):
LGTM, just a nitpick for the error message.

hnykda:
(opened a new issue https://github.com/supabase/cli/issues/4142)

PR #4159, opened 2025-09-15
feat: support generating JWT auth token

sweatybridge:

feature

## What is the new behavior?

Easier to use than 3rd party implementation.

```bash
supabase gen bearer-jwt --role service_role --valid-for 1h
```

## Additional context

### DIFF
--- a/cmd/gen.go
+++ b/cmd/gen.go
@@ -10,9 +10,9 @@ import (
 	"github.com/golang-jwt/jwt/v5"
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
+	"github.com/supabase/cli/internal/gen/bearerjwt"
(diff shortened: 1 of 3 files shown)


avallete (approved):

cmd/gen.go
```diff
@@ -118,6 +120,23 @@ Supported algorithms:
 			return signingkeys.Run(cmd.Context(), algorithm.Value, appendKeys, afero.NewOsFs())
 		},
 	}
+
+	claims   config.CustomClaims
+	expiry   time.Time
+	validFor time.Duration
+
+	genTokenCmd = &cobra.Command{
+		Use:   "auth-token",
```
hf, line 129: Not sure I love `auth-token`... We should maybe stick to the "API key" name?

hf (approved):

Issue #4042, opened 2025-08-15
[Bug]: Local anon key persists after full reset, causing persistent "Invalid JWT" errors


shine83:
Hello Supabase Team,

Describe the bug

The local development server (supabase start) consistently generates the same anon key even after a full reset using supabase stop --no-backup, and even after a complete docker system prune -a --volumes.

This stale key is not valid for the local API gateway, causing all requests to Edge Functions to fail with an {"msg":"Invalid JWT"} error. This makes local testing of Edge Functions completely impossible.

To Reproduce

Run supabase start in a project. Note the outputted anon key.

Run supabase stop --no-backup to stop the server and remove the database.

(We also tried a more drastic reset) Run docker system prune -a --volumes and restart Docker Desktop.

Run supabase start again.

Observe that the "new" anon key printed to the console is identical to the key from step 1.

Attempt to call any Edge Function using this key via curl. The request fails with {"msg":"Invalid JWT"}.

Expected behavior

After a full reset (supabase stop --no-backup or a Docker prune), supabase start should generate a completely new and valid anon key that is accepted by the local API gateway.

System information

OS: macOS (Apple Silicon)

Supabase CLI version: [2.34.3]
Additional context

This issue seems to be caused by the JWT secret not being properly reset for the local services, making local development impossible. Thank you for looking into this.

PR #4177, opened 2025-09-18
fix: reset local JWT secret after stop --no-backup to ensure fresh anon key (#4042)

srini-abhiram:
- Remove .supabase/config.json after `supabase stop --no-backup` to reset the local JWT secret.
- Add test to verify config removal and JWT secret reset.
- Ensure `supabase start` generates a new anon key after a full reset.
- Fixes issue #4042 (Edge Functions fail with stale anon key after reset).

Bug fix

Running supabase stop --no-backup only stopped containers and removed Docker volumes.
The local JWT secret in .supabase/config.json was not deleted or reset.
On subsequent supabase start, the CLI reused the old JWT secret, so the anon key stayed the same.
This caused Edge Functions to reject requests with {"msg":"Invalid JWT"} after a full reset refer #4042 

## What is the new behavior?

Running supabase stop --no-backup now also deletes .supabase/config.json, removing the local JWT secret.
On the next supabase start, the CLI generates a new JWT secret and a new anon key.
The anon key is now valid for the local API gateway and Edge Functions after a full reset.
Local development and Edge Functions work reliably after a stop/start cycle.

## Additional context

Added a test case for the jwt secret reset.

### DIFF
--- a/internal/stop/stop.go
+++ b/internal/stop/stop.go
@@ -32,6 +32,14 @@ func Run(ctx context.Context, backup bool, projectId string, all bool, fsys afer
 		return err
 	}
 
+	// Remove local JWT secret if --no-backup is used
+	if !backup {
+		configPath := utils.ConfigPath // typically ".supabase/config.json"
+		if err := fsys.Remove(configPath); err != nil && err != afero.ErrFileNotFound {
+			fmt.Printf("Warning: failed to remove JWT secret config: %v\n", err)
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


sweatybridge:
> The local JWT secret in .supabase/config.json was not deleted or reset.

Are you referring to config.toml? It shouldn't be deleted.

The local auth tokens are always static with a 10 year expiry. Can you reproduce the problem reported originally?

srini-abhiram:
> Can you reproduce the problem reported originally?

I could see that the anon key was the same, but I couldn't reproduce the jwt error.

PR #4179, opened 2025-09-18
feat(auth): introduce `EmailOptional` config for providers

fadymak:
Introduces `EmailOptional` configuration to allow sign-ins/ups when the provider does not return an email address.

Related: https://github.com/supabase/supabase/pull/38788


sweatybridge (commented):
Looks good. Just need to add the new field to default config.toml as documentation https://github.com/supabase/cli/blob/develop/pkg/config/templates/config.toml#L266

fadymak:
Thanks @sweatybridge! Just added the config to the template ✅

PR #4180, opened 2025-09-18
fix: Update signed upload url default expiration to match prod

itslenny:

Bug fix

SIGNED_UPLOAD_URL_EXPIRATION_TIME defaults to 60 seconds, but in production the default is 7200

## What is the new behavior?

set SIGNED_UPLOAD_URL_EXPIRATION_TIME to 7200 to match production environment


itslenny:
duplicate... oops.

itslenny:
Re-opened because the [other PR](https://github.com/supabase/cli/pull/4172) that I thought this was a duplicate of changed the ENV in the migrate job container instead of the storage api container.

PR #4236, opened 2025-10-01
feat: populate pg-meta crypto key in pg-meta and studio

mattrossman:
Companion to https://github.com/supabase/supabase/pull/39041

- Adds `PgmetaCryptoKey` to the studio config, defaulting to "SAMPLE_KEY" to match the [postgres-meta fallback](https://github.com/supabase/postgres-meta/blob/dcb8e9ba9a70f05aab4865144743233ed285db59/src/server/constants.ts#L9)
- Populates studio's `PG_META_CRYPTO_KEY` and pg-meta's `CRYPTO_KEY` with this

Ref AI-153

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -169,12 +169,13 @@ type (
 	}
 
 	studio struct {
-		Enabled      bool   `toml:"enabled" json:"enabled"`
-		Image        string `toml:"-" json:"-"`
-		Port         uint16 `toml:"port" json:"port"`
-		ApiUrl       string `toml:"api_url" json:"api_url"`
-		OpenaiApiKey Secret `toml:"openai_api_key" json:"openai_api_key"`
-		PgmetaImage  string `toml:"-" json:"-"`
+		Enabled         bool   `toml:"enabled" json:"enabled"`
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -169,12 +169,13 @@ type (
...
-		ApiUrl       string `toml:"api_url"`
-		OpenaiApiKey Secret `toml:"openai_api_key"`
-		PgmetaImage  string `toml:"-"`
+		Enabled         bool   `toml:"enabled"`
+		Image           string `toml:"-"`
+		Port            uint16 `toml:"port"`
+		ApiUrl          string `toml:"api_url"`
+		OpenaiApiKey    Secret `toml:"openai_api_key"`
+		PgmetaImage     string `toml:"-"`
+		PgmetaCryptoKey string `toml:"-"`
```
mattrossman, line 178: I figure `PgmetaCryptoKey` belongs here since `PgmetaImage` is already there, but was wondering if it's better to split the pgmeta stuff into its own struct.

mattrossman:
Clearing out old PRs, this one is still relevant I think. Updated to latest changes on `develop`

Issue #4211, opened 2025-09-26
`anon key` & `service_role key` unavailable with `supabase start` and `supabase status`


tomgreef:
**Describe the bug**

Previously, in version `2.40.X`, every time I executed the `supabase start` or `supabase status` commands, I observed the `anon key` and `service_role key`. However, upon upgrading to `2.45.5`, I no longer encounter these values in the terminal. Despite [the documentation](https://supabase.com/docs/guides/local-development/cli/getting-started?queryGroups=access-method&access-method=studio&queryGroups=platform&platform=macos#access-your-projects-services) indicating that these values should be present,I have been unable to locate these keys.

Are there any new methods of accessing these keys?

**To Reproduce**
1. `supabase init`
2. `supabase start`
3. You can also try `supabase status`

**Expected behavior**

I expect to see both the `anon key` and `service_role key` so that I can add them to my environment variables as before.

**Screenshots**

<img width="669" height="307" alt="Image" src="https://github.com/user-attachments/assets/88229043-360c-4f38-bd72-63b75f12ccc3" />

Marley-Mulvin-Broome:
As mentioned [here](https://github.com/supabase/supabase-js/issues/1568#issuecomment-3338107800) you can use `npx supabase status -o env` in the meantime to get the service key.

tomgreef:
Ahh! Thanks for clearing that up @Marley-Mulvin-Broome! Should I close the issue?

sweatybridge:
Yes, please use the publishable and secret keys in place of deprecated anon and service role keys. It's a drop in replacement.

PR #4320, opened 2025-10-16
show api keys with deprecated flag

stylessh:

A lot of users still use Anon and Service Role key, they find it difficult to get those while developing with the CLI as they expect to see them in the `supabase status` command.

e.g [2231](https://github.com/orgs/supabase/discussions/2231), [4211](https://github.com/supabase/cli/issues/4211)

The CLI skips the deprecated fields instead of showing a tag.
Screenshot 2025-10-16 at 18 48 21

## What is the new behavior?

The CLI shows the keys with a `soon deprecating` tag.
Screenshot 2025-10-16 at 19 00 21

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -145,11 +145,11 @@ func Execute() {
 	if err != nil {
 		fmt.Fprintln(utils.GetDebugLogger(), err)
 	}
-	if semver.Compare(version, "v"+utils.Version) > 0 {
-		fmt.Fprintln(os.Stderr, suggestUpgrade(version))
-	}
 	if len(utils.CmdSuggestion) > 0 {
 		fmt.Fprintln(os.Stderr, utils.CmdSuggestion)
+	}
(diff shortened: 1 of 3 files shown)


sweatybridge:
The publishable and secrets are drop in replacement for deprecated anon and served role keys. I don't think there's a reason to use old keys?

Can we set `utils.CmdSuggestion` instead? This way a suggestion to use the new keys is printed on status command, ie.

```golang
utils.CmdSuggestion = "Please use publishable key instead of anon key."' 

```

stylessh:
Screenshot 2025-10-20 at 20 31 48

how about this?

sweatybridge:
That's very overwhelming from a DX perspective. Let's not print show the deprecated keys at all.

stylessh:
You're right, my bad!

I updated the code, deprecated keys are removed now @sweatybridge

PR #4368, opened 2025-10-27
feat: add deploy command to push all changes to linked project

matlin:

Feature

Have to seperately deploy changes to migrations, config, and functions.

## What is the new behavior?

A single command that will deploy all changes to a remote proejct. 

## Additional context

- Adds  `--remote` flag to `status` to get the health of the remote project. 
- Adds `--dry-run` support to all deploy commands (added to functions and config)


pkg/function/batch.go
```diff
@@ -25,6 +25,10 @@ const (
 )
 
 func (s *EdgeRuntimeAPI) UpsertFunctions(ctx context.Context, functionConfig config.FunctionConfig, filter ...func(string) bool) error {
+	return s.upsertFunctions(ctx, functionConfig, false, filter...)
+}
+
+func (s *EdgeRuntimeAPI) upsertFunctions(ctx context.Context, functionConfig config.FunctionConfig, dryRun bool, filter ...func(string) bool) error {
```
sweatybridge: Similarly for UpsertFunctions, I'd prefer we continue using filter variadic arg instead of introducing a new dryRun arg.

sweatybridge (commented):
Just a few high level comments before I dive into the implementation detail.

pkg/function/batch.go
matlin: All the dry-run functionality should use the variadic filter param but let me know if my approach isn't what you had in mind

pkg/function/batch.go
```diff
@@ -25,6 +25,10 @@ const (
 )
 
 func (s *EdgeRuntimeAPI) UpsertFunctions(ctx context.Context, functionConfig config.FunctionConfig, filter ...func(string) bool) error {
+	return s.upsertFunctions(ctx, functionConfig, filter...)
+}
+
+func (s *EdgeRuntimeAPI) upsertFunctions(ctx context.Context, functionConfig config.FunctionConfig, filter ...func(string) bool) error {
```
sweatybridge: ok to undo changes in this file?

cmd/deploy.go
```diff
@@ -0,0 +1,192 @@
...
+	// What to deploy - use direct Bool() since we check via cmd.Flags().Changed()
+	cmdFlags.Bool("include-db", true, "Include database migrations (default: true)")
+	cmdFlags.Bool("include-functions", true, "Include edge functions (default: true)")
+	cmdFlags.Bool("include-config", true, "Include config.toml settings (default: true)")
+
+	// DB push options (from db push command)
+	cmdFlags.BoolVar(&deployDryRun, "dry-run", false, "Print operations that would be performed without executing them")
+	cmdFlags.BoolVar(&deployIncludeAll, "include-all", false, "Include all migrations not found on remote history table")
+	cmdFlags.BoolVar(&deployIncludeRoles, "include-roles", false, "Include custom roles from "+utils.CustomRolesPath)
+	cmdFlags.BoolVar(&deployIncludeSeed, "include-seed", false, "Include seed data from your config")
```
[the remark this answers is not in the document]
matlin: For some of these flags (db-url, linked, password, etc) it seems required for some of the sub commands. Is there a better way to utilize these without explicitly defining them?

PR #4372, opened 2025-10-28
fix: toggle `DENO_NO_PACKAGE_JSON` conditionally

nyannyacha:

Enhancement

## Description

Make CLI conditionally toggle `DENO_NO_PACKAGE_JSON` based on what files are in each function folder.
This only applies when deploying functions locally/Docker and not using the deploy api.

Related supabase/supabase#38402
Towards FUNC-304

### DIFF
--- a/internal/functions/deploy/bundle.go
+++ b/internal/functions/deploy/bundle.go
@@ -61,8 +61,10 @@ func (b *dockerBundler) Bundle(ctx context.Context, slug, entrypoint, importMap 
 	}
 	cmd = append(cmd, function.BundleFlags...)
 
-	usePackageJson := function.ShouldUsePackageJsonDiscovery(entrypoint, importMap, afero.NewIOFS(b.fsys))
-	env := []string{fmt.Sprintf("DENO_NO_PACKAGE_JSON=%t", !usePackageJson)}
+	env := []string{}
+	if !function.ShouldUsePackageJsonDiscovery(entrypoint, importMap, afero.NewIOFS(b.fsys)) {
+		env = append(env, "DENO_NO_PACKAGE_JSON=1")
+	}
 	if custom_registry := os.Getenv("NPM_CONFIG_REGISTRY"); custom_registry != "" {
 		env = append(env, "NPM_CONFIG_REGISTRY="+custom_registry)
 	}
--- a/pkg/function/bundle.go
+++ b/pkg/function/bundle.go
@@ -68,9 +68,8 @@ func (b *nativeBundler) Bundle(ctx context.Context, slug, entrypoint, importMap 
 	cmd := exec.CommandContext(ctx, edgeRuntimeBin, args...)
 	cmd.Stderr = os.Stderr
 	cmd.Stdout = os.Stdout
-	if fsys, ok := b.fsys.(fs.StatFS); ok {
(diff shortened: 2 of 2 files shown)


internal/functions/new/templates/config.toml
```diff
@@ -2,6 +2,7 @@
 [functions.{{ . }}]
 enabled = true
 verify_jwt = true
+use_package_json = false
```
sweatybridge: I'm not sure if this is the right fix. Based on [their example](https://github.com/supabase/supabase/issues/38402#issuecomment-3298859419), the error seems to be caused by [health function](https://github.com/Johnny5AtGithub/supabase-todos-demo/blob/master/supabase/functions/health/index.ts#L4) failing to import package.json on Windows.

Most likely we have a bug in path resolution that's preventing package.json at root repo directory from being volume mounted into the edge runtime container.

internal/functions/deploy/bundle.go
```diff
@@ -25,7 +25,7 @@ func NewDockerBundler(fsys afero.Fs) function.EszipBundler {
 	return &dockerBundler{fsys: fsys}
 }
 
-func (b *dockerBundler) Bundle(ctx context.Context, slug, entrypoint, importMap string, staticFiles []string, output io.Writer) (function.FunctionDeployMetadata, error) {
+func (b *dockerBundler) Bundle(ctx context.Context, slug, entrypoint, importMap string, staticFiles []string, usePackageJson bool, output io.Writer) (function.FunctionDeployMetadata, error) {
```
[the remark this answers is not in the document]
nyannyacha: Hey @sweatybridge 

I also dislike the lengthening iface signature, but I can't think of a suitable way to pass `usePackageJson`.
Should I define it directly as a field in the `dockerBundler` struct (and any struct that implements the Bundle interface)?

pkg/config/config.go
```diff
@@ -200,11 +200,12 @@ type (
 	FunctionConfig map[string]function
 
 	function struct {
-		Enabled     bool   `toml:"enabled" json:"-"`
-		VerifyJWT   bool   `toml:"verify_jwt" json:"verifyJWT"`
-		ImportMap   string `toml:"import_map" json:"importMapPath,omitempty"`
-		Entrypoint  string `toml:"entrypoint" json:"entrypointPath,omitempty"`
-		StaticFiles Glob   `toml:"static_files" json:"staticFiles,omitempty"`
+		Enabled        bool   `toml:"enabled" json:"-"`
+		UsePackageJson bool   `toml:"-" json:"-"`
```
sweatybridge: ~looks like an old change that can be reverted~ nvm, now i understand

nyannyacha:
Could you guide me on how to correctly reference `usePackageJson` within the `Bundle` method?
You linked https://github.com/supabase/cli/blob/develop/pkg/config/config.go#L324
But ultimately, to use it within the Bundle method, wouldn't we need to add a variadic argument after the arg list? Is that acceptable?

I just need to be able to reference it within the Bundle, and I'd appreciate it if you could let me know any way to do that. 🥹

sweatybridge:
> But ultimately, to use it within the Bundle method, wouldn't we need to add a variadic argument after the arg list? Is that acceptable?

Yea kinda of. Because adding variadic args is backwards compatible.

Let me think about it a bit more too.

sweatybridge (approved):

PR #4381, opened 2025-10-29
feat: `functions download foo --use-api`

johnstonmatt:

Feature [DEVWF-756](https://linear.app/supabase/issue/DEVWF-756/download-functions-without-docker)

- `supabase functions download foo` spins up an edge-runtime container to download the eszip bundle and writes it to the filesystem

## What is the new behavior?

- [x] `supabase functions download foo --use-api` calls the mgmt-api to unbundle the eszip on the server, then returns it as a multipart response and writes it to the filesystem

## Parity

We expect that whether you choose to `--use-api` or not you should end up with the same files on disk.

## Additional context

### DIFF
--- a/internal/functions/download/download.go
+++ b/internal/functions/download/download.go
@@ -273,7 +273,11 @@ func downloadWithServerSideUnbundle(ctx context.Context, slug, projectRef string
 	if err != nil {
 		return err
 	}
-	defer form.RemoveAll()
+	defer func() {
+		if err := form.RemoveAll(); err != nil {
+			fmt.Fprintln(os.Stderr, err)
+		}
+	}()
 
 	// Read entrypoint path from deno2 bundles
 	metadata := bundleMetadata{}
--- a/internal/functions/download/download_test.go
+++ b/internal/functions/download/download_test.go
@@ -174,115 +174,6 @@ func TestRunLegacyUnbundle(t *testing.T) {
 	})
 }
 
-func TestDownloadFunction(t *testing.T) {
-	const slug = "test-func"
-	// Setup valid project ref
-	project := apitest.RandomProjectRef()
-	// Setup valid access token
-	token := apitest.RandomAccessToken(t)
-	t.Setenv("SUPABASE_ACCESS_TOKEN", string(token))
-
-	t.Run("throws error on network error", func(t *testing.T) {
-		// Setup mock api
-		defer gock.OffAll()
-		gock.New(utils.DefaultApiHost).
-			Get("/v1/projects/" + project + "/functions/" + slug).
-			Reply(http.StatusOK).
-			JSON(api.FunctionResponse{Id: "1"})
-		gock.New(utils.DefaultApiHost).
-			Get("/v1/projects/" + project + "/functions/" + slug + "/body").
-			ReplyError(errors.New("network error"))
-		// Run test
-		err := downloadFunction(context.Background(), project, slug, "")
-		// Check error
-		assert.ErrorContains(t, err, "network error")
-	})
-
-	t.Run("throws error on service unavailable", func(t *testing.T) {
-		// Setup mock api
-		defer gock.OffAll()
-		gock.New(utils.DefaultApiHost).
-			Get("/v1/projects/" + project + "/functions/" + slug).
-			Reply(http.StatusOK).
-			JSON(api.FunctionResponse{Id: "1"})
-		gock.New(utils.DefaultApiHost).
-			Get("/v1/projects/" + project + "/functions/" + slug + "/body").
-			Reply(http.StatusServiceUnavailable)
-		// Run test
-		err := downloadFunction(context.Background(), project, slug, "")
-		// Check error
-		assert.ErrorContains(t, err, "Unexpected error downloading Function:")
-	})
-
-	t.Run("throws error on extract failure", func(t *testing.T) {
-		// Setup deno error
-		t.Setenv("TEST_DENO_ERROR", "extract failed")
-		// Setup mock api
-		defer gock.OffAll()
-		gock.New(utils.DefaultApiHost).
-			Get("/v1/projects/" + project + "/functions/" + slug).
-			Reply(http.StatusOK).
-			JSON(api.FunctionResponse{Id: "1"})
-		gock.New(utils.DefaultApiHost).
-			Get("/v1/projects/" + project + "/functions/" + slug + "/body").
-			Reply(http.StatusOK)
-		// Run test
-		err := downloadFunction(context.Background(), project, slug, "")
-		// Check error
-		assert.ErrorContains(t, err, "Error downloading function: exit status 1\nextract failed\n")
-		assert.Empty(t, apitest.ListUnmatchedRequests())
-	})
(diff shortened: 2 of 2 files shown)


internal/functions/download/download.go
```diff
@@ -238,3 +246,134 @@ deno_version = 2
...
+
+	if !strings.HasPrefix(mediaType, "multipart/") {
+		return errors.Errorf("expected multipart response, got %s", mediaType)
+	}
+
+	// Create function directory
+	funcDir := filepath.Join(utils.FunctionsDir, slug)
+	if err := fsys.MkdirAll(funcDir, 0755); err != nil {
+		return errors.Errorf("failed to create function directory: %w", err)
+	}
```
sweatybridge:
```suggestion
	if err := utils.MkdirIfNotExistFS(fsys, funcDir); err != nil {
		return err
	}
```

we have a little util to handle errors consistently.

internal/functions/download/download.go
```diff
@@ -238,3 +260,128 @@ deno_version = 2
...
+		if err != nil {
+			return err
+		}
+		if relPath == "" {
+			// Skip parts without filenames
+			continue
+		}
+
+		// Create the full path for the file
+		filePath := filepath.Join(funcDir, relPath)
```
sweatybridge: There seems to be an edge that results in double nested download directory.

1. Create a single `supabase/functions/demo/index.ts` file.

```ts
console.info('server started');
Deno.serve(async (req)=>{
  const { name } = await req.json();
  const data = {
    message: `Hello ${name}!`
  };
  return new Response(JSON.stringify(data), {
    headers: {
      'Content-Type': 'application/json',
      'Connection': 'keep-alive'
    }
  });
});
```

2. Deploy using docker

```bash
go run . functions deploy demo
```

3. Download using api

```bash
go run . functions download --use-api demo
```

4. Observe that we have the following file created.

```bash
supabase/functions/hello-world/functions/hello-world/index.ts
```

This is not reproducible if we include `supabase/functions/demo/deno.json` file together with index.ts

internal/functions/download/download.go
```diff
@@ -238,3 +260,128 @@ deno_version = 2
...
+			return errors.Errorf("failed to read multipart: %w", err)
+		}
+
+		// Determine the relative path from headers to preserve directory structure.
+		relPath, err := resolvedPartPath(slug, part)
+		if err != nil {
+			return err
+		}
+		if relPath == "" {
+			// Skip parts without filenames
```
sweatybridge:
```suggestion
			fmt.Fprintln(utils.GetDebugLogger(), "Skipping parts without filename")
```

Could be worth printing to debug logs when such parts are skipped. It's not actionable and kinda indicative of a bug on our side imo.

internal/functions/download/download.go
```diff
@@ -238,3 +260,128 @@ deno_version = 2
...
+// ensure path in a given function's part metadata stays confined within that function's directory
+func sanitizeRelativePath(slug, raw string) (string, error) {
+	candidate := strings.TrimSpace(raw)
+	if candidate == "" {
+		return "", nil
+	}
+
+	// Normalize to forward slashes before cleaning for consistent handling.
+	candidate = strings.ReplaceAll(candidate, "\\", "/")
+	cleaned := path.Clean(candidate)
```
sweatybridge:
```suggestion
	cleaned := filepath.ToSlash(candidate)
```

any reason not to use ToSlash?

internal/functions/download/download.go
```diff
@@ -238,3 +260,128 @@ deno_version = 2
...
+	// part.FileName() does not allow us to handle relative paths, so we parse Content-Disposition manually
+	cd := part.Header.Get("Content-Disposition")
+	if cd == "" {
+		return "", nil
+	}
+
+	_, params, err := mime.ParseMediaType(cd)
+	if err != nil {
+		return "", errors.Errorf("failed to parse content disposition: %w", err)
+	}
```
sweatybridge:
```suggestion
	_, params, err := mime.ParseMediaType(part.Header.Get("Content-Disposition"))
	if err != nil {
		return "", errors.Errorf("failed to parse content disposition: %w", err)
	}
```

Just a nitpick since we used similar logic in https://github.com/supabase/cli/pull/4381/files#diff-96027856dbd1e73e2cb0c5d85365f4875ef299086d21ceaf98e068477afa14e8R286

internal/functions/download/download.go
johnstonmatt: still no repro for me but I've made this PR more robust now anyway, should we try to track this bug separately?

internal/functions/download/download.go
```diff
@@ -238,3 +256,138 @@ deno_version = 2
...
+
+// joinWithinDir safely joins base and rel ensuring the result stays within base directory
+func joinWithinDir(base, rel string) (string, error) {
+	cleanRel := filepath.Clean(rel)
+	// Be forgiving: treat a rooted path as relative to base (e.g. "/foo" -> "foo")
+	if filepath.IsAbs(cleanRel) {
+		cleanRel = strings.TrimLeft(cleanRel, "/\\")
+	}
+	if cleanRel == ".." || strings.HasPrefix(cleanRel, "../") {
+		return "", errors.Errorf("invalid file path outside function directory: %s", rel)
```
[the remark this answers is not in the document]
johnstonmatt: Confirmed with test case here: [e79c831](https://github.com/supabase/cli/pull/4381/commits/e79c8312264f8814693e9cd8f7947590d60e29cf)

internal/functions/download/download_test.go
```diff
@@ -235,3 +320,413 @@ func TestGetMetadata(t *testing.T) {
...
+		{
+			name:       "prefers relative match",
+			entrypoint: "file:///source/index.ts",
+			filenames:  []string{"source/index.ts", "source/utils.ts"},
+			want:       "source",
+		},
+		{
+			name:       "falls back to absolute match",
+			entrypoint: "file:///src/index.ts",
+			filenames:  []string{filepath.FromSlash("/tmp/project/src/index.ts")},
```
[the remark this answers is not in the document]
laktek: yep, technically possible to happen as users can set entrypoint_path at the time of deploy (either via API or config.toml)

internal/functions/download/download.go
```diff
@@ -238,3 +256,342 @@ deno_version = 2
...
+		return errors.Errorf("expected multipart response, got %s", mediaType)
+	}
+
+	// Root directory on disk: supabase/functions/<slug>
+	funcDir := filepath.Join(utils.FunctionsDir, slug)
+	if err := utils.MkdirIfNotExistFS(fsys, funcDir); err != nil {
+		return err
+	}
+
+	bufferedParts, cleanupTemp, err := bufferMultipartParts(resp.Body, params["boundary"], slug, fsys)
```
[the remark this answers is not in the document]
johnstonmatt: each final file path is determined by using the complete list of all filepaths in conjunction with the entrypoint

so we could keep the file contents in memory while we decide where to write them, but this solution saves on memory and ensures that any path related failures prevent writes to non-tmp directories

PR #4391, opened 2025-10-31
fix: drop schema specifier when enabling pgmq

sweatybridge:

Bug fix

## What is the new behavior?

When running db dump, the `pgmq` schema is excluded because it's managed by extension.

Dropping the specifier allows pgmq extension to be created without running into missing schema error.

## Additional context


jgoux (approved):
What bug triggered this change?

sweatybridge:
added PR description, basically the same fix as pgsodium.

PR #4394, opened 2025-11-01
feat(cli): add concurrent images pulling

avallete:

- When running `supabase start` all images are pulled at the time the service is starting, so each services need to wait until it's started before pulling something else. This change the logic into two steps:

1. List all services/images that will be needed concurrently (with a 1s delay between each to reduce the likeliness of hitting AWS ecr rate limit issue), prioritizing the largest images first.
2. Actually start the service with docker (when the image is already there)

On my machine/network, this reduce the "cold start" (no images at all) of `supabase start` from ~2:43.19 total to ~1:30.33 total.

Note that even with all images already present there is a ~30s uncompressible delay to start all the containers. So this only bring the cold start download overhead from ~2.13min to ~1min.


sweatybridge (commented):
Can we check how docker compose downloads images and import that as a library instead? https://github.com/docker/compose/blob/main/pkg/compose/images.go

avallete:
> Can we check how docker compose downloads images and import that as a library instead? https://github.com/docker/compose/blob/main/pkg/compose/images.go

Didn't found any pull related things on the linked file, but here: https://github.com/docker/compose/blob/b80bb0586e42dbb6adc2190cb84881138b61eb53/pkg/compose/pull.go#L312-L331

It seems quite similar to what I've done, I can see they introduce some `maxConcurrency` and doesn't optimize by pull largest images files first (makes sense, we do know which image we pull and what are the approximate size for each, they don't).

I'll see if we can re-use this directly rather than the custom implementation.

sweatybridge:
What about the console logs when pulling images concurrently? Docker compose has a nice display for each image and clears up when downloads complete.

avallete:
> What about the console logs when pulling images concurrently? Docker compose has a nice display for each image and clears up when downloads complete.

Handled this with a single spinner showing progress (nb of remaining images to pull) and a checkmark at the end to show which image failed/succeed.
Screenshot 2025-11-03 at 09 46 47

Screenshot 2025-11-03 at 09 47 49


I've looked at the docker-compose repo, and it's not built as a library, the pull logic seems purely internal so I think we need to come up with our own implementation here.

sweatybridge:
Hmm ok, I will explore a bit more if you don't mind. Their progress logs is something I've always wanted to try.

avallete:
> Hmm ok, I will explore a bit more if you don't mind. Their progress logs is something I've always wanted to try.

If it's only the progress bar you're interested in, then maybe that's doable, it seems like it's handled on it's own separate package in their codebase.

avallete:
Made a new version keeping the concurrency logic in our code (since it's not exposed by compose). But improving the implementation by looking how compose does it (errgroup, channel).

Also added a 1s delay at the start between each image pull, this reduce greatly the number of "api rate limit" encountered which overall reduce the total pulling time.

And re-use part of compose for better logging (docker-compose style):

<div>
    <a href="https://www.loom.com/share/6a775f3733f3417cb23dca250f82bde0">
      <p>Parallel image pull - Docker-compose style - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/6a775f3733f3417cb23dca250f82bde0">
      
    </a>
  </div>
  
  
 Also thank's for this new more detailed log (time for each image) we can see the largest image (postgres) takes ~1min to download, which match our total time of running from cached image (30s) + network speed (1min to download the largest).

avallete:
Closing in favor of: https://github.com/supabase/cli/pull/4447

Which branch from https://github.com/supabase/cli/pull/4394/commits/3b6ecd499b3fafa7a21122c8cf8c40f6aff95cec before the attempt to directly use the docker-compose approach.

Reason are:

1. Couldn't figure out a way to have the docker-compose used as a library to have image pulling retry atomics (only allowed to retry pulling all images of service if one of them encounter an error during pulling)
2. Couldn't distinguish between "fatal" and "retryable" errors while I can with custom implementation.
3. Couldn't have the "pull the largest ones in priority" even though this is just a nitpick, it still makes things between 5-10% faster compare to non prioritized pulling of docker-compose during my testing.

sweatybridge:
Revived this PR since we've found a solution to the following issues

> 1. Couldn't figure out a way to have the docker-compose used as a library to have image pulling retry atomics (only allowed to retry pulling all images of service if one of them encounter an error during pulling)

This was due to `ImageInspect` not wrapped in retry client. Adding retries helped avoid early exits due to rate limited error.

> 2. Couldn't distinguish between "fatal" and "retryable" errors while I can with custom implementation.

We now distinguish fatal and retryable error by string matching on the error message.

> 3. Couldn't have the "pull the largest ones in priority" even though this is just a nitpick, it still makes things between 5-10% faster compare to non prioritized pulling of docker-compose during my testing.

I think in the long run, we want to parallelise starting of container with downloading of images. So the e2e latency could be further improved when we move to using docker compose up.

sweatybridge (approved):

PR #4420, opened 2025-11-07
feat: support append mode when updating network restrictions

maxmetcalfe:

API improvement

Currently, running `network-restrictions update` will replace all existing restrictions with the provided list. This makes it easy to wipe all existing restrictions when adding a new one.

## What is the new behavior?

`network-restrictions update` will now add the provided restrictions to the existing list. This matches the behavior of the "Add restriction" in the Supabase UI.

### DIFF
--- a/internal/restrictions/update/update.go
+++ b/internal/restrictions/update/update.go
@@ -12,41 +12,31 @@ import (
 
 // Run updates the network restriction lists using the provided CIDRs.
 func Run(ctx context.Context, projectRef string, dbCidrsToAllow []string, bypassCidrChecks bool, appendMode bool) error {
+	// 1. separate CIDR to v4 and v6
+	body := api.V1UpdateNetworkRestrictionsJSONRequestBody{
+		DbAllowedCidrs:   &[]string{},
+		DbAllowedCidrsV6: &[]string{},
+	}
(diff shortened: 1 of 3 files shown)


sweatybridge (changes requested):
This would be a breaking change for those who rely on the current behaviour. Could you introduce a new `--append` flag to the update command instead?

maxmetcalfe:
thanks @sweatybridge. updated!

internal/restrictions/update/update.go
```diff
@@ -4,44 +4,109 @@ import (
...
+// Run updates the network restriction lists using the provided CIDRs.
+func Run(ctx context.Context, projectRef string, dbCidrsToAllow []string, bypassCidrChecks bool, appendMode bool) error {
+	newCidrsBody, err := buildRequestBody(dbCidrsToAllow, bypassCidrChecks)
+	if err != nil {
+		return err
+	}
+
+	var body api.V1UpdateNetworkRestrictionsJSONRequestBody
+
+	if appendMode {
```
sweatybridge: Instead of updating the network list client side, let's use the PATCH endpoint for updating network restrictions atomically https://api.supabase.com/api/v1#tag/projects/patch/v1/projects/{ref}/network-restrictions

PR #4471, opened 2025-11-19
fix(reset): ensure pgmq extension is fully rebuilt during remote resets

7ttp:
```
ERROR: relation "pgmq.q_<queue>_msg_id_seq" does not exist
```

**Related issue:** https://github.com/supabase/cli/issues/4492

---

## 🐛 Root cause

Preview-branch resets can leave the `pgmq` extension in a **partial / inconsistent state**, where some extension-managed objects (tables or sequences) are missing.  
When migrations later execute `pgmq.create(...)`, the extension expects these objects to exist and fails with missing-relation errors.

---

## ✅ What this change does

During **remote resets only**, this PR:

- Checks whether the `pgmq` extension exists  
- If present, runs:  
  `DROP EXTENSION IF EXISTS pgmq CASCADE`
- Recreates it with:  
  `CREATE EXTENSION IF NOT EXISTS pgmq`
- Uses a short bounded retry to tolerate transient lock or catalog-contention issues  
- Then continues with the standard `down.ResetAll(...)` migration flow

This ensures the PGMQ extension is always in a **known-good, consistent state** before migrations run, eliminating missing-sequence errors.

---

## 📌 Notes

- Change is scoped **only** to the *remote / preview-branch* reset path  
- Local development resets **are unaffected**  
- If the remote Postgres instance disallows extension creation or removal, the reset will fail with a clear error — this is expected and indicates a platform-side configuration issue

---

## 🔧 What kind of change does this PR introduce?

**Bug fix**

---

## 🚨 Current behavior

Preview-branch resets may leave the PGMQ extension in a partial state, causing migrations such as:

```sql
SELECT pgmq.create('my_queue');
SELECT pgmq.create('my_queue_dlq');
```

to fail with missing-relation errors.

More details: https://github.com/supabase/cli/issues/4492

---

## 🎉 New behavior

Preview-branch resets now consistently rebuild the PGMQ extension, ensuring all `pgmq.create(...)` migrations run successfully without errors.

---

## 📎 Additional context

- Issue occurs **only** on preview-branch resets (first builds and new commits work normally)  
- Verified locally using a disposable Postgres container

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/fsnotify/fsnotify v1.9.0
 	github.com/getsentry/sentry-go v0.38.0
 	github.com/go-errors/errors v1.5.1
-	github.com/go-git/go-git/v5 v5.16.3
+	github.com/go-git/go-git/v5 v5.16.4
 	github.com/go-playground/validator/v10 v10.28.0
 	github.com/go-viper/mapstructure/v2 v2.4.0
 	github.com/go-xmlfmt/xmlfmt v1.1.3
--- a/go.sum
+++ b/go.sum
@@ -316,8 +316,8 @@ github.com/go-git/go-billy/v5 v5.6.2 h1:6Q86EsPXMa7c3YZ3aLAQsMA0VlWmy43r6FHqa/UN
 github.com/go-git/go-billy/v5 v5.6.2/go.mod h1:rcFC2rAsp/erv7CMz9GczHcuD0D32fWzH+MJAU+jaUU=
 github.com/go-git/go-git-fixtures/v4 v4.3.2-0.20231010084843-55a94097c399 h1:eMje31YglSBqCdIqdhKBW8lokaMrL3uTkpGYlE2OOT4=
 github.com/go-git/go-git-fixtures/v4 v4.3.2-0.20231010084843-55a94097c399/go.mod h1:1OCfN199q1Jm3HZlxleg+Dw/mwps2Wbk9frAWm+4FII=
-github.com/go-git/go-git/v5 v5.16.3 h1:Z8BtvxZ09bYm/yYNgPKCzgWtaRqDTgIKRgIRHBfU6Z8=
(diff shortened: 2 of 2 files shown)


sweatybridge:
You should enable the pgmq extension from your migration files. For eg. `supabase/migrations/<timestamp>_name.sql`

```sql
create extension if not exists pgmq;

select pgmq.create('my_queue');
```

7ttp:
thanks @sweatybridge for the heads-up! I haven’t added the migration yet but I’ll include the pgmq extension setup:

CREATE EXTENSION IF NOT EXISTS pgmq;
SELECT pgmq.create('my_queue');

I’ll push the update shortly.

sweatybridge:
Cool, all extension related setup should be enabled through user created migration files. I will probably close this PR if you can confirm that the above comment works.

ToJen:
@sweatybridge unfortunately adding create extension in the migrations is not enough. it will still fail after a reset. @7ttp what was the outcome for you?

7ttp:
i tested it again just to be sure, adding `create extension if not exists pgmq` in the migrations wasn’t enough on my side either. after a remote/preview reset the extension objects were still in a partial state, so the next `pgmq.create(...)` call failed with the same missing-relation error.  @ToJen @sweatybridge

ToJen:
@sweatybridge

7ttp:
@ToJen @sweatybridge  I’m working on the fixes now. I’ll clean up the gofmt formatting, add the missing migration so the recommendation is covered, and update the PR to align with the feedback. I’ll push the updates shortly.

7ttp:
@sweatybridge Could you take a look at this when you have a moment?

sweatybridge:
I don't think it's the CLI's responsibility to customise user's migrations for any specific extensions. So I will go ahead and close this PR.

I will follow up on the original issue to resolve any errors you are facing https://github.com/supabase/cli/issues/4492

PR #4489, opened 2025-11-23
fix: enable es256 jwt signing algorithm support for gotrue

### LINKED ISSUE #4488: Local Supabase setup unable to take in `ES256` signing keys configuration
**Describe the bug**
I'm trying to setup a process to impersonate users locally. reading this source: https://supabase.com/docs/guides/auth/signing-keys#how-to-create-mint-jwts-if-access-to-the-private-key-or-shared-secret-is-not-possible
it seems that the general steps are:
- generate signing keys with `supabase gen signing-key --algorithm ES256`
- update `config.toml` to point to file
- restart Supabase
- observation: sessions in web app are actually invalid now and I have to re-login (so it seems like Supabase taking the new signing keys?)
- using the jwt generated from `supbase gen bearer-jwt` in the code blow, returns the Supabase error: `AuthApiError: invalid JWT: unable to parse or verify signature, token signature is invalid: signing method ES256 is invalid`
  code:
  ```js
  const supabase = createClient(
    env.SUPABASE_URL,
    env.SUPABASE_ANON_KEY,
    {
      global: {
        headers: {
          Authorization: `Bearer ${token}`,
        },
      },
    },
  )
  const {
    data: { user },
    error,
  } = await supabase.auth.getUser()
  console.log(error) // AuthApiError: invalid JWT: unable to parse or verify signature, token signature is invalid: signing method ES256 is invalid
  ```
- it seems that Supabase is still using `HS256`?

reading https://github.com/supabase/cli/issues/4098#issuecomment-3237731745, it seems there are some issues in Supabase around this?

to validate, I wrote a script to generate JWTs using `HS256`, and the generated JWT actually worked!

can you confirm that this is an issue with Supabase?

**System information**
 - Version of OS: MacOS, 15.6.1
 - Version of CLI: 2.58.5
 - Version of Docker: 28.4.0
 - Versions of services:
 ```
  ------------------------|------------------------|------------
   supabase/postgres      | 17.6.1.021             | 17.6.1.021 
   supabase/gotrue        | v2.180.0               | v2.182.1   
   postgrest/postgrest    | v13.0.5                | v13.0.5    
   supabase/realtime      | v2.63.0                | -          
   supabase/storage-api   | v1.29.0                | -          
   supabase/edge-runtime  | v1.69.23               | -          
   supabase/studio        | 2025.11.10-sha-5291fe3 | -          
   supabase/postgres-meta | v0.93.1                | -          
   supabase/logflare      | 1.25.3                 | -          
   supabase/supavisor     | 2.7.4                  | -  
 ```
- @supabase/supabase-js: 2.81.1
- node: v22.18.0

7ttp:
bug fix

gotrue rejects es256 signed jwts with "signing method es256 is invalid" error despite proper signing key configuration in config.toml. the issue occurs because gotrue defaults to only accepting hs256 tokens without explicit algorithm configuration.

gotrue now receives gotrue_jwt_valid_methods environment variable during startup with all configured signing algorithms (hs256, rs256, es256). the fix detects algorithms from signing keys configuration and automatically includes them in the valid methods list. hs256 is always included for backward compatibility.

resolves #4488


internal/start/start.go
```diff
@@ -531,9 +531,22 @@ EOF
...
+					algSet["RS256"] = true
+				case config.AlgES256:
+					algSet["ES256"] = true
+				}
+			}
+			algorithms := make([]string, 0, len(algSet))
+			for alg := range algSet {
+				algorithms = append(algorithms, alg)
+			}
+			env = append(env, "GOTRUE_JWT_VALID_METHODS="+strings.Join(algorithms, ","))
```
sweatybridge:
```suggestion
			if len(utils.Config.Auth.SigningKeys) > 0 {
				env = append(env, "GOTRUE_JWT_VALID_METHODS=RS256,ES256")
			}
```

I think it's simpler to list all supported algorithms.

Also `HS256` only applies to legacy JWT keys which has been deprecated. So we don't need to include it in the list of valid algs.

PR #4513, opened 2025-11-25
fix: add schema qualification hints for extension type errors

### LINKED ISSUE #4572: Ltree extension should be in extension or public schema?
# Bug report

## Describe the bug

I have a list of migrations where in the begining the 1st migration enables all the extensions. Among several extensions I needed `ltree` extention. 
Now unlike other extension where it is supposed to defined in extension schema, there is ambiguity in where `tree` should be defined. 
In my second migration I use create a table using one col being type ltree.

**Case 1: Define `ltree` in extension schema** 
- `supabase db reset` works in local db
- `supabase db reset --linked` fails saying `ERROR: type "ltree" does not exist (SQLSTATE 42704)`

**Case 2: Define `ltree` in public schema** 
- `supabase db reset` fails in local db. At the end there is a 502 code

```
2025/11/25 19:30:09 HTTP GET: http://127.0.0.1:54321/storage/v1/bucket
Error status 502: An invalid response was received from the upstream server
```

`--debug` flag doesn't give much information

**Case 3: Define `ltree` in public and extension schema** 
- Both local and remote migrations works!
- Remote DB throws a warning saying 
```
Issue
Extension ltree is installed in the public schema. Move it to another schema.
```

## To Reproduce

1. Create 1st migration file:

```
create extension if not exists "ltree"
with
  schema "extensions"; // Or use `public` to test out the other case
```

2. Create 2nd migration file:
 

```
create table product_taxonomy.categories (
  id text not null,
  locale text not null, 
  version text not null, 
  name text not null, 
  parent_id text,
  tree_path ltree not null, -- hierarchical path (hyphens converted to underscores)
  level integer not null default 0, -- depth in hierarchy (0 = root)
  breadcrumb text not null,
  primary key (id, locale, version),
  -- Self-referential FK ensuring parent exists in same locale & version
  foreign key (parent_id, locale, version) references product_taxonomy.categories (id, locale, version) on delete cascade
);
```

## Expected behavior

Ideally the docs should mention which schema to keep ltree extension in. 

## Screenshots

NaN

## System information

- OS: macOs
- Supabase CLI Version: 2.62.5

7ttp:
## Summary
Improve CLI migration errors for undefined PostgreSQL extension types by adding a context aware hint for unqualified type references.

## Problem
Migrations can fail with:
```
ERROR: type "ltree" does not exist (SQLSTATE 42704)
```

This error does not indicate whether the type is missing due to schema visibility,
 extension installation, or configuration differences. 

## Solution
Inspect pgconn.PgError during migration execution and conditionally provide a schema-qualification hint only when the referenced type is not already schema-qualified.

## Related
 - Closes #4572

### DIFF
--- a/pkg/migration/file_test.go
+++ b/pkg/migration/file_test.go
@@ -120,6 +120,38 @@ func TestMigrationFile(t *testing.T) {
 	})
 }
 
+func TestExtractTypeName(t *testing.T) {
+	t.Run("extracts type name from standard error message", func(t *testing.T) {
+		result := extractTypeName(`type "ltree" does not exist`)
+		assert.Equal(t, "ltree", result)
+	})
+
+	t.Run("extracts schema-qualified type name", func(t *testing.T) {
+		result := extractTypeName(`type "extensions.ltree" does not exist`)
+		assert.Equal(t, "extensions.ltree", result)
+	})
+
+	t.Run("extracts type with underscores", func(t *testing.T) {
+		result := extractTypeName(`type "my_custom_type" does not exist`)
+		assert.Equal(t, "my_custom_type", result)
+	})
+
+	t.Run("returns empty string for non-matching message", func(t *testing.T) {
+		result := extractTypeName(`column "name" does not exist`)
+		assert.Equal(t, "", result)
+	})
+
+	t.Run("returns empty string for empty message", func(t *testing.T) {
+		result := extractTypeName("")
+		assert.Equal(t, "", result)
+	})
+
+	t.Run("handles type names with numbers", func(t *testing.T) {
+		result := extractTypeName(`type "type123" does not exist`)
+		assert.Equal(t, "type123", result)
+	})
+}
+
 func TestIsSchemaQualified(t *testing.T) {
 	assert.True(t, IsSchemaQualified("extensions.ltree"))
 	assert.True(t, IsSchemaQualified("public.my_type"))


7ttp:
@sweatybridge can you confirm if adding extensions to the search_path after each RESET ALL is the right approach to keep types like ltree accessible during migrations? Just want to make sure this aligns with expected behavior before taking the PR out of draft. If you have any suggestions, I’m happy to adjust.

sweatybridge:
Thanks for checking. Our current recommendation is to always use schema qualified references in your migration file. For eg.

```sql
CREATE TABLE test (path extensions.ltree NOT NULL);
```

This avoids ambiguity in default search path resolution which can be configured per session, per role, or per cluster.

The default search path for Supabase postgres role is currently set to [`$user, public, extensions`](https://github.com/supabase/postgres/blob/develop/migrations/db/init-scripts/00000000000003-post-setup.sql#L4). We can't guarantee that it won't be changed in the future. Therefore, I think it's not ideal to make such assumptions in the CLI migration tool.

7ttp:
> Thanks for checking. Our current recommendation is to always use schema qualified references in your migration file. For eg.
> 
> ```sql
> CREATE TABLE test (path extensions.ltree NOT NULL);
> ```
> 
> This avoids ambiguity in default search path resolution which can be configured per session, per role, or per cluster.
> 
> The default search path for Supabase postgres role is currently set to [`$user, public, extensions`](https://github.com/supabase/postgres/blob/develop/migrations/db/init-scripts/00000000000003-post-setup.sql#L4). We can't guarantee that it won't be changed in the future. Therefore, I think it's not ideal to make such assumptions in the CLI migration tool.

@sweatybridge Thanks for the elaboration! I get that relying on a fixed search_path isn’t future-proof.
That said, the current type "ltree" does not exist errors are pretty opaque especially since local and remote migrations behave differently.
instead I can:
update docs to recommend schema qualified types
improve the error message when an extension type isn’t found
optionally add a CLI hint suggesting extensions.ltree
Let me know if that approach is fine and frameable for a pr, I can revise this patch accordingly.

sweatybridge:
> improve the error message when an extension type isn’t found
> optionally add a CLI hint suggesting extensions.ltree

Do you plan to do this by inspecting the error object returned from pgx? If so, that sounds good to me.

Thanks for your help.

7ttp:
> > improve the error message when an extension type isn’t found
> > optionally add a CLI hint suggesting extensions.ltree
> 
> Do you plan to do this by inspecting the error object returned from pgx? If so, that sounds good to me.
> 
> Thanks for your help.

Yep, I’ll inspect the pgx error and add the clearer message + schema qualification hint. On it

7ttp:
@sweatybridge whenever you have a moment, a quick review would be appreciated. 💚

pkg/migration/file.go
```diff
@@ -96,6 +96,20 @@ func (m *MigrationFile) ExecBatch(ctx context.Context, conn *pgx.Conn) error {
...
+				// Extract type name from error message (e.g., 'type "ltree" does not exist')
+				typeName := extractTypeName(pgErr.Message)
+				msg = append(msg, "")
+				msg = append(msg, "Hint: This type may be defined in a schema that's not in your search_path.")
+				msg = append(msg, "      Use schema-qualified type references to avoid this error:")
+				if typeName != "" {
+					msg = append(msg, fmt.Sprintf("        CREATE TABLE example (col extensions.%s);", typeName))
+				} else {
+					msg = append(msg, "        CREATE TABLE example (col extensions.<type_name>);")
+				}
```
sweatybridge:
```suggestion
			if typeName := extractTypeName(pgErr.Message); len(typeName) > 0 && pgErr.Code == "42704" {
				msg = append(msg, "")
				msg = append(msg, "Hint: This type may be defined in a schema that's not in your search_path.")
				msg = append(msg, "      Use schema-qualified type references to avoid this error:")
				msg = append(msg, fmt.Sprintf("        CREATE TABLE example (col extensions.%s);", typeName))
```

7ttp:
@sweatybridge All requested changes have been applied. Ready for rereview.

pkg/migration/file.go
```diff
@@ -26,7 +26,10 @@ type MigrationFile struct {
 	Statements []string
 }
 
-var migrateFilePattern = regexp.MustCompile(`^([0-9]+)_(.*)\.sql$`)
+var (
+	migrateFilePattern = regexp.MustCompile(`^([0-9]+)_(.*)\.sql$`)
+	typeNamePattern    = regexp.MustCompile(`type "([^"]+)" does not exist`)
```
sweatybridge: We should distinguish type not found errors where the schema is already specified. For eg.

```bash
postgres=> CREATE TABLE test (path extensions.ltree NOT NULL);
ERROR:  type "extensions.ltree" does not exist
LINE 1: CREATE TABLE test (path extensions.ltree NOT NULL);
```

In this case, user may not have enabled the `ltree` extension or the extension might be installed in a different schema. So our hint should not suggest using schema qualified reference.

sweatybridge: probably better to define a new method `func IsSchemaQualified(typename string) bool` that accepts the value returned by `extractTypeName`.

7ttp:
@sweatybridge heyy! 
(sorry for the ping on an old PR)
Can u please have another look at this aswell?? 😁💚

sweatybridge:
Can we add some unit tests for the new code please?

7ttp:
> Can we add some unit tests for the new code please?

Added unit tests for extractTypeName function

PR #4517, opened 2025-11-26
fix: ensure app token is correctly scoped

staaldraad:

Bug fix

In [actions/create-github-app-token](https://github.com/actions/create-github-app-token?tab=readme-ov-file#repositories)
> If owner and repositories are empty, access will be scoped to only the current repository.

## What is the new behavior?

Token is scoped to be valid for the required repositories

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,6 +35,7 @@ jobs:
             cli
             homebrew-tap
             scoop-bucket
+            supabase
 
   commit:
     name: Publish Brew and Scoop


.github/workflows/release.yml
```diff
@@ -30,6 +30,11 @@ jobs:
         with:
           app-id: ${{ secrets.APP_ID }}
           private-key: ${{ secrets.GH_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          repositories: |
+            cli
+            homebrew-tap
+            scoop-bucket
```
sweatybridge:
```suggestion
            scoop-bucket
            supabase
```

PR #4545, opened 2025-11-28
feat: support storage s3 protocol config

sweatybridge:

feature

## What is the new behavior?

Enable s3 protocol via config as code.

## Additional context

### DIFF
--- a/internal/start/start.go
+++ b/internal/start/start.go
@@ -244,9 +244,10 @@ func run(ctx context.Context, fsys afero.Fs, excludedContainers []string, dbConf
 	}
 
 	var started []string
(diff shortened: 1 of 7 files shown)


riderx:
@sweatybridge my setup got broken by this change. I like the idea, of making it explicit but a minor version bump with a breaking change in the config is not so nice.

sweatybridge:
Thanks for letting me know and sorry about the breaking change. I will definitely take note for future changes.

I can also revert to enabling s3 protocol by default if that helps.

riderx:
I got it fixed so  not a problem for me, i hope it will not break others too.

PR #4555, opened 2025-11-30
fix: throw error when static_files are used with --use-docker but Docker isn't running

### LINKED ISSUE #4554: static_files ignored when deploying without docker
**Describe the bug**

**To Reproduce**
set `static_files` in config.toml
deploy supabase edge function --use-docker
terminal will say 
```bash
WARNING: Docker is not running
```
if docker is not running

the function deployed does not have the static files.
if docker is runnign it will have them

**Expected behavior**
either throw if static_files is set and docker is not running or support static_files in non docker mode

7ttp:
## prob:
when deploying edge functions with static_files configured and --use-docker is specified while docker is not running, the cli silently falls back to the native bundler. this causes static files to be ignored, since the native bundler requires the edge runtime to be installed locally for static file bundling.

`without --use-docker: ` static files are bundled correctly. with --use-docker and docker stopped, the deploy still succeeds but the static files are missing. so the issue only appears when a user explicitly requests docker bundling but docker isn’t available

## sol: 
this change adds a check that returns an explicit error when static_files is configured, --use-docker is used, and docker is unavailable, instead of silently producing an incomplete deploy.

## ref
- closes #4554


7ttp:
@sweatybridge Please review, when available!

sweatybridge:
Deploying static files in non-docker mode is supported. Can you reproduce the issue and verify?

7ttp:
> Deploying static files in non-docker mode is supported. Can you reproduce the issue and verify?

Thanks for the review @sweatybridge !! retested with the correct paths:

- Without `--use-docker`: static files bundled correctly 
- With `--use-docker` and Docker stopped: deploy succeeds but static files are missing 

So the problem is specifically when a user explicitly requests Docker bundling but Docker isn’t available. The current behaviour silently produces an incomplete deploy. This PR makes that failure explicit by returning an error when `--use-docker` is specified, Docker isn't running, and `static_files` is configured.

sweatybridge:
> With --use-docker and Docker stopped: deploy succeeds but static files are missing

This code path falls back to deploy using API, which is the same as running `supabase functions deploy --use-api`.

If static files don't work with `--use-api` flag, we should find the bug and fix it. Because as I said, this is a supported feature.

sweatybridge (changes requested):
blocking until further investigation

PR #4567, opened 2025-12-03
feat(studio): functions management api - attaching functions folder

kallebysantos:

Feature

## What is the new behaviour?

Its allows Edge Functions Management API to use folder introspection and serve function metadata.
ref: https://github.com/supabase/supabase/pull/40690, [internal RFC](https://www.notion.so/supabase/Self-hosted-Edge-functions-UI-2aa5004b775f80a787f2d80aba78375d?source=copy_link)

## Additional context

<details>
<summary>screenshots</summary>

image

image

</details>

---

I'm keeping it very simple for now. 
Please let me know if a more complex volume binding like this one is needed https://github.com/supabase/cli/blob/cd22bfec95e396d533d2642e9c03d200a447e361/internal/functions/serve/serve.go#L253-L270

cc @sweatybridge


internal/start/start.go
```diff
@@ -1154,6 +1159,7 @@ EOF
 				},
 			},
 			container.HostConfig{
+				Binds: []string{filepath.Join(cwd, utils.FunctionsDir) + ":/app/edge-functions:ro"},
```
sweatybridge: Let's try to use the `binds` directories returned from [populatePerFunctionConfigs](https://github.com/supabase/cli/blob/develop/internal/functions/serve/serve.go#L156) instead. It handles relative imports from `../../../parent.ts` for eg.

```bash
.
+-- parent/index.ts
+-- supabase
|   +-- functions/slug/index.ts
|   +-- config.toml
```

Could you also double check that such functions work in the local studio UI?

internal/start/start.go
```diff
@@ -1125,6 +1125,10 @@ EOF
 
 	// Start Studio.
 	if utils.Config.Studio.Enabled && !isContainerExcluded(utils.Config.Studio.Image, excluded) {
+		cwd, err := os.Getwd()
+		if err != nil {
+			return errors.Errorf("failed to get working directory: %w", err)
+		}
```
sweatybridge: This can be refactored and reused with https://github.com/supabase/cli/blob/develop/internal/start/start.go#L275

sweatybridge (changes requested):

internal/start/start.go
kallebysantos: Hey Qiao! I was able to get the volume binds 💚
But currently, the studio is strict to only look functions in `./**/index.ts` way  — _I didn't happen to think of that other possibility at the time_ 😅

Since you mentioned this behaviour, 
it totally makes sense to add `SUPABASE_INTERNAL_FUNCTIONS_CONFIG` support for Studio, and I should probably do it in near future 🥹

For now, I think that its ok to only show what's inside `supabase/functions` folder. 
Thx a lot for pointing it 💚

PR #4591, opened 2025-12-07
fix: remove inline `jsr:` import from generated functions template

7ttp:
## Summary

The template generated by `supabase functions new` produces code that triggers Deno lint errors due to an inline `jsr:` import without a version specifier.

## Problem

The generated template includes:
```ts
```

This violates two default Deno lint rules:
- **no-import-prefix** – inline `jsr:` dependencies are not allowed
- **no-unversioned-import** – missing version in specifier

## Related

- Closes https://github.com/supabase/cli/issues/4424


woodside-david-francoeur:
Thanks, for that, it works for me

IdellHeaney:
+1 thanks @sweatybridge @7ttp

Issue #4266, opened 2025-10-08
Deno 2 deprecation warnings during edge function deployment (dx: low priority)


sbs44:
**Describe the bug**

When deploying edge functions, the CLI outputs Deno 2 deprecation warnings about `--import-map` and `--decorator` flags. While deployments succeed, these warnings clutter the output and indicate the CLI is using deprecated Deno command-line flags.

**To Reproduce**

1. Create any edge function with a `deno.json` containing imports
2. Configure `import_map` in `supabase/config.toml`:

   ```toml
   [functions.awesome-function]
   enabled = true
   verify_jwt = false
   import_map = "./functions/awesome-function/deno.json"
   entrypoint = "./functions/awesome-function/index.ts"
   ```
3. Deploy the function: `supabase functions deploy awesome-function --no-verify-jwt`
4. Observe deprecation warnings in output (see "Actual output", below)

**Expected behavior**

Edge functions should deploy without deprecation warnings (modern Deno 2 reads configuration from `deno.json` files automatically).

**Actual output**

```zsh
Bundling Function: awesome-function
Specifying import_map through flags is no longer supported. Please use deno.json instead.
Specifying decorator through flags is no longer supported. Please use deno.json instead.
Deployed Functions on project abcdefghijklmnopq: awesome-function
```

**System information**

Rerun the failing command with `--create-ticket` flag.
 - Ticket ID: (FYI: adding "--create-ticket" flag does not output ticket ID)
 - Version of OS: macOS Tahoe 26.0.1
 - Version of CLI: 2.50.2 (beta)
 - Version of Docker: 28.4.0
 - Versions of services:

```zsh
   SERVICE IMAGE          | LOCAL                  | LINKED     
  ------------------------|------------------------|------------
   supabase/postgres      | 17.6.1.005             | 17.6.1.005 
   supabase/gotrue        | v2.179.0               | v2.179.0   
   postgrest/postgrest    | v13.0.5                | v13.0.5    
   supabase/realtime      | v2.52.0                | -          
   supabase/storage-api   | v1.28.0                | -          
   supabase/edge-runtime  | v1.69.12               | -          
   supabase/studio        | 2025.10.06-sha-426fda2 | -          
   supabase/postgres-meta | v0.91.6                | -          
   supabase/logflare      | 1.22.4                 | -          
   supabase/supavisor     | 2.7.1                  | -          
```

**Root Cause**

The issue originates from hardcoded flags in `pkg/function/bundle.go`:

[...]

PR #4599, opened 2025-12-09
fix: remove deprecated Deno 2 flags to eliminate warnings

avallete:
Deno 2 emits deprecation warnings when using --import-map and --decorator flags, as it now auto-discovers deno.json/deno.jsonc files and uses TC39 decorators by default.

Changes:
- Remove --decorator tc39 flag (TC39 is now the default in Deno 2)
- Skip --import-map flag for deno.json/deno.jsonc files (Deno 2 auto-discovers)
- Keep --import-map flag for legacy import_map.json files (backwards compatible)

Fixes: #4266

### DIFF
--- a/.github/workflows/api-sync.yml
+++ b/.github/workflows/api-sync.yml
@@ -47,7 +47,7 @@ jobs:
       - name: Create Pull Request
         if: steps.check.outputs.has_changes == 'true'
         id: cpr
-        uses: peter-evans/create-pull-request@v7
+        uses: peter-evans/create-pull-request@v8
         with:
           token: ${{ steps.app-token.outputs.token }}
           commit-message: "chore: sync API types from infrastructure"
--- a/cmd/gen.go
+++ b/cmd/gen.go
@@ -60,6 +60,7 @@ var (
 			types.LangTypescript,
 			types.LangGo,
 			types.LangSwift,
+			types.LangPython,
 		},
 		Value: types.LangTypescript,
 	}
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -2,8 +2,6 @@ package cmd
 
 import (
 	"fmt"
-	"os"
-	"os/signal"
 
 	"github.com/spf13/afero"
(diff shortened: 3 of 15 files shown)


internal/functions/serve/templates/main.ts
```diff
@@ -187,11 +187,6 @@ Deno.serve({
     const cpuTimeSoftLimitMs = 1000;
     const cpuTimeHardLimitMs = 2000;
 
-    // NOTE(Nyannyacha): Decorator type has been set to tc39 by Lakshan's request,
-    // but in my opinion, we should probably expose this to customers at some
-    // point, as their migration process will not be easy.
-    const decoratorType = "tc39";
```
sweatybridge: Feel like we should leave it in as both deno 1 and 2 share the same main.ts file for serving locally.

cc @nyannyacha wdyt?

internal/functions/deploy/bundle.go
```diff
@@ -50,8 +50,17 @@ func (b *dockerBundler) Bundle(ctx context.Context, slug, entrypoint, importMap
 	hostOutputPath := filepath.Join(hostOutputDir, "output.eszip")
 	// Create exec command
 	cmd := []string{"bundle", "--entrypoint", utils.ToDockerPath(entrypoint), "--output", utils.ToDockerPath(hostOutputPath)}
+	// Handle import map/config flags based on Deno version
+	// Deno 2: use --config for deno.json files, --import-map for legacy import_map.json
+	// Deno 1: use --import-map for all import map files
 	if len(importMap) > 0 {
-		cmd = append(cmd, "--import-map", utils.ToDockerPath(importMap))
+		if utils.Config.EdgeRuntime.DenoVersion > 1 && function.IsDenoJsonImportMap(importMap) {
```
sweatybridge: So on closer look, the import map struct is a subset of deno.json file. So we only need to check the deno version here.

sweatybridge (approved):

PR #4605, opened 2025-12-10
feat(DEVWF-935): add deno script for running pg-delta

sweatybridge:

feature

## What is the new behavior?

```bash
$ supabase db diff --use-pg-delta
```

## Additional context


cmd/db.go
```diff
@@ -257,7 +260,8 @@ func init() {
 	diffFlags.BoolVar(&useMigra, "use-migra", true, "Use migra to generate schema diff.")
 	diffFlags.BoolVar(&usePgAdmin, "use-pgadmin", false, "Use pgAdmin to generate schema diff.")
 	diffFlags.BoolVar(&usePgSchema, "use-pg-schema", false, "Use pg-schema-diff to generate schema diff.")
-	dbDiffCmd.MarkFlagsMutuallyExclusive("use-migra", "use-pgadmin")
+	diffFlags.BoolVar(&usePgDelta, "use-pg-delta", false, "Use pg-delta to generate schema diff.")
+	dbDiffCmd.MarkFlagsMutuallyExclusive("use-migra", "use-pgadmin", "use-pg-schema", "use-pg-delta")
```
avallete: **question**

I guess `pg-schema` was just missing from the list before ?

internal/db/diff/templates/pgdelta.ts
```diff
@@ -0,0 +1,23 @@
...
+try {
+  const result = await createPlan(source, target, supabase);
+  const statements = result?.plan.statements ?? [];
+  for (const sql of statements) {
+    console.log(`${sql};`);
+  }
+} catch (e) {
+  console.error(e);
+  // Force close event loop
+  throw new Error("");
```
sweatybridge: @avallete btw do you know if pgdelta starts anything in the background that doesn't get killed on error? The deno worker exits successfully with migra.ts but not pgdelta.ts

avallete: Had a look up to it but nothing stands out.
Maybe in case of error the db connection needs to be manually closed ? 🤔

sweatybridge: It only hangs on error if I don't bubble it up, for eg. trying to diff a linked project (permission error due to temp role).

Issue #1542, opened 2023-10-02
Listen on 0.0.0.0 instead of localhost


piccinnigius:
**Describe the bug**
I've seed that this issue #103 disabled the possibility to access to supabase from other devices on the network.

**To Reproduce**
1. supabase start
2. try to reach your supabase project from another device
3. Timeout

**Expected behavior**
Supabase instance created by "supabase start" needs to be reachable from any device on the network (maybe with a flag or something inside config.toml)

**Screenshots**

**Desktop (please complete the following information):**
 - Browser (if applicable) [e.g. chrome, safari]
 - Version of CLI [e.g. v1.99.5]
 - Version of supabase-js (if applicable) [e.g. v2.36.0]
 - Version of Node.js (if applicable) [e.g. v18.16.0]

**Additional context**

piccinnigius:
Hi @sweatybridge,

this is my output:

```
CONTAINER ID   IMAGE                                             COMMAND                   CREATED        STATUS                     PORTS                                                                                            NAMES
3adcfa5b4ccc   public.ecr.aws/supabase/studio:20230912-748fd33   "docker-entrypoint.s…"    9 days ago     Up 11 minutes (healthy)    0.0.0.0:54323->3000/tcp, :::54323->3000/tcp                                                      supabase_studio_Pyramid
0bc25ec645d1   public.ecr.aws/supabase/postgres-meta:v0.68.0     "docker-entrypoint.s…"    9 days ago     Up 11 minutes (healthy)    8080/tcp                                                                                         supabase_pg_meta_Pyramid
245895a6c2ae   public.ecr.aws/supabase/edge-runtime:v1.18.1      "sh -c 'mkdir -p /ho…"    9 days ago     Exited (255) 8 days ago    8081/tcp                                                                                         supabase_edge_runtime_Pyramid
f62c72c9507b   public.ecr.aws/supabase/imgproxy:v3.8.0           "imgproxy"                9 days ago     Up 11 minutes (healthy)    8080/tcp                                                                                         storage_imgproxy_Pyramid
5cc0315f3ace   public.ecr.aws/supabase/storage-api:v0.40.4       "docker-entrypoint.s…"    9 days ago     Up 11 minutes (healthy)    5000/tcp                                                                                         supabase_storage_Pyramid
[...]

Issue #1397, opened 2023-08-21
Change docker ip in port from 0.0.0.0 for supabase start local development.


0xBigBoss:
is there anyway to change the binding IP for local supabase from 0.0.0.0 to 127.0.0.1? 

![image](https://github.com/supabase/cli/assets/95193764/8f0ce163-7527-4443-abf9-11fafb66511f)
 
supabase cli is also not respecting the /etc/docker/daemon.json default IP

![image](https://github.com/supabase/cli/assets/95193764/2a8772bd-8d78-4d0a-85ec-0959e1d24c4a)

daniel-j-h:
Coming from https://github.com/supabase/cli/issues/1977

When running supabase start or supabase db start it looks like services are getting exposed on the network, such as the postgres database. I believe this is because when we start containers they bind to all network interfaces instead of just localhost

**This is both unexpected and dangerous to have services exposed publicly on the network.**

Especially because the locally running services are not secured in any way: someone in the same cafe Wifi as I am can run `psql` against my host and is in without any credentials

    $ psql --host 192.168.2.101 --port 54322 --user postgres

    psql (15.5 (Ubuntu 15.5-1.pgdg20.04+1), server 15.1 (Ubuntu 15.1-1.pgdg20.04+1))
    Type "help" for help.

    postgres=>

This works, I just tested it from a second computer to get access to my laptop's supabase stack running from `supabase start`

---

Ideally we'd expose services only on localhost - the equivalent of

    docker run -p 127.0.0.1:54322:54322

Here's an idea how it could work
- By default all sub-commands starting services start them on 127.0.0.1 only
- Sub-commands starting services get a flag like `--host 0.0.0.0` or a binary flag like `--expose-publicly`

jmath3912:
any updates regarding this? i find it odd that supabase says its listening on localhost when its really listening on all interfaces

sweatybridge:
Binding only to localhost interface is now possible in beta release with custom docker network support. For eg.

```bash
docker network create -o 'com.docker.network.bridge.host_binding_ipv4=127.0.0.1' local-network
npx supabase@beta start --network-id local-network
```

ChanningJacobs:
I'm concerned that this documentation may be incorrect for Docker Desktop on MacOS. I followed the Supabase Local Dev & CLI [guide](https://supabase.com/docs/guides/local-development). I'm still seeing ports are bound to 0.0.0.0 on my host after running:

```
[...]

PR #4613, opened 2025-12-12
feat/bind to local

doublethink:
Proposal to bind CLI services to local IPs (while being configurable). See Linear ticket for context.

Fixes SEC-620

### DIFF
--- a/internal/db/start/start.go
+++ b/internal/db/start/start.go
@@ -118,7 +118,7 @@ EOF`}
 func NewHostConfig() container.HostConfig {
 	hostPort := strconv.FormatUint(uint64(utils.Config.Db.Port), 10)
 	hostConfig := container.HostConfig{
-		PortBindings:  nat.PortMap{"5432/tcp": []nat.PortBinding{{HostPort: hostPort}}},
+		PortBindings:  nat.PortMap{"5432/tcp": []nat.PortBinding{{HostIP: utils.Config.Db.HostIP, HostPort: hostPort}}},
 		RestartPolicy: container.RestartPolicy{Name: container.RestartPolicyUnlessStopped},
 		Binds: []string{
 			utils.DbId + ":/var/lib/postgresql/data",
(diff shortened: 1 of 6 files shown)


sweatybridge:
We addressed this before. Just need to document how it works https://github.com/supabase/cli/issues/1397#issuecomment-2153837111

doublethink:
> We addressed this before. Just need to document how it works [#1397 (comment)](https://github.com/supabase/cli/issues/1397#issuecomment-2153837111)

@sweatybridge is the behavior on by default?

sweatybridge:
If you mean whether we are only binding to localhost by default, the decision is no. It was done for convenience of testing local mobile apps. https://github.com/supabase/cli/issues/1542 And it would be a breaking change to revert this for existing users.

The CVE you mentioned is mostly concerned with folks exposing CLI to the internet. This is a misuse of CLI. In local environments, user laptops are connected to their private trusted network most of the time. Hence, I'd suggest keeping the current behaviour and adding a warning to our docs:

**If you are running CLI in a public network that you don't trust, you should create a separate docker network and bind to 127.0.0.1 before starting the local stack.**

```bash
docker network create -o 'com.docker.network.bridge.host_binding_ipv4=127.0.0.1' local-network
npx supabase@beta start --network-id local-network
```

I believe we should lean on docker network because it's more robust than adding our own custom networking config.

doublethink:
@sweatybridge was security asked about this change or was this made in isolation?
its important to note that there is nothing in our documentation warning user about exposing cli services externally, i don't think we can say this is a misuse if we aren't cautioning users.
keen for @aantti to weigh in on this decision?

sweatybridge:
Binding to 0.0.0.0 is not a new change. The "new" change introduced 2 years ago was to support custom docker network for users who don't want to bind to 0.0.0.0.

I'm happy to review this decision since we now have a security team. But I think the first step should be to update our docs to warn users about exposing services externally if it hasn't been done before.

Bear in mind that it's not just mobile apps that need to bind to 0.0.0.0 for local dev, it's also a breaking change for Lens users.

sweatybridge:
Are we happy to close this as we discussed offline to change the default only when working on cli v3?

Feel free to reopen if plans change.

Issue #1705, opened 2023-08-07
running webhooks on local development


holdehn:
### **Bug Report**
*  I confirm this is a bug with Supabase, not with my own application.
*  I confirm I have searched the [Docs](https://docs.supabase.com/), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com/).

 **Describe the bug**

When developing locally with Supabase webhooks, I'm unable to run the edge-function webhook using a local URL. When setting up the webhook in the remote DB with an edge function, it's intended to call an edge function anytime a row is inserted into a specified table. However, during local development, this behavior isn't replicated; instead, the webhook defaults to an HTTPS request. The option to use a native edge function (which is essentially an HTTPS URL) disappears. I'm looking for a seamless way to set the webhook URL to a local edge function during local development without manual intervention.

**To Reproduce**

* 		Set up Supabase webhook in remote DB with an edge function.
* 		Ensure the edge function is called whenever a row is added to a particular table.
* 		Run the application in local development.
* 		Observe that the webhook defaults to an HTTPS request instead of the intended edge function.

**Expected behavior**

When running local development, the webhook should be able to point to a local edge function, similar to how it operates in a production environment. The switch between local and production webhooks should be smooth, without needing manual adjustments.

**Additional context**

The primary goal is to maintain consistency between local and production environments. The current behavior makes local testing challenging due to the difference in how webhooks are handled.

Suzan-Dev:
Facing same thing on local enviroment.

CalvinWilkinson:
I am having issues myself but it is not related to **_https_**.

When running `Supabase status` I get:
```txt
API URL: http://127.0.0.1:64321
GraphQL URL: http://127.0.0.1:64321/graphql/v1
DB URL: postgresql://postgres:postgres@127.0.0.1:64322/postgres
Studio URL: http://127.0.0.1:64323
Inbucket URL: http://127.0.0.1:64324
```

The base URL that I am using in my edge function is `http://127.0.0.1:64321`.  My test edge function that is running locally when I run the `supabase functions serve` command has a URL path of `functions/v1/update-instructor-ui`.  So for the webhook URL, I am using `http://127.0.0.1:64321/functions/v1/update-instructor-ui`.

I did the same setup and pattern in my cloud instance of supabase and it executed fine.  The only difference is that in my cloud instance are the choices of webhook types to choose from.

**Cloud Instance:**
[...]

PR #4629, opened 2025-12-16
feat(db): support edge function names in webhook triggers

JEETDESAI25:

Feature

Refs: #1705, supabase/supabase#34139

Webhooks require full URLs, breaking when migrating between environments.

## What is the new behavior?

- `supabase_functions.http_request()` auto-detects slugs vs URLs, reads secrets from vault, injects Authorization header
-  Vault secrets synced in `start`, `db push`, `db reset`, `migration up/down`
-  Works for PG14 and PG15+ stacks

### DIFF
--- a/internal/db/diff/diff_test.go
+++ b/internal/db/diff/diff_test.go
@@ -42,6 +42,9 @@ func TestRun(t *testing.T) {
 		// Setup in-memory fs
 		fsys := afero.NewMemMapFs()
 		require.NoError(t, flags.LoadConfig(fsys))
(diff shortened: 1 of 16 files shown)


sweatybridge:
I think this is a good idea. But we also need to be consistent with the webhooks definition on platform side.

Let me check with our team and get back on whether something similar can be rolled out on all hosted projects.

JEETDESAI25:
Thanks for the review! Let me know what changes are needed to better align with platform side definition.

PR #4671, opened 2025-12-30
fix(db): execute seed sql for modified seed files

### LINKED ISSUE #4670: `supabase db push --include-seed` doesn't actually run seed files
**Describe the bug**
`supabase db push --include-seed` detects changes to seed files, updates hashes but it doesn't actually run the seed files.

**To Reproduce**
1. Modify a seed file
2. Add a migration (seed changes don't seem detected without a migration -- another issue)
3. run `supabase db push --include-seed` and answer Y to prompt to update seed
4. Check remote database, seed data is not updated

**Expected behavior**
The seed data on the remote database is updated

**System information**
 - Version of OS: Darwin 25.1.0 arm64 arm
 - Version of CLI: 2.70.5
 - Version of Docker: Docker version 29.1.3, build f52814d
 - Versions of services: 
 - 
  ------------------------|------------------------|---------------------------
   supabase/postgres      | 17.6.1.054             | 17.6.1.054                
   supabase/gotrue        | v2.184.0               | v2.184.0                  
   postgrest/postgrest    | v14.2                  | postgrest/postgrest:v14.2 
   supabase/realtime      | v2.69.0                | -                         
   supabase/storage-api   | v1.33.2                | v1.33.0                   
   supabase/edge-runtime  | v1.68.4                | -                         
   supabase/studio        | 2025.12.23-sha-f3af2c3 | -                         
   supabase/postgres-meta | v0.95.1                | -                         
   supabase/logflare      | 1.27.3                 | -                         
   supabase/supavisor     | 2.7.4                  | -

7ttp:
when running `supabase db push --include-seed`, modified seed files were detected and their hashes updated, but the actual seed sql was not executed. this happened because `ExecBatchWithCache` only queued sql statements when `Dirty` was false, but modified seeds have `Dirty` set to true.

closes https://github.com/supabase/cli/issues/4670

### DIFF
--- a/pkg/migration/file.go
+++ b/pkg/migration/file.go
@@ -177,10 +177,8 @@ func (m *SeedFile) ExecBatchWithCache(ctx context.Context, conn *pgx.Conn, fsys 
 	}
 	// Data statements don't mutate schemas, safe to use statement cache
 	batch := pgx.Batch{}
-	if !m.Dirty {
-		for _, line := range lines {
-			batch.Queue(line)
-		}
+	for _, line := range lines {
+		batch.Queue(line)
 	}
 	batch.Queue(UPSERT_SEED_FILE, m.Path, m.Hash)
 	// No need to track version here because there are no schema changes
--- a/pkg/migration/seed_test.go
+++ b/pkg/migration/seed_test.go
@@ -98,10 +98,12 @@ func TestSeedData(t *testing.T) {
 		conn := pgtest.NewConn()
 		defer conn.Close(t)
 		mockSeedHistory(conn).
-			Query(UPSERT_SEED_FILE, seed.Path, seed.Hash).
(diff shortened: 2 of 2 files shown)


mfairley:
@7ttp Could we get this landed soon?

7ttp:
@sweatybridge

sweatybridge:
This is intentional because most seed scripts are not idempotent. My suggestion would be to add a new seed file. Wdyt?

mfairley:
> This is intentional because most seed scripts are not idempotent. My suggestion would be to add a new seed file. Wdyt?

@sweatybridge The CLI prompts the user about updated seed files and then updates the hash for the file references stored in the database in supabse_migrations.seed_files. This seems to imply the files are run and it also informs the user prior to running them, so it's confusing if they aren't run.

Shouldn't it be the user's responsibility not to update a seed file if it's not idempotent? If I have to make new seed files then that feels no different to migrations.

PR #4680, opened 2026-01-04
feat(seed): Add new flag option for setting custom seed file path

coolusaHD:
Feature

Currently, [supabase db reset](vscode-file://vscode-app/c:/Users/alber/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) only supports two seed behaviors:

Run the default seed file(s) defined in config (e.g. `supabase/seed.sql`)
Skip seeding entirely with `--no-seed`
There is no way to specify an alternative seed file at runtime. Users who want to reset their database with different initial data (e.g., an empty state with minimal setup vs. a full product example) must either:

Manually swap seed files before running reset
Skip seeding and manually run the desired seed file afterward

## What is the new behavior?
Added a new `--override-seed-file` flag to `supabase db reset` that allows specifying a custom seed SQL file path:

The `--override-seed-file` and `--no-seed`  flags are mutually exclusive since using both doesn't make sense.

## Additional context

### DIFF
--- a/cmd/db.go
+++ b/cmd/db.go
@@ -193,8 +193,9 @@ var (
 		},
 	}
 
-	noSeed      bool
-	lastVersion uint
+	noSeed           bool
+	lastVersion      uint
+	overrideSeedFile string
 
 	dbResetCmd = &cobra.Command{
 		Use:   "reset",
@@ -202,6 +203,8 @@ var (
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if noSeed {
 				utils.Config.Db.Seed.Enabled = false
+			} else if len(overrideSeedFile) > 0 {
+				utils.Config.Db.Seed.SqlPaths = []string{overrideSeedFile}
 			}
 			return reset.Run(cmd.Context(), migrationVersion, lastVersion, flags.DbConfig, afero.NewOsFs())
 		},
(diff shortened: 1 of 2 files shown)


sweatybridge:
You can specify custom seed paths via config.toml

```config
[db.seed]
# If enabled, seeds the database after migrations during a db reset.
enabled = true
# Specifies an ordered list of seed files to load during db reset.
# Supports glob patterns relative to supabase directory: "./seeds/*.sql"
sql_paths = ["./seed.sql"]
```

coolusaHD:
> You can specify custom seed paths via config.toml

@sweatybridge Thanks for having a look at my PR. The reason why I want to add an seed file flag is that the config.toml support only one state of seed files. I also currently use a project with musltiple seed files but they dont cancel each other out. 

What I need for my project is that I have a seed file which contains a complete filled databse with test data for everthing. But I want to have a second seed state where just the basic outline data is filled into the database to like simulate the beginning of a new client, etc.

Let me know what you think about that 🙌

coolusaHD:
@sweatybridge Any update on this ?

kevin-brotcke:
We need this feature too. I'm not really seeing a good alternative if you run a lot of tests with specific seed data. The workarounds I can think of all seem kind of hacky:

1. Run a sed command to update the `config.toml` file
2. Move/copy seed files via shell commands
3. Use the `--workdir` param to point to another supabase directory with different `config.toml` and seed data. I like this but now you have dups of the entire supabase infra folder that's bound to get out of sync.

I actually think a more reusable solution would be to set a `config.toml` path via runtime. Or a config override path that will merge with base config. The logic looks like it's already there for [branch specific config values](https://supabase.com/docs/guides/local-development/cli/config#branching-config).

coolusaHD:
@kevin-brotcke I invited you to my fork
Feel free to add your suggestions
I'm more than happy to see others are also struggel with that

jgoux:
Hey, thanks for your contribution! 🙏 

I opened https://github.com/supabase/cli/pull/5691 to address the issue with a new repeatable `--sql-path` flag which overrides the value in `config.toml`.

In the future we will maybe look into having seeds profiles.

Thanks again for addressing the issue!

PR #4686, opened 2026-01-06
feat: add -f/--foreground flag to supabase start

### LINKED ISSUE #4681: Add 'supabase start -f' so you don't need to remember to stop it everytime
Yes, the problem is having supabase containers running all day on my pc because I forget to stop it. This option to me makes sense because I don't mind having to run supabase start everytime but to me is worse to forget to run supabase stop and then have it taking resources on my pc for the whole day.

Adding an ```'-f'``` option so you can specify that you want to stop it when terminal is closed like ```supabase start -f```

```npx supabase start && trap 'npx supabase stop' EXIT INT TERM HUP && sleep infinity```

**Additional context**
I'm using supabase for many full stack projects.

Ayush1404:
### Feature

Add foreground mode to supabase start command that keeps the process
running and automatically stops services when the terminal exits.

supabase start : 

- Starts all Supabase services and exits immediately
- Services continue running in background
- User must manually run 'supabase stop' to clean up

Closes #4681 

## What is the new behavior?

supabase start -f : 

- Starts all Supabase services
- Stays running in foreground 
- Automatically stops services when terminal exits (Ctrl+C, close window, etc.)
- Same graceful shutdown as manual 'supabase stop'

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -40,6 +40,7 @@ var (
 	excludedContainers []string
 	ignoreHealthCheck  bool
 	preview            bool
+	foreground         bool
 
 	startCmd = &cobra.Command{
 		GroupID: groupLocalDev,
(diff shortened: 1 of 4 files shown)


lukebussey:
Can this PR be reopened? It would really help while running the cli with turbo

Issue #4614, opened 2025-11-30
JWKS Endpoint Returns Empty Key Set for Local Supabase


whollacsek:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

When following the guide to verify a jwt key https://supabase.com/docs/guides/auth/jwts#verifying-a-jwt-from-supabase on the local Supabase instance (https://127.0.0.1:8000/auth/v1/.well-known/jwks.json) the result is:
```json
{
"keys": []
}
```
Running `supabase status` will display the publishable and secret keys.

This does work on the cloud Supabase instance, the issue is only affecting local dev.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Run `supabase start`
2. Go to `/auth/v1/.well-known/jwks.json`
3. See empty keys array

## Expected behavior

Public key should be returned.

## System information

- Version of supabase cli: 2.63.1

## Additional context

`/auth/v1/.well-known/jwks.json` works on hosted Supabase.

Adityakk9031:
can i work on this issue

singh-inder:
I'm guessing you haven't added signing key in config.toml. 
1. You will have to generate a new signing key using the command in this guide: https://supabase.com/docs/guides/auth/signing-keys#how-to-create-mint-jwts-if-access-to-the-private-key-or-shared-secret-is-not-possible
2. Save the key in a file and add the path in config.toml (`signing_keys_path` field is already present in config.toml, you just need to uncomment it and update the signing key file path if necessary)
	```toml
	[auth]
	signing_keys_path = "./signing_key.json"
	```
Restart and now you should get a public key when you visit `https://127.0.0.1:8000/auth/v1/.well-known/jwks.json`

whollacsek:
Yes I figured it out after a lot of searching. This should be automated as
it breaks the local dev experience

harshtiwari2022:
@whollacsek  Sir, I have made improvements and would like to continue working on this. Could you please assign the issue to me?

aantti:
@sweatybridge

sweatybridge:
Moving to cli repo. We can look into generating a signing key by default.

joelpramos:
> I'm guessing you haven't added signing key in config.toml.
> 
[...]

PR #4662, opened 2025-12-23
fix JWK endpoints returning empty keys

olamilekan000:
## What kind of change does this PR introduce?

Bug fix

## What is the current behavior?

The JWKS endpoint (`/auth/v1/.well-known/jwks.json`) returns an empty keys array (`{"keys": []})` for local Supabase instances because signing keys are not automatically generated during supabase init. Users must manually generate signing keys using supabase gen signingkeys and configure `signing_keys_path` in `config.toml.`
Issue: [#4614](https://github.com/supabase/cli/issues/4614)

## What is the new behavior?
- Automatically generates a default RS256 signing key during supabase init
- Saves the key to supabase/signing_keys.json (already in .gitignore)
- Skips generation if the signing key file already exists
- The JWKS endpoint now returns the public key, matching cloud behavior

## Additional context
Add any other context or screenshots.
Screenshot 2025-12-23 at 23 02 57

PR #4688, opened 2026-01-07
fix: use asymmetric signing key by default

sweatybridge:

Bug fix https://github.com/supabase/cli/issues/4614

## What is the new behavior?

Initializes default signing key for `curl http://127.0.0.1:54321/auth/v1/.well-known/jwks.json`

## Additional context

supersedes https://github.com/supabase/cli/pull/4662

### DIFF
--- a/internal/start/start.go
+++ b/internal/start/start.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	_ "embed"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net"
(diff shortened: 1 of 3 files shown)


mateuszbasiak:
Hi! 
In most recent version of Supabase CLI (2.71.1) including these changes I started to encounter following error during `npx supabase db reset`:

```
Error status 400: {"statusCode":"403","error":"Unauthorized","message":"Key for the ES256 algorithm must be one of type CryptoKey, KeyObject, or JSON Web Key. Received an instance of Uint8Array"}
```

I tried running the same command in previous version as well (2.71.0) and it seems to work fine for without any issues.

Is this error related to these changes? Is there anything I could do to avoid this error?

sweatybridge:
It's a bug that I missed. I've opened a PR to address it in v2.71.2. Thanks for reporting!

PR #4692, opened 2026-01-07
fix: use static list of project regions

### LINKED ISSUE #4689: Create project via the cli not working with --region flag
**Describe the bug**

**To Reproduce**
Go to CLI and and create a new project in a specific region as follows `supabase projects create "test-project" --org-id  "<org-id>" --region us-east-1`

**Expected behavior**
Should produce a new project on the platform and  produce details of that new project in the terminal output

```
   ORG ID               | REFERENCE ID         | NAME         | REGION                   | CREATED AT (UTC)    
  ----------------------|----------------------|--------------|--------------------------|---------------------
   <org-id> | bafwjygnsxvxoagycnnz | test-project | East US (North Virginia) | 2026-01-07 04:11:36 
```

**Screenshots**
The actual behaviour is the following error
Image

**System information**
 - Version of CLI: 2.67.1

**Additional context**

singh-inder:

Bug fix #4689

## What is the new behavior?

LoadProfile is called before awsRegions are loaded.

## Additional context

`region` var is initialized at package scope in projects.go. As regions are scoped to profile now (https://github.com/supabase/cli/pull/4596), at this point the current profile isn't loaded yet so the `region.Allowed` is set to empty slice. `LoadProfile` is called inside `rootCmd.PersistentPreRunE` which runs after flag parsing and flag validation.

Marking this as a draft for now to get early feedback from maintainers. Happy to adjust the approach if there's a preferred pattern here.

### DIFF
--- a/cmd/projects.go
+++ b/cmd/projects.go
@@ -28,8 +28,10 @@ var (
 	projectName string
 	orgId       string
 	dbPassword  string
-	region      utils.EnumFlag
-	size        = utils.EnumFlag{
+	region      = utils.EnumFlag{
+		Allowed: utils.AwsRegions(),
+	}
+	size = utils.EnumFlag{
 		Allowed: []string{
 			string(api.V1CreateProjectBodyDesiredInstanceSizeLarge),
 			string(api.V1CreateProjectBodyDesiredInstanceSizeMedium),
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -8,7 +8,6 @@ import (
 	"net/url"
 	"os"
 	"os/signal"
-	"sort"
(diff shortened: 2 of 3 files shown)


sweatybridge:
I did more testing and found that it doesn't actually work. Reason is we need to parse the `--profile` flag to know which profile to load, but if we let cobra parse all flags, it will also include the enum flag that depends on the current profile. so it's a catch-22.

let me see if there's another alternative

currently settling on a static list of regions to unblock

singh-inder:
🤦‍♂️ Yeah, I completely overlooked `--profile` flag. sorry about that.

PR #4693, opened 2026-01-07
fix: use hardcoded signing key

sweatybridge:

Follow up https://github.com/supabase/cli/pull/4688

## What is the new behavior?

Initialise a hardcoded signing key so it doesn't change between commands.

Allows bearer jwt to be generated from local signing key.

```bash
$ supabase gen bearer-jwt --role service_role
Enter your signing key in JWK format (or leave blank to use local default):
```

## Additional context


mateuszbasiak:
Thank you @sweatybridge for quick action! Error is now gone and the reset command works for me as expected ✅

Issue #1894, opened 2024-01-30
"database not healthy" when running `db diff`


probablykasper:
**Describe the bug**
I get the below error whenever I run `supabase db diff`. Even removing all images and containers doesn't fix it.

Other commands work fine, such as `start`, `stop` and `reset`.

```
Connecting to local database...
Creating shadow database...
Failed to remove container: e672a047315cd867543cde6ae1ce037b4a69e761116e5ce7621483468a182075 Error response from daemon: No such container: e672a047315cd867543cde6ae1ce037b4a69e761116e5ce7621483468a182075
database is not healthy
Try rerunning the command with --debug to troubleshoot the error.
```

<details>
<summary>Logs with --debug`</summary>

```
Supabase CLI 1.136.3
Connecting to local database...
2024/01/30 09:40:11 Connect Start: tcp 127.0.0.1:54322
2024/01/30 09:40:11 Connect Done: tcp 127.0.0.1:54322
2024/01/30 09:40:11 PG Send: {"Type":"StartupMessage","ProtocolVersion":196608,"Parameters":{"database":"postgres","user":"postgres"}}
2024/01/30 09:40:11 PG Recv: {"Type":"AuthenticationSASL","AuthMechanisms":["SCRAM-SHA-256"]}
2024/01/30 09:40:11 PG Send: {"Type":"SASLInitialResponse","AuthMechanism":"SCRAM-SHA-256","Data":"n,,n=,r=GLRIk8WY+aEYM9cvggJKRLws"}
2024/01/30 09:40:11 PG Recv: {"Type":"AuthenticationSASLContinue","Data":"r=GLRIk8WY+aEYM9cvggJKRLwsHjj2XhMitmA3njoB9Ur/qEEu,s=Flsuj2nZzUpCA2ZSVHVmQA==,i=4096"}
2024/01/30 09:40:11 PG Send: {"Type":"SASLResponse","Data":"c=biws,r=GLRIk8WY+aEYM9cvggJKRLwsHjj2XhMitmA3njoB9Ur/qEEu,p=z+powyfZHSeNEBzmSp7s8OO0sk24OXDyzo66pzZ4BZ0="}
2024/01/30 09:40:11 PG Recv: {"Type":"AuthenticationSASLFinal","Data":"v=JvkNWprPCDi5PM6a7ZFP4aru26gBveySZ0HnfT8tuhY="}
2024/01/30 09:40:11 PG Recv: {"Type":"AuthenticationOK"}
[...]

PR #4699, opened 2026-01-08
feat: allow configuring local health check timeout

sweatybridge:

closes https://github.com/supabase/cli/issues/1894

## What is the new behavior?

Allow users to configure database health check timeout via config.toml.

## Additional context


sai-sy:
should this not also apply that to `db pull` and not just `db diff`

PR #4707, opened 2026-01-10
feat(cli): add --no-seed flag to supabase start

MuzzaiyyanHussain:
Adds a --no-seed flag to supabase start to skip storage bucket seeding during startup.
This allows faster restarts and avoids unnecessary seeding when it’s not needed.
Default behavior remains unchanged when the flag is not used.

### DIFF
--- a/internal/start/start.go
+++ b/internal/start/start.go
@@ -249,9 +249,6 @@ func run(ctx context.Context, fsys afero.Fs, excludedContainers []string, dbConf
 		utils.Config.Storage.ImageTransformation.Enabled && !isContainerExcluded(utils.Config.Storage.ImgProxyImage, excluded)
 	isS3ProtocolEnabled := utils.Config.Storage.S3Protocol != nil && utils.Config.Storage.S3Protocol.Enabled
 	fmt.Fprintln(os.Stderr, "Starting containers...")
-	if skipSeed {
-	fmt.Fprintln(os.Stderr, "Skipping storage seeding (--no-seed enabled)")
-}
 	workdir, err := os.Getwd()
 	if err != nil {
 		return errors.Errorf("failed to get working directory: %w", err)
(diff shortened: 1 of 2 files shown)


sweatybridge:
Is it not sufficient to use config.toml to turn off seeding?

```toml
[db.seed]
enabled = false
```

MuzzaiyyanHussain:
Good question.
db.seed.enabled = false only controls SQL/database seeding, but this issue is about storage bucket seeding, which currently runs unconditionally during supabase start.
The --no-seed flag is meant as a one-off CLI override (useful for local dev, CI, or recovery scenarios) without requiring config changes.
Happy to adjust this to reuse config if you prefer, but at the moment config.toml doesn’t cover storage seeding.

MuzzaiyyanHussain:
Hi @sweatybridge Just to confirm, should storage seeding fully follow `db.seed.enabled`, and `--no-seed` act only as a temporary CLI override?

If that works for you, I’ll update the implementation and push a follow-up commit.

sweatybridge:
> Just to confirm, should storage seeding fully follow db.seed.enabled, and --no-seed act only as a temporary CLI override?

Generally I'd avoid adding more flags because it's hard to maintain. But I agree `--no-seed` is desired in this case for adhoc invocation of supabase start command.

My only suggestion is to reuse the [noSeed](https://github.com/supabase/cli/blob/develop/cmd/db.go#L193-L205) variable instead of declaring a new one.

And yes, let's have storage seeding follow `db.seed.enabled` for simplicity. Technically seeding storage objects also means inserting metadata entry into `storage.objects` table, so it's not wrong to use the same config.

MuzzaiyyanHussain:
That makes sense 

I’ll reuse the existing `noSeed` variable instead of introducing a new one, and update the logic so storage seeding follows `db.seed.enabled`, with `--no-seed` acting as an ad-hoc override for `supabase start`.

I’ll push a follow-up commit with these changes shortly.

MuzzaiyyanHussain:
@sweatybridge  I’ve updated storage seeding to follow db.seed.enabled by default, with --no-seed acting as a CLI override.
Added a test to cover the config-based behavior.
Tests pass and I verified manually with  go run . start --no-seed on Windows.

MuzzaiyyanHussain:
@sweatybridge ?

internal/start/start.go
```diff
@@ -1279,14 +1277,23 @@ EOF
...
+		if skipSeed {
+			fmt.Fprintln(os.Stderr, "Skipping storage seeding (--no-seed enabled)")
+		} else if !utils.Config.Db.Seed.Enabled {
+			fmt.Fprintln(os.Stderr, "Skipping storage seeding (db.seed.enabled = false)")
+		} else {
+			if err := buckets.Run(ctx, "", false, fsys); err != nil {
+				return err
+			}
 		}
+
```
sweatybridge:
```suggestion
		if !utils.Config.Db.Seed.Enabled {
			fmt.Fprintln(os.Stderr, "Skipping storage seeding...")
		} else if err := buckets.Run(ctx, "", false, fsys); err != nil {
			return err
		}
```

sweatybridge (commented):
Thanks a lot for creating this PR. I just want to suggest a minor change. Don't worry if you are busy because I can also pick it up next week.

MuzzaiyyanHussain:
HI @sweatybridge, are you picking this up ??

MuzzaiyyanHussain:
closing this for now

PR #4721, opened 2026-01-14
feat: add hybrid jwt verification

kallebysantos:

Bug fix, feature

Since API keys did change, users can't call functions without manually disabling `verify_jwt`.

## What is the new behavior?

JWKs are now default exposed https://github.com/supabase/cli/pull/4688, so It allows to verify both new asymmetric tokens as well legacy ones.

This way it applies a temporary fix while migrating API keys. 

- closes supabase/edge-runtime/issues/654
- solves supabase/supabase/issues/41929


joelpramos:
@kallebysantos `Deno.env.get` doesn't seem to have access to `"SUPABASE_INTERNAL_*"` for me. Is there any property that needs enablement?

```typescript
const JWT_SECRET = Deno.env.get('SUPABASE_INTERNAL_JWT_SECRET')!;
const HOST_PORT = Deno.env.get('SUPABASE_INTERNAL_HOST_PORT')!;
const DEBUG = Deno.env.get('SUPABASE_INTERNAL_DEBUG') === 'true';
const FUNCTIONS_CONFIG_STRING = Deno.env.get(
  'SUPABASE_INTERNAL_FUNCTIONS_CONFIG'
)!;

async function verifyLegacyJWT(
  jwt: string
): Promise<JWTVerifyResult<JWTPayload> | undefined> {
  // const JWT_SECRET = Deno.env.get('SUPABASE_INTERNAL_JWT_SECRET')!;
  logger.debug('JWT secret: ' + JWT_SECRET);
  logger.debug('Host port: ' + HOST_PORT);
  logger.debug('Debug: ' + DEBUG);
  logger.debug('Functions config string: ' + FUNCTIONS_CONFIG_STRING);
  const encoder = new TextEncoder();
  const secretKey = encoder.encode(JWT_SECRET);
  try {
    return await jwtVerify(jwt, secretKey);
  } catch (e) {
    logger.error('Symmetric Legacy JWT verification error', e);
    return undefined;
  }
}
```
image

kallebysantos:
Hi @joelpramos 💚
Could you give me more context about your version? Are you calling it inside an edge function or just trying my fork?

joelpramos:
hi @kallebysantos:

- Supabase version 2.75.0
- Local (self hosted)
- Just trying to replicate your middleware change, not using your fork
- Getting the var from within a Supabase edge function (my middleware, trying to replicate what you have here as I have this problem locally)
- I can see the vars injected in the container itself and confirmed by opening the shell and printing it

kallebysantos:
Hey @joelpramos 💚
The `SUPABASE_INTERNAL_*` are only available inside CLI context as internal envs for the docker local-dev stack — So it only works in this fork / `supabase start` command (only for internal `main.ts` scope)

Since you're using self-hosted, the `main/index.ts` template already implements the [legacy verification](https://github.com/supabase/supabase/blob/4651480f28f0939b7fad339e91045575e76ad5df/docker/volumes/functions/main/index.ts#L21) and the env is named as [`JWT_SECRET`](https://github.com/supabase/supabase/blob/4651480f28f0939b7fad339e91045575e76ad5df/docker/volumes/functions/main/index.ts#L6).

I'm not really sure, but new JWK are not available for self-hosting yet, so you "can't" replicate this hybrid replication — At least without extra manual steps. 

cc @aantti can give you more details about Self-Host situation of new Asymmetric Keys.

aantti:
> can give you more details about Self-Host situation of new Asymmetric Keys

It's WIP :) Hoping to add it soon.

joelpramos:
@aantti @kallebysantos at least locally the new solution works totally fine for users but not for service_user (e.g., functions triggered from Cron). I am also having some issues in the deployed instance with functions triggered from cron jobs but haven't triaged yet the root cause and missing some logs.

fwiw other than the fact the variable wasn't available, this hybrid solution you posted @kallebysantos  works for me i.e. seems like the service_role from the cron user is using the old auth method. Is that something on both your radars?

kallebysantos:
Hey @joelpramos 

>  service_role...is using the old auth method

Yes, `ANON_KEY` and `SERVICE_ROLE` uses old auth (symmetric JWT)  which can be verified from `*JWT_SECRET`

> ...fine for users but not for the service_role...

I don't think so, user's tokens is now being issued as new JWT (Asymmetric JWK) and the purpose of this PR is to handle both kinds of tokens, using legacy as fallback. 
So it should work for any valid token... but I can do more testing to make sure if cron requests is working fine.

FredrikWendt:
Looking forward to see this happen. Currently, we're lowering security on our apps to work around this issue. No action here in two weeks isn't impressive for a bug impacting security/auth.

joelpramos:
> Hey @joelpramos
> 
> > service_role...is using the old auth method
> 
> Yes, `ANON_KEY` and `SERVICE_ROLE` uses old auth (symmetric JWT) which can be verified from `*JWT_SECRET`
> 
> > ...fine for users but not for the service_role...
> 
> I don't think so, user's tokens is now being issued as new JWT (Asymmetric JWK) and the purpose of this PR is to handle both kinds of tokens, using legacy as fallback. So it should work for any valid token... but I can do more testing to make sure if cron requests is working fine.

@kallebysantos 

So I guess what I am not yet 100% clear is what is now the flow for my Postgres Functions (`pg_net`) calling Edge Functions. I have tried  new API Keys it doesn't appear like the Bearer token is "exchanged" into an actual JWT token (new or legacy), as it happens locally (self hosted). Also tried old tokens which obviously aren't active anymore.

Not sure if there's a different network route that it isn't sending it through Kong to 'transform' the Bearer token from secret API key into a valid JWT?

At the end of the day it's not clear yet to me what is the appropriate configuration for this set up... so that I can wrap these function calls with cron. Clearly behaviours diverge.

kallebysantos:
Hey guys!
Thanks for your feedbacks, we're working daily on this situation.
Hope to have a proper fix soon 🙏

> @FredrikWendt: Looking forward to see this happen....No action here in two weeks isn't impressive for a bug impacting security/auth.

We do understand your disappointment 🥲
Even this PR not being active for some time, keep in mind that its only for local CLI. We're active working on internal changes to fix this situation at all environments. 

> @joelpramos:  as it happens locally (self hosted)... 

Unfortunately there's a different behaviour on how Local CLI is handling the `sb_xxx` keys, and I believe this gateway injection doesn't happen at Platform.

> t's not clear yet to me... wrap these function calls with cron...

Its a good catch 👍 
I still need to validate the best approach for it, then I can loop back to you with a recommended solution.

kallebysantos:
Hey @joelpramos 💚

While we're still working on a better official solution
I believe that should be possible to invoke from webhooks/cron with following the steps:

## Using new API Keys to create webhooks:
1. Create a new publishable/secret key at `https://supabase.com/dashboard/project/[ref-id]/settings/api-keys`
<details>
image
</details>

2. Add it as custom secret like: `SB_CRON_WEBHOOK`
3. Use the incoming `apiKey` to check if it matches your secret as well to initialise the `createClient()`
<details>

```ts
import { createClient } from 'npm:@supabase/supabase-js@2'

console.info('server started');

Deno.serve(async (req: Request) => {
  const apiKey = req.headers.get('apiKey')
  if (apiKey !== Deno.env.get('SB_CRON_WEBHOOK')!) {
    return new Response(null, { status: 401 });
  }

  // Use the request 'apiKey' to init supabase client
  const supabase = createClient(Deno.env.get('SUPABASE_URL')!, apiKey);

  // Do somenthing

  return new Response();
});
```

</details>

4. Call it from your cron or webhook, sending the `apiKey` header. — You don't need JWT in this situation...

PR #4730, opened 2026-01-17
DRAFT: replace inbucket with minimal implementation

kiwicopple:
Implements an embedded SMTP server that replaces the Docker-based inbucket container for local email testing. This will make the CLI lighter and faster to start.

Emails are stored as .eml files on disk for easy access.

Features:

- SMTP server using github.com/emersion/go-smtp
- File-based storage with configurable path via SUPABASE_MAIL_PATH
- Email parsing using github.com/jhillyerd/enmime/v2
- API to list mailboxes, list emails, read and delete emails

Storage format:

$SUPABASE_MAIL_PATH/
└── {recipient-email}/
    └── {timestamp}-{id}.eml

### DIFF
--- a/pkg/mail/storage.go
+++ b/pkg/mail/storage.go
@@ -29,14 +29,44 @@ type Storage struct {
 
 // NewStorage creates a new file-based storage at the given path.
 func NewStorage(basePath string) (*Storage, error) {
+	// Resolve to absolute path for consistent path validation
+	absPath, err := filepath.Abs(basePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve storage path: %w", err)
+	}
+
 	// Create the base directory if it doesn't exist
-	if err := os.MkdirAll(basePath, 0755); err != nil {
+	if err := os.MkdirAll(absPath, 0755); err != nil {
 		return nil, fmt.Errorf("failed to create storage directory: %w", err)
 	}
 
 	return &Storage{
-		basePath: basePath,
+		basePath: absPath,
 	}, nil
+}
+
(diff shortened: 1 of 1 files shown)


internal/start/start.go
```diff
@@ -81,6 +94,22 @@ func Run(ctx context.Context, fsys afero.Fs, excludedContainers []string, ignore
...
+		// Wait for context cancellation (Ctrl+C)
+		<-ctx.Done()
+		fmt.Fprintln(os.Stderr, "\nStopping embedded mail server...")
+		StopEmbeddedMailServer()
+		// Also stop Docker containers
+		if err := utils.DockerRemoveAll(context.Background(), os.Stderr, utils.Config.ProjectId); err != nil {
+			fmt.Fprintln(os.Stderr, err)
+		}
+		fmt.Fprintln(os.Stderr, "Stopped all services.")
+	}
```
sweatybridge: My concern here is that it relies on cli to be running on foreground. For anyone running local stack in background, we still need to fallback to mailpit container.

For ease of maintenance, I would avoid duplicating the implementation.

You can also measure the overhead of mailpit container by setting `[inbucket] enabled = false`. I suspect the impact is negligible.

The best outcome IMO is to use docker compose for starting all services. We already import that as a library. Remaining work is move individual `DockerStart` API calls to a single compose spec.

kiwicopple: this is a good point - i think my main idea was to minimize the surface area. I didn't realize we are using mailpit (because the container name was still `inbucket` but the docker image is [only 13MB](https://hub.docker.com/r/axllent/mailpit) so already quite small

I'll close for now

PR #4743, opened 2026-01-20
feat: use smaller supabase postgres and restrict started projects

filipecabaco:

adds a new `--small` flag to only startup postgres, kong, auth and postgrest as a PoC on how a smaller setup could look like.

Note: relies on the existance of the image built by https://github.com/supabase/postgres/pull/2009


jgoux:
Smaller images will be addressed after we finish the port.
Closing this PR for now. 🙏

PR #4747, opened 2026-01-21
fix: parser treats 'atomic' in identifiers as BEGIN ATOMIC keyword

7ttp:
.

### DIFF
--- a/pkg/parser/state.go
+++ b/pkg/parser/state.go
@@ -47,7 +47,6 @@ func (s *ReadyState) Next(r rune, data []byte) State {
 		fallthrough
 	case 'C':
 		offset := len(data) - len(BEGIN_ATOMIC)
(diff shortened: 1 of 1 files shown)


xdmden (commented):
Nice fix! The whitespace boundary check is the right approach. Two suggestions:

1. **Simplify and fix `offset == 0` edge case** — when `offset == 0`, there's no `BEGIN` before `ATOMIC`, so it should never enter `AtomicState`. Changing `offset >= 0` to `offset > 0` handles this and lets us collapse the nested ifs into one condition. Since whitespace characters are all ASCII, we can also skip the rune decode and just cast the byte directly.

2. **More test coverage** — the current test covers `atomic_test` (prefix), but it's worth covering suffix (`my_function_atomic`), mid-identifier (`is_atomic`), and digit-boundary (`col1atomic`) cases too, since these are all patterns users are likely to hit.

PR #4755, opened 2026-01-25
fix: mount notification email templates

### LINKED ISSUE #4783: notification email not sent
**Describe the bug**
Hi team 👋
I’m following the supabase guide on customizing email templates for local development: https://supabase.com/docs/guides/local-development/customizing-email-templates

Custom email templates are working, but I’ve noticed that the "password changed" notification email is never sent after password is successfully updated. This is also the case for other notification emails.

**To Reproduce**
1. My config.toml
```toml
[auth.email.template.recovery]
subject = "Reset password"
content_path = "./supabase/templates/recovery.html"

[auth.email.notification.password_changed]
enabled = true
subject = "Your password has been changed"
content_path = "./templates/password_changed_notification.html"
```
2. Trigger a password reset flow. 
3. Complete reset password flow.
4. no password changed notification email is received

**Expected behavior**
notification email should have been sent

**System information**
 - Version of CLI: v2.74.4

singh-inder:

Bug fix #4783 

Follow up: #4400

When running `supabase start`, only auth email templates were mounted in kong and configured in auth service.
Notification email templates were ignored.

## What is the new behavior?

Notification email templates are now mounted in Kong, and the required env vars for notification templates are configured in the auth service.


singh-inder:
@sweatybridge sorry for the tag, would you mind taking a look at this as well? A couple of users are stuck on it.

Also, kind of related: https://github.com/supabase/supabase/issues/41860#issuecomment-3810223360, its about how auth template paths (if not absolute) are expected to be relative to a level above `builder.SupabaseDirPath` but not notification templates. I also notice there is a FIXME comment next to code building auth template path. I assume that auth template path will only be fixed in the next major version but until then it might be better if notification template paths used the same base for consistency.

https://github.com/supabase/cli/blob/3fe548f00b6c80ee3666b5cafba51b6684637868/pkg/config/config.go#L697-L710

Issue #4725, opened 2026-01-16
Feature Request: Support Authenticated Image Registries (e.g. Dependency Proxy)


ffabss:
**Is your feature request related to a problem? Please describe.**
The Supabase CLI allows configuring a custom image registry for pulling Supabase Docker images via `SUPABASE_INTERNAL_IMAGE_REGISTRY`. However, this mechanism only works with public registries. There is currently no way to provide authentication credentials for registries that require login.

This limitation makes it impossible to use private or authenticated registries, such as dependency proxies, in CI environments.

**Describe the solution you’d like**
The Supabase CLI should support authenticated image registries by allowing users to provide registry credentials. This could be done by:

* Reusing existing Docker credentials (e.g. from `docker login`)
* Accepting credentials via environment variables
* Or providing a CLI configuration option for registry authentication

With this, the Supabase CLI could pull images from registries that require authentication.

**Use case**
In CI pipelines, I would like to use the GitLab Dependency Proxy to cache Supabase images and reduce the startup time of the Supabase environment. For example, Docker authentication in GitLab CI is typically handled via:

```sh
echo "$CI_REGISTRY_PASSWORD" | docker login -u "$CI_REGISTRY_USER" --password-stdin "$CI_REGISTRY"
```

The Supabase CLI currently cannot take advantage of this authenticated registry setup.

**Describe alternatives you’ve considered**
There is no practical alternative other than relying on public registries with generous rate limits, which is not ideal for CI environments.

ffabss:
I looked into the issue further and it turns out this is partially possible already, but with an important caveat.

Authentication itself works if the login is done against the GitLab Dependency Proxy, not the regular container registry, e.g.:

```sh
echo "$CI_DEPENDENCY_PROXY_PASSWORD" | docker login -u "$CI_DEPENDENCY_PROXY_USER" --password-stdin "$CI_DEPENDENCY_PROXY_SERVER"
```

With this in place, Docker credentials are available and the CLI can authenticate.
However, it still fails for images like kong, because the CLI does not know that the dependency proxy is effectively a **mirror of docker.io**.

When `SUPABASE_INTERNAL_IMAGE_REGISTRY` is set to the proxy host, the CLI rewrites image references and ends up resolving images from the wrong namespace instead of preserving the original docker.io paths. This behavior seems to come from the image reference handling logic here:

[...]

PR #4762, opened 2026-01-27
feat: allow namespace override per image (#4725)

ffabss:

Feature

There is currently no support for overriding the namespace used when resolving container image references.  
This makes it impossible to pull certain images (e.g. `kong`) via the GitLab CI Dependency Proxy, which only supports images from `docker.io`.

Related issue: https://github.com/supabase/cli/issues/4725

## What is the new behavior?

This PR introduces support for per-image namespace configuration via environment variables such as `SUPABASE_INTERNAL_IMAGE_NAMESPACE_KONG=library`.

With this change, users can explicitly control the resolved image namespace, enabling images like `kong` to be pulled through the GitLab CI Dependency Proxy.

## Additional context

The image resolution logic was updated to apply the configured namespace when constructing the full image reference, and tests were added to cover the new behavior.

### DIFF
--- a/internal/utils/docker.go
+++ b/internal/utils/docker.go
@@ -194,15 +194,49 @@ func GetRegistry() string {
 	return strings.ToLower(registry)
 }
 
-func GetRegistryImageUrl(imageName string) string {
+// GetRegistryImageUrl builds a fully-qualified image reference.
+//
+// Expected input format:
+//
+//	NAMESPACE/IMAGE:TAG
+//
+// Behavior:
+//   - Uses the configured registry (via GetRegistry()).
+//   - Allows per-image namespace overrides via
+//     INTERNAL_IMAGE_NAMESPACE_"IMAGE_NAME".
+//   - Falls back to the "supabase" namespace for non-docker.io registries
+//     when no override is configured.
+func GetRegistryImageUrl(imageRef string) string {
 	registry := GetRegistry()
-	if registry == "docker.io" {
-		return imageName
+
+	// Split path segments (supports nested namespaces)
+	parts := strings.Split(imageRef, "/")
+	if len(parts) < 2 {
+		return imageRef // invalid format
 	}
-	// Configure mirror registry
-	parts := strings.Split(imageName, "/")
-	imageName = parts[len(parts)-1]
-	return registry + "/supabase/" + imageName
+
+	imageWithTag := parts[len(parts)-1]
+	namespace := strings.Join(parts[:len(parts)-1], "/")
+
+	// Split image and tag
+	imageParts := strings.SplitN(imageWithTag, ":", 2)
+	if len(imageParts) != 2 {
+		return imageRef // invalid format
+	}
+
+	imageName := imageParts[0]
+	imageTag := imageParts[1]
+
+	// Check for per-image namespace override
+	overrideKey := "INTERNAL_IMAGE_NAMESPACE_" + strings.ToUpper(imageName)
+	if override := viper.GetString(overrideKey); override != "" {
+		namespace = override
+	} else if registry != "docker.io" {
+		// Default namespace for non-docker.io registries
+		namespace = "supabase"
+	}
+
+	return fmt.Sprintf("%s/%s/%s:%s", registry, namespace, imageName, imageTag)
 }
 
 func DockerImagePull(ctx context.Context, imageTag string, w io.Writer) error {
--- a/internal/utils/docker_test.go
+++ b/internal/utils/docker_test.go
@@ -306,3 +306,49 @@ func TestExecOnce(t *testing.T) {
 
 	// TODO: mock tcp hijack
 }
+
+func TestGetRegistryImageUrl(t *testing.T) {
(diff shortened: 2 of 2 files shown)


sweatybridge:
The env var you are looking for is `SUPABASE_INTERNAL_IMAGE_REGISTRY=docker.io`

append this before supabase start and it will pull images from dockerhub

ffabss:
The issue is **not** that I can’t set the registry.

The issue is that the method I changed **explicitly checks for `docker.io`** and then adjusts the Docker image namespace differently.

This behavior is inherently necessary because **Supabase does not re-publish all images under the `supabase/*` namespace** (for example: `kong` is not published as `supabase/kong`).

For the `Kong` image on `docker.io`, there is **no image** under the namespace `supabase/kong`.
On `docker.io`, the correct namespace would be either:

* `library/kong`, or
* `kong`

So if someone is using a **custom pull-through cache** (still using `docker.io`) or the **GitLab Dependency Proxy**, the method `GetRegistryImageUrl` currently returns a URL for an image that **does not exist**.

### What my changes add

My changes introduce the possibility to **override the namespace per image**, so users can configure the correct namespace for cases like GitLab Dependency Proxy.

For example, Kong can be fixed by setting:

```text
SUPABASE_INTERNAL_IMAGE_NAMESPACE_KONG=library
```

### Example

The method `GetRegistryImageUrl` receives the input:

```text
library/kong:2.8.1
```

* If `SUPABASE_INTERNAL_IMAGE_REGISTRY` is set to `docker.io`, the method returns:

```text
docker.io/library/kong:2.8.1
```

* For any other value of `SUPABASE_INTERNAL_IMAGE_REGISTRY`, the method returns basically:

```text
$SUPABASE_INTERNAL_IMAGE_REGISTRY/supabase/kong:2.8.1
```

With the new namespace override, users can ensure the method returns valid image URLs even when images are **not available under `supabase/*`**.

sweatybridge:
> This behavior is inherently necessary because Supabase does not re-publish all images under the supabase/* namespace (for example: kong is not published as supabase/kong).

Could you let me know which registry are you using? We have kong published under supabase namespace for [GHCR](https://github.com/supabase/cli/pkgs/container/kong) and [ECR](https://gallery.ecr.aws/supabase/kong). These are the only ones we support officially right now.

We are only making an exception for `docker.io` because we don't control some of the namespaces there.

ffabss:
Im using the GitLab Dependency Proxy, which acts as a pull-through cache for `docker.io`. As far as I know, its not possible to use a different upstream registry with it.
See: https://docs.gitlab.com/user/packages/dependency_proxy/#use-the-dependency-proxy-for-docker-images

The only real workaround would be to set up and maintain a **separate pull-through cache** that targets a different upstream registry (e.g. GHCR or AWS ECR). However, that adds operational overhead (infrastructure + configuration + ongoing maintenance).

### Images required for running all services locally

I checked which images are needed to run all services locally and found the following list:

* `axllent/mailpit:v1.22.3`
* `darthsim/imgproxy:v3.8.0`
* `library/kong:2.8.1`
* `postgrest/postgrest:v14.3`
* `supabase/edge-runtime:v1.70.0`
* `supabase/gotrue:v2.185.0`
* `supabase/logflare:1.30.3`
* `supabase/postgres-meta:v0.95.2`
* `supabase/postgres:17.6.1.074`
* `supabase/postgres:17.6.1.074`
* `supabase/realtime:v2.72.0`
* `supabase/storage-api:v1.33.5`
* `supabase/studio:2026.01.26-sha-1b6582e`
* `supabase/supavisor:2.7.4`
* `timberio/vector:0.28.1-alpine`

These are the configured image names the CLI uses.

### Why the current behavior breaks with GitLab Dependency Proxy

The old logic for constructing the image URL effectively enforced the namespace `supabase` whenever the registry was **not** `docker.io`.

### Proposed solution

With the proposed code changes, the following environment variables can be used to ensure images are pulled from the correct namespaces:

* `SUPABASE_INTERNAL_IMAGE_NAMESPACE_MAILPIT=axllent`
* `SUPABASE_INTERNAL_IMAGE_NAMESPACE_IMGPROXY=darthsim`
* `SUPABASE_INTERNAL_IMAGE_NAMESPACE_KONG=library`
* `SUPABASE_INTERNAL_IMAGE_NAMESPACE_POSTGREST=postgrest`
* `SUPABASE_INTERNAL_IMAGE_NAMESPACE_VECTOR=timberio`

Additionally, the following would be used to direct Supabase to use the GitLab Dependency Proxy:

```text
SUPABASE_INTERNAL_IMAGE_REGISTRY=${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}
```

This will set the registry to something like:

```text
gitlab.example.com:443/my-group/dependency_proxy/containers
```

See: https://docs.gitlab.com/user/packages/dependency_proxy/#authenticate-within-cicd

ffabss:
@sweatybridge 
Could you please take a quick look at this and let me know what you think?

PR #4768, opened 2026-01-28
fix: remove dots from allowed project_id characters to prevent Vector container crashes

### LINKED ISSUE #4767: supabase start fails silently when project_id contains dots - Vector container crashes without helpful error message
### Problem
When a `project_id` in `supabase/config.toml` contains dots (e.g., `myapp-2.0`), the Vector analytics container crashes during startup with a vague error message about an invalid IPv4 address. The error message does not mention the `project_id` as the culprit, making it difficult to diagnose.

### Error Message
```
supabase_vector_myapp-2.0 container logs:
2026-01-28T08:24:27.453956Z  INFO vector::app: Internal log rate limit configured. internal_log_rate_secs=10
2026-01-28T08:24:27.454168Z  INFO vector::app: Log level is enabled. level="vector=info,codec=info,vrl=info,file_source=info,tower_limit=trace,rdkafka=info,buffers=info,lapin=info,kube=info"
2026-01-28T08:24:27.454268Z  INFO vector::app: Loading configs. paths=["/etc/vector/vector.yaml"]
thread 'main' panicked at 'invalid authority: InvalidIpv4Address', src/sinks/util/uri.rs:122:69
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The panic repeats multiple times, then the container becomes unhealthy and Supabase startup aborts with `Error status 500`.

### Root Cause
The `project_id` value is used internally to construct container names and URIs. When it contains dots (invalid characters for URI authorities), Vector's sink configuration receives a malformed authority string, causing the panic.

### Reproduction Steps

1. Create a new Supabase project or initialize one with:
```bash
supabase init myapp
```

2. Edit `supabase/config.toml` and modify the `project_id` to include a dot:
```toml
project_id = "myapp-2.0"
```

3. Run:
```bash
supabase start --debug
```

4. Observe the Vector container repeatedly crashing with the `invalid authority: InvalidIpv4Address` panic.

### Expected Behavior
- The CLI should validate the `project_id` format before starting containers
- If invalid characters are detected, it should show a clear error message like: `"project_id 'myapp-2.0' contains invalid characters. Use only alphanumerics, hyphens, and underscores."`
- Startup should fail cleanly with helpful guidance

### Workaround
Replace dots with underscores in the `project_id`:
```diff
- project_id = "myapp-2.0"
+ project_id = "myapp-2_0"
```

Then `supabase start` works correctly.

### Related
- #2538 (similar Vector container startup failures) - see [comment with workaround](https://github.com/supabase/cli/issues/2538#issuecomment-2544553520)

### Expected Fix
Add validation in the CLI to:
1. Check `project_id` format when parsing `config.toml`
2. Enforce alphanumeric characters, hyphens, and underscores only
3. Reject dots and other invalid characters early with a clear error message
4. Prevent container startup with invalid IDs

7ttp:
### Summary
Dots in `project_id` cause Vector analytics container to crash with `invalid authority: InvalidIpv4Address` panic,
 preventing Supabase from starting.

### Problem
Many  users where facing an issue that when `project_id` contains dots (e.g., `myapp-2.0`),  the value is used to construct Docker hostnames like `http://myapp-2.0:port`.
 Vector's URI parser interprets `2.0` as an incomplete IPv4 address,  causing a panic and crash loop. The existing regex allowed dots as valid characters.

### Solution
Remove `.` from the regex pattern. Now dots are automatically replaced with underscores during sanitization,
with a warning message to inform users. 

### Related

- closes https://github.com/supabase/cli/issues/4767

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -435,7 +435,7 @@ var (
 	initConfigEmbed    string
 	initConfigTemplate = template.Must(template.New("initConfig").Parse(initConfigEmbed))
 
-	invalidProjectId = regexp.MustCompile("[^a-zA-Z0-9_.-]+")
+	invalidProjectId = regexp.MustCompile("[^a-zA-Z0-9_-]+")
 	refPattern       = regexp.MustCompile(`^[a-z]{20}$`)
 )
 
@@ -975,10 +975,10 @@ func truncateText(text string, maxLen int) string {
 const maxProjectIdLength = 40
 
 func sanitizeProjectId(src string) string {
-	// A valid project ID must only contain alphanumeric and special characters _.-
+	// A valid project ID must only contain alphanumeric and special characters _-
 	sanitized := invalidProjectId.ReplaceAllString(src, "_")
 	// It must also start with an alphanumeric character
-	sanitized = strings.TrimLeft(sanitized, "_.-")
+	sanitized = strings.TrimLeft(sanitized, "_-")
 	// Truncate sanitized ID to 40 characters since docker hostnames cannot exceed
 	// 63 characters, and we need to save space for padding supabase_*_edge_runtime.
 	return truncateText(sanitized, maxProjectIdLength)


7ttp:
@sweatybridge 
(sry for the ping)
Can u have a quick look on this aswell 😁💚

sweatybridge (commented):
I will punt on this because it's a breaking change for existing config files.

Does new version of vector fix it?

PR #4778, opened 2026-01-30
fix: defaults to legacy anon key

sweatybridge:

fixes: https://github.com/supabase/cli/issues/4751
follow up https://github.com/supabase/cli/pull/4688

## What is the new behavior?

Keep legacy anon and service_role keys around unless user explicitly enables custom signing key.

## Additional context

supersedes https://github.com/supabase/cli/pull/4752

### DIFF
--- a/pkg/config/apikeys.go
+++ b/pkg/config/apikeys.go
@@ -74,7 +74,7 @@ func (a *auth) generateAPIKeys() error {
 
 func (a auth) generateJWT(role string) (string, error) {
 	claims := CustomClaims{Issuer: "supabase-demo", Role: role}
-	if len(a.SigningKeys) > 0 {
+	if len(a.SigningKeysPath) > 0 {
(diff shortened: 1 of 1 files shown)


mateuszbasiak:
Hello!
After this change I started to encounter following error while calling `supabase.auth.admin` using secret (`sb_secret_`) key from `supabase status`: 
```
[AuthApiError: invalid JWT: unable to parse or verify signature, token signature is invalid: signing method HS256 is invalid]
```

While using previous version of CI I'm able to avoid this issue. I didn't configure custom signing key. Is this error expected? Is there anything I could do to avoid it besides switching fully to asymmetric keys?

sweatybridge:
Sorry this became a bit of a whack a mole. There's another fix that we need to land https://github.com/supabase/cli/pull/4781

I will test more before merging.

sweatybridge:
Could you give v2.74.4 a try? It should accept `sb_secret_` key even if you are still using legacy api keys.

mateuszbasiak:
@sweatybridge Seems to be working now. Thank you for quick action

Issue #3593, opened 2025-05-21
Mounts denied for /socket_mnt/.../docker.sock with Docker Desktop when running supabase start


Skirja:
**Describe the bug**
When running supabase start on Fedora with Docker Desktop, the process fails with a "Mounts denied" error specifically for the Docker Desktop socket path (/socket_mnt/home/<user>/.docker/desktop/docker.sock). This occurs even though Docker Desktop's file sharing is configured to share the parent directory (/home/<user>/.docker/desktop), and basic Docker operations (like docker run hello-world and mounting regular files from the same directory) work correctly. The issue seems to be specific to Docker Desktop denying the mount of its own socket when requested by Supabase CLI.

**To Reproduce**

1. Initialize a new Supabase project: supabase init (accept defaults).
2. Attempt to start the Supabase services: supabase start.
3. Observe the process, which proceeds through database migrations and seeding.
4. See error: failed to start docker container: Error response from daemon: Mounts denied: The path /socket_mnt/home/skirja/.docker/desktop/docker.sock is not shared from the host and is not known to Docker

**Expected behavior**
supabase start should successfully start all local Supabase services without any "Mounts denied" errors, allowing for local development.

**System information**
Rerun the failing command with `--create-ticket` flag.
 - Ticket ID: 8d5d10aac33249ab92b8d2aa1eb050d5
 - Version of OS: Fedora Linux 42 (Workstation Edition)
 - Version of CLI: 2.23.4
 - Version of Docker: Docker version 28.1.1, build 4eba377

![Image](https://github.com/user-attachments/assets/f42f4e25-9d9c-4f13-b6a8-53ce0a5b5b1a)

![Image](https://github.com/user-attachments/assets/48bb76ae-567a-4046-914a-49ca1db7255b)

Melochaotic:
A slightly more secure solution is to `mkdir /socket_mnt` then add that to the file share.

Worked on Linux Mint 22.1

Thanks for the inspiration @zacwalls

btzo:
Complementing the above: `socket_mnt` is created at `/`, 
so you’ll need `sudo` to create that folder and set its group to docker to avoid permission issues. 

```bash
sudo mkdir /socket_mnt
sudo chown $USER:docker /socket_mnt
```

threehappypenguins:
> Complementing the above: `socket_mnt` is created at `/`, so you’ll need `sudo` to create that folder and set its group to docker to avoid permission issues.
> 
[...]

Issue #3512, opened 2025-05-03
Containers: postgres-meta and studio will not start on Docker Desktop 4.40.0 (187762) (MAC)


ThingEngineer:
**Describe the bug**
Supabase CLI starts and runs all containers/services on Docker Desktop 4.40.0 (187762) as expected. After upgrading to Docker Desktop 4.41.1 (191279) the `supabase start` command starts most of the containers without fail but near the end of the startup process, prints the logs for supabase_pg_meta_dolphin-ir then the logs for supabase_studio_dolphin-ir then immediately stops all containers and reports the pg_meta container as unhealthy and studio as not ready. Watching the Docker Desktop Containers dashboard postgres-meta and studio never start. The postgres-meta volume does go green on the Images dashboard but studio does not.

**To Reproduce**
1. Update to Docker Desktop 4.41.1 (191279)
2. Verify settings, quit Docker Desktop, full reboot
3. Start Docker Desktop and wait docker engine to start
4. Run `supabase start` on a known working project
5. See error:
```
supabase start --create-ticket

WARN: no SMS provider is enabled. Disabling phone login
WARNING: You are running different service versions locally than your linked project:
supabase/storage-api:v1.22.6 => v1.22.4
Run supabase link to update them.
WARNING: analytics requires mounting default docker socket: /var/run/docker.sock
supabase_pg_meta_dolphin-ir container logs:
{"level":"info","time":"2025-05-03T00:43:22.902Z","pid":1,"hostname":"6447abeb9142","msg":"Server listening at http://0.0.0.0:8080"}
{"level":"info","time":"2025-05-03T00:43:22.910Z","pid":1,"hostname":"6447abeb9142","msg":"Server listening at http://0.0.0.0:8081"}
supabase_studio_dolphin-ir container logs:
  ▲ Next.js 14.2.26
  - Local:        http://localhost:3000
  - Network:      http://0.0.0.0:3000

 ✓ Starting...
  ▲ Next.js 14.2.26
  - Local:        http://localhost:3000
  - Network:      http://0.0.0.0:3000

 ✓ Starting...
 ✓ Ready in 1421ms
Stopping containers...
supabase_pg_meta_dolphin-ir container is not ready: unhealthy
supabase_studio_dolphin-ir container is not ready: starting  
Try rerunning the command with --debug to troubleshoot the error.
Sent crash report: b17d028f771a460789ffb87641d3f18a
```

**Expected behavior**
[...]

PR #4791, opened 2026-02-02
fix: use correct docker.sock binding with vector (fixes #3127, #3512, #3593)

### LINKED ISSUE #3127: Add docker socket mount path to config.toml
The problem is that on rootless docker in Linux (In my case, I'm on nixos) the docker socket isn't in the default path of `/var/run/docker.sock`, but instead at `/run/user/1000/docker.sock`. 

Add an option to the config.toml file to be able to change the location of the mounted docker socket

**Additional context**
This concerns users on Linux running the supabase cli - since the vector service mounts the docker socket, it crashes when it's not able to.
I can probably name 4-5 issues + a bunch of discord issues regarding the issue of the vector service crashing because of the `permissions os 13 error` regarding this issue. A fix like the one above could potentially solve all of these.

unlair:
This fixes Docker Desktop on Linux failing to start the vector container. The container tries to bind to the Docker Desktop socket directly, which is unsupported. This fix resolves #3127, #3512, and #3593.

Bug fix.

When using Docker Desktop on Linux, Vector fails to start when doing `supabase start`.

## What is the new behavior?

When using Docker Desktop on Linux, Vector successfully starts when doing `supabase start`.

## Additional context

Using Docker or Docker Desktop on Linux, the specified host binding path to the Docker socket should be `/var/run/docker.sock`, as that path has special handling under the hood (see https://github.com/docker/for-mac/issues/6545#issuecomment-1295599122). **Note that it does not directly use the socket sitting at `/var/run/docker.sock` on the host machine.** My understanding is that what it ends up doing under the hood is essentially using what you get with `/var/run/docker.sock.raw` from outside Docker, as mentioned here: https://docs.docker.com/extensions/extensions-sdk/guides/use-docker-socket-from-backend/

### DIFF
--- a/.github/workflows/api-sync.yml
+++ b/.github/workflows/api-sync.yml
@@ -18,7 +18,7 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
-      - uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
+      - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
           go-version-file: go.mod
           cache: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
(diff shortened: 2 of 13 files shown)


unlair:
@avallete @jgoux 
I'd like to know how I can get these 3 PRs looked at, especially this one here. They solve some long-standing issues and are required for me to be able to use supabase-cli on Debian (I've been using my own fork in the meantime).

unlair:
> @avallete @jgoux I'd like to know how I can get these 3 PRs looked at, especially this one here. They solve some long-standing issues and are required for me to be able to use supabase-cli on Debian (I've been using my own fork in the meantime).

@avallete pinging again, would really like someone to take a look at this or tell me why it's being passed over.

Issue #4817, opened 2026-02-07
JWT validation broken: GOTRUE_JWT_VALIDMETHODS env var not recognized by GoTrue v2.186.0


Roozbehy:
### Description

CLI v2.74.4+ sets the environment variable `GOTRUE_JWT_VALIDMETHODS` (no underscore between VALID and METHODS), but the GoTrue v2.186.0 Docker image bundled with recent CLI versions expects `GOTRUE_JWT_VALID_METHODS` (with underscore) due to the `split_words` struct tag added in [gotrue#2334](https://github.com/supabase/auth/pull/2334).

This means GoTrue never receives the configured valid methods, defaults to ES256-only signing, and edge functions with `verify_jwt = true` reject all tokens with "Invalid JWT" because the edge runtime only has the HS256 secret (`SUPABASE_INTERNAL_JWT_SECRET`).

### Steps to Reproduce

1. Use CLI v2.75.0 or later (or `version: latest` in CI)
2. Run `supabase start`
3. Create a test user and sign in to get an access token
4. Call any edge function that has `verify_jwt = true`
5. Observe `401 {"msg":"Invalid JWT"}`

Functions with `verify_jwt = false` work fine.

### Expected Behavior

GoTrue should accept HS256 tokens and edge functions should validate them successfully, as they did with CLI v2.72.7 + GoTrue v2.185.0.

### Actual Behavior

Every authenticated edge function request returns `401 {"msg":"Invalid JWT"}`.

### Root Cause

| Version | Env var set by CLI | GoTrue expects | Result |
|---------|-------------------|----------------|--------|
| CLI v2.72.7 + GoTrue v2.185.0 | `GOTRUE_JWT_VALID_METHODS` | Couldn't parse it (no `split_words`), fell back to permissive defaults | **Works** |
| CLI v2.74.4+ + GoTrue v2.186.0 | `GOTRUE_JWT_VALIDMETHODS` | `GOTRUE_JWT_VALID_METHODS` (has `split_words` from auth#2334) | **Broken** — variable ignored, defaults to ES256 only |

The fix in [CLI PR #4781](https://github.com/supabase/cli/pull/4781) (v2.74.4) removed the underscore to match the old GoTrue behavior, but GoTrue v2.186.0 added `split_words` which re-introduces the underscore expectation. The two fixes worked at cross-purposes.

### Fix

In `internal/start/start.go`, change:
```go
"GOTRUE_JWT_VALIDMETHODS=HS256,RS256,ES256"
[...]

PR #4818, opened 2026-02-07
fix: set correct valid methods env var

sweatybridge:
Reverts supabase/cli#4781


singh-inder:
Hi @sweatybridge, just wanted to flag that this env var hasn’t been updated in auth v2.186.0 yet. Related: https://github.com/supabase/cli/issues/4820

sweatybridge:
lol that's unfortunate. Let me revert the revert. Would you like to close the gotrue PR so it doesn't confuse more folks?

sweatybridge:
For reference https://github.com/supabase/cli/issues/4817

Issue #4839, opened 2026-02-10
TLS error with self-hosted database despite sslmode=disable (CLI 2.75.0) - regression of #4288


anthonyg2003:
**Describe the bug**
I'm experiencing a TLS connection error when using `supabase db push` and other supabase commands with a self-hosted Supabase instance, despite specifying `sslmode=disable` in the connection string. This appears to be a regression of the issue reported in #4288, which was marked as resolved.

When running database commands without the --debug flag, the CLI fails with:
```
failed to connect to postgres: failed to connect to `host=<my database's local ip> user=supabase_admin database=postgres`: tls error (server refused TLS connection)
```

However, the exact same command succeeds when the `--debug` flag is added.

**To Reproduce**

1. Set up a self-hosted Supabase instance with PostgreSQL exposed directly (not through Supavisor)

2. Run the following command:   
```
supabase db push --db-url "postgresql://supabase_admin:<password>@<ip>:5433/postgres?sslmode=disable"
```

3. Observe the TLS error

4. Run the same command with --debug:
```
supabase db push --db-url "postgresql://supabase_admin:password@<ip>:5433/postgres?sslmode=disable" --debug
```

5. Observe that it succeeds

**Expected behavior**
The command should respect the sslmode=disable parameter and connect successfully without requiring the --debug flag.

**Actual Behavior**
- Without `--debug`: Fails with "tls error (server refused TLS connection)"
- With `--debug`: Works perfectly, connects and applies migrations successfully

The CLI appears to ignore the `sslmode=disable` parameter unless the `--debug` flag is present

**Screenshots**
NA

**System information**
Rerun the failing command with `--create-ticket` flag.
 - Ticket ID: 1c217e156c264a7985ed08b5a76f06d9
 - Version of OS: Windows 11 Pro (Build 26220)
 - Version of CLI: 2.75.0 (installed with scoop)
 - Version of Docker: 29.1.3, build f52814d
 - Versions of services: 
```
   SERVICE IMAGE          | LOCAL                  | LINKED
  ------------------------|------------------------|--------
   supabase/postgres      | 15.8.1.085             | -
   supabase/gotrue        | v2.186.0               | -
   postgrest/postgrest    | v14.3                  | -
   supabase/realtime      | v2.73.2                | -
   supabase/storage-api   | v1.35.3                | -
   supabase/edge-runtime  | v1.70.0                | -
   supabase/studio        | 2026.01.27-sha-2a37755 | -
[...]

Issue #4288, opened 2025-10-11
failed to connect to postgres: failed to connect to `host=127.0.0.1 user=postgres database=contrib_regression`: tls error (server refused TLS connection)


NixBiks:
**Describe the bug**
I get the following error when running `supabase db diff --local`

```
Diffing schemas...
Connecting to remote database...
failed to connect to postgres: failed to connect to `host=127.0.0.1 user=postgres database=contrib_regression`: tls error (server refused TLS connection)
Try rerunning the command with --debug to troubleshoot the error.
```

If I run `supabase db diff --debug` then it works just fine

**System information**
Rerun the failing command with `--create-ticket` flag.
 - Ticket ID: 2ba80b05e78343938254ec85ab0b5a97
 - Version of OS: macOS
 - Version of CLI: 2.48.3
 - Version of Docker: 28.3.2
 - Versions of services: 
```
  
   SERVICE IMAGE          | LOCAL                  | LINKED     
  ------------------------|------------------------|------------
   supabase/postgres      | 17.6.1.003             | 17.6.1.003 
   supabase/gotrue        | v2.179.0               | v2.179.0   
   postgrest/postgrest    | v13.0.5                | v13.0.5    
   supabase/realtime      | v2.51.11               | -          
   supabase/storage-api   | v1.28.0                | -          
   supabase/edge-runtime  | v1.68.3                | -          
   supabase/studio        | 2025.10.01-sha-8460121 | -          
   supabase/postgres-meta | v0.91.6                | -          
   supabase/logflare      | 1.22.4                 | -          
   supabase/supavisor     | 2.7.0                  | -      
```

sweatybridge:
We fixed this in the beta release. Could you give that a try?

```bash
npx supabase@beta db diff
```

NixBiks:
> We fixed this in the beta release. Could you give that a try?
> 
> npx supabase@beta db diff

Indeed it works - thank you.

sweatybridge:
Thanks for confirming. I will make release tmr to stable channel.

PR #4841, opened 2026-02-10
Honor sslmode=disable for db connections (fix TLS regression)

Flamki:

Bug fix.

CLI 2.75.x attempts a TLS connection even when `sslmode=disable` is set in `--db-url`, causing `tls error (server refused TLS connection)`. The same command succeeds with `--debug` because TLS is disabled there.  
Issue: #4839 (regression of #4288)

## What is the new behavior?

When `sslmode=disable` is present in the connection string, the CLI explicitly disables TLS, so non‑TLS self‑hosted databases connect successfully without needing `--debug`.

## Additional context

This aligns normal behavior with the `--debug` path, which already disables TLS. No user‑visible changes beyond respecting `sslmode=disable`.

### DIFF
--- a/pkg/pgxv5/connect.go
+++ b/pkg/pgxv5/connect.go
@@ -18,6 +18,9 @@ func Connect(ctx context.Context, connString string, options ...func(*pgx.ConnCo
 	if err != nil {
 		return nil, errors.Errorf("failed to parse connection string: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


Flamki:
CI status note:
- CI / Codegen` fails during `go generate` with `DiskRequestBody` discriminator mapping (pre-existing issue).
- CI / Link` fails due to missing `project-ref` flag in CI.

All build/lint/test checks for this change are green.

PR #4847, opened 2026-02-11
fix: only upload functions declared in config.toml

singh-inder:

Bug fix

1. If a function dir with index.ts is present on disk but not defined in config.toml, it is still deployed. I assume this isn't expected behavior.
  	To Reproduce:
  	- run: `supabase functions new should-not-upload`
  	- comment out/remove this function's config from config.toml
  	- run: `supabase functions deploy --use-api`
  	- function is still deployed.
 
	image
	
	Screenshot from 2026-02-12 03-07-38-final

 
2. deploy success msg is shown for all function slugs, even the ones which aren't deployed. Also the slugs slice contains duplicates, so does the msg derived from slugs slice
  
	  `should-not-upload` is disabled `functions.should-not-upload.enabled=false` but the deploy success msg still prints this slug and msg contains duplicates.
	  	Screenshot from 2026-02-12 01-07-17-final

## What is the new behavior?

- dont upload functions if not defined in config.toml. `<function>.enabled` behavior remains same.
- dedup slugs returned from `GetFunctionSlugs` and only print deploy success msg for functions actually deployed.

### DIFF
--- a/internal/functions/deploy/deploy.go
+++ b/internal/functions/deploy/deploy.go
@@ -110,7 +110,7 @@ func GetFunctionConfig(slugs []string, importMapPath string, noVerifyJWT *bool, 
 	for _, name := range slugs {
 		function, ok := utils.Config.Functions[name]
 		if !ok {
-			function.Enabled = true
+			function.Enabled = false
 			function.VerifyJWT = true
 		}
 		// Precedence order: flag > config > fallback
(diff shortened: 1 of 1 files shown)


singh-inder:
I think loop iteration can be skipped here.

```
if !ok {
			function.Enabled = false
			function.VerifyJWT = true
     		functionConfig[name] = function
     		continue
		}
```

PR #4857, opened 2026-02-15
fix: kong stops responding

unlair:
kong easily gets overwhelmed when receiving many requests, resulting in an error complaining about not having enough workers. This was due to KONG_NGINX_WORKER_PROCESSES=1 being set. Kong automatically determines the right number of worker processes (1 per allocated CPU) when this is not specified, which resolves the issue.

Bug fix (removing a bad env var for the Kong container).

If many parallel requests are made to the Storage API, Kong will quickly stop responding and start terminating socket connections.

## What is the new behavior?

Kong is able to handle many parallel connections without choking.

### DIFF
--- a/apps/cli-go/internal/start/start.go
+++ b/apps/cli-go/internal/start/start.go
@@ -526,7 +526,6 @@ vector --config /etc/vector/vector.yaml
 					// Ref: https://github.com/Kong/kong/issues/3974#issuecomment-482105126
 					"KONG_NGINX_PROXY_PROXY_BUFFER_SIZE=160k",
 					"KONG_NGINX_PROXY_PROXY_BUFFERS=64 160k",
-					"KONG_NGINX_WORKER_PROCESSES=1",
 					// Use modern TLS certificate
 					"KONG_SSL_CERT=/home/kong/localhost.crt",
 					"KONG_SSL_CERT_KEY=/home/kong/localhost.key",


avallete (changes requested):
Thanks for the contribution and for digging into this! The diagnosis is spot on a single Nginx worker can definitely get saturated under parallel load (bulk Storage uploads being the classic case).

For context, the `KONG_NGINX_WORKER_PROCESSES=1` was set intentionally to reduce Kong's memory footprint in the local dev stack. This is a common pattern for projects embedding Kong in Docker see https://github.com/edgexfoundry/edgex-compose/issues/177 for a similar rationale. With `auto`, Kong spawns one worker per visible CPU core, which on a modern dev machine (8–16 cores) means 8–16 Nginx workers each with their own Lua VM and connection pools. Given that `supabase start` already launches ~12 containers, the extra memory pressure adds up.

Rather than removing the line entirely, would you consider setting it to `2` instead? That should resolve the concurrency bottleneck you're hitting (two workers can handle a solid amount of parallel connections) while keeping the local stack reasonably lightweight for the majority of users who aren't doing heavy parallel workloads.

For users who genuinely need `auto` (e.g., load testing against the local stack), this could also be a good candidate for a `config.toml` override down the line (or a passed down env config) but that's a separate effort and shouldn't block this fix.

What do you think?

unlair:
Thanks for the thoughtful reply and added context. Giving Docker Desktop all of their system's resources seems inherently risky, and Kong should only see the number of CPUs made available to it.
It does look like Kong is the heaviest container by about 100MB, but it's still <15% of the total memory usage after `supabase start` on my machine.

Kong's developer docs suggest:

> If you run Kong Gateway pods on Kubernetes worker nodes with n CPUs, allocate n-2 or n-1 to Kong Gateway, and configure a worker process count equal to this number. This ensures that any configured daemons and Kubernetes processes, like kubelet, don’t contend for resources with Kong Gateway.
([source](https://developer.konghq.com/gateway/performance/optimize/))

Perhaps a better solution here would be settings container limits for the Kong container to min(2,n-2) CPUs based on Docker Desktop's allocated resources? It seems like resource limits should be something imposed on the container, rather than configured _inside_ the container. Please let me know what you think.

avallete:
Thanks for the detailed write-up and for digging into the Kong docs — really helpful context.

A bit of history on that line: `KONG_NGINX_WORKER_PROCESSES=1` wasn't arbitrary. It was added in #1271 specifically to minimize the local stack's memory usage, since nginx otherwise spawns one worker per CPU core. It originally shipped as a config option (default `1`, with autodetect when unset) and was later hardcoded to `1` in 2bcc9c7 when we trimmed the config surface. So removing the line outright would re-introduce the exact behavior that change was guarding against worker count (and memory) scaling with core count on larger dev machines.

Rather than computing container CPU limits (which gets messy across Docker Desktop's VM vs. native Linux), I'd suggest making this an opt-in shell override and keeping `1` as the default. We already do this for the storage container (see `appendStorageVectorEnv` in `internal/start/start.go`), so it fits an existing pattern  read `KONG_NGINX_WORKER_PROCESSES` from the environment, falling back to `1` when unset.

That way anyone hitting the worker bottleneck can run with `KONG_NGINX_WORKER_PROCESSES=auto` (one worker per core) or a fixed count, while the default footprint stays unchanged for everyone else and there's no `config.toml` change needed. It's essentially a one-line change at the Kong env block.

Would you be up for updating the PR along those lines? Happy to help if useful.

avallete:
Applied the env override approach here: https://github.com/supabase/cli/pull/5648
It should be enough for anyone to unblock himself, while preserving concervative defaults.

unlair:
Thanks, that works! Might be a good idea to include this in https://supabase.com/docs/guides/troubleshooting for anyone who hits the same issue (Kong locking up when it gets overwhelmed)

avallete:
Good idea, ongoing PR here: supabase/supabase#47198

PR #4911, opened 2026-02-28
chore: remove unmocked keyring test

tobert:
## Summary
- Remove `internal/utils/credentials/keyring_test.go` which calls
  `zalando/go-keyring` directly against the system D-Bus secret service
- `TestDeleteAll` fails on headless Linux/CI with "The name is not activatable"
  because no `org.freedesktop.secrets` provider is running
- The identical coverage (Get, Set, Delete, DeleteAll + error paths) already
  exists in `store_test.go` using `keyring.MockInit()`, per CONTRIBUTING.md's
  guidance to mock external systems

## Test plan
- [x] `go test ./internal/utils/credentials/... -race -v -count=1 -failfast` passes
- [x] No coverage lost — `store_test.go` covers all KeyringStore methods with mocks


tobert:
Another option would be to tag it as an integration test that so it doesn't run in the default `go test ./...`.

PR #4957, opened 2026-03-15
feat(db): add `supabase db advisors` command for checking security and performance issues

Rodriguespn:
## Summary

Add `supabase db advisors` command that checks the database for security and performance issues — the CLI equivalent of the `get_advisors` MCP tool.

### Why do we need this if we already have `get_advisors` in the MCP server?

Same reasoning as `supabase db query` ([#4955](https://github.com/supabase/cli/pull/4955)): the MCP server works well for agents that go through the OAuth flow for remote projects, but for local development the human still has to reload the agent session to load MCP tools into context. This adds friction to a path that should be fully agentic (no human in the loop).

With this command, an agent can check the security posture of the local database using just the CLI — no MCP configuration needed.

**Example use case: security-aware migrations.** The agent applies a migration that creates new tables, then runs `supabase db advisors --type security` to verify RLS is enabled and policies exist. If issues are found, the agent can fix them before committing the migration. Today this requires the `get_advisors` MCP tool to be configured. With this command, the entire flow only needs the CLI.

### Implementation

- **Local** (`supabase db advisors`, default): Executes the same [splinter](https://github.com/supabase/splinter) SQL lints that the Studio dashboard runs. Uses **pgx** to connect directly to the local Postgres — consistent with every other local `db` subcommand (`push`, `pull`, `diff`, `lint`, `test`, `reset`, `dump`). Runs in a read-only transaction (always rolled back). No extensions required.
- **Remote** (`supabase db advisors --linked`): Calls the Management API endpoints (`GET /v1/projects/{ref}/advisors/security` and `GET /v1/projects/{ref}/advisors/performance`), authenticated with the access token from `supabase login`.

### Differences from `db lint`

| | `db lint` | `db advisors` |
|---|---|---|
| **Purpose** | Checks PL/pgSQL function bodies for type errors | Checks database security posture and performance issues |
| **Scope** | Function-level code quality | Schema-level security/performance (RLS, indexes, auth config) |
| **Extension** | Requires `plpgsql_check` | No extensions needed |
| **Checks** | Syntax/type errors in functions | 23 checks: missing RLS, exposed auth.users, unused indexes, missing primary keys, etc. |

### Usage

```bash
# Local (default) — checks the local database
supabase db advisors

# Security checks only
supabase db advisors --type security

# Performance checks only
supabase db advisors --type performance

# Remote — checks the linked project via Management API
supabase db advisors --linked

# Include INFO-level issues (default is WARN+)
supabase db advisors --level info

# CI: exit non-zero on errors
supabase db advisors --fail-on error
```

## Test plan

`go test ./internal/db/advisors/...` — 15 unit tests covering:
  - Local: query parsing, empty results, query errors
  - Filtering: by type (security/performance/all), by level (info/warn/error), combined
  - Output: JSON encoding, no-issues message, fail-on thresholds
  - API: response conversion, security/performance fetch, error handling

### DIFF
--- a/internal/db/advisors/advisors.go
+++ b/internal/db/advisors/advisors.go
@@ -1,0 +1,259 @@
+package advisors
+
+import (
+	"context"
+	_ "embed"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/go-errors/errors"
+	"github.com/jackc/pgconn"
+	"github.com/jackc/pgx/v4"
+	"github.com/supabase/cli/internal/utils"
+	"github.com/supabase/cli/pkg/api"
+)
+
+var (
+	AllowedLevels = []string{
+		"info",
+		"warn",
+		"error",
+	}
+
+	AllowedTypes = []string{
+		"all",
+		"security",
+		"performance",
+	}
+
+	//go:embed templates/lints.sql
+	lintsSQL string
+)
+
+type LintLevel int
+
+func toEnum(level string) LintLevel {
+	switch level {
+	case "INFO", "info":
+		return 0
+	case "WARN", "warn":
+		return 1
+	case "ERROR", "error":
+		return 2
+	}
+	return -1
+}
+
+type Lint struct {
+	Name        string           `json:"name"`
+	Title       string           `json:"title"`
+	Level       string           `json:"level"`
+	Facing      string           `json:"facing"`
+	Categories  []string         `json:"categories"`
+	Description string           `json:"description"`
+	Detail      string           `json:"detail"`
+	Remediation string           `json:"remediation"`
+	Metadata    *json.RawMessage `json:"metadata,omitempty"`
+	CacheKey    string           `json:"cache_key"`
+}
+
+func RunLocal(ctx context.Context, advisorType string, level string, failOn string, config pgconn.Config, options ...func(*pgx.ConnConfig)) error {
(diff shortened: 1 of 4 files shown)


internal/db/advisors/advisors.go
```diff
@@ -0,0 +1,259 @@
...
+func filterByLevel(lints []Lint, level string) []Lint {
+	minLevel := toEnum(level)
+	var filtered []Lint
+	for _, l := range lints {
+		if toEnum(l.Level) >= minLevel {
+			filtered = append(filtered, l)
+		}
+	}
+	return filtered
+}
```
avallete: **nitpick**

Those two seems to do relatively the same work. Consider calling `filterByLevel` within `filterLints` for re-usability, or merge them all in favor of a single `filterLints` which would `no-op` if the lint type is "all" for remote ?

avallete (approved):
Overall LGTM, just a minor nit.

internal/db/advisors/advisors.go
Rodriguespn: Went with merging them into a single filterLints. Since RunLinked already selects which API endpoints to call based on the type, the type filter on the result set is effectively a no-op - matchesType with "all" returns true immediately. So we just pass "all" as the type from RunLinked and drop filterByLevel entirely. One function, two parameters, either can be permissive

PR #4966, opened 2026-03-17
feat: add pg delta declarative sync command

avallete:
## Summary

This PR introduces an experimental pg-delta powered declarative schema workflow to the CLI.

It adds a new `supabase db declarative` command group so users can generate declarative SQL files from a database, edit those files as the source of truth, and sync them back into the migration workflow by generating a new migration from the declarative diff.

In addition to the new command surface, this PR expands the underlying pg-delta integration with declarative export/apply support, reusable catalog caching, explicit source/target diffing, configurable declarative paths and formatting, and better debug artifacts for failed sync/apply flows.

## Main Changes

- Adds `supabase db declarative generate` to export a database into structured declarative SQL files.
- Adds `supabase db declarative sync` to diff declarative files against the migrations state, generate a migration, warn on destructive statements, and optionally apply it locally.
- Adds pg-delta declarative export and apply support through shared edge-runtime execution.
- Extends `supabase db diff` with explicit `--from` / `--to` refs and optional `--output`.
- Extends `supabase db pull` so it can export declarative schema files through pg-delta instead of only generating timestamped migration SQL.
- Introduces hashed catalog caching for both migrations and declarative schemas to reduce repeated shadow-db work.
- Adds debug bundle generation for sync/apply failures to make pg-delta issues easier to report and reproduce.
- Adds `[experimental.pgdelta]` config support, including:
  - `enabled`
  - `declarative_schema_path`
  - `format_options`

## End-to-End Flow

1. Enable `[experimental.pgdelta]` in `config.toml`.
2. Run `supabase db declarative generate` to export the current database state into declarative SQL files.
3. Edit the declarative schema files locally.
4. Run `supabase db declarative sync`.
5. The CLI builds a source catalog from migrations and a target catalog from the declarative files, reusing cached snapshots when possible.
6. pg-delta computes the SQL diff and the CLI writes a new migration file.
7. The CLI surfaces drop warnings and can optionally apply the generated migration to the local database.

## Notes

- Related flows such as `db pull`, `db diff`, `db start`, `db push`, and migration rollback/setup were updated so pg-delta-backed catalog state stays warm and consistent across commands.
- The declarative flow is gated behind the experimental pg-delta configuration.


jgoux (commented):
I'm just not fully convinced by the command name with "declarative". I would prefer we rely on `supabase db schema` namespace even if it means a new `--declarative` flag or a new `supabase db schema declarative` subcommand as we're really treating the same subject here.

avallete:
> I'm just not fully convinced by the command name with "declarative". I would prefer we rely on `supabase db schema` namespace even if it means a new `--declarative` flag or a new `supabase db schema declarative` subcommand as we're really treating the same subject here.

Updated in favor of `supabase db schema declarative`, must either run with explicit `--experimental` flag our have:

```
[experimental.pgdelta]
enabled = true
```

In the config.

jgoux (approved):

PR #4967, opened 2026-03-17
fix(db): split advisors session setup from lint query

### LINKED ISSUE #4965: db advisors fails for local projects with custom ports and multi-statement lint SQL
`supabase db advisors` fails when executing the embedded local lint SQL.

Tagging @Rodriguespn since the original advisors implementation came from commit `43ce0a6a`.

## Repro

Run the command against a local database with an explicit Postgres URL:

```bash
go run . db advisors --db-url 'postgresql://postgres:postgres@127.0.0.1:55322/postgres?sslmode=disable' --debug
```

## Actual Behavior

The command connects successfully, starts a transaction, and then fails when executing the embedded lint SQL:

```text
```

The debug trace points at `internal/db/advisors/advisors.go` in the `tx.Query(ctx, lintsSQL)` path.

## Root Cause

`lintsSQL` begins with a separate session statement:

```sql
set local search_path = '';
```

That setup statement and the actual lint query are bundled together and sent through the prepared-statement path, which Postgres rejects.

## Expected Behavior

`supabase db advisors` should execute the session setup and the lint query successfully against a local database.

## Suggested Fix

Split the session setup from the lint query, for example:

1. `tx.Exec(ctx, "set local search_path = ''")`
2. `tx.Query(ctx, <remaining lint query>)`

jsj:
Fixes #4965.

## Summary
- split the local advisors session setup from the embedded lint query
- execute `set local search_path = ''` separately before running the lint SQL
- update advisors tests to cover the split query path and setup failure handling

## Validation
- `go test ./internal/db/advisors`
- verified against a live local Supabase project

### DIFF
--- a/internal/db/advisors/advisors.go
+++ b/internal/db/advisors/advisors.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	"github.com/go-errors/errors"
 	"github.com/jackc/pgconn"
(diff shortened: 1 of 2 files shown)


Rodriguespn (approved):
Thanks for taking a look at this @jsj

avallete (approved):

Issue #4788, opened 2026-02-02
Adding a download all option to CLI functions would be great


a0m0rajab:
**Is your feature request related to a problem? Please describe.**

When i create a new project for supabase i might need to update or download all functions from supabase and it's super annoying to do that by writing each function name, it would be greate to be able to download all functions

**Describe the solution you'd like**
If someone used: `supabase functions download`, it should be able to download all functions, not just one.

**Describe alternatives you've considered**

i thought about having options like: 
`supabase functions download --all`, but the no-argument option felt better.

PR #5005, opened 2026-03-28
fix(functions): support `--all` flag to download all functions

### LINKED ISSUE #5004: Dump All Functions At Once
**Problem**
Moving functions from a development db into a production db will take a lot of time manually dumping function by function

**Solution**
It would be nice if we had a flag to dump all functions together, is it some thing you don't want to have for some reason or I can open a PR introducing this feature?

Rooffeell:
Previously, the command `supabase functions download --all` won't work, most of cli tools use the `--all` flag to do all, so I added it
Closes #5004

### DIFF
--- a/cmd/functions.go
+++ b/cmd/functions.go
@@ -53,7 +53,7 @@ var (
 				useDocker = false
 			}
 			slug := ""
-			if len(args) > 0 {
+			if len(args) > 0 && !downloadAll {
 				slug = args[0]
 			}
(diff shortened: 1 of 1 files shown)


singh-inder:
I believe not having the `--all` flag is expected behavior, and changing it now could break existing workflows that depend on this #4788

Rooffeell:
> I believe not having the --all flag is expected behavior, and changing it now could break existing workflows that depend on this #4788

I feel that using --all will be better though because this is what most cli tools do, what about adding support for --all without breaking the no flag one?

PR #5019, opened 2026-04-01
feat: add posthog telemetry

avallete:
## What

Adds product analytics to the Supabase CLI via PostHog. This gives us structured, privacy-conscious telemetry to understand how the CLI is actually being used which commands are run, in what environments (CI, TTY, AI agents), and with what outcomes.

Closes: CLI-1243

## Why

We currently have limited visibility into CLI usage patterns. Sentry gives us crash data, but nothing about normal usage: which commands are popular, how long they take, whether users are running in CI vs. interactive terminals, or whether AI coding agents are driving invocations. PostHog fills that gap and lets us make data-informed decisions about where to invest CLI effort.

### New `internal/telemetry/` package

A self-contained telemetry subsystem with clean separation of concerns:

- **`client.go`**  PostHog HTTP client, handles batching and flushing events
- **`service.go`**  High-level `Service` type that owns the client lifecycle and `Capture()` method; stored in context via `WithService` / `FromContext`
- **`state.go`**  Persistent anonymous distinct ID (generated once per machine, stored on disk)
- **`project.go`**  Captures project-scoped telemetry events (on `link`, `login`, `start`)
- **`events.go`**  Central registry of all event names and property keys (addresses the "telemetry-constants" pattern used on the frontend)

### Analytics context layer (`cmd/root_analytics.go`)

All analytics signal gathering lives in one file:

- **Command context**: full command path, unique run ID, and changed flags
- **Flag value redaction**: flag values are redacted by default. Only flags explicitly annotated as telemetry-safe (e.g. `--project-ref`), boolean flags, and enum flags pass their actual values
- **Environment detection**: CI (`GITHUB_ACTIONS`, `BUILDKITE`, `GITLAB_CI`, etc.), TTY status, AI agent detection
- **Env signal collection**: presence-only keys (just reports `true` if set) and value keys (reports the actual value, truncated to 80 chars)

### Integration into the command lifecycle

- `PersistentPreRunE` on the root command initializes the telemetry service and attaches it to the command context
- `Execute()` now uses `ExecuteC()` to get the executed command back, then captures a `command_executed` event with exit code and duration before flushing
- PostHog API key and endpoint are injected at build time via `ldflags` in `.goreleaser.yml`
- Release workflow (`release-beta.yml`) updated to pass `POSTHOG_API_KEY` and `POSTHOG_ENDPOINT` secrets

### Privacy

- Respects the existing `supabase telemetry disable` opt-out
- Anonymous distinct ID no PII is collected
- Free-text flag values (passwords, db-urls, file paths, etc.) are redacted unless the flag is explicitly allow-listed via `markFlagTelemetrySafe()`
- Env variable values are truncated; presence-only keys just report a boolean

## Events captured

| Event | When | Key properties |
|---|---|---|
| `command_executed` | After every CLI invocation | command, flags, exit_code, duration_ms, is_ci, is_tty, is_agent, env_signals |
| `project_linked` | On `supabase link` | project_ref |
| `login_completed` | On `supabase login` | — |
| `local_dev_started` | On `supabase start` | — |

## What to review

- The flag redaction logic in `root_analytics.go` the allow-list approach (safe annotation on specific flags) is intentionally conservative
- The `env_signals` design presence-only vs. value keys split
- Whether the set of telemetry-safe flags (`--project-ref`, `--project-id`, `--org-id`, `--version`) covers the right surface area without over-collecting


seanoliver (approved):
Tested locally with PH staging project and everything looks great!

The `env_signals` property is 🔥  — this is going to be _super_ useful for segmentation.

pamelachia:
one thing from the telemetry side: the 4 events are defined as inline strings at each call site. on the frontend we have telemetry-constants.ts that acts as a registry for every event with documented properties for each event, which makes it easy to see what's being tracked, keep naming consistent, and lets data people know what to expect without reading through the Go code. 

would it be possible to add something similar for the CLI? even just a constants file in internal/telemetry/ with all event names and expected properties in one place. makes it way easier for us to track what exists and catch naming issues before they ship

jgoux (approved):

PR #5050, opened 2026-04-08
fix(docker): bump mailpit to an ARM-safe image

jimboswankster:
## Summary
Updates the local development Mailpit image from `axllent/mailpit:v1.22.3` to `axllent/mailpit:v1.29.6`.

## Why
On Apple Silicon, the Mailpit image currently used by the CLI can resolve to an amd64 payload that segfaults under emulation during `supabase start`, which leaves the local `supabase_inbucket_*` container restarting and breaks the default local startup path.

I reproduced this on an M1 Mac with Supabase CLI 2.84.2. The failing local image path was the Supabase Mailpit tag derived from this Dockerfile template. Direct tests showed:

- the older Mailpit image crashed with `SIGSEGV: segmentation violation`
- upstream `axllent/mailpit:v1.29.6` pulled and ran as `linux/arm64`
- substituting the newer image restored a healthy Mailpit container and allowed full `supabase start` to complete successfully

## Validation
- confirmed the older Mailpit image crashed on Apple Silicon during container startup
- confirmed `axllent/mailpit:v1.29.6` runs cleanly as a native ARM image
- confirmed the local Supabase stack starts successfully once Mailpit uses the newer image


jgoux:
Hey, thanks for the contribution!
I just updated `mailpit` to its latest version. It should be available on our `beta` channel shortly!

PR #5054, opened 2026-04-08
feat(telemetry): stitch CLI identity from API response header

seanoliver:
## Summary

- Adds an `identityTransport` RoundTripper that reads `X-Gotrue-Id` from management API response headers
- When the header is present and the device hasn't been identified yet, calls existing `StitchLogin()` to link `device_id` → `gotrue_id` in PostHog
- One-time per device, cached in `telemetry.json` after first stitch

**Companion PR:** supabase/platform#31378 — platform-side interceptor that sets the `X-Gotrue-Id` response header. Safe to merge in either order; without the header, this change is a transparent no-op.

**Context:** [GROWTH-756](https://linear.app/supabase/issue/GROWTH-756/cli-telemetry-send-dollaridentify-with-gotrue-id-for-token-based-auth) — ~89% of CLI devices are anonymous. CLI tokens are opaque (`sbp_*`), so `gotrue_id` can't be extracted locally. This piggybacks on existing API calls to passively resolve identity, taking coverage from 11% → ~85%.

## Test plan

- [ ] `go test ./internal/utils/ -run TestIdentityTransport -v` — transport captures header when present, ignores when absent
- [ ] `go test ./internal/telemetry/ -run TestServiceNeedsIdentityStitch -v` — returns true when DistinctID empty, false after stitch
- [ ] `go test ./...` — no regressions
- [ ] Manual: run any authenticated command with `DEBUG=true` — confirms no errors (header not present yet, so no stitch attempt)

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -140,6 +140,15 @@ var (
 			} else {
 				ctx = telemetry.WithService(ctx, service)
 			}
+			if service != nil {
+				utils.OnGotrueID = func(gotrueID string) {
+					if service.NeedsIdentityStitch() {
+						if err := service.StitchLogin(gotrueID); err != nil {
+							fmt.Fprintln(utils.GetDebugLogger(), err)
+						}
+					}
+				}
+			}
 			ctx = telemetry.WithCommandContext(ctx, commandAnalyticsContext(cmd))
 			cmd.SetContext(ctx)
 			// Setup sentry last to ignore errors from parsing cli flags
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/signal"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/getsentry/sentry-go"
@@ -141,11 +142,14 @@ var (
 				ctx = telemetry.WithService(ctx, service)
 			}
 			if service != nil {
+				var stitchOnce sync.Once
 				utils.OnGotrueID = func(gotrueID string) {
 					if service.NeedsIdentityStitch() {
-						if err := service.StitchLogin(gotrueID); err != nil {
-							fmt.Fprintln(utils.GetDebugLogger(), err)
-						}
+						stitchOnce.Do(func() {
+							if err := service.StitchLogin(gotrueID); err != nil {
+								fmt.Fprintln(utils.GetDebugLogger(), err)
+							}
+						})
 					}
 				}
 			}
--- a/internal/utils/api.go
+++ b/internal/utils/api.go
@@ -21,6 +21,8 @@ const (
 	DNS_OVER_HTTPS = "https"
 )
 
+var OnGotrueID func(string)
+
 var (
 	clientOnce sync.Once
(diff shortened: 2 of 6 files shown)


internal/utils/api.go
```diff
@@ -21,6 +21,8 @@ const (
 	DNS_OVER_HTTPS = "https"
 )
 
+var OnGotrueID func(string)
+
```
avallete, line 25: **nitpick**
This works but creates an implicit, timing-dependent coupling:
The value is captured by reference at GetSupabase() init time:
```go
transport := &identityTransport{
    RoundTripper: http.DefaultTransport,
    onGotrueID:   OnGotrueID,   // captured once via clientOnce.Do
}
```

If anything calls GetSupabase() before PersistentPreRunE sets OnGotrueID, the transport permanently gets nil and stitching silently never fires. Today the ordering is safe (PersistentPreRunE → command handler → first GetSupabase() call), but it's fragile against future refactors.

Consider making the transport read through the pointer indirectly rather than capturing the value:

```go
// In identityTransport, change the field to:
onGotrueID *func(string)

// In RoundTrip:
if id := resp.Header.Get(HeaderGotrueID); id != "" && t.onGotrueID != nil && *t.onGotrueID != nil {
    (*t.onGotrueID)(id)
}

// In api.go:
transport := &identityTransport{
    RoundTripper: http.DefaultTransport,
    onGotrueID:   &OnGotrueID,
}
```
This decouples initialization order entirely.

cmd/root.go
```diff
@@ -140,6 +141,18 @@ var (
 			} else {
 				ctx = telemetry.WithService(ctx, service)
 			}
+			if service != nil {
+				var stitchOnce sync.Once
```
avallete, line 145: **nitpick**
This protects against duplicate `StitchLogin` calls from concurrent responses, but the OnGotrueID variable itself has no synchronization. It's written in PersistentPreRunE (main goroutine) and read in identityTransport.RoundTrip (potentially a different goroutine if any prefetch or background request exists). In practice the current flow is sequential, but strictly speaking this is a data race under the Go memory model.

If you take the pointer-indirection approach above, this goes away.

avallete (approved):
Minor nitpicks, nothing blocking.

internal/utils/api.go
seanoliver: Good call, took the pointer indirection approach exactly as you suggested. Transport now holds `&OnGotrueID` and dereferences at call time, so initialization order doesn't matter. Also added a double-nil guard in `RoundTrip` (`t.onGotrueID != nil && *t.onGotrueID != nil`) to cover the case where the pointer is live but the func hasn't been assigned yet. 271bcc93

cmd/root.go
seanoliver: Yep, addressed by the pointer indirection in 271bcc93 — the transport no longer reads the `OnGotrueID` variable directly, it holds `&OnGotrueID` and dereferences at call time. Kept the `sync.Once` since it still guards against duplicate `StitchLogin` calls from concurrent responses.

PR #5063, opened 2026-04-10
feat(functions) update functions template to match new 'server' style

kallebysantos:

Feature

## What is the new behavior?

Apply `withSupabase` helper from `@supabase/server`

## Additional context

A new optional flag `--auth` was introduced for the `functions new` command, chose between the auth access template:
- `none` No credentials required - Open endpoints, wrappers that handle their own auth
- `apikey` Valid publishable key or secret key - Use respectively for Client-facing, key-validated endpoints or Server-to-server, internal calls
- `user` User JWT required - Authenticated user endpoints

Current default `--auth apikey`

--
Towards FUNC-563


kallebysantos:
WIP: Do not merge until `@supabase/server` be announced

cmd/functions.go
```diff
@@ -165,13 +174,13 @@ func init() {
...
 	downloadFlags := functionsDownloadCmd.Flags()
 	downloadFlags.StringVar(&flags.ProjectRef, "project-ref", "", "Project ref of the Supabase project.")
-	markFlagTelemetrySafe(downloadFlags.Lookup("project-ref"))
 	downloadFlags.BoolVar(&useLegacyBundle, "legacy-bundle", false, "Use legacy bundling mechanism.")
 	downloadFlags.BoolVar(&useApi, "use-api", false, "Unbundle functions server-side without using Docker.")
 	downloadFlags.BoolVar(&useDocker, "use-docker", true, "Use Docker to unbundle functions client-side.")
 	functionsDownloadCmd.MarkFlagsMutuallyExclusive("use-api", "use-docker", "legacy-bundle")
 	cobra.CheckErr(downloadFlags.MarkHidden("legacy-bundle"))
 	cobra.CheckErr(downloadFlags.MarkHidden("use-docker"))
+	functionsNewCmd.Flags().Var(&authAccessMode, "auth", "use a specific auth access (default always)")
```
tomaspozo: Wondering if we should default to `user` as a more secure default... wdyt? @kallebysantos

internal/functions/new/templates/index_always_access.ts
```diff
@@ -0,0 +1,32 @@
...
+import { withSupabase } from "@supabase/server";
+
+console.log("Hello from Functions!");
+
+// This endpoint uses 'always' access, no credentials required, every request is accepted.
+// Use it for health checks, public APIs, or when you handle auth yourself inside the handler.
+export default {
+  fetch: withSupabase({ allow: "always" }, async (req, ctx) => {
+    const { name } = await req.json();
+
```
tomaspozo, line 16: Or, instead of defaulting to "user" as proposed in the other comment, maybe we can add a comment here: "implement your own authentication logic if needed"

kallebysantos: I made the description more explicit, [d6fa851](https://github.com/supabase/cli/pull/5063/commits/d6fa851d9c89f40ea1f4f1b231f4676367428d6f)

cmd/functions.go
kallebysantos: You have the final word on it 💚 

I set `always` as default in order to make it easier for try-out. So users can just create the function and start to request it while developing. 

Change it to `user` will require create a **complete auth flow** or instruct users to **mint fake user's tokens** with:
```
supabase gen bearer-jwt --role authenticated --sub <id> \
 --payload '{ ...<user fields> }'
```

Which one you think should be kept?
- **always**: Simpler to try
- **user**: Follow same pattern as docs and lib defaults

kallebysantos: Changed to `apikey` auth [6b96df4](https://github.com/supabase/cli/pull/5063/commits/6b96df41dec4f585ec3b0855ebdfc104af005e6e)

internal/functions/new/templates/index_user_access.ts
```diff
@@ -4,29 +4,27 @@
...
 
   1. Run `supabase start` (see: https://supabase.com/docs/reference/cli/supabase-start)
   2. Make an HTTP request:
 
   curl -i --location --request POST '{{ .URL }}' \
-    --header 'Authorization: Bearer {{ .Token }}' \
-    --header 'Content-Type: application/json' \
-    --data '{"name":"Functions"}'
-
+    --header 'apiKey: {{ .PublishableKey }}'
```
avallete: **issue**

This wont execute if copy/pasted because of missing backslash

```suggestion
    --header 'apiKey: {{ .PublishableKey }}' \
```

internal/functions/new/templates/index_user_access.ts
```diff
@@ -4,29 +4,27 @@
 
 // Setup type definitions for built-in Supabase Runtime APIs
 import "@supabase/functions-js/edge-runtime.d.ts"
+import { withSupabase } from '@supabase/server'
```
avallete: **nitpick**
For code style consistency with the other templates it should be:

```suggestion
import { withSupabase } from "@supabase/server";
```

internal/functions/new/templates/deno.json
```diff
@@ -1,5 +1,6 @@
 {
   "imports": {
-    "@supabase/functions-js": "jsr:@supabase/functions-js@^2"
+    "@supabase/functions-js": "jsr:@supabase/functions-js@^2",
+    "@supabase/server": "npm:@supabase/server"
```
avallete: **question**

Should this be pinned to a major version like for `functions-js` ?

cmd/functions.go
```diff
@@ -94,7 +101,12 @@ var (
 			return cmd.Root().PersistentPreRunE(cmd, args)
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return new_.Run(cmd.Context(), args[0], afero.NewOsFs())
+			authMode := new_.AuthAccessModeApiKey
+			if len(authAccessMode.Value) > 0 {
+				authMode = new_.AuthAccessMode(authAccessMode.Value)
+			}
```
avallete: **nitpick**

Instead of manually setting this default could use the `Value` param in the flag definition like we do here: https://github.com/kallebysantos/supabase-cli/blob/6b96df41dec4f585ec3b0855ebdfc104af005e6e/cmd/gen.go#L64

```
	authAccessMode = utils.EnumFlag{
		Allowed: []string{
			string(new_.AuthAccessModeAlways),
			string(new_.AuthAccessModeApiKey),
			string(new_.AuthAccessModeUser),
		},
        Value: string(new_.AuthAccessModeApiKey)
	}
```

That'll also make `--help` show the real default automatically and let us drop the `"(default apikey)"` in the flag description: https://github.com/kallebysantos/supabase-cli/blob/6b96df41dec4f585ec3b0855ebdfc104af005e6e/cmd/functions.go#L183

avallete (commented):
A few comments. Otherwise LGTM.

internal/functions/new/templates/deno.json
tomaspozo: also maybe we use always `jsr:` or `npm:` for all imports on these sample codes...

internal/functions/new/templates/index_user_access.ts
kallebysantos: Solved at ef4edbc778bb

internal/functions/new/templates/index_user_access.ts
kallebysantos: Thx, fixed at 830a4afface7

cmd/functions.go
kallebysantos: Ohh this is so useful, fixed on 1701a220ead5

internal/functions/new/templates/deno.json
kallebysantos: It would be nice, but I remember a previous issue where was decided to use the `deno.json` file to import packages.

Not sure if I should just add it as in-line import directly from `jsr:` same way I keep for dashboard templates

avallete: I don't get it 🤔 

The current `@supabase/functions-js` is in `deno.json` but still import from `jsr:` with a fixed major version. Not a deno expert though but it seems that we could do the both that way ?

Also, from our docs, the `deno.json` approach is the preferred one: https://supabase.com/docs/guides/functions/dependencies#using-denojson-recommended

Which make sense, it's much easier and reliable for us to bundle the functions locally with that approach (I think that's the reason why we made this the recommended way).

kallebysantos:
Last update: Towards FUNC-581

internal/functions/new/templates/deno.json
kallebysantos: Version pinned to v1 https://github.com/supabase/cli/pull/5063/commits/733155bade73a890fe25fd28911fa7a641c10fae

tomaspozo (approved):
Looks good! 🔥

avallete (approved):

PR #5064, opened 2026-04-10
fix: atomic parser 

### LINKED ISSUE #4746: Database migration fails on local server when table or column name contains 'atomic'
**Describe the bug**

When running database migrations locally using the Supabase CLI (`supabase db reset --local`), a migration file that contains multiple SQL statements fails to execute if any table or identifier name contains the substring 'atomic'.

This issue seems to originate in the parser (`/pkg/parser/state.go`), which incorrectly interprets the substring ATOMIC inside identifiers as the start of a BEGIN ATOMIC block. Once this state is entered, the parser stops treating semicolons as statement terminators until an END token is encountered, causing multiple statements in the migration file to be merged and executed as one.

**To Reproduce**

1. Create a new DB migration
2. Write SQL script to create a table with atomic in the name
3. Execute another query (eg. select 1)
4. Push migration to local DB or reset local database and run migrations with 

***Migration file:***

```
create table public.atomic_test (id int);
select 1;
```

**Expected behaviour**

The migration to execute successfully.

**Observed behaviour**

The migration fails to execute. When running with the —debug flag, this is the output:

```
2026/01/21 11:04:18 PG Send: {"Type":"Parse","Name":"","Query":"create table public.atomic_test (id int);\nselect 1","ParameterOIDs":null}
2026/01/21 11:04:18 PG Send: {"Type":"Bind","DestinationPortal":"","PreparedStatement":"","ParameterFormatCodes":null,"Parameters":[],"ResultFormatCodes":[]}
2026/01/21 11:04:18 PG Send: {"Type":"Describe","ObjectType":"P","Name":""}
2026/01/21 11:04:18 PG Send: {"Type":"Execute","Portal":"","MaxRows":0}
2026/01/21 11:04:18 PG Send: {"Type":"Parse","Name":"","Query":"INSERT INTO supabase_migrations.schema_migrations(version, name, statements) VALUES($1, $2, $3)","ParameterOIDs":[25,25,1009]}
2026/01/21 11:04:18 PG Send: {"Type":"Bind","DestinationPortal":"","PreparedStatement":"","ParameterFormatCodes":[0,0,1],"Parameters":[{"text":"20260120173701"},{"text":"bug_report"},{"binary":"000000010000000000000019000000010000000100000032637265617465207461626c65207075626c69632e61746f6d69635f746573742028696420696e74293b0a73656c6563742031"}],"ResultFormatCodes":[]}
2026/01/21 11:04:18 PG Send: {"Type":"Describe","ObjectType":"P","Name":""}
2026/01/21 11:04:18 PG Send: {"Type":"Execute","Portal":"","MaxRows":0}
2026/01/21 11:04:18 PG Send: {"Type":"Sync"}
2026/01/21 11:04:18 PG Recv: {"Type":"ErrorResponse","Severity":"ERROR","SeverityUnlocalized":"ERROR","Code":"42601","Message":"cannot insert multiple commands into a prepared statement","Detail":"","Hint":"","Position":0,"InternalPosition":0,"InternalQuery":"","Where":"","SchemaName":"","TableName":"","ColumnName":"","DataTypeName":"","ConstraintName":"","File":"postgres.c","Line":1478,"Routine":"exec_parse_message","UnknownFields":null}
2026/01/21 11:04:18 PG Recv: {"Type":"ReadyForQuery","TxStatus":"I"}
2026/01/21 11:04:18 PG Send: {"Type":"Terminate"}
At statement: 0                                                                  
create table public.atomic_test (id int);                                        
select 1
```

**System information**
- CLI v2.72.8

**Workaround**
Splitting each statement with the word 'atomic' into its own migration causes the migrations to succeed.

### LINKED ISSUE #5020: Statement splitter misparses `atomic` in function names as BEGIN ATOMIC keyword
## Bug

The Supabase CLI statement splitter fails to parse multi-statement migration files when a function name contains the substring `atomic` (case-insensitive). It appears to misinterpret it as the SQL-standard `BEGIN ATOMIC ... END;` syntax, breaking `$$` dollar-quoting detection.

## Minimal reproduction

This 8-line migration fails on `supabase db reset` and `supabase db push`:

```sql
CREATE OR REPLACE FUNCTION fn_atomic()
RETURNS void LANGUAGE plpgsql AS $$
BEGIN
  NULL;
END;
$$;

SELECT 1;
```

```
```

Change one character in the name and it passes:

```sql
CREATE OR REPLACE FUNCTION fn_atomix()
RETURNS void LANGUAGE plpgsql AS $$
BEGIN
  NULL;
END;
$$;

SELECT 1;
```

## Verified behavior

- `atomic` anywhere in the identifier triggers the bug: `fn_atomic`, `atomic_fn`, `my_atomic_thing`, `xatomicx`
- Case-insensitive: `ATOMIC`, `Atomic`, `atomiC` all trigger it
- Single-statement migrations (no `SELECT 1` / `REVOKE` / `GRANT` after `$$;`) work fine regardless of name
- The SQL is valid PostgreSQL — running the same file via `psql` works perfectly
- Tested on CLI **v2.75.0** and **v2.84.2** on macOS (Darwin 24.6.0) — both affected

## Expected behavior

The statement splitter should not treat `atomic` inside an identifier (e.g. `fn_atomic`) as the `BEGIN ATOMIC` keyword. `BEGIN ATOMIC` is only valid as a function body delimiter in SQL-standard syntax, not inside `$$` dollar-quoted blocks or as part of an identifier.

## Workaround

Split multi-statement migrations so each file contains a single statement, or avoid the substring `atomic` in function names.

### LINKED ISSUE #5062: SQL splitter matches 'atomic' too aggressively, triggering SQLSTATE 42601
### CLI version

`2.84.2` (also reproduced on `1.226.4`, at a different migration)

```
$ supabase --version
2.84.2
```

### Environment

- macOS 26.4 (Darwin 25.4.0)
- Apple Silicon (`arm64`)
- Docker Desktop 29.2.1
- PostgreSQL 17.6.1.063 (Supabase official image)
- Installation: `brew install supabase/tap/supabase`

### Summary

The migration runner's SQL splitter (or its interaction with pgx's extended
query protocol) raises `ERROR: cannot insert multiple commands into a prepared
statement (SQLSTATE 42601)` when a migration file contains a
`CREATE [OR REPLACE] FUNCTION` statement whose unquoted identifier contains
the substring `atomic` (case-insensitive) **and** any further SQL statement
after that function body's closing `$$;`.

My working hypothesis is that the CLI's splitter is performing a naive text
search for the token `ATOMIC` (as in PostgreSQL 14's
`CREATE FUNCTION ... BEGIN ATOMIC ... END` alternative-body syntax) and
treating everything up to the end of the file — or the next top-level
`END` — as the function body. The subsequent SQL statements are then
concatenated into a single `Parse` message, which Postgres rejects because
the extended query protocol permits only one command per Parse.

The splitter does **not** tokenize SQL properly — comments, string literals,
and quoted/unquoted identifiers are all treated as raw text.

### Minimal reproducer

Create a fresh Supabase project and put this one file in
`supabase/migrations/`:

```sql
-- 001_repro.sql — triggers SQLSTATE 42601 on `supabase db reset`

CREATE OR REPLACE FUNCTION public.atomic_example()
RETURNS INTEGER
LANGUAGE plpgsql
AS $$
BEGIN
  RETURN 1;
END;
$$;

-- ^^ any statement below this line triggers the bug
GRANT EXECUTE ON FUNCTION public.atomic_example() TO authenticated;
```

Then:

```bash
supabase start
supabase db reset
```

**Expected:** migration applies cleanly. Postgres parses
`CREATE FUNCTION ... atomic_example ... $$;` as one statement and
`GRANT EXECUTE ...` as a separate statement, since `atomic_example` is an
identifier, not the `BEGIN ATOMIC` keyword sequence.

**Actual:**

```
Applying migration 001_repro.sql...
GRANT EXECUTE ON FUNCTION public.atomic_example() TO authenticated;
```

With `--debug`:

```
PG Recv: {"Type":"ErrorResponse","Severity":"ERROR","Code":"42601",
         "Message":"cannot insert multiple commands into a prepared statement",
         "File":"postgres.c","Line":1478,"Routine":"exec_parse_message"}
```

### Variations confirmed with the same reproducer project

| Variation | Result |
|---|---|
| Function body contains the word `atomic` in a comment, with no trailing statement | ✅ passes |
| Function body contains `atomic` in a comment, with trailing `GRANT` | ❌ fails |
| Function name contains `atomic` (`atomic_foo`, `foo_atomic`, `abc_atomic_bar`) + trailing statement | ❌ fails |
| Function name `atomic_foo` with no trailing statement (nothing after `$$;`) | ✅ passes |
| Whole file wrapped in `BEGIN; ... COMMIT;` | ❌ still fails |
| Top-level `-- atomic` comment (no function with `atomic` in its name) + CREATE FUNCTION + GRANT | ✅ passes |

**Summary**: the presence of `atomic` *in a CREATE FUNCTION identifier*,
combined with a following statement, is the trigger. `atomic` in comments
alone is fine.

### Workaround

Rename the function to avoid the substring `atomic`. A `_tx` suffix works
well semantically for transactional RPCs:

```sql
CREATE OR REPLACE FUNCTION public.example_tx()  -- was: atomic_example
...
```

For codebases with many such functions, a simple lint rule on
`CREATE [OR REPLACE] FUNCTION` identifiers catches the pattern at commit
time. Example (bash):

```bash
grep -rEn 'CREATE[[:space:]]+(OR[[:space:]]+REPLACE[[:space:]]+)?FUNCTION[[:space:]]+([a-zA-Z_][a-zA-Z_0-9]*\.)?[a-zA-Z_0-9]*[aA][tT][oO][mM][iI][cC][a-zA-Z_0-9]*[[:space:]]*\(' supabase/migrations/
```

### Suggested fix

The splitter should tokenize SQL with awareness of:

1. **Dollar-quoted strings** (`$$...$$`, `$tag$...$tag$`) as opaque bodies
2. **Line comments** (`--`) and **block comments** (`/* */`)
3. **Quoted identifiers** (`"ATOMIC"` is an identifier, not a keyword)
4. **Context for `ATOMIC`**: only match it as a keyword when it appears
   **immediately after a closing `)` of a function signature and before an
   opening `BEGIN`** (the actual `BEGIN ATOMIC` syntax), not inside
   identifiers

A straightforward fix is to replace any regex-based statement splitting
with a real SQL tokenizer (e.g., go-pg's or pgx's built-in parser)
that respects PostgreSQL's lexer rules.

### Impact

This is a silent-until-triggered bug. A codebase with 178+ migrations built
over several years (we've been maintaining a production Supabase project since 2023) can contain perfectly valid migrations for a
long time before someone adds a trailing `GRANT` that suddenly breaks
`supabase db reset`. The resulting error message points nowhere near the
root cause (the function definition is many hundreds of lines away) and
the word `atomic` is not obviously a reserved or dangerous token to the
developer.

Teams often hit this only when:

- Upgrading the CLI (when the splitter was rewritten), or
- Running `db reset` for the first time in a while, or
- Bringing a new contributor onto the project who tries to set up local dev

### Additional context

- `supabase migration repair --status reverted` is needed between failed
  attempts because the CLI records failed migrations as "applied" in
  `schema_migrations` before the error bubbles up.
- Pinning older CLI versions (`1.226.4`) doesn't help for all projects —
  the same bug triggers on a different SQL pattern in that line.
- The case sensitivity of the match is inconsistent: `ATOMIC` in a comment
  passes, `atomic` in a function name fails, `ATOMIC` in a function name
  also fails. A single case-insensitive raw text search for `atomic` is
  the likely mechanism.

### Happy to help

I have a minimal reproducer project that isolates the bug in 15 lines of
SQL. I can share it privately if useful. I also have a bisection log
showing exactly which SQL patterns do and don't trigger the error.

7ttp:
## TL;DR

fixes SQL parser incorrectly treating function names containing "atomic" as BEGIN ATOMIC syntax

## problem

saw multiple reports that: 

The SQL parser treats any occurrence of "atomic" as BEGIN ATOMIC syntax, 
even when it's part of a func somewhat like `atomic_example()`
which causes migrations to fail when a func with "atomic" in its name is followed by another statement

```sql
CREATE FUNCTION atomic_example() RETURNS INTEGER AS $$ ... $$;
GRANT EXECUTE ON FUNCTION atomic_example() TO authenticated;
-- ERROR: cannot insert multiple commands into a prepared statement
```

## sol

check that "atomic" is actually part of `BEGIN ATOMIC` syntax, 
not just a substring in an identifier
parser now verifies that both BEGIN and ATOMIC are standalone keywords, not part of function names or other identifiers.

## ref:

- closes https://github.com/supabase/cli/issues/5062
- closes https://github.com/supabase/cli/issues/4746
- closes https://github.com/supabase/cli/issues/5020

### DIFF
--- a/pkg/parser/state.go
+++ b/pkg/parser/state.go
@@ -46,12 +46,38 @@ func (s *ReadyState) Next(r rune, data []byte) State {
 	case 'c':
 		fallthrough
 	case 'C':
-		offset := len(data) - len(BEGIN_ATOMIC)
-		if offset >= 0 && strings.EqualFold(string(data[offset:]), BEGIN_ATOMIC) {
+		if isBeginAtomic(data) {
 			return &AtomicState{prev: s, delimiter: []byte(END_ATOMIC)}
 		}
 	}
 	return s
+}
+
+func isBeginAtomic(data []byte) bool {
+	offset := len(data) - len(BEGIN_ATOMIC)
+	if offset < 0 || !strings.EqualFold(string(data[offset:]), BEGIN_ATOMIC) {
+		return false
+	}
+	if offset > 0 {
+		r, _ := utf8.DecodeLastRune(data[:offset])
+		if isIdentifierRune(r) {
+			return false
+		}
+	}
+	prefix := bytes.TrimRightFunc(data[:offset], unicode.IsSpace)
+	offset = len(prefix) - len("BEGIN")
+	if offset < 0 || !strings.EqualFold(string(prefix[offset:]), "BEGIN") {
+		return false
+	}
+	if offset == 0 {
+		return true
+	}
+	r, _ := utf8.DecodeLastRune(prefix[:offset])
+	return !isIdentifierRune(r)
+}
+
+func isIdentifierRune(r rune) bool {
+	return unicode.IsLetter(r) || unicode.IsDigit(r) || r == '_' || r == '$'
 }
 
 // Opened a line comment
--- a/pkg/parser/state_test.go
+++ b/pkg/parser/state_test.go
@@ -167,4 +167,44 @@ END
 ;`}
 		checkSplit(t, sql)
 	})
+
+	t.Run("ignores atomic in identifiers", func(t *testing.T) {
+		names := []string{
+			"fn_atomic",
+			"atomic_fn",
+			"my_atomic_thing",
+			"xatomicx",
+			"fn_ATomiC",
+		}
+		for _, name := range names {
+			t.Run(name, func(t *testing.T) {
+				sql := []string{
+					`CREATE OR REPLACE FUNCTION ` + name + `()
+RETURNS void LANGUAGE plpgsql AS $$
+BEGIN
+  NULL;
+END;
+$$;`,
+					`
+SELECT 1;`,
+				}
+				checkSplit(t, sql)
+			})
+		}
+	})
+
+	t.Run("does not treat schema-qualified atomic function names as begin atomic", func(t *testing.T) {
+		sql := []string{`CREATE OR REPLACE FUNCTION public.atomic_example()
+RETURNS INTEGER
+LANGUAGE plpgsql
+AS $$
+BEGIN
+  RETURN 1;
+END;
+$$;`,
+			`
+GRANT EXECUTE ON FUNCTION public.atomic_example() TO authenticated;`,
+		}
+		checkSplit(t, sql)
+	})
 }


avallete (approved):

avallete:
Thank's for the fix !

PR #5066, opened 2026-04-11
feat(plan-gate): wire billing links into remaining plan-gated commands

pamelachia:
## Summary

Extends `SuggestUpgradeOnError` to check org entitlements on any non-2xx API response (not just 402), then shows a billing upgrade link only when the entitlements API confirms `hasAccess == false`. Wires this into SSO and vanity-subdomain commands.

Previously, only `branches create` and `branches update` had billing links. Many plan-gated endpoints return 400 or 404 instead of 402, so the old 402-only check missed them. The new approach catches all error codes while preventing false positives via entitlements confirmation.

## Changes

- Refactor `SuggestUpgradeOnError` to trigger on any 4xx, confirm gate via entitlements API before showing billing link
- Remove generic fallback messages ("may require a plan upgrade") that caused false positives
- Wire `sso create/update/list/remove` with `auth.saml_2` entitlement check
- Wire `vanity-subdomains activate/check` with `vanity_subdomain` entitlement check
- Extract shared `telemetry.TrackUpgradeSuggested` helper, replacing 8 identical per-package copies
- Update existing tests to handle the new entitlements probe on non-2xx responses

## Testing

Tested locally:
- [x] `go build ./...` passes
- [x] All affected package tests pass (`branches`, `sso`, `vanity_subdomains`, `utils`)
- [x] Existing 402 behavior preserved for `branches create/update`
- [x] SSO commands: 404 from GoTrue + gated entitlement shows billing link
- [x] SSO commands: 404 from GoTrue + ungated entitlement shows existing "enable through dashboard" message
- [x] Vanity subdomain: 400 + gated entitlement shows billing link
- [x] Non-2xx errors with failed entitlements lookup show no suggestion (no false positives)
- [x] 2xx responses skip the check entirely

### Descoped (intentionally)
- `custom_domain`, `pitr.available_variants`: addon-gated, not plan-gated (GROWTH-741)
- `function.max_count`: `hasAccess: true` for all plans (numeric limit only)
- `network-restrictions`: not plan-gated (infra flag)
- `branches delete`: intentionally not gated

## Linear
- fixes GROWTH-740

### DIFF
--- a/internal/vanity_subdomains/check/check.go
+++ b/internal/vanity_subdomains/check/check.go
@@ -20,7 +20,7 @@ func Run(ctx context.Context, projectRef string, desiredSubdomain string, fsys a
 		return errors.Errorf("failed to check vanity subdomain: %w", err)
 	} else if resp.JSON201 == nil {
 		if orgSlug, isGated := utils.SuggestUpgradeOnError(ctx, projectRef, "vanity_subdomain", resp.StatusCode()); isGated {
-			trackUpgradeSuggested(ctx, "vanity_subdomain", orgSlug)
+			telemetry.TrackUpgradeSuggested(ctx, "vanity_subdomain", orgSlug)
 		}
 		return errors.Errorf("unexpected check vanity subdomain status %d: %s", resp.StatusCode(), string(resp.Body))
 	}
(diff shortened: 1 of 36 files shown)


internal/vanity_subdomains/check/check.go
```diff
@@ -26,3 +30,12 @@ func Run(ctx context.Context, projectRef string, desiredSubdomain string, fsys a
...
 }
+
+func trackUpgradeSuggested(ctx context.Context, featureKey, orgSlug string) {
+	if svc := telemetry.FromContext(ctx); svc != nil {
+		_ = svc.Capture(ctx, telemetry.EventUpgradeSuggested, map[string]any{
+			telemetry.PropFeatureKey: featureKey,
+			telemetry.PropOrgSlug:    orgSlug,
+		}, nil)
+	}
+}
```
avallete: **remark**

This function is repeated a lot, we should extract it into a shared location, maybe within `internal/utils/plan_gate.go` ? Or even better have the `SuggestUpgradeOnError` handle the telemetry. It already receive the `ctx` + status code.

We already know if "is gated" in there, that seems like the right place to put telemetry on.

pamelachia: Fixed. Extracted a shared `telemetry.TrackUpgradeSuggested` helper and replaced all 8 per-package copies. Couldn't move it into `SuggestUpgradeOnError` directly due to a utils->telemetry import cycle, but the telemetry package is the natural home for it.

avallete (approved):

PR #5083, opened 2026-04-14
fix(config): add PG17 image mapping for db dump

### LINKED ISSUE #5007: `db dump` fails against remote PostgreSQL 17: config.go missing `case 17` image mapping
**Describe the bug**
`supabase db dump` fails with "server version mismatch" when the remote database is PostgreSQL 17. The CLI's Docker container uses pg_dump 15.6, which refuses to connect to a PG 17.6 server.

Root cause: `internal/utils/config.go` `Load()` switch only handles `case 13`, `case 14`, `case 15` — there is no `case 17`. The `Image` field is tagged `toml:"-"` so users cannot override it via `config.toml`. The `Validate()` function accepts `case 15, 17:` as valid, so `major_version = 17` is accepted in config — it just never assigns the correct Docker image.

The PG 17 Docker image (`supabase/postgres:17.6.1.095`) already exists on ECR and is referenced in the `main` branch Dockerfile, but it's not wired into `config.go`.

**To Reproduce**
1. Have a Supabase project running PostgreSQL 17
2. Set `major_version = 17` in `supabase/config.toml`
3. Run `supabase db dump`
4. Error: `pg_dump: error: aborting because of server version mismatch`
   - `server version: 17.6; pg_dump version: 15.6`

**Expected behavior**
`db dump` should use a Docker image with pg_dump 17.x when `major_version = 17` is configured, matching the remote server version.

**Screenshots**
N/A

**System information**
 - Ticket ID: f08c1ad7d1624aeb8aa21b9cfe1debfd
 - Version of OS: macOS 26.4
 - Version of CLI: v2.84.5
 - Version of Docker: v29.0.1
 - Versions of services: N/A (remote-only, not using local dev)

**Additional context**
- `config.go` source references:
  - `Load()` image switch: https://github.com/supabase/cli/blob/main/internal/utils/config.go (search for `case 15:` in the `DbImage` / major version switch)
  - `Validate()` accepts 17: https://github.com/supabase/cli/blob/main/internal/utils/config.go (search `case 15, 17:`)
  - `Image` field: tagged `toml:"-"` — not user-configurable
- The fix appears to be ~2 lines: add a `Pg17Image` constant to `constants.go` and a `case 17:` branch in `Load()`
- Possibly related to #3748, though that issue is about the inverse direction (`db diff` with local PG 17 vs remote PG 15)

nanookclaw:
## Summary

Add `case 17` to the `MajorVersion` switch in `config.Load()` so that `major_version = 17` resolves to the correct Docker image (`supabase/postgres:17.6.1.106`).

## Problem

`supabase db dump` fails with "server version mismatch" when the remote database is PostgreSQL 17. The CLI's Docker container uses pg_dump 15.6, which refuses to connect to a PG 17.6 server.

Root cause: `Load()` only handles `case 13`, `case 14`, `case 15` — there is no `case 17`. The `Image` field is tagged `toml:"-"` so users cannot override it via `config.toml`. The `Validate()` function already accepts `case 15, 17:` as valid, so `major_version = 17` is accepted — it just never assigns the correct Docker image.

## Changes

- **`pkg/config/constants.go`**: Add `pg17 = "supabase/postgres:17.6.1.106"` constant (matches the version in the embedded Dockerfile template).
- **`pkg/config/config.go`**: Add `case 17: c.Db.Image = pg17` to the `MajorVersion` switch in `Load()`.

## Testing

- `go build ./...` passes
- `go test -short ./...` passes (only pre-existing keyring test failure unrelated to this change)

Fixes #5007


jgoux:
Hello, thanks for your contributions.
Since this PR was opened we ported the command from Go to TypeScript and the bug should be fixed now. 👍

Issue #5080, opened 2026-04-14
declarative: When pgmq is enabled, the declarative schema `generate` and `sync` commands fail.


tolgap:
**Describe the bug**
There seems to be an issue with the experimental `declarative` command 

**To Reproduce**
1. Ensure pgmq extension is enabled in your existing DB
2. Run the declarative command
3. Observe error about pgmq
4. The `generate` command still succeeds in generating the files though
5. Run `sync` command, and observe error about any function that refers to `pgmq` schema not existing

Commands:

```console
supabase db schema declarative generate --local --experimental --schema pgmq,public,<all my schemas>
supabase db schema declarative sync --experimental --schema pgmq,public,<all my schemas>
```

**Expected behavior**
I would expect the `pgmq` schema to be created in the shadow DB, because the `cluster/extensions/pgmq.sql` does contain pgmq enabling.

**Screenshots**

<img width="794" height="426" alt="Image" src="https://github.com/user-attachments/assets/f536c030-e853-4722-bb92-dd9e8c703962" />

Here's the output of `errors.txt`:

```
[...]

PR #5085, opened 2026-04-15
fix(pgdelta): unmarshal stuckStatements as objects, not strings

SAY-5:
Per supabase/cli#5080, `ApplyResult.StuckStatements` was typed as `[]string` but `pgdelta_declarative_apply.ts` emits an array of objects with shape `{ statement, code, message, isDependencyError }`. The mismatch caused `json.Unmarshal` to fail with:

```
failed to parse pg-delta apply output: json: cannot unmarshal object
into Go struct field ApplyResult.stuckStatements of type string
```

This obscured the real underlying error (the `pgmq` schema not being present in the shadow DB) behind an unrelated parsing failure, and left users with no way to inspect *which* statements got stuck.

This PR introduces a `StuckStatement` struct that captures the documented fields, retains the inner statement payload as `json.RawMessage` so future schema changes don't break parsing, and updates the stderr output to render `code + message` for each stuck statement instead of dumping a Go-struct repr.

Closes supabase/cli#5080

### DIFF
--- a/internal/db/declarative/declarative.go
+++ b/internal/db/declarative/declarative.go
@@ -235,10 +235,9 @@ func WriteDeclarativeSchemas(output diff.DeclarativeOutput, fsys afero.Fs) error
 			return err
 		}
 	}
-	// When pg-delta has its own config section, the declarative path is the single
-	// source of truth there; do not overwrite [db.migrations] schema_paths.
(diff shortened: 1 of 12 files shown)


avallete:
Thank's for the contribution, but that should already be taken care of in this PR: https://github.com/supabase/cli/pull/5082

PR #5087, opened 2026-04-15
feat(telemetry): attach org/project groups to all CLI events

seanoliver:
## Problem

Only ~19% of CLI events have PostHog group properties set (`$group_0` for org, `$group_1` for project). Groups are only written during `supabase link`, so any command using `--project-ref` without linking first sends events that are invisible to PostHog group analytics. This means we can't filter or aggregate ~80% of CLI activity by org or project.

Back in early April when CLI telemetry shipped, the plumbing for groups was built out (the `CommandContext.Groups` field, `mergeGroups()` layering, `linkedProjectGroups()` reader) but only the link flow actually populates the cache.

## Changes

- Add `EnsureProjectGroupsCached()` in `internal/telemetry/project.go` — when a project ref is available but no matching `linked-project.json` exists, fetches the project from the management API and caches it. Subsequent commands with the same ref skip the API call entirely.
- Call it from `Execute()` in `cmd/root.go` right before firing `cli_command_executed`, so the existing `linkedProjectGroups()` reader picks up the cached data.
- Best-effort throughout — errors log to debug and never break commands.
- New test file with 8 cases covering cache hit, cache miss, ref change, API errors, and 404s.

## Testing

- All 25 telemetry tests pass (`go test ./internal/telemetry/... -v`)
- Full build passes (`go build ./...`)
- Verified the API response type (`V1ProjectWithDatabaseResponse`) matches what `SaveLinkedProject` expects

Closes GROWTH-761

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -173,6 +173,7 @@ func Execute() {
 	executedCmd, err := rootCmd.ExecuteC()
 	if executedCmd != nil {
 		if service := telemetry.FromContext(executedCmd.Context()); service != nil {
+			telemetry.EnsureProjectGroupsCached(executedCmd.Context(), flags.ProjectRef, afero.NewOsFs())
 			_ = service.Capture(executedCmd.Context(), telemetry.EventCommandExecuted, map[string]any{
 				telemetry.PropExitCode:   exitCode(err),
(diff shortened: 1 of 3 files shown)


jgoux (approved):
LGTM

PR #5126, opened 2026-04-25
fix(start): guard db_logs vector transform against null regex capture

rebasecase:
## What

In the `db_logs` transform of the Vector config, `upcase!()` aborts
with `expected string, got null` when `parse_regex` matches an event
message but the `level` named group resolves to null. The existing
fallback branch only covers regex failure
(`err != null || parsed == null`), so a third path (match succeeded,
capture null) overwrites the `"info"` fallback with null and crashes
on upcase.

## Why

Running a Next.js dev server against the local Supabase stack with
routine service-role Postgres queries produces 3,000+ aborted
`db_logs` transforms in ~2 minutes. Vector retries the failures,
Logflare's ingest queue backs up (`ErlSysMon` warnings for
`:long_message_queue` and `:long_schedule` at 746ms), and the
analytics container sustains 100% CPU. On a resource-constrained
host (laptop, Docker at 8 GB) this cascades into swap thrashing and
host unresponsiveness.

## How

Extend the fallback condition to also fire when `parsed.level` is
null, and guard the match-path assignment symmetrically, so
`error_severity` always holds a non-null string before `upcase!()`.

## Verification

Before: thousands of `function call error for "upcase" at (476:516):
expected string, got null` errors from `db_logs` in the Vector
container logs during dev activity.

After: zero transform failures under the same workload.

### DIFF
--- a/internal/start/templates/vector.yaml
+++ b/internal/start/templates/vector.yaml
@@ -154,11 +154,10 @@ transforms:
 
       parsed, err = parse_regex(.event_message, r'.*(?P<level>INFO|NOTICE|WARNING|ERROR|LOG|FATAL|PANIC?):.*', numeric_groups: true)
 
-      if err != null || parsed == null {
+      if err != null || parsed == null || parsed.level == null {
         .metadata.parsed.error_severity = "info"
-      }
-      if parsed != null {
-       .metadata.parsed.error_severity = parsed.level
+      } else {
+        .metadata.parsed.error_severity = parsed.level
       }
       if .metadata.parsed.error_severity == "info" {
           .metadata.parsed.error_severity = "log"


rebasecase:
Verified the fix locally against the same vector image (`public.ecr.aws/supabase/vector:0.53.0-alpine`), running the original and patched VRL side by side against a synthetic event where `parsed.level` is null.

<details>
<summary>Test config (<code>vector-test.yaml</code>)</summary>

```yaml
sources:
  stdin:
    type: stdin
    decoding:
      codec: json

transforms:
  broken:
    type: remap
    inputs: [stdin]
    source: |-
      parsed = {"some_other_key": "value"}
      err = null
      if err != null || parsed == null {
        .metadata.parsed.error_severity = "info"
      }
      if parsed != null {
        .metadata.parsed.error_severity = parsed.level
      }
      if .metadata.parsed.error_severity == "info" {
        .metadata.parsed.error_severity = "log"
      }
      .metadata.parsed.error_severity = upcase!(.metadata.parsed.error_severity)
      .label = "broken"

  fixed:
    type: remap
    inputs: [stdin]
    source: |-
      parsed = {"some_other_key": "value"}
      err = null
      if err != null || parsed == null || parsed.level == null {
        .metadata.parsed.error_severity = "info"
      } else {
        .metadata.parsed.error_severity = parsed.level
      }
      if .metadata.parsed.error_severity == "info" {
        .metadata.parsed.error_severity = "log"
      }
      .metadata.parsed.error_severity = upcase!(.metadata.parsed.error_severity)
      .label = "fixed"

sinks:
  out:
    type: console
    inputs: [broken, fixed]
    encoding:
      codec: json
```

</details>

<details>
<summary>Run</summary>

```bash
echo '{"event_message":"test"}' | docker run --rm -i \
  -v /tmp/vector-test.yaml:/etc/vector/vector.yaml:ro \
  public.ecr.aws/supabase/vector:0.53.0-alpine \
  --config /etc/vector/vector.yaml
```

`parsed` is hardcoded to a non-null object with no `level` key, which is the state empirically produced by the real regex under live log input.

</details>

**Original code (`broken` transform):**

```
ERROR vector::internal_events::remap: Mapping failed with event.
  error="function call error for \"upcase\" at (324:364):
         expected string, got null"
  error_type="conversion_failed"
  stage="processing"
```

Event is dropped, VRL aborts on `upcase!(null)`. Same signature as the `function call error for "upcase" ... expected string, got null` entries that fire thousands of times per minute under dev load in the real `supabase_vector` container.

**Patched code (`fixed` transform):**

```json
{
  "event_message": "test",
  "label": "fixed",
  "metadata": {"parsed": {"error_severity": "LOG"}}
}
```

Event passes cleanly, `error_severity` resolves to `"LOG"` via the `info` fallback then the `log` to `LOG` uppercase step, no VRL abort.

avallete (approved):
Thanks for your contribution !

avallete (approved):

PR #5145, opened 2026-04-28
Consistently use local database as the default

icopp:

The codebase currently has a truly nightmare state of commands randomly defaulting to the linked project or the local database depending on specific command, including some commands that mutate the database like `migration repair`. This change makes them all consistently default to the local database only, explicitly requiring `--linked` or `--db-url` to do anything destructive to another database. The change in default behavior will likely take a major version bump, but should save a lot of people from pain in the future.

The only way to know which commands default to local behavior and which default to remote behavior is to individually look at the docs of every command.

See for example, https://github.com/supabase/cli/issues/3493

## What is the new behavior?

Every command defaults to local. A user must explicitly use `--linked` or `--db-url` to do anything that reads or affects anything but the local dev database.


jgoux:
Hello, thanks for your contribution.

We're currently porting all the commands to TypeScript so I'll close this PR but be sure that we heard the feedback and we agree with your proposal.

We're working on simplifying the API surface of the CLI and we will have a way to chose between local-first vs remote-first globally when running commands.

PR #5153, opened 2026-04-30
feat: emit Claude Code plugin hint from CLI

Rodriguespn:
## Summary

- Adds `SuggestClaudePlugin()` to `internal/utils/misc.go` that returns the `<claude-code-hint />` tag when the `CLAUDECODE` env var is set (no-op otherwise)
- Calls it in `Execute()` (`cmd/root.go`) so every `supabase` invocation can trigger the hint
- Calls it after login success in `internal/login/login.go` (both token and browser paths) to catch users in a setup flow

Claude Code deduplicates by plugin per session, so emitting broadly is safe. The tag is stripped from output before it reaches the model and never counts toward token usage.

Closes [AI-689](https://linear.app/supabase/issue/AI-689/investigate-claude-code-plugin-hints)

image

## References

- Protocol docs: https://code.claude.com/docs/en/plugin-hints
- Plugin source: https://github.com/Rodriguespn/claude-plugins-official

### DIFF
--- a/internal/utils/misc.go
+++ b/internal/utils/misc.go
@@ -38,6 +38,15 @@ func ShortContainerImageName(imageName string) string {
 
 const SuggestDebugFlag = "Try rerunning the command with --debug to troubleshoot the error."
 
+const claudeCodeHint = `<claude-code-hint v="1" type="plugin" value="supabase@claude-plugins-official" />`
+
+func SuggestClaudePlugin() string {
+	if os.Getenv("CLAUDECODE") != "" {
+		return claudeCodeHint
+	}
+	return ""
+}
+
 var (
 	CmdSuggestion string
 	CurrentDirAbs string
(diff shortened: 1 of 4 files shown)


internal/utils/misc.go
```diff
@@ -38,6 +38,15 @@ func ShortContainerImageName(imageName string) string {
 
 const SuggestDebugFlag = "Try rerunning the command with --debug to troubleshoot the error."
 
+const claudeCodeHint = `<claude-code-hint v="1" type="plugin" value="supabase@claude-plugins-official" />`
+
+func SuggestClaudePlugin() string {
+	if os.Getenv("CLAUDECODE") != "" {
+		return claudeCodeHint
+	}
```
avallete: **nitpick**

This is different from the check we do here:

https://github.com/supabase/cli/blob/2a4b558fa6461822e9efdf00bc8b2b1f73700a5c/internal/utils/agent/agent.go#L39-L42

Maybe we could centralize both checks within a function like `IsClaudeCode()` that we can re-use in both places ?

Rodriguespn: Great idea, thanks for the suggestion. Created the `IsClaudeCode()` to centralize this part.

avallete (approved):

PR #5156, opened 2026-05-01
fix(migration): handle pipeline-incompatible statements in ExecBatch

### LINKED ISSUE #5139: supabase db reset fails on multi-statement migrations (42601) and CONCURRENTLY in pipeline (25001)
﻿## Environment` down goes in the body. Nothing in this document needs editing before filing.

---

## Title

`supabase db reset` fails on multi-statement migrations: SQLSTATE 42601 "cannot insert multiple commands into a prepared statement"

---

## Environment

- **CLI versions tested (all reproduce the issue, see version matrix below):** v2.89.0, v2.90.0, v2.91.0, v2.92.1, v2.95.3
- **OS:** Windows 11 Home 10.0.26200
- **Docker:** Docker version 29.2.1, build a5c7197
- **Postgres (in the Supabase-managed `supabase_db_*` container):** PostgreSQL 17.6
- **Shell:** Git Bash on Windows

---

## Summary

`supabase db reset` (and by extension `supabase migration up`) cannot replay a real-world Supabase migrations directory end-to-end. Two distinct CLI-side failures are involved:

1. **v2.89.0â€“v2.91.0:** Any migration file containing more than one SQL statement is rejected by Postgres with `SQLSTATE 42601` ("cannot insert multiple commands into a prepared statement"). The CLI is sending the entire file as a single Parse message in the extended query protocol, which Postgres does not allow.
2. **v2.92.1, v2.95.3:** The multi-statement issue above is fixed, but the CLI now wraps each migration in a pipeline. Any migration containing `CREATE INDEX CONCURRENTLY` (or any other statement that cannot run inside a pipeline / transaction) is rejected with `SQLSTATE 25001` ("CREATE INDEX CONCURRENTLY cannot be executed within a pipeline").

Either bug alone is sufficient to break local replay; together they mean **no released CLI version can replay this codebase**. The same migration files apply correctly via `psql` and via the Supabase platform / dashboard / production CI path.

---

### Bug 1 â€” multi-statement file rejected (v2.89.0 â€“ v2.91.0)

Create `supabase/migrations/20260101000000_multi_statement_repro.sql` with the following 14 lines:

```sql
-- Reproduces SQLSTATE 42601 on `supabase db reset` (CLI v2.89-v2.91).
-- This file is valid SQL and applies cleanly via `psql`.

CREATE TABLE IF NOT EXISTS public.cli_repro_a (id INT);
CREATE TABLE IF NOT EXISTS public.cli_repro_b (id INT);

DO $$
BEGIN
  INSERT INTO public.cli_repro_a VALUES (1);
  INSERT INTO public.cli_repro_b VALUES (2);
END
$$;

-- end of file
```

Then run:

```bash
supabase db reset
```

### Bug 2 â€” pipeline wrapper breaks CONCURRENTLY (v2.92.1, v2.95.3)

Create `supabase/migrations/20260101000001_concurrently_repro.sql`:

```sql
-- Reproduces SQLSTATE 25001 on `supabase db reset` (CLI v2.92+).
CREATE TABLE IF NOT EXISTS public.cli_repro_c (id INT);
CREATE INDEX CONCURRENTLY IF NOT EXISTS cli_repro_c_idx ON public.cli_repro_c (id);
```

Then run:

```bash
supabase db reset
```

---

## Expected behaviour

`supabase db reset` should apply both repro files successfully, since they:

- are valid SQL,
- apply cleanly via `psql` (single-statement protocol path),
- apply cleanly when sent to the Supabase platform's migrations endpoint,
- match patterns found in many real production migrations (multi-statement DDL, idempotent index creation, `DO` blocks).

---

### Bug 1

```
Applying migration 20260101000000_multi_statement_repro.sql...
```

`supabase db reset --debug` shows the protocol-level cause:

```
PG Recv: {"Type":"ErrorResponse","Severity":"ERROR","Code":"42601",
         "Message":"cannot insert multiple commands into a prepared statement",
         "File":"postgres.c","Line":1478,"Routine":"exec_parse_message"}
```

`exec_parse_message` is the Postgres backend handler for the **Parse** message in the extended query protocol. The CLI is sending the entire multi-statement file as a single Parse message, which Postgres rejects by design â€” Parse only accepts a single command. The simple-query protocol (used by `psql` by default) is what supports multiple commands per message.

### Bug 2

```
Applying migration 20260101000001_concurrently_repro.sql...
ERROR: CREATE INDEX CONCURRENTLY cannot be executed within a pipeline (SQLSTATE 25001)
```

This is a regression from the v2.91â†’v2.92 change that fixed Bug 1. Earlier CLI versions ran each file in a per-file session without an implicit pipeline/transaction wrapper, which is what `CREATE INDEX CONCURRENTLY` requires.

---

## Workaround currently in use

We bypass the CLI's migration runner entirely with a small Node script that pipes each migration through `docker exec ... psql`:

- File: `scripts/load/apply-migrations-via-psql.mjs`
- Behaviour: drops + recreates the `public` schema (preserving the supabase-managed `auth`, `storage`, `realtime`, `extensions`, and `supabase_migrations` schemas), then iterates `supabase/migrations/*.sql` in chronological order and applies each one via `docker exec -i <supabase_db_container> psql -v ON_ERROR_STOP=1`. Each applied file is recorded in `supabase_migrations.schema_migrations` so `supabase migration list` reflects reality afterward.
- Result: applies all 137 migrations in ~30 seconds, no errors.

The relevant excerpt:

```js
// scripts/load/apply-migrations-via-psql.mjs (excerpt)
for (const file of migrationFiles) {
  const sql = readFileSync(file, "utf8");
  execSync(
    `docker exec -i -e PGPASSWORD=postgres ${dbContainer} psql -U postgres -d postgres -v ON_ERROR_STOP=1`,
    { input: sql, stdio: ["pipe", "inherit", "inherit"] },
  );
  // ...record in supabase_migrations.schema_migrations
}
```

This works because `psql` reads from stdin using the simple-query protocol, which accepts multiple commands in a single message and does not implicitly wrap statements in a pipeline.

---

## Scope of impact in our codebase

A static-analysis sweep of our `supabase/migrations/` directory (137 files total, generated by `scripts/load/enumerate-multistmt-migrations.mjs`) shows:

- **Bug 1 (multi-statement) affects 97 of 137 migration files.** Examples include multi-statement DDL, accompanying RLS policy creation, idempotent grants, and `DO $$ ... $$` bodies that share a file with other statements. None of these files are pathological â€” they're idiomatic migration files.
- **Bug 2 (CONCURRENTLY in a pipeline) affects 1 of 137 migration files.** That file's own header comment even anticipates the constraint:
  > "CONCURRENTLY cannot run inside a transaction; this migration relies on Supabase CLI's per-file session (which does not auto-wrap in BEGIN/COMMIT)."
  That assumption was correct on v2.91 and earlier. v2.92+ broke it.

The largest multi-statement files (purely as a sense-check of "this is normal real-world migration content"):

| Statements | File |
|------------:|------|
| 43 | `20260212100000_user_taxonomy.sql` |
| 42 | `20260203000000_initial_schema.sql` |
| 39 | `20260316200001_harden_security_definer_privileges_part2.sql` |
| 33 | `20260305200000_v1_content_model.sql` |
| 30 | `20260221000000_import_v2_staging.sql` |

---

## Production migration application is unaffected

Migrations applied through the Supabase platform / Dashboard / production CI path apply cleanly â€” that path uses the simple-query protocol and does not wrap statements in a pipeline. This issue is **local-CLI-only**, but it means every developer running `supabase db reset` against a real Supabase project that uses multi-statement migrations or CONCURRENTLY indexes will hit it.

---

## Version matrix

| CLI version | Bug 1 (multi-statement) | Bug 2 (CONCURRENTLY in pipeline) | Replays this codebase end-to-end? |
|-------------|-------------------------|----------------------------------|-----------------------------------|
| v2.89.0     | âŒ FAIL                 | (not reached)                    | No                                |
| v2.90.0     | âŒ FAIL                 | (not reached)                    | No                                |
| v2.91.0     | âŒ FAIL                 | (not reached)                    | No                                |
| v2.92.1     | âœ… pass                 | âŒ FAIL                          | No                                |
| v2.95.3     | âœ… pass                 | âŒ FAIL                          | No                                |

No tested version replays a real-world migrations directory containing both multi-statement files and `CREATE INDEX CONCURRENTLY` migrations.

---

## Suggested direction (optional)

For Bug 1: send each file using the simple-query protocol (`PQexec`-equivalent, single Query message containing the file body), or split each file at top-level statement boundaries before sending each one through the extended protocol.

For Bug 2: do not implicitly wrap each migration file in a pipeline. Either run each statement as its own round-trip, or detect statements that cannot run in a pipeline (`CREATE INDEX CONCURRENTLY`, `REINDEX CONCURRENTLY`, `VACUUM`, `ALTER SYSTEM`, etc.) and run them outside the wrapper.

These are suggestions only â€” happy to test a candidate fix on the affected codebase if useful.

---

## Reference files (in our codebase, for context only â€” not required to triage)

- Workaround script: `scripts/load/apply-migrations-via-psql.mjs`
- Static enumeration of affected files: `scripts/load/enumerate-multistmt-migrations.mjs`, output in `scripts/load/multistmt-migrations.txt`
- Failure captures: `scripts/load/db-reset-output.txt`, `scripts/load/db-reset-output-2.txt`

wucm667:
Fixes #5139

## What type of PR is this?

- bug

## What this PR does / why it is needed:

When running `supabase db reset`, migrations containing `CREATE INDEX CONCURRENTLY` fail with SQLSTATE 25001 ("cannot be executed within a pipeline"). This is because `ExecBatch` in `pkg/migration/file.go` sends all statements through `conn.PgConn().ExecBatch()`, which uses PostgreSQL's extended query protocol pipeline — incompatible with CONCURRENTLY, VACUUM, REINDEX CONCURRENTLY, ALTER SYSTEM, and CLUSTER statements.

The fix:
- Added `isPipelineIncompatible()` to detect statements that cannot run in a pipeline
- Modified `ExecBatch` to flush the batch before these statements and execute them individually via `conn.Exec` (simple query protocol), then resume batching for subsequent statements
- The version record is inserted at the end of the flush, only after all statements succeed

## Testing

Manual testing with a migration file containing `CREATE INDEX CONCURRENTLY` confirms the fix — the statement now executes successfully via simple protocol while other statements continue to benefit from pipelining.

### DIFF
--- a/apps/cli-go/pkg/migration/file.go
+++ b/apps/cli-go/pkg/migration/file.go
@@ -72,23 +72,22 @@ func NewMigrationFromReader(sql io.Reader) (*MigrationFile, error) {
 	return &MigrationFile{Statements: lines}, nil
 }
 
+func isPipelineIncompatible(sql string) bool {
+	upper := strings.ToUpper(strings.TrimSpace(sql))
+	return strings.Contains(upper, "INDEX CONCURRENTLY") ||
+		strings.Contains(upper, "REINDEX CONCURRENTLY") ||
+		strings.HasPrefix(upper, "VACUUM ") ||
+		strings.HasPrefix(upper, "ALTER SYSTEM") ||
+		strings.HasPrefix(upper, "CLUSTER ")
+}
+
 func (m *MigrationFile) ExecBatch(ctx context.Context, conn *pgx.Conn) error {
-	// Batch migration commands, without using statement cache
 	batch := &pgconn.Batch{}
-	for _, line := range m.Statements {
-		batch.ExecParams(line, nil, nil, nil, nil)
-	}
-	// Insert into migration history
-	if len(m.Version) > 0 {
-		if err := m.insertVersionSQL(conn, batch); err != nil {
-			return err
-		}
-	}
-	// ExecBatch is implicitly transactional
-	if result, err := conn.PgConn().ExecBatch(ctx, batch).ReadAll(); err != nil {
-		// Defaults to printing the last statement on error
+	batchSize := 0
+	executed := 0
+
+	formatError := func(err error, i int) error {
 		stat := INSERT_MIGRATION_VERSION
-		i := len(result)
 		if i < len(m.Statements) {
 			stat = m.Statements[i]
 		}
@@ -99,7 +98,6 @@ func (m *MigrationFile) ExecBatch(ctx context.Context, conn *pgx.Conn) error {
 			if len(pgErr.Detail) > 0 {
 				msg = append(msg, pgErr.Detail)
 			}
-			// Provide helpful hint for extension type errors (SQLSTATE 42704: undefined_object)
 			if typeName := extractTypeName(pgErr.Message); len(typeName) > 0 && pgErr.Code == "42704" && !IsSchemaQualified(typeName) {
 				msg = append(msg, "")
 				msg = append(msg, "Hint: This type may be defined in a schema that's not in your search_path.")
@@ -111,7 +109,43 @@ func (m *MigrationFile) ExecBatch(ctx context.Context, conn *pgx.Conn) error {
 		msg = append(msg, fmt.Sprintf("At statement: %d", i), stat)
 		return errors.Errorf("%w\n%s", err, strings.Join(msg, "\n"))
 	}
-	return nil
+
+	flushBatch := func() error {
+		if batchSize == 0 {
+			return nil
+		}
+		if result, err := conn.PgConn().ExecBatch(ctx, batch).ReadAll(); err != nil {
+			return formatError(err, executed+len(result))
+		}
+		executed += batchSize
+		batch = &pgconn.Batch{}
+		batchSize = 0
+		return nil
+	}
+
+	for _, line := range m.Statements {
+		if isPipelineIncompatible(line) {
+			if err := flushBatch(); err != nil {
+				return err
+			}
+			if _, err := conn.Exec(ctx, line); err != nil {
+				return formatError(err, executed)
+			}
+			executed++
+		} else {
+			batch.ExecParams(line, nil, nil, nil, nil)
+			batchSize++
+		}
+	}
+
+	// Insert into migration history
+	if len(m.Version) > 0 {
+		if err := m.insertVersionSQL(conn, batch); err != nil {
+			return err
+		}
+	}
+
+	return flushBatch()
 }
 
 func markError(stat string, pos int) string {
--- a/apps/cli-go/pkg/migration/file.go
+++ b/apps/cli-go/pkg/migration/file.go
@@ -29,6 +29,11 @@ type MigrationFile struct {
 var (
 	migrateFilePattern = regexp.MustCompile(`^([0-9]+)_(.*)\.sql$`)
 	typeNamePattern    = regexp.MustCompile(`type "([^"]+)" does not exist`)
+	createIndexPattern = regexp.MustCompile(`^CREATE\s+(UNIQUE\s+)?INDEX\s+CONCURRENTLY(\s|\z)`)
+	reindexPattern     = regexp.MustCompile(`^REINDEX(\s|\().*\sCONCURRENTLY(\s|\z)`)
+	vacuumPattern      = regexp.MustCompile(`^VACUUM(\s|\(|\z)`)
+	alterSystemPattern = regexp.MustCompile(`^ALTER\s+SYSTEM(\s|\z)`)
+	clusterPattern     = regexp.MustCompile(`^CLUSTER(\s|\z)`)
 )
 
 func NewMigrationFromFile(path string, fsys fs.FS) (*MigrationFile, error) {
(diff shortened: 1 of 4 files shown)


Coly010:
Thanks for this @wucm667, and for the thorough writeup — the pipeline / `CONCURRENTLY` diagnosis and the flush-then-run-standalone approach are spot on.

Heads up on where this lands: the CLI is mid-migration from Go to TypeScript, and the `supabase migration` commands that go through this `ExecBatch` path have just been ported to a native TypeScript implementation (#5671). Rather than carry the fix in the Go path that's being retired for these commands, we've adopted your approach directly into the TS apply:

- `legacyIsPipelineIncompatible` ports your detection set (`CREATE [UNIQUE] INDEX CONCURRENTLY`, `REINDEX … CONCURRENTLY`, `VACUUM`, `ALTER SYSTEM`, `CLUSTER`), including the leading-comment stripping;
- the apply flushes the open batch, runs the incompatible statement standalone, then resumes batching, with the history insert in the final batch so the migration is recorded only after every statement succeeds;
- your `TestIsPipelineIncompatible` cases are mirrored as TS unit tests.

Commit: supabase/cli@29d3fb0e (in #5671).

I'm closing this in favour of that implementation, since these commands no longer run through the Go path. Really appreciate the contribution — the design carried over directly. 🙏

PR #5168, opened 2026-05-05
chore(monorepo): consolidate repos CLI-1408

Coly010:
Merge diverging CLI repos


.github/workflows/api-sync.yml
```diff
@@ -4,9 +4,9 @@ on:
   repository_dispatch:
     types:
       - api-sync
-  workflow_dispatch: # allow manual triggering
-
-# Add explicit permissions
```
avallete, line 9: **question**

Woudn't that turn it into something that run at any change to the go-cli source code is that what we want ?

It's meant to be a manual triggering as it get "manually triggered" by an automated action each time the `/v1` api schema change is successfully deployed on staging.

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,30 @@
...
+      run: npm install --global corepack && corepack enable
+
+    - name: Configure dependency cache
+      uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6
+      with:
+        cache: pnpm
+
+    - name: Install dependencies
+      shell: bash
+      run: pnpm install --frozen-lockfile
```
avallete, line 30: **question**

Do we use pnpm / bun ? Seems like we use both I guess that's expected could use the context why we don't just use Bun everywhere.

avallete: Claude gave me this answer, just double checking it's correct:

Found the file you're looking at — it's `.github/actions/setup/action.yml`. Here's what's actually going on:

```yaml
- Install Bun                  (oven-sh/setup-bun)
- Install Node.js              (actions/setup-node, .nvmrc)
- Enable Corepack              (provides pnpm@10.33.0 from "packageManager")
- Configure dependency cache   (cache: pnpm)
- Install dependencies         (pnpm install --frozen-lockfile)
```

The two tools play different roles, they're not redundant:

**pnpm = package/workspace manager.** The repo is a pnpm workspace (`pnpm-workspace.yaml`, `pnpm-lock.yaml`, `packageManager: pnpm@10.33.0`). It's pinned because:
- It uses `catalog:` versions (e.g. `"@swc-node/register": "catalog:"` in the root `package.json`) — a pnpm workspace feature.
- The lockfile committed to the repo is `pnpm-lock.yaml`; switching the installer would invalidate reproducibility and CI caching (`cache: pnpm`).
- `CLAUDE.md` is explicit: "`pnpm` is the package manager. Use `pnpm <script>`. Do not use `bun run` or `npm run`."

**Bun = TypeScript runtime for executing scripts/tests.** It's not used to install deps. It's needed because:
- Workspaces extend `@tsconfig/bun/tsconfig.json` and pull in `@types/bun` — code is written against Bun's runtime/types.
- Root scripts shell out to it directly, e.g. `"local-registry": "bun tools/release/local-registry.ts"` and `"cli-release": "bun tools/release/local-release.ts"` — running `.ts` without a build step.
- Tests use `@effect/vitest` with `it.live`, and the harness expects the Bun runtime to be present.

So: pnpm owns dependency graph + workspace plumbing + lockfile; Bun owns "execute TypeScript directly." Going bun-only would mean migrating the lockfile, dropping pnpm catalogs, and rewriting workspace tooling — a much bigger change than the consolidation PR is doing. It's a deliberate split, not an oversight.

.github/workflows/api-sync.yml
Coly010: ^ makes sense - i think this was done erroneously to ensure that it pointed to the correct location in the repo. i'll double check what it should be

.github/actions/setup/action.yml
Coly010: Yeah we're using PNPM as package manager and Bun as runtime

pnpm has some better handling of publish story, and Nx in particular works better with pnpm than Bun

avallete (approved):

PR #5170, opened 2026-05-05
fix: error when Docker is not running for static_files bundle (#5169)

### LINKED ISSUE #5169: static_files: bundle silently omits files when Docker daemon is not running (no warning, no error)
**Describe the bug**
Hi,

I used a edge function to import html files to build emails.
Since I updated my CLI to 2.90.0, my html files are not imported in my edge function anymore.

how I did in my config.toml

``[functions.send-auth-email]`
verify_jwt = false
static_files = ["./functions/send-auth-email/templates/*.html"]
my path : supabase/functions/send-auth-email/templates/[my html files]``

in my index.ts:

`const rawHtml = await Deno.readTextFile( new URL(./templates/${localTemplate.file}, import.meta.url), );`

I think there is an issue with the CLI, silently failing without warning the user that is will not bundle templates when docker is not running on the machine

[discussion](https://github.com/orgs/supabase/discussions/45518)

[issue related](https://github.com/supabase/cli/issues/4554)

**To Reproduce**
1. Go to an edge action on your codebase eg :send-auth-email
2. create a folder "templates" and put html file in it
setup config.toml to accept static_files for example:
[functions.send-auth-email]
static_files = ["./functions/send-auth-email/templates/*.html"]
3. read these html files with DENO in your index.ts like that : const rawHtml = await Deno.readTextFile( new URL(./templates/${localTemplate.file}, import.meta.url), );
call supabase functions deploy [your function]

**Expected behavior**
Supabase edge function is bundled with html templates and you see them on your the dashboard or a throw error

**System information**
- OS: macOS
- Version of supabase-js: 2.98.0
- Version of Node.js: 24.14.1

ChrisJr404:
## Summary

`supabase functions deploy` defaults to `--use-docker` (hidden flag, default `true`). When the Docker daemon is unreachable, the CLI prints `WARNING: Docker is not running` and silently falls back to the multipart API upload path. That path streams function source but does not produce the eszip bundle the edge runtime needs to interpret `static_files`, so any templates declared in `config.toml` are dropped server-side. The function deploys successfully, then fails at runtime when `Deno.readTextFile()` can't find its template.

This patch detects the case before any upload happens. If any enabled function has `static_files` configured and Docker is not running, the deploy aborts with:

```
Docker is not running, but static_files is configured for: send-auth-email.
Start Docker and re-run, or remove static_files from config.toml.
```

Functions without `static_files` keep the existing warning + API-fallback behaviour, so this is only stricter for the configurations that would otherwise ship broken.

A hard error is the right response here because the alternative is a silent regression that's only observable at runtime in production. The OP in #5169 explicitly accepts "throw an error" as the expected fix, and #4554 reported the same silent-drop behavior previously.

Closes #5169. Also addresses the silent-drop reported in #4554. Related discussion: https://github.com/orgs/supabase/discussions/45518

## Files changed

- `internal/functions/deploy/deploy.go` — pre-flight check before constructing the deploy options.
- `internal/functions/deploy/deploy_test.go` — `TestDeployCommand/throws_error_when_static_files_configured_but_Docker_not_running` simulates an unreachable Docker daemon (TCP listener bound then closed → `IsErrConnectionFailed`) and asserts the deploy returns an error mentioning `Docker is not running`, `static_files`, and the affected slug.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/functions/deploy/...` (8/8 PASS, including the new case)
- [x] `go test ./internal/functions/...` (all packages green)

### DIFF
--- a/internal/functions/deploy/deploy.go
+++ b/internal/functions/deploy/deploy.go
@@ -57,6 +57,20 @@ func Run(ctx context.Context, slugs []string, useDocker bool, noVerifyJWT *bool,
 		if utils.IsDockerRunning(ctx) {
 			opt = function.WithBundler(NewDockerBundler(fsys))
 		} else {
+			// Bundling static_files requires the eszip bundler, which only runs
+			// inside the edge-runtime Docker image. Without Docker, the API
+			// upload path silently drops static_files and the deployed function
+			// ships without them. Fail fast so the user can start Docker or
+			// switch to --use-api with static_files removed.
+			var withStatic []string
+			for slug, fc := range functionConfig {
+				if fc.Enabled && len(fc.StaticFiles) > 0 {
+					withStatic = append(withStatic, slug)
+				}
+			}
+			if len(withStatic) > 0 {
+				return errors.Errorf("Docker is not running, but static_files is configured for: %s.\nStart Docker and re-run, or remove static_files from config.toml.", strings.Join(withStatic, ", "))
+			}
 			fmt.Fprintln(os.Stderr, utils.Yellow("WARNING:"), "Docker is not running")
 		}
 	}
--- a/internal/functions/deploy/deploy_test.go
+++ b/internal/functions/deploy/deploy_test.go
@@ -3,12 +3,14 @@ package deploy
 import (
 	"context"
 	"fmt"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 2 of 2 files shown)


jgoux:

This was fixed as part of our port to TypeScript from Golang. 🎊

PR #5206, opened 2026-05-07
chore(ci): shard release build into per-target matrix

avallete:
## Summary

Splits the single serial `build` job in [`release-shared.yml`](.github/workflows/release-shared.yml) into an 8-way matrix on `large-linux-x86` (one shard per published platform package), and broadens the path filter on [`smoke-test-pr.yml`](.github/workflows/smoke-test-pr.yml) so PRs targeting `develop` run the full build + smoke validation whenever they touch anything that affects the build phase or runtime behaviour.

## Why

- The release `build` job runs all 8 `bun --compile` targets, the 6 Go cross-compiles (legacy), and the 6 nfpm packages serially on a single runner. Wall-clock time is bounded by the sum of every target.
- The PR-time path filter on `smoke-test-pr.yml` was scoped to `apps/cli/**` + `release-shared.yml`. A regression in `apps/cli-go/**`, `packages/cli-*/**`, root workspace files, or the shared setup action would not trigger a PR build, so build breakages could land on `develop` undetected.

### Sharded build matrix in [`release-shared.yml`](.github/workflows/release-shared.yml)

- 8 parallel shards on `large-linux-x86`, one per platform package: `cli-darwin-arm64`, `cli-darwin-x64`, `cli-linux-{arm64,x64}`, `cli-linux-{arm64,x64}-musl`, `cli-windows-{arm64,x64}`.
- Each shard uploads its own slice as `cli-build-shard-${shell}-${version}-${target}`.
- **No merge job, no re-upload of bulk bytes.** `smoke-test` / `publish` / `publish-homebrew` / `publish-scoop` switch from `name:` to `pattern: cli-build-shard-${shell}-${version}-* + merge-multiple: true`. Artifact bytes hit storage exactly once.
- `Setup Go` and `Pre-download Go modules` are gated on `inputs.shell == 'legacy'`; `Install nfpm` is gated on `startsWith(matrix.target, 'cli-linux-')`, so darwin/windows shards skip irrelevant setup.
- New `build_only` input (default `false`, so existing callers are unaffected). When `true`, `smoke-test` is skipped via `if:` and `publish` / `publish-homebrew` / `publish-scoop` auto-skip via their `needs:` chain. Useful for ad-hoc `workflow_dispatch` runs that just want to validate the build matrix.

### [`apps/cli/scripts/build.ts`](apps/cli/scripts/build.ts) refactor

- New `--target <pkg-name>` flag builds only one shard's outputs (bun binary + Go binary for legacy + archive for standard targets + deb/rpm for glibc Linux + apk for musl).
- Without `--target`, the script's end-to-end behaviour is preserved for local dev.
- Each musl shard now cross-compiles its own Go binary directly (`CGO_ENABLED=0` produces output identical to the glibc build), so the cross-shard `copyFile` from glibc is gone.

### [`apps/cli/scripts/checksums.ts`](apps/cli/scripts/checksums.ts)

New self-contained script that walks `dist/supabase_${version}_*.{tar.gz,zip,deb,rpm,apk}` and writes `dist/checksums.txt`. Called by `build.ts` in all-targets mode and by the three consumer CI jobs (`publish`, `publish-homebrew`, `publish-scoop`) right after their pattern-download.

### Broader path filter on [`smoke-test-pr.yml`](.github/workflows/smoke-test-pr.yml)

Replaces the old `apps/cli/**` + workflow paths with an explicit list of every path that can affect the build phase or how the built artifacts behave at runtime:

- `apps/cli/scripts/{build,checksums,sync-versions}.ts`
- `apps/cli/src/**`, `apps/cli/package.json`
- `apps/cli-go/**`
- `packages/cli-*/**`
- `package.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`
- `.github/actions/setup/**`
- `.github/workflows/{release-shared,smoke-test-pr}.yml`

Now any PR to `develop` that could break the release pipeline runs the full build + smoke matrix before merge.

## Compatibility

- `release-shared.yml`'s public input contract stays backward-compatible: the new `build_only` input has `default: false`.
- Local `pnpm exec bun apps/cli/scripts/build.ts --version <v>` still produces every binary, archive, package, and `checksums.txt` in `dist/`.
- The published GitHub Release contents are byte-identical (same file list, same checksum format).

## Test plan

- [ ] On this PR, confirm `smoke-test-pr.yml` runs against the new sharded build matrix and all 8 shards are green.
- [ ] On the next release-from-`develop`, confirm the matrix-fanout `build`, the pattern-download in `smoke-test` / `publish` / `brew` / `scoop`, and the inline `Generate checksums` step all behave as expected.
- [ ] Verify the published `dist/` artifact set on GitHub Releases matches the current release's file list (12 archives + packages + `checksums.txt`).


jgoux (approved):

avallete:
Dropping the PR as none of it really improve the overall CI time.

PR #5207, opened 2026-05-07
fix(cli): make npm publish idempotent for partial-failure re-runs

Coly010:
## Current Behavior

Re-running the release workflow after a partial failure produces a hard 403 from npm:

```
npm error code E403
npm error 403 Forbidden - PUT https://registry.npmjs.org/@supabase%2fcli-darwin-arm64
npm error 403 You cannot publish over the previously published versions: 2.99.0-beta.1
```

`apps/cli/scripts/publish.ts` publishes 8 platform packages in parallel and then the umbrella `supabase` package, with no error handling and no "already published" check. `semantic-release` runs in `dry_run: true` mode (`release.yml`) — it computes a version but never tags git — so a re-run on `develop` (no new commits) computes the same version. Failure modes that hit the 403:

- Some platform packages publish, then a transient error fails the rest → re-run 403s on the already-published ones.
- All platform packages publish but the umbrella build/publish fails → re-run 403s on every platform package before reaching the umbrella.
- Publish job succeeds but GH-release-creation fails → no tag, semantic-release re-computes the same version on the next push → 403.

## Expected Behavior

Re-runs of the same version are safe: already-published packages are skipped, missing ones publish, and the script exits 0 so downstream steps (GH release, brew, scoop) can converge.

## Summary

- New `isAlreadyPublished(name, version)` — anonymous HTTP GET against the active npm registry. 200 → already published, 404 → not, anything else throws. Reads the registry URL from `npm config get registry` so the `pnpm local-registry` Verdaccio harness keeps working faithfully.
- New `publishPackage({ name, version, cwd, extraFlags })` — pre-flight skip + post-failure reconciliation around the existing `pnpm publish` invocation. On publish-time conflict the probe re-runs, in case of registry CDN cache lag.
- Added a startup assertion that every platform `package.json` matches the umbrella version, so running `publish.ts` without `sync-versions.ts` first fails loud instead of publishing an inconsistent set.
- Kept `Promise.all`, all existing flags, all existing logging conventions.

Out of scope: `update-homebrew.ts` / `update-scoop.ts` have a separate latent "nothing to commit" idempotency gap on re-run after a successful brew/scoop commit. That's downstream of `publish` so it doesn't bite the failure mode here. Tracked as a follow-up.

## Test Plan

- [x] `pnpm check:all` (types, lint, fmt, knip) passes.
- [x] `bun build apps/cli/scripts/publish.ts` parses cleanly.
- [ ] Local Verdaccio re-publish smoke (manual): start `pnpm local-registry`, run `publish.ts` twice with the same version, confirm second run logs `[skip] @supabase/...@<v> already published.` for every package and exits 0.
- [ ] Production smoke (optional, after merge): `workflow_dispatch` against a known-published beta version with `dry_run=false`, confirm every package skipped and workflow green.

### DIFF
--- a/apps/cli/scripts/publish.ts
+++ b/apps/cli/scripts/publish.ts
@@ -37,11 +37,64 @@ if (!VALID_TAGS.has(tag)) {
 const cliDir = path.join(root, "apps/cli");
 const cliPkgJson = await Bun.file(path.join(cliDir, "package.json")).json();
 const umbrellaName: string = cliPkgJson.name;
+const umbrellaVersion: string = cliPkgJson.version;
 
 const dryRunFlag = dryRun ? ["--dry-run"] : [];
 const tagFlag = ["--tag", tag];
 const provenanceFlag = ["--provenance"];
 const noGitChecksFlag = ["--no-git-checks"];
+
+// Reads the active npm registry once. We honour `npm config get registry`
+// rather than hard-coding registry.npmjs.org so the existence probe and the
+// publish target stay aligned — important for the local Verdaccio harness
+// (`pnpm local-registry`), which rewrites the global npm/pnpm registry config.
+const registryUrl = (await $`npm config get registry`.quiet().text()).trim().replace(/\/$/, "");
+
+// Probes the registry for `<name>@<version>`:
+//   200 → already published, 404 → not, anything else throws.
+// Used both as a pre-flight skip check and as a post-failure reconciliation
+// when the actual publish errors with E403 (registry CDN cache may lag).
+async function isAlreadyPublished(name: string, version: string): Promise<boolean> {
+  const encodedName = name.replace("/", "%2F");
+  const res = await fetch(`${registryUrl}/${encodedName}/${version}`, { method: "GET" });
+  if (res.status === 200) return true;
+  if (res.status === 404) return false;
+  throw new Error(`npm registry probe for ${name}@${version} returned HTTP ${res.status}`);
+}
+
+// Publishes one workspace package idempotently. If the version is already
+// on the registry — either before we start or after a publish-time conflict
+// — we skip and return. Any other failure propagates.
+async function publishPackage(opts: {
+  name: string;
+  version: string;
+  cwd: string;
+  extraFlags?: string[];
+}): Promise<void> {
+  const { name, version, cwd, extraFlags = [] } = opts;
+  const label = `${name}@${version}`;
+
+  if (await isAlreadyPublished(name, version)) {
+    console.log(`  [skip] ${label} already published.`);
+    return;
+  }
+
+  console.log(`  Publishing ${label}...`);
+  try {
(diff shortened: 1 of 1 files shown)


apps/cli/scripts/publish.ts
```diff
@@ -37,37 +37,107 @@ if (!VALID_TAGS.has(tag)) {
...
+  } catch (error) {
+    if (await isAlreadyPublished(name, version)) {
+      console.log(
+        `  [skip] ${label} reported a conflict but is now present on the registry; treating as success.`,
+      );
+      return;
+    }
+    throw error;
+  }
+}
```
avallete, line 97: **question**

Won't this be an issue where we think some changes has been published under a version but it didn't ?

In the case of "already published" shouldn't we bump up the version to the next available one ?

Coly010: Good question — the skip is intentional and bumping would actually be worse here.

The version is computed deterministically by `semantic-release` from the commit history (`dry_run: true` in `release.yml`, no git tag), so re-running on the same commits produces the same version representing the same bytes. There's no "different content, same version" path unless `package.json` is hand-edited mid-flight, which the startup assertion (every platform pkg must match the umbrella version) catches as a hard failure.

Bumping on conflict breaks the lockstep contract: the umbrella pins all 8 platform deps to its own version, so if one platform was already at `v` and the rest published at `v+1`, the umbrella at `v+1` would reference a non-existent platform package. It'd also desync from the version semantic-release computes from git on the next run.

The skip exists specifically to recover from the failure modes in the description (transient flake mid-publish, umbrella build fails after platforms, GH-release step fails after publish job) — in all of those, the version already on the registry *is* the artifact we'd produce now, so converging to "all 9 published" is the right outcome.

avallete (approved):

apps/cli/scripts/publish.ts
Coly010: Following up — you were right. Tested it on the post-merge run (`25501868685`) and the failure mode you described is exactly what happened:

- semantic-release runs in `dry_run: true` and sees `v2.98.2` as the last *released* tag — it doesn't recognise the `v2.99.0-beta.1` tag from the previous attempt because dry-run never landed the GH release.
- It computed `2.99.0-beta.1` again from the 110 commits since `v2.98.2`.
- All 9 packages probed `200` on npm (the bytes from this morning's earlier publish) and the script skipped → exit 0.
- Net effect: today's fixes (#5205, #5207 itself) didn't ship, but the workflow reported green.

So my premise of "no new commits = same bytes" was wrong: the commit set grew, semantic-release just didn't advance the version because the `v2.99.0-beta.1` tag wasn't registered as a release. Same version, different intended bytes — and the idempotent skip silently dropped them.

Plan, matching @ColumFerry's read in Slack:

1. Manual version override on `workflow_dispatch` so we can re-cut as `2.99.0-beta.2` against the current commit without involving semantic-release.
2. Reliable tag landing — push the tag (and create the GH release) from the publish job even on partial failure, so semantic-release advances on the next push.

I'll take both in a follow-up PR.

PR #5210, opened 2026-05-07
fix(cli): drop full-request timeout from service gateway uploads

avallete:
## Summary

- Removes the 10s `http.Client.Timeout` from `NewServiceGateway` and `NewKongClient` so streaming storage uploads no longer fail with `Client.Timeout exceeded while awaiting headers` under load.
- The cap was inherited from the tenant-API health-check client when `NewServiceGateway` was extracted in #3929; it was never chosen for storage uploads. The active client for local seed (`newLocalClient` → `NewKongClient`) carries the same anti-pattern, so both are addressed.
- Connection setup is still bounded by the default transport's dial / TLS-handshake timeouts, and per-call deadlines continue to flow through the request `Context`.

Fixes [CLI-1441](https://linear.app/supabase/issue/CLI-1441/storage-uploads-can-time-out-after-10-seconds).

## Why

`http.Client.Timeout` covers dial + send + waiting for response, so a single small upload taking >10s for the server to ack would surface as a hard failure. `supabase seed buckets --linked` already worked because `newRemoteClient` overrides the gateway client with `http.DefaultClient` (no timeout); the local path through Kong did not.

## Test plan

- [x] `go test ./internal/utils/tenant/ ./internal/storage/client/ ./internal/status/` (all pass)
- [x] `go vet` clean for both touched packages
- [ ] Manual repro: re-run the proxy-delay scenario from the issue and confirm the upload completes instead of timing out at 10s.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QFScBK1kF6twDYjraZ2hQm)_

### DIFF
--- a/.github/workflows/cli-go-ci.yml
+++ b/.github/workflows/cli-go-ci.yml
@@ -38,7 +38,7 @@ jobs:
       - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: code-coverage-report
-          path: coverage.out
+          path: apps/cli-go/coverage.out
 
   coverage:
     name: Coverage
@@ -71,6 +71,7 @@ jobs:
           args: --timeout 5m --verbose
           version: latest
           only-new-issues: true
+          working-directory: apps/cli-go
 
   start:
     name: Start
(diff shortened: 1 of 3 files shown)


avallete (commented):
LGTM

jgoux (approved):

PR #5217, opened 2026-05-09
fix(functions): prevent silent hangs in local edge runtime

AndroidPoet:
## Summary
- add a timeout wrapper around worker.fetch(req) in functions serve runtime template
- return 504 Gateway Timeout with explicit message when function execution never resolves

## Why
Issue #3879 reports JWT.getAccessToken() hanging silently in local edge runtime.
This change prevents indefinite request hangs and surfaces a deterministic error response.

## Testing
- go test ./internal/functions/serve -run TestServe -count=1

### DIFF
--- a/apps/cli-go/internal/functions/serve/templates/main.ts
+++ b/apps/cli-go/internal/functions/serve/templates/main.ts
@@ -60,6 +60,34 @@ interface FunctionConfig {
   verifyJWT: boolean;
 }
 
+async function fetchWithTimeout(
+  worker: Awaited<ReturnType<typeof EdgeRuntime.userWorkers.create>>,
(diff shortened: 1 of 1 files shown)


jgoux:
Thanks for your contribution.

I tried to reproduce the original issue with the latest version of the CLI but it seemed to work fine.

Don't hesitate to post a reproduction if you still experience the issue and we'll look into the root cause and fix it. 👍

In the meantime I'm closing this PR.

AndroidPoet:
Haha Thanks @jgoux for reviewing this i guess i already have a lot on my plate https://github.com/AndroidPoet/supabase-kmp

PR #5219, opened 2026-05-10
fix(cli): suppress Claude Code hint when stderr is not a terminal

### LINKED ISSUE #5212: `supabase gen types` corrupts redirected output with `<claude-code-hint />` trailer when CLAUDECODE=1
### Describe the bug

When the Supabase CLI is invoked with stdout redirected to a file inside an environment where `CLAUDECODE=1` is set (e.g. inside Claude Code, or any agent harness that sets that env var), the resulting file is contaminated with a Claude Code plugin-hint trailer:

```
... legitimate generated TypeScript types ...

<claude-code-hint v="1" type="plugin" value="supabase@claude-plugins-official" />
```

That trailer makes the generated `database.types.ts` (or any other redirected output) fail to parse as TypeScript, breaking builds until the trailer is manually stripped.

This appears to have started when the CLI added Claude Code plugin-hint emission in v2.98.0+.

### To Reproduce

1. In any shell with `CLAUDECODE=1` set (or running inside Claude Code), run:
   ```bash
   npx supabase@latest gen types typescript --project-id <some-project-id> > database.types.ts
   ```
2. Open `database.types.ts`. The bottom of the file contains:
   ```
   <claude-code-hint v="1" type="plugin" value="supabase@claude-plugins-official" />
   ```
3. `tsc` (or any TypeScript compiler) reports a parse error on that line.

### Expected behavior

The hint should not contaminate the file written by `>` redirection. Two ways to satisfy this:

1. **Emit the hint to stderr** (per the [Claude Code plugin-hints documentation](https://code.claude.com/docs/en/plugin-hints), which uses `process.stderr.write(...)` in its example), so shell redirection of stdout never picks it up.
2. **Skip emission when stdout is non-TTY** (i.e. when output is being piped or redirected). Detecting non-interactive stdout is the standard CLI pattern for this.

Either fix would resolve the corruption without losing the install-prompt UX in the interactive case.

### Workarounds

For anyone hitting this in the meantime:

- `CLAUDECODE='' npx supabase gen types typescript --project-id <id> > database.types.ts` — unset the env var for the single command.
- Pipe through `sed` to strip the hint: `npx supabase gen types typescript --project-id <id> | sed '/<claude-code-hint/d' > database.types.ts`
- Strip after the fact: `sed -i '/<claude-code-hint/d' database.types.ts`

### Environment

- CLI version: 2.98.0+ (any version since the hint emission landed)
- Affected commands: confirmed on `gen types typescript`; likely affects any other command whose stdout is the primary output (e.g. `db dump`, `db diff`)
- Reproduces on: macOS, Linux, Windows (it's a shell-redirection issue, not platform-specific)

### Reference

Per the [Claude Code plugin-hints documentation](https://code.claude.com/docs/en/plugin-hints):

> When a CLI detects it's running in Claude Code (via the `CLAUDECODE` environment variable), it writes a self-closing `<claude-code-hint />` tag **to stderr**.

Example from those docs:

```js
if (process.env.CLAUDECODE) {
  process.stderr.write(
    '<claude-code-hint v="1" type="plugin" value="example-cli@claude-plugins-official" />\n',
  )
}
```

The Supabase CLI appears to be writing it to stdout instead, which is the source of the corruption.

### Additional context

Claude Code does scan both stdout and stderr for these markers and strips them before they reach the model — so the trailer is invisible inside Claude Code's interactive flow. The corruption only surfaces when the user redirects stdout to a file with `>`, because the redirect captures the bytes before Claude Code's stripping pass runs.

1wos:
Fixes #5212.

`SuggestClaudePlugin()` now skips the hint when stderr is not a terminal, so it can no longer contaminate redirected output such as `supabase gen types > out.ts 2>&1`. Verified inside a Claude Code session and added a unit test.

### DIFF
--- a/apps/cli-go/internal/utils/misc.go
+++ b/apps/cli-go/internal/utils/misc.go
@@ -19,6 +19,7 @@ import (
 	"github.com/spf13/viper"
 	"github.com/supabase/cli/internal/utils/agent"
 	"github.com/supabase/cli/pkg/migration"
+	"golang.org/x/term"
 )
 
 // Assigned using `-ldflags` https://stackoverflow.com/q/11354518
@@ -42,10 +43,16 @@ const SuggestDebugFlag = "Try rerunning the command with --debug to troubleshoot
 const claudeCodeHint = `<claude-code-hint v="1" type="plugin" value="supabase@claude-plugins-official" />`
 
 func SuggestClaudePlugin() string {
-	if agent.IsClaudeCode() {
-		return claudeCodeHint
+	if !agent.IsClaudeCode() {
+		return ""
 	}
-	return ""
+	// Suppress the hint when stderr is not a terminal. This avoids contaminating
+	// captured or redirected output (for example `supabase gen types > out.ts 2>&1`,
+	// or any tool runner that pipes both streams to a file). See #5212.
+	if !term.IsTerminal(int(os.Stderr.Fd())) {
+		return ""
+	}
+	return claudeCodeHint
 }
 
 var (
--- a/apps/cli-go/internal/utils/misc_test.go
+++ b/apps/cli-go/internal/utils/misc_test.go
@@ -215,3 +215,21 @@ func TestGetDeclarativeDir(t *testing.T) {
 		assert.Equal(t, DeclarativeDir, GetDeclarativeDir())
 	})
 }
+
+func TestSuggestClaudePlugin(t *testing.T) {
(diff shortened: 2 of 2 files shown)


avallete:
Hey there !

Thank's for your contribution, however we now try to migrated toward a more e2e testing approach. We need to evolve our contributing guidelines to match it.

This will be superseded by: https://github.com/supabase/cli/pull/5223

PR #5220, opened 2026-05-10
feat(cli): Add --no-apply flag for db schema declarative sync command

### LINKED ISSUE #5218: feat(cli): non-interactive flag for db schema declarative sync
## Summary

`supabase db schema declarative sync` always prompts "Apply this migration to local database? [Y/n]" when run in an interactive terminal. There's currently no flag to opt out of the prompt without also force-applying via `--apply`.

## Current behavior (CLI 2.95.4)

```
$ supabase db schema declarative sync --name test
...
Created new migration at supabase/migrations/20260428093833_test.sql
Apply this migration to local database? [Y/n]
```

The only escape hatches today:
- `--apply` — force-applies the migration (opposite of what's wanted)
- Closing stdin via piping, e.g. `true | supabase db schema declarative sync ... 2>&1 | cat`

## Requested behavior

A flag (e.g. `--no-apply`) that generates the migration file and exits without prompting or applying. This preserves the two-step flow where an agent or human can review the generated migration against specs and apply it later via `supabase migration up`.

## Use case

Agentic workflows and CI scripts that want the generated migration as an artifact for review before any database mutation. Forcing users to pipe `true |` into the CLI to suppress the prompt is awkward and non-obvious.

## References

- Reported by @star26bsd in https://github.com/orgs/supabase/discussions/44938#discussioncomment-16741419
- Confirmed as a gap by @avallete in https://github.com/orgs/supabase/discussions/44938#discussioncomment-16755690

mittal-parth:
## Summary

This change introduces a new flag called `no-apply` to the `db schema declarative sync command` , allowing users to generate migration files without applying them to the local database.

## Current behavior

```
$ supabase db schema declarative sync --name test
...
Created new migration at supabase/migrations/20260428093833_test.sql
Apply this migration to local database? [Y/n]
```

The only escape hatches today:
- `--apply` — force-applies the migration (opposite of what's wanted)
- Closing stdin via piping, e.g. `true | supabase db schema declarative sync ... 2>&1 | cat`

Closes #5218

## New behavior

- If `--no-apply` is set, the command writes the migration file and skips the apply step without any prompt
- `--no-apply` overrides global `--yes` and cannot be combined with `--apply`.
- Parity with legacy TS is maintained


apps/cli/docs/go-cli-porting-status.md
```diff
@@ -210,105 +210,105 @@ Legend:
...
+| `db branch switch`                     | `wrapped`     | [`../src/legacy/commands/db/branch/switch/switch.command.ts`](../src/legacy/commands/db/branch/switch/switch.command.ts)                                                                     |
+| `db remote changes`                    | `wrapped`     | [`../src/legacy/commands/db/remote/changes/changes.command.ts`](../src/legacy/commands/db/remote/changes/changes.command.ts)                                                                 |
+| `db remote commit`                     | `wrapped`     | [`../src/legacy/commands/db/remote/commit/commit.command.ts`](../src/legacy/commands/db/remote/commit/commit.command.ts)                                                                     |
+| `db schema declarative sync`           | `wrapped`     | [`../src/legacy/commands/db/schema/declarative/sync/sync.command.ts`](../src/legacy/commands/db/schema/declarative/sync/sync.command.ts) — `--apply` and `--no-apply` are mutually exclusive |
```
avallete: **nipick**

Worth flagging here that `--no-apply` is not compatible with `--yes` flag.

apps/cli/src/legacy/commands/db/schema/declarative/sync/sync.integration.test.ts
```diff
@@ -0,0 +1,80 @@
...
+          "--schema",
+          "public",
+          "--name",
+          "foo",
+          "--no-apply",
+        ],
+      ]);
+    }).pipe(Effect.provide(go.layer));
+  });
+});
```
avallete: **nitpick**

The integration test for the TS wrapper duplicates argv-construction coverage that's effectively a pure transformation. Per `CLAUDE.md`, "If a test is mostly validating a pure transformation … it should usually be a unit test instead."

mittal-parth: Two ways to solve this: 

1. We extract a pure function out in the sync.handler.ts which can be unit tested separately, without the need for building a mock. This would however, follow a different pattern than all other legacy proxy handlers. 
Eg: 

```go

export function buildLegacyDbSchemaDeclarativeSyncArgs(
  flags: LegacyDbSchemaDeclarativeSyncFlags,
): readonly string[] {
  const args: string[] = ["db", "schema", "declarative", "sync"];
  if (flags.noCache) args.push("--no-cache");
 ....
}

export const legacyDbSchemaDeclarativeSync = Effect.fn("legacy.db.schema.declarative.sync")(
  function* (flags: LegacyDbSchemaDeclarativeSyncFlags) {
    const proxy = yield* LegacyGoProxy;
    yield* proxy.exec(buildLegacyDbSchemaDeclarativeSyncArgs(flags));
  },
);

```

2. The other legacy proxy handlers do not have any tests. So we could get rid of this sync.integration.test.ts entirely, similar to other handlers.

avallete: Sorry I should've been more explicit, I think you can drop this test 👍

mittal-parth: Done

mittal-parth:
My understanding is that the failing [CI / Codegen (pull_request)](https://github.com/supabase/cli/actions/runs/25853255519/job/75965033347?pr=5220) will get resolved once we merge the deploy branch into this branch?

Or do I need to run a `go generate` to include those?

avallete:
> My understanding is that the failing [CI / Codegen (pull_request)](https://github.com/supabase/cli/actions/runs/25853255519/job/75965033347?pr=5220) will get resolved once we merge the deploy branch into this branch?
> 
> Or do I need to run a `go generate` to include those?

Your understanding is correct, no action needed on your side 👍

avallete (approved):

PR #5240, opened 2026-05-12
fix(cli): support compiled next cli e2e runtime

jgoux:
## Summary

- Run the next CLI e2e suite against the compiled `dist/supabase-next` binary so tests exercise the same artifact users run.
- Fix the two compiled-Bun blockers tracked by [CLI-1452](https://linear.app/supabase/issue/CLI-1452/compiled-bun-compile-next-binary-cant-run-supabase-functions-dev) and [CLI-1453](https://linear.app/supabase/issue/CLI-1453/compiled-bun-compile-next-binary-start-detach-daemon-fork-ignores).

## CLI-1452: `functions dev` and native/runtime assets

The compiled next binary could not boot commands that pulled in `@parcel/watcher`, because the native optional package is not embedded by `bun --compile` and the fallback `.node` path does not exist inside Bun's embedded filesystem.

This PR keeps the compiled binary path viable by making the next CLI e2e harness run through `dist/supabase-next`, by statically referencing the platform-specific watcher packages, and by embedding the Edge Runtime bootstrap with Bun's native text import attribute instead of resolving the bootstrap through a runtime source-file path. The Edge Runtime service writes that embedded bootstrap source into the runtime temp directory before launching the function runtime.

## CLI-1453: compiled binary daemon/supervisor dispatch

Compiled Bun ignores the script-path argv shape that works in JIT mode, so `child_process.fork(daemonEntryPoint)` re-entered the compiled CLI entrypoint instead of running the daemon script.

This PR makes compiled-binary dispatch explicit: `forkDaemon` marks daemon children through the environment, and the compiled next CLI entrypoint routes that child process directly to the daemon runner. The same pattern is used for process supervision: `@supabase/process-compose` forks through `process.execPath`, uses a tiny env protocol only for compiled-Bun self-dispatch, and otherwise remains a generic Node/Bun process supervision library. The supervisor runtime/protocol are native `.ts` files, so Bun forks Bun and Node forks Node.

## Reviewer Notes

The process-compose changes are intentionally runtime-neutral: there are no Supabase-specific paths or command names in the library. Supabase-specific compiled-binary dispatch lives in the CLI/stack entrypoints that own that behavior.

The compiled Bun runtime-dispatch decision is now documented in [ADR 0012](https://github.com/supabase/cli/blob/julien/cli-1452-compiled-bun-compile-next-binary-cant-run-supabase-functions/docs/adr/0012-compiled-bun-runtime-dispatch.md), and the detached-mode docs now describe the compiled-binary daemon dispatch path. The watcher loader was also reshaped into nested platform/arch branches per review feedback.


apps/cli/src/shared/runtime/parcel-file-watcher.layer.ts
```diff
@@ -8,6 +9,40 @@ import {
...
+      return wrapBinding(require("@parcel/watcher-linux-x64-musl"));
+    }
+    return wrapBinding(require("@parcel/watcher-linux-x64-glibc"));
+  }
+  if (process.platform === "win32" && process.arch === "arm64") {
+    return wrapBinding(require("@parcel/watcher-win32-arm64"));
+  }
+  if (process.platform === "win32" && process.arch === "x64") {
+    return wrapBinding(require("@parcel/watcher-win32-x64"));
+  }
```
avallete: **nitpick**

Might wanna have things nested so we have a 2 level test each maching the branch:

```
if (platform === "darwin") {
    if (process.arch === "arm64") {
    }
    if (process.arch === "x64") {
    }
}
if (platform === "linux") {
   ...
}
```

avallete (approved):
I think given the specificity and complexity of the whole fork/process/supavisor thinggy, it might be worth having this well documented in an ADR so we can more easily grasp context / understand why / how things glue together.

avallete (approved):

PR #5252, opened 2026-05-15
feat(cli): add org creation survey prompts

dnywh:
Adds optional survey capture to the Go-backed `supabase orgs create` flow. Resolves GROWTH-772.

The command now creates the organization first, then asks interactive, pretty-output users where they heard about Supabase and what they are building as an optional follow-up. Org creation still sends only the required org name, and non-empty survey answers are submitted separately to the onboarding survey endpoint with the created org slug. Non-interactive and structured-output runs continue to send only the required org name so scripts and agents are not blocked by prompts.


dnywh:
Cancelled as per [this comment](https://linear.app/supabase/issue/GROWTH-772/cli-add-survey-fields-to-org-creation-via-cli#comment-33600233).

PR #5264, opened 2026-05-18
chore: move packaged dependencies to dev dependencies

jgoux:
## What changed

This updates the published `supabase` npm package so it behaves as a lightweight shim package. The package now ships `dist/supabase.js` and declares only the platform-specific `@supabase/cli-*` packages as optional runtime dependencies. The CLI runtime dependencies are kept as development dependencies because the compiled Bun binaries already contain what they need.

The dependency catalog and lockfile were refreshed alongside that packaging change, including aligning the Effect platform packages on the same beta. To keep the workspace green with those updates, the TypeScript code was migrated to the newer Effect service APIs (`Context.Service`), schema decoding defaults were updated for the current Schema API, and mock child process handles now implement the required `unref` member.

The PR also updates the vendored `.repos/*` snapshots that were staged for this branch.


apps/cli/package.json
jgoux, line 1: This is the important change, we're moving all dependencies as dev dependencies as they're packaged in the final Bun binary and we don't want them to end up in node_modules when the users do `npm install`

avallete (approved):

PR #5265, opened 2026-05-18
fix(cli): restore --jobs flag

### LINKED ISSUE #5263: v2.99.0: `--jobs` flag dropped from `functions deploy` legacy proxy (Unrecognized flag)
**Describe the bug**

  After v2.99.0 (released 2026-05-18), `supabase functions deploy --use-api --jobs <n>` fails immediately with `Unrecognized flag: --jobs in command supabase functions deploy`. The flag worked on v2.98.x and earlier.

  The underlying Go CLI still accepts `--jobs` — it appears the new TS shim does not declare the flag on its `functions deploy` command, so it is rejected at the shim's arg-parsing layer before the Go binary is ever invoked.

  This looks like the same class of bug as #5254 (which restored `--auth` on `functions new` for the same reason): the TS shim must explicitly declare each flag it forwards to the legacy Go binary.

  **To Reproduce**

  1. Install or update the Supabase CLI to v2.99.0 (e.g. `supabase/setup-cli@v2` with `version: latest`, or `version: 2.99.0`).
  2. Run:
     ```
     supabase functions deploy --project-ref <ref> --use-api --jobs 5
     ```
  3. Observe:
     ```
     ERROR
     USAGE
       Unrecognized flag: --jobs in command supabase functions deploy
       supabase functions deploy [flags] <Function name...>
     ```

  The same command on v2.98.0 succeeds and deploys in parallel.

  **Expected behavior**

  The `--jobs` flag is accepted by the shim and forwarded to the underlying Go CLI, restoring pre-v2.99.0 behaviour (parallel deploys when combined with `--use-api`).

  **Screenshots**

  N/A — the failure is in CLI output, included above.

  **System information**

  - Version of OS: Ubuntu (GitHub-hosted `ubuntu-latest` runner)
  - Version of CLI: 2.99.0
  - Version of Docker: N/A (`--use-api` skips Docker)

  **Additional context**

  - Regression introduced in v2.99.0 (2026-05-18).
  - Workaround: pin `supabase/setup-cli@v2` to `version: 2.98.0`.
  - The v2.99.0 shim's `functions deploy --help` lists only: `--project-ref`, `--no-verify-jwt`, `--use-api`, `--import-map`, `--prune`. `--jobs` is missing.
  - Likely affects every flag the Go CLI exposes that the new TS shim does not explicitly mirror. #5254 is the precedent fix for `--auth` on `functions new`

7ttp:
- closes https://github.com/supabase/cli/issues/5263

### DIFF
--- a/apps/cli/src/legacy/commands/functions/deploy/deploy.command.ts
+++ b/apps/cli/src/legacy/commands/functions/deploy/deploy.command.ts
@@ -16,6 +16,11 @@ const config = {
   useApi: Flag.boolean("use-api").pipe(
     Flag.withDescription("Bundle functions server-side without using Docker."),
   ),
+  jobs: Flag.integer("jobs").pipe(
+    Flag.withAlias("j"),
+    Flag.withDescription("Maximum number of parallel jobs."),
+    Flag.optional,
+  ),
   importMap: Flag.string("import-map").pipe(
     Flag.withDescription("Path to import map file."),
     Flag.optional,
@@ -34,6 +39,7 @@ export const legacyFunctionsDeployCommand = Command.make("deploy", config).pipe(
       projectRef: flags.projectRef,
       noVerifyJwt: flags.noVerifyJwt,
       useApi: flags.useApi,
+      jobs: flags.jobs,
       importMap: flags.importMap,
       prune: flags.prune,
     }),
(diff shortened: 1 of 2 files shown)


jgoux:
Thanks @7ttp! Seems like it was fixed in https://github.com/supabase/cli/pull/5266. You're too fast. 😄

PR #5295, opened 2026-05-19
fix: hide legacy hidden subcommands

7ttp:
## TL;DR

hide legacy subcommands from help output when the go cli marks them hidden, while keeping them callable

## prob

the legacy ts cli still showed a few subcommands in `--help` that the go cli explicitly marks as hidden:

- `branches disable`
- `db branch`
- `db remote`
- `db test`

this happened because the existing hidden help mechanism only covered flags, not subcommands

change extends the same pattern to hidden subcommands and 
annotates the affected parent commands accordingly

execution is unchanged, so the subcommands remain callable directly

## ref: 

- extends https://github.com/supabase/cli/pull/5278


avallete:
We have an upstream PR to effect so this can be baked in. But meanwhile this seems like a good PR: https://github.com/Effect-TS/effect-smol/pull/2228

avallete (approved):
LGTM !

Thank's for helping bring things 1-1.

Issue #5283, opened 2026-05-18
gen types --db-url still requires local supabase start state (fails with "supabase start is not running")


mikkelam:
# Describe the bug
`supabase gen types --lang typescript --db-url "postgresql://postgres:postgres@127.0.0.1:54322/postgres"` fails with `supabase start is not running` even though the database is reachable and we intentionally do not use supabase start (containers are managed with Docker Compose). Debug output shows it still resolves local profile/project state (Using profile: supabase) and fails in internal/gen/types/types.go:65.

# To Reproduce

1. Run local Postgres with Docker Compose (no supabase start), reachable at 127.0.0.1:54322.
2. Run: (with appropriate db-url)

`supabase gen types --lang typescript --db-url "postgresql://postgres:postgres@127.0.0.1:54322/postgres" --debug`

3. See error and stack trace:

```
supabase gen types --lang typescript --db-url "postgresql://postgres:postgres@127.0.0.1:54322/postgres" --debug
open /Users/user/.supabase/profile: no such file or directory
Supabase CLI
Using profile: supabase (supabase.co)
github.com/supabase/cli/internal/utils/misc.go:151 (0x103875c68)
github.com/supabase/cli/internal/utils/misc.go:145 (0x10388030c)
github.com/supabase/cli/internal/gen/types/types.go:65 (0x10388030d)
github.com/supabase/cli/cmd/gen.go:99 (0x104f3adcc)
github.com/spf13/cobra@v1.10.2/command.go:1015 (0x102dc1fd4)
github.com/spf13/cobra@v1.10.2/command.go:1148 (0x102dc2750)
github.com/supabase/cli/cmd/root.go:173 (0x104f497a8)
github.com/supabase/cli/main.go:11 (0x104f4e15c)
internal/runtime/atomic/types.go:194 (0x102c0f168)
runtime/asm_arm64.s:1268 (0x102c4ad14)

supabase start is not running.
```

# Expected behavior
gen types should connect directly to the provided --db-url and generate types without requiring local supabase start state.

System information

- Ticket ID: `3748954abc3e406190420d9f5c015245`
- Version of OS: macOS (Apple Silicon)
- Version of CLI: reproduces on 2.90.0 and 2.100.0
- Version of Docker: local Docker daemon via OrbStack
- Versions of services: custom compose stack (Postgres + PostgREST + Storage + imgproxy + Kong)

Additional context

- We do not use Supabase CLI for container lifecycle; only Docker Compose.
[...]

PR #5304, opened 2026-05-19
fix(cli-go): avoid requiring local stack for gen types with explicit --db-url

mikkelam:
fix(cli-go): avoid requiring local stack for gen types with explicit --db-url

Bug fix.

`supabase gen types --db-url ...` can fail with supabase start is not running when the provided DB URL points to localhost local stack defaults, even though `--db-url` is explicitly set and the target DB is reachable without supabase start.

Issue: #5283

## What is the new behavior?

When `--db-url` is explicitly provided, gen types no longer enforces local stack running checks.
It uses the provided connection directly, including localhost targets managed outside supabase start (for example Docker Compose).

Added regression coverage in `internal/gen/types/types_test.go` for explicit `--db-url` on localhost.

## Additional context

Scope is intentionally narrow:

- behavior change applies only to explicit `--db-url` in gen types
- default local/linked flows are unchanged

### DIFF
--- a/apps/cli-go/cmd/gen.go
+++ b/apps/cli-go/cmd/gen.go
@@ -88,7 +88,8 @@ var (
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := cmd.Context()
-			if flags.DbConfig.Host == "" {
+			dbURLChanged := cmd.Flags().Changed("db-url")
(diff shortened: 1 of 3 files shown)


jgoux:
This was fixed as part of our port from Go to TypeScript! ✅

PR #5315, opened 2026-05-20
chore(ci): add setup-cli smoke test workflow

avallete:
Add a new GitHub Actions workflow to smoke test the `supabase/setup-cli` action against published CLI versions. This catches regressions like setup-cli#427 where musl libc and archive layout changes broke the action on Alpine silently.

**Key changes:**

- New `.github/workflows/setup-cli-smoke-test.yml` workflow that:
  - Tests CLI installation across Ubuntu, macOS, and Windows runners
  - Includes a dedicated Alpine container job to catch musl libc regressions
  - Verifies the installed CLI version matches the expected version
  - Can be triggered manually via `workflow_dispatch` or called from other workflows
  
- Updated `.github/workflows/release-shared.yml` to:
  - Add a `setup-cli-smoke` job that runs after beta releases
  - Runs as a post-publish signal without gating other release jobs
  - Uses `always() && needs.publish.result == 'success'` to run even when homebrew/scoop jobs are skipped or fail

The Alpine job is kept separate because GitHub Actions only honors the `container:` field on Linux runners, and the musl-vs-glibc regressions only reproduce inside a real Alpine container.

Closes: CLI-1507

https://claude.ai/code/session_01NzzeYET8sYhpCVdSvBBJhC

### DIFF
--- a/.github/workflows/release-shared.yml
+++ b/.github/workflows/release-shared.yml
@@ -329,23 +329,6 @@ jobs:
           GH_TOKEN: ${{ github.token }}
         run: gh release edit "v${VERSION}" --draft=false
 
-  # Post-publish smoke test for the `supabase/setup-cli` GitHub Action against
-  # the just-released beta. The action fetches the binary from the GitHub
-  # release we just published, so it must run after `publish`.
-  #
-  # Gated to beta releases only: beta is where setup-cli regressions need to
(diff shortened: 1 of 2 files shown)


jgoux (approved):
Nice, thanks Andrew!

PR #5331, opened 2026-05-21
feat(cli): port backups list and restore to native TypeScript

Coly010:
## Summary

Replaces the Phase-0 Go-proxy handlers for `supabase backups list` and `supabase backups restore` with native Effect-based implementations. Adds the supporting legacy infrastructure (`legacy/auth`, `legacy/config`, project-ref resolver, Glamour table renderer) that subsequent ports will reuse.

## Highlights

- **Strict Go parity on the wire.** Byte-identical `--output json` (alphabetical struct-field order, `backups: null` for empty slices to match Go's nil-slice semantics), Glamour-styled tables verified byte-for-byte against Go test fixtures, restore stderr line preserved.
- **`Output.raw(text, stream)` service method.** Handlers now route stdout/stderr writes through the `Output` service instead of calling `process.stdout/stderr.write` directly. `mockOutput` captures these into `rawChunks` + `stdoutText` / `stderrText` getters, eliminating ~30 lines of `process.*.write` monkey-patching per integration test file.
- **Shared backups infrastructure.** New `backups.layers.ts` exposes a `legacyBackupsRuntimeLayer(subcommand)` factory so each subcommand wires the platform-API + project-ref stack identically. New `mapLegacyBackupHttpError` factory in `backups.errors.ts` consolidates `RESPONSE_ERROR_TAGS` + HTTP-error dispatch and truncates response bodies to 1024 chars before embedding them in tagged errors.
- **Flag-type discipline.** Both `*.command.ts` files mark `config as const` and `export type LegacyBackups*Flags = CliCommand.Command.Config.Infer<typeof config>` (canonical `login.command.ts` pattern); handlers import the type instead of duplicating private interfaces.
- **Spinner suppressed in non-text modes.** `output.task("Fetching backups...")` / `"Initiating PITR restore..."` only run when `output.format === "text"`, eliminating dangling `[task] start:` lines on stderr in JSON / stream-json modes.
- **API contracts regenerated.** `packages/api/src/generated/contracts.ts` rebuilt from upstream OpenAPI — adds the missing `id` field on backup items, plus broader spec drift since the last sync.
- **Tests + checks pass.** Unit + integration suites green, including the new `backups.encoders.unit.test.ts` and the byte-stable `--output json` assertion (against the Go fixture from `apps/cli-go/internal/backups/list/list_test.go`). Targeted e2e `--help` smoke tests for both `list` and `restore`.

## Known Gaps (documented, not blocking)

- `V1RestorePitrBackupInput.recovery_time_target_unix` retains an upstream `>= 0` constraint that Go's `int64` does not enforce. A negative timestamp surfaces a local schema-decode error rather than the API's own error. Noted in `restore/SIDE_EFFECTS.md`; resolving requires an upstream OpenAPI change.

## Reviewer Notes

- The handlers do not log any token, error body, or response field that isn't already part of the documented Go output. Bodies are capped at 1024 chars even though the Management API is trusted, to set the right precedent for future ports against less-trusted endpoints.
- The OpenAPI regen also touches `effect-client.ts` and `openapi.json`. Diff scope is large but mechanical — all the meaningful schema deltas land in `contracts.ts`. One unrelated `next/` snapshot expectation (`platform-schema.integration.test.ts`) updated to match the new upstream description text for `v1ListAllProjects`.

Closes CLI-1301


apps/cli/src/legacy/commands/backups/list/list.e2e.test.ts
jgoux, line 1: I'm not sure if these e2e tests are useful. The golden e2e test for this command would be to actually hit the backend and assert we see the list of backups.
I hope we can run Supabase as a platform for this kind of test soon in CI. 😄 
But for now I would simply don't have an e2e test at all.

jgoux (approved):
Just a remark about e2e tests, solid work! 👍 

Don't hesitate to add more instructions in AGENTS.md if you think they are helping the agent nailing the port (and we can all benefit from it later if we help!)

PR #5333, opened 2026-05-21
feat(cli): add local_smtp config

### LINKED ISSUE #5222: Update old references to Inbucket with Mailpit
**Describe the bug**
The config.toml file still refers to inbucket as the local development server when it should be Mailpit.

**To Reproduce**
1. Create a new supabase cli setup
2. See the generated config.toml file with incorrect references

**Expected behavior**
References should be updated

**Screenshots**

**Additional context**
If applicable, add any other context about the problem here.
 - Version of supabase-cli: 2.98.2

7ttp:
## TL;DR

adds `local_smtp` as the new preferred config key, keeps `inbucket` valid with a deprecation warning and updates the default template to use `local_smtp`

## what's introduced:

`local_smtp` is now the preferred config/schema key, while existing
 `inbucket` configs continue to work for now....

## why

basically this moves the config away from an implementation specific name without breaking existing setups in phase 1 :) 

## ref

- closes https://github.com/supabase/cli/issues/5222


7ttp:
basically the phase 1 of 
- https://github.com/supabase/cli/issues/5222#issuecomment-4432418797

cc @avallete

jgoux (approved):

PR #5352, opened 2026-05-26
fix(cli): fallback function downloads

### LINKED ISSUE #5349: supabase functions download fails with 'invalid eszip v2' — appears to be eszip version skew in the extraction container
Hitting this on CLI 2.101.0. Deploying a function with `supabase functions deploy ping` works fine — it pulls the bundler image, bundles, uploads, and the function runs correctly when I invoke it over HTTP. But trying to download the same function back fails immediately:

```
$ supabase functions download ping --project-ref <ref>
Downloading function: ping
invalid eszip v2
error running container: exit 1

If your function is deployed using CLI < 1.120.0, trying running supabase functions download --legacy-bundle ping instead.
```

`--legacy-bundle` doesn't help either — it just blows up inside the wasm parser:

```
panicked at 'unreachable'
   at https://deno.land/x/eszip@v0.30.0/eszip_wasm_bg.wasm:1:1862894
   at __wbg_adapter_18 (eszip_wasm.generated.js:146:6)
   at real (eszip_wasm.generated.js:130:14)
```

I poked around to figure out what was actually going wrong. I grabbed the multipart response from `GET /v1/projects/<ref>/functions/<slug>/body` directly with curl and pulled out just the file part, then compared it byte-for-byte to the bundle the server has on disk. Identical — magic bytes `ESZIP2.3` either way:

```
$ head -c 16 bundle.eszip | xxd
00000000: 4553 5a49 5032 2e33 0000 0004 0000 0100  ESZIP2.3........
```

So nothing's getting mangled in transit. The CLI's own bundler image (`supabase/edge-runtime:v1.73.13`, which the CLI pulls itself when I run `deploy`) just produced these bytes minutes earlier. But when `download` runs and tries to read them — whether through the docker container path or the bundled `deno.land/x/eszip@v0.30.0` wasm path — both reject them as invalid v2.

Looks like the writer side moved to ESZIP2.3 but the reader side (whichever runtime image `download` spins up, plus the in-process wasm fallback) is still on something older that only goes up to 2.2. The CLI's writer and reader are out of sync within the same release.

Tried this against both self-hosted (running edge-runtime v1.74.0, latest) and the symptom is the same. From the related issues below, it sounds like Cloud users hit similar download issues — though their failure modes are different so probably not the same root cause.

Sanity: my CLI version is 2.101.0 (homebrew), Docker is running (29.2.1), function deploys cleanly with all current tooling. Just can't get it back.

Workarounds I considered if anyone else hits this:
- `--legacy-bundle` — no, also fails (wasm panic above)
- `--use-api` deploy path (raw source) — download presumably works there since the server holds plain `.ts` files; but you'd have had to deploy that way originally
- Don't lose your source 🙃

Possibly related but different symptoms:
- #1912 — old issue, silent extraction to `/home/deno` (different shape, already closed)
- #5000, #5086 — `_shared/` not extracted with Docker (partial extraction, parse succeeds — mine fails the parse)

7ttp:
## TL;DR

fixes `functions download` by falling back 

## prob: 

`functions download` tries docker unbundle first, and that path can fail, even though server side unbundle works. because of that, normal downloads can fail even when deploy succeeded :D

## sol:

basically what i did was keep the current behavior as it is , but if that docker path fails with the known mismatch, 
the cli now retries with the existing server side unbundle path instead of failing immediately

## ref

- closes https://github.com/supabase/cli/issues/5349

### DIFF
--- a/apps/cli-go/cmd/functions.go
+++ b/apps/cli-go/cmd/functions.go
@@ -56,7 +56,15 @@ var (
 			if len(args) > 0 {
 				slug = args[0]
 			}
-			return download.Run(cmd.Context(), slug, flags.ProjectRef, useLegacyBundle, useDocker, afero.NewOsFs())
+			return download.Run(
+				cmd.Context(),
+				slug,
+				flags.ProjectRef,
+				useLegacyBundle,
+				useDocker,
+				afero.NewOsFs(),
+				cmd.Flags().Changed("use-docker"),
+			)
 		},
 	}
 
--- a/apps/cli-go/internal/functions/download/download.go
+++ b/apps/cli-go/internal/functions/download/download.go
@@ -118,11 +118,12 @@ func downloadFunction(ctx context.Context, projectRef, slug, extractScriptPath s
 	return nil
 }
 
-func Run(ctx context.Context, slug, projectRef string, useLegacyBundle, useDocker bool, fsys afero.Fs) error {
+func Run(ctx context.Context, slug, projectRef string, useLegacyBundle, useDocker bool, fsys afero.Fs, useDockerChanged ...bool) error {
 	// Sanity check
 	if err := flags.LoadConfig(fsys); err != nil {
 		return err
 	}
+	dockerFlagChanged := len(useDockerChanged) > 0 && useDockerChanged[0]
 
 	// Defaults to server-side unbundling with multipart/form-data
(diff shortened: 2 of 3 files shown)


7ttp:
should be resolved in:
- https://github.com/supabase/cli/pull/5527

PR #5357, opened 2026-05-26
feat(cli): port secrets commands to native TypeScript

Coly010:
Replaces the Phase-0 Go-binary proxies for `supabase secrets list`, `secrets set`, and `secrets unset` with native Effect handlers. Output, error messages, exit codes, and filesystem side effects match the Go CLI.

## What changed

- **`secrets list/set/unset` are now native TS** — typed Management API client with `withCommandInstrumentation` + `mapLegacyHttpError`, honoring all five Go `--output {pretty,json,yaml,toml,env}` encoders plus the TS `--output-format {text,json,stream-json}`. `secrets list` emits the Glamour-table ASCII output byte-for-byte.

- **`secrets set` reads `[edge_runtime.secrets]` from `config.toml`** via `@supabase/config`'s `loadProjectConfig` + `resolveProjectSubtree`. Resolved secret values come back wrapped in `Redacted<string>`; unresolved `env(VAR)` literals (env var unset) stay as plain strings and are filtered at the handler — matches Go's `set.go:48-52` which skips entries whose SHA256 is empty (the SHA256 is empty when `DecryptSecretHookFunc` sees a still-literal `env(VAR)`).

- **No foundational `LegacyProjectConfig` wrapper.** The original port plan (CLI-1522) introduced a tolerant subtree reader at `legacy/config/legacy-project-config.{service,layer}.ts` to bypass CLI-1489's strict-decode crash. With CLI-1489 fixed upstream in #5341, that workaround is no longer needed — handlers consume the canonical loader directly. CLI-1522 is superseded.

## Reviewer-relevant context

- **Config-source filter rule:** the handler only includes secrets where `Redacted.isRedacted(value)` is true. This naturally excludes both unresolved `env(VAR)` literals (which stay as plain strings per CLI-1489's verbatim-on-missing semantics) and any pre-existing literal whose env var couldn't be resolved — equivalent to Go's empty-SHA256 filter without re-implementing the hashing pipeline.

- **`LegacySecretsConfigParseError` is kept as a domain-specific wrapper.** The handler catches `ProjectConfigParseError` from `@supabase/config` and re-raises as `LegacySecretsConfigParseError` so the legacy error tag remains stable for error-message parity and the integration test's tag assertion.

- **TOML format change:** integration test fixtures previously used the encoded `{ sha256, value }` envelope (modeling the original tolerant reader's storage shape). They now use the plain-string form (`KEY = "value"` / `KEY = "env(VAR)"`) which is what `@supabase/config` decodes and what users actually write.

- **Smoke-tested the bundled `dist/supabase-legacy` binary** (per the `Layer.provide` sibling-sharing footgun) against a fixture with both literal and unresolved-env secret entries — no service-not-found panic, POST body contains only resolved entries, unset env() refs filtered out.

- `docs/go-cli-porting-status.md` flips all three secrets rows from `wrapped` to `ported`.

Fixes CLI-1291
Closes CLI-1522


apps/cli/src/legacy/commands/secrets/list/list.command.ts
```diff
@@ -21,5 +28,8 @@ export const legacySecretsListCommand = Command.make("list", config).pipe(
       description: "List secrets for a specific project",
     },
   ]),
-  Command.withHandler((flags) => legacySecretsList(flags)),
+  Command.withHandler((flags) =>
+    legacySecretsList(flags).pipe(withCommandInstrumentation(), withJsonErrorHandling),
```
jgoux, line 32: I'm wondering something. I think I saw a few PRs from the growth team fixing telemetry on the Golang side and I don't think these fixes were ported to TS. We may want to check that the telemetry story is the same in both Golang and TS and that we reuse the same state / files to track it so it's transparent when using either side.

jgoux (approved):
LGTM, just a question about telemetry and side effects.

PR #5359, opened 2026-05-26
feat(cli): align legacy telemetry payload with Go CLI

Coly010:
## Summary

Audit (CLI-1531) found the native legacy ports were emitting a `cli_command_executed` payload that diverged from `apps/cli-go/` in three ways:

- `flags_used` (string array) + `flag_values` (allowlisted) instead of Go's single `flags` map with `"<redacted>"` for non-safe values.
- `ai_tool: string` instead of `is_agent: boolean`.
- Missing `env_signals` map.

Since the legacy/ shell sends events to the same PostHog pipeline as the Go binary, this drift was breaking dashboards and funnels silently.

## What changed

- **`shared/telemetry/event-catalog.ts`** — single literal-string source mirroring `apps/cli-go/internal/telemetry/events.go` (5 events, 18 props, 35 env-presence + 6 env-value keys, 80-char cap).
- **`shared/telemetry/posthog-config.ts`** — extracted PostHog host/key constants out of `next/config/cli-config.layer.ts` so both shells' analytics layers can share them without a cross-shell import.
- **`legacy/telemetry/legacy-analytics.layer.ts`** — implements `Analytics` with Go-shape base properties (`is_agent`, `env_signals`, no `ai_tool`). Loads linked-project cache from `SUPABASE_WORKDIR ?? process.cwd()` so it doesn't depend on per-command flag services.
- **`legacy/telemetry/legacy-command-instrumentation.ts`** — `withLegacyCommandInstrumentation({ flags, safeFlags })` emits a single `flags` map. Booleans + explicit `safeFlags` pass through verbatim (mirrors Go's `markFlagTelemetrySafe` annotation); everything else becomes `"<redacted>"`. Flag-name sort matches Go's `sort.Slice`.
- **`shared/cli/run.ts`** no longer hard-wires the analytics layer; `legacy/cli/main.ts` and `next/cli/main.ts` each provide their own.
- **7 native command files** (backups/{list,restore}, secrets/{list,set,unset}, ssl-enforcement/{get,update}) switched to `withLegacyCommandInstrumentation({ flags })`.
- The `next/` shell intentionally keeps the richer `ai_tool` / `flags_used` / `flag_values` payload — parity is scoped to `legacy/` only.

## Scope decision: custom events deferred

Every command that emits a custom event in Go — `login` (`cli_login_completed` + Alias/Identify stitching), `link` (`cli_project_linked`), `start` (`cli_stack_started`), `sso/*` and `branches/*` (`cli_upgrade_suggested`) — is still a Phase 0 `LegacyGoProxy` in the TS shell. The Go subprocess fires those events on its own; a TS wrapper would double-count. The native ports of those commands will need to reproduce the Go calls when they land — the canonical table is now documented in `AGENTS.md § Legacy Port: Telemetry Parity` and `apps/cli/src/legacy/SIDE_EFFECTS_TEMPLATE.md`.

## Reviewer notes

- Native handlers wrap with `withLegacyCommandInstrumentation`; proxy handlers stay unwrapped.
- `event-catalog.ts` is exempted from `knip:check` so future-use constants (e.g. `EventLoginCompleted`) don't get pruned before they're consumed.
- The legacy `AnalyticsContext` shape now carries both `flags` (legacy) and `flags_used`/`flag_values` (next); each shell's analytics layer reads only its own fields.
- Bundled `dist/supabase-legacy` was smoke-tested against both a native command (`backups list --help`) and a proxy command (`init --help`) per the [[feedback-layer-provide-semantics]] memory.

Fixes CLI-1531

### DIFF
--- a/apps/cli/AGENTS.md
+++ b/apps/cli/AGENTS.md
@@ -269,6 +269,32 @@ When porting a Management-API-style command, verify each item before marking the
 
 6. **Both `--output` (Go) and `--output-format` (TS) must be honored** — Go's `--output` (`pretty|json|yaml|toml|env`) takes priority when set. Pattern in `backups/list/list.handler.ts:85-113`: branch on `goOutputFlag` first, then fall through to TS `--output-format` text/json/stream-json.
 
+7. **PostHog telemetry payload matches Go 1:1** — see the next section.
+
+---
+
+## Legacy Port: Telemetry Parity
+
+The legacy shell sends the same PostHog events to the same product analytics pipeline as the Go CLI. Drift is silent (no test will catch it) and breaks dashboards. The rules:
+
+- **The canonical catalog is `shared/telemetry/event-catalog.ts`** — a 1:1 mirror of `apps/cli-go/internal/telemetry/events.go`. Reference its exported constants (`EventCommandExecuted`, `PropFlags`, `EnvSignalPresenceKeys`, …) instead of writing bare strings. When the Go catalog changes, update the TS catalog in the same PR.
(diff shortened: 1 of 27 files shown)


jgoux (approved):
Nice work, thanks!

PR #5366, opened 2026-05-27
fix(cli): skip identity stitch in CI to stop ephemeral-env identify spam

pamelachia:
## What

Gate the auto-stitch path on `!isCI` so CI runners and other ephemeral environments stop firing `$identify` / `$create_alias` once per CLI invocation.

```go
func (s *Service) NeedsIdentityStitch() bool {
    return s != nil && s.state.DistinctID == "" && s.canSend() && !s.isCI
}
```

`canSend()` is unchanged, so `cli_*` capture events still fire from CI. `login.go` calls `StitchLogin` directly without this guard, so an explicit `supabase login` still identifies in CI.

## Why

PostHog alert "total identify events increase" fired 2026-05-22 (244.5% day-over-day). I traced it to the Go CLI's first credentialed production deploy (#5329 at 2026-05-21 08:24 UTC, the binary that combined #5054's identity-stitch logic with #5314's credential wiring). Hour-by-hour change-point matches that deploy within minutes; the spike is 100% from `$lib = 'posthog-go'`.

The persistence code is correct. `SaveState` writes `~/.supabase/telemetry.json` synchronously after a successful stitch. The break is environmental: CI runners, Docker containers, and `npx supabase` wrappers wipe the home directory between invocations, so every fresh process re-stitches.

Cohort breakdown over 6 days post-deploy:

| Days active in window | Users | Identifies / user |
| -- | -- | -- |
| 1 day only | 48,387 | 6.5 |
| 6 days (daily) | 1,747 | 95.7 |

Single-day users look like CI runs. The daily, 96-identifies-per-user cohort looks like engineers whose own CI runs many Supabase workflows.

Daily `posthog-go $identify` volume went from ~15K to 638K/day and was still growing.

## Why not gate `canSend()` itself

`cli_*` capture events are heavily used:

| Event | CI share |
| -- | -- |
| `cli_stack_started` | 85% |
| `cli_project_linked` | 57% |
| `cli_command_executed` | 31% |
| `cli_login_completed` | 14% |

Six existing PostHog insights consume `cli_command_executed`, including the Agent-Led Growth dashboards. Killing CI capture would blind us to the dominant CLI use case.

Identity stitches in CI have no analytical value because each ephemeral run mints a fresh `device_id` and immediately discards it. Capture events ARE valuable because the `is_ci` property already segments them cleanly in PostHog.

## Test plan

- [x] `TestServiceNeedsIdentityStitch` adds a subtest covering `IsCI: true` (passes locally)
- [x] Full `internal/telemetry/...` package tests pass
- [x] `gofmt -d` clean, `go vet ./internal/telemetry/...` clean
- [ ] Watch `posthog-go $identify` daily volume drop back toward the pre-spike ~15K/day baseline within a day of release

## Linked

GROWTH-886


jgoux:
Hey @pamelachia thanks for that. As we're porting commands, we also have the same telemetry implementation on the TypeScript side. Can you also apply this change to the telemetry implementation we're using in TypeScript so it's 1:1 for ported and non ported commands?

apps/cli-go/internal/telemetry/service.go
```diff
@@ -154,7 +154,7 @@ func (s *Service) ClearDistinctID() error {
 }
 
 func (s *Service) NeedsIdentityStitch() bool {
-	return s != nil && s.state.DistinctID == "" && s.canSend()
+	return s != nil && s.state.DistinctID == "" && s.canSend() && !s.isCI
```
seanoliver: Do we only need to suppress auto-stitch for detected CI here? The PR description also calls out Docker/npx-style ephemeral homes, and those won't necessarily set `isCI`, so they could keep aliasing/identifying once per invocation.

(Not blocking IMO, this is probably fine if the observed spike is CI-only.)

seanoliver (approved):
LGTM! Left one non-blocking question inline!

apps/cli-go/internal/telemetry/service.go
jgoux: I tried to address it in https://github.com/supabase/cli/pull/5366/commits/efc02997cb986b298800b5084d53522c38d9bc18, tell me what you think. 🙏

seanoliver: This LGTM — Ship it!

jgoux (approved):

PR #5369, opened 2026-05-27
fix(functions): pass deno config as import map

subhwastaken:
## What changed

Pass the resolved import map to edge-runtime whenever one is configured, including `deno.json` and `deno.jsonc` files that live beside the function entrypoint.

This fixes the regression described in #5069 where bare imports from `deno.json` were no longer resolved during `supabase functions deploy` because the CLI skipped `--import-map` and relied on edge-runtime auto-discovery.

## Why

The CLI already resolves the correct import map path before bundling. Passing it explicitly keeps legacy `import_map.json` behavior and restores `deno.json`/`deno.jsonc` import resolution for both native and Docker bundling.

## Verification

- `GOCACHE=... GOMODCACHE=... go test ./function` from `apps/cli-go/pkg` passed
- `GOCACHE=... GOMODCACHE=... go test ./internal/functions/deploy` from `apps/cli-go` was attempted, but the runner volume ran out of temp build space while compiling dependencies before package tests could run


jgoux:

This bug was fixed during the TypeScript port of the command so I'm closing this PR.

PR #5374, opened 2026-05-28
feat(cli): port branches commands to native TypeScript

Coly010:
## What

Replaces the Phase 0 Go-proxy shims for all eight `supabase branches` subcommands (`list`, `create`, `get`, `update`, `pause`, `unpause`, `delete`, `disable`) with native Effect handlers under `apps/cli/src/legacy/commands/branches/`. Behavior, output formatting, exit codes, and telemetry are 1:1 with `apps/cli-go/internal/branches/*`.

## Why

Fifth batch in the Go-to-TS port (CLI-1289). Removing the subprocess fork-exec from the hot path of every `supabase branches` invocation is the immediate motivation; the broader goal is shrinking the Go-binary surface so the TS shell can ship standalone.

## Highlights

**Per-family colocated helpers**

- `branches.errors.ts` — tagged errors for every HTTP call (15 pairs) plus validation/cancellation errors.
- `branches.format.ts` — list-table and get-table renderers, `toStandardEnvs` env-map projection (`POSTGRES_URL` / `SUPABASE_URL` / `SUPABASE_<NAME>_KEY` / `SUPABASE_JWT_SECRET`), Go-byte-exact `toPostgresUrl` and `parsePoolerConnectionString`.
- `branches.resolver.ts` — `GetBranchProjectRef` equivalent (project-ref / UUID / named-lookup dispatch). Exports the shared `LEGACY_BRANCH_PROJECT_REF_PATTERN` and `LEGACY_BRANCH_UUID_PATTERN`.
- `branches.prompt.ts` — TTY list-picker and non-TTY stdin prompt with git-branch default; emits `Selected branch ID: <ref>` to stderr after a TTY pick. Cyan-styles the git-branch hint and surfaces the `Create your first branch with: supabase branches create` suggestion when branching is disabled.

**Hoisted shared infrastructure**

- `shared/git/git-branch.ts` — `detectGitBranch` (reads `$GITHUB_HEAD_REF`, then walks up CWD reading `.git/HEAD`). `next/branches/create` switches to this helper in the same change.
- `legacy/shared/legacy-profile.ts` — consolidates `profile → project_host`, `profile → dashboard_url`, and `(profile, org_slug) → billing_url` lookups that were duplicated between formatters and the upgrade-suggest helper.
- `legacy/shared/legacy-upgrade-suggest.ts` — replays Go's `SuggestUpgradeOnError` + `TrackUpgradeSuggested` flow. Fires `cli_upgrade_suggested` with `{feature_key, org_slug}` on 4xx-gated entitlement failures from `create` (`branching_limit`) and `update` (`branching_persistent`). Writes the bold billing-URL suggestion to stderr in text mode only.

**Notable Go-parity decisions**

- `create` / `update` use `Option<boolean>` for `--persistent` and `--with-data` so explicit-false reaches the API. Mirrors Go's `cmdFlags.Changed("persistent")` check; previously these would have been silently dropped.
- `create` confirm prompt fires unconditionally (TTY or non-TTY) — Go's `PromptYesNo` defaults to true on EOF, the non-interactive `Output` layer matches that.
- `update` upgrade-suggest passes the resolved `branchRef` (Go parity with `update.go:26`); the test path uses a new `mockLegacyPlatformApiService` direct-service mock that sidesteps the upstream `branch_id_or_ref` `oneOf` schema constraint.
- Pooler warning sanitizes the parse error so the connection string (pooler user/host/port) is never echoed to stderr.
- `WARNING:` prefix is yellow-styled to match `utils.Yellow`; the git-branch default in the prompt label is cyan to match `utils.Aqua`.

**Tests**

- 56 integration tests (handler-level, mocked Effect services) + 34 unit tests.
- All 982 unit + integration tests pass.
- New direct-service mock (`mockLegacyPlatformApiService`) and `legacyStatusCodeFailure` helper in `tests/helpers/legacy-mocks.ts` for handler tests that need to bypass schema validation.

Updated `docs/go-cli-porting-status.md` — eight `branches *` rows flipped from `wrapped` to `ported`. Each subcommand directory has a refreshed `SIDE_EFFECTS.md`.

Fixes CLI-1289


jgoux (approved):
LGTM, big one!

PR #5375, opened 2026-05-28
feat(cli): port sso commands to native TypeScript

Coly010:
## Summary

Replaces the six Phase 0 Go-binary proxies (`sso list`, `add`, `remove`, `update`, `show`, `info`) with native Effect implementations and flips them to `ported` in `apps/cli/docs/go-cli-porting-status.md`.

The port matches Go byte-for-byte on: error strings, exit codes, text output (Glamour AsciiStyle tables + `## SAML 2.0 Metadata XML` pretty-print via a [go-xmlfmt v1.1.3 port](apps/cli/src/legacy/commands/sso/sso.format.ts)), Go-flavoured `--output {pretty,json,yaml,toml,env}`, TS-flavoured `--output-format {text,json,stream-json}`, telemetry (`cli_command_executed` plus `cli_upgrade_suggested` on 4xx/gated for `list`/`add`/`update`/`remove`, intentionally omitted for `show`/`info` per Go), filesystem side effects (`~/.supabase/telemetry.json`, `<workdir>/supabase/.temp/linked-project.json`), and the Management API request/response contract.

## Notable changes

- **`fix(api):` first commit** repairs an unrelated codegen bug in `@supabase/api`: the OpenAPI sanitiser was stripping any property literally named `default` — including the SAML `attribute_mapping.keys.<name>.default` field. Regenerated `contracts.ts` now preserves that field across all five SSO schemas. Without this fix, `sso show`/`sso remove` would silently drop user-supplied `default` values on read.
- **Defence-in-depth over Go for the user-supplied `--metadata-url` fetch**: refuse 3xx redirects (so HTTPS-only can't be bypassed by `https://… → http://169.254.169.254/`), cap response body at 5 MiB, strict-UTF-8 decode.
- **Shared helpers hoisted for the next port (`branches/*`)**:
  - `legacy/telemetry/legacy-upgrade-suggested.ts` — `cli_upgrade_suggested` event helper.
  - `legacy/shared/legacy-resolve-token.ts` — env-vs-credentials token fallback used by raw-HTTP handlers.
  - `legacy-http-errors.ts` exports `sanitizeLegacyErrorBody` so the raw-HTTP paths in `sso add`/`sso update` get the same body-cap + control-char strip the typed-client error mapper already applies.
- **`shared/output/normalize-error.ts`** now unwraps Effect CLI's `ShowHelp` envelope and maps `MissingOption` to Cobra-format `Error: required flag(s) "X" not set` — so scripts grepping for that exact Go string keep working.

## Test coverage

- **28 unit tests** across `sso.format`, `sso.saml`, `legacy-upgrade-suggested` — including a verified-against-Go SAML fragment for `formatSsoMetadataXml` and a parity-guard test pinning the intentional Go bug where the `UPDATED AT (UTC)` row renders `created_at`.
- **90+ integration tests** per-subcommand achieving 100% branch coverage (per `apps/cli/CLAUDE.md`), exercising every output format, error branch, telemetry path, and the upgrade-gate side-call.
- **5 e2e tests** for golden paths that don't need a live API (info JSON output, UUID validation negatives across show/update/remove).

## Reviewer notes

- The `## Attribute Mapping` / `## SAML 2.0 Metadata XML` section headings + fences are emitted as plain markdown — Glamour AsciiStyle equivalence for arbitrary markdown would require a full markdown renderer. The XML body inside the fence is byte-parity via `formatSsoMetadataXml`. Section table data is byte-parity.
- Effect CLI's parser dumps the full `--help` block to stdout before a missing-required-flag error; Cobra doesn't. The error string itself is parity (Cobra wording) but the surrounding usage dump would require forking Effect CLI to suppress. Documented in `add/SIDE_EFFECTS.md`.
- `sso add` and `sso update` deliberately POST/PUT via the raw `HttpClient` rather than the typed client. The plan was written before the codegen fix in the first commit; switching to typed now is a low-risk follow-up but kept out of scope to limit blast radius.

Fixes CLI-1292


apps/cli/src/legacy/commands/sso/sso.format.ts
```diff
@@ -0,0 +1,321 @@
...
+
+// Tag matcher mirroring go-xmlfmt's `reg`: captures opening, closing, comment,
+// declaration, and self-closing tags. Non-greedy inner body so we don't span
+// across consecutive `<…>` clusters.
+const XMLFMT_TAG_RE = /<([/!]?)([^>]+?)(\/?)>/g;
+const XMLFMT_INTERTAG_SPACES_RE = />\s+</g;
+
+/**
+ * Pretty-prints an XML document by inserting newlines and indentation between
+ * tags. Behaviour-compatible port of `github.com/go-xmlfmt/xmlfmt@v1.1.3`
```
jgoux, line 149: I hope we won't have to port too many go libs. 😅

jgoux (approved):
Great work on this!

apps/cli/src/legacy/commands/sso/sso.format.ts
Coly010: agreed, it might be worth a final pass after some of these commands are ported to see if we can switch to an npm library equivalent, however, in this case i was concerned about scripted/programmatic usage that might be expecting a specific output shape

PR #5378, opened 2026-05-28
chore(cli): replace legacy hidden subcommands

7ttp:
## TL;DR

Switch legacy hidden subcommands to native `Command.withHidden`

## What's introduced:

- hide legacy `branches disable` with `Command.withHidden`
- hide legacy `db test`, `db branch`, and `db remote` with `Command.withHidden`
- remove the local hidden subcommand annotation/filtering path
- adds coverage to ensure stay callable by exact name while staying out of help output

## Why:

Effect now supports hidden subcommands natively, so legacy commands can express that behavior directly at the subcommand definition 
this removes custom local handling and leaves the path smaller, clearer, and easier to maintain :D

## ref:

- extends: https://github.com/Effect-TS/effect-smol/pull/2228
- followup to: https://github.com/supabase/cli/pull/5295


7ttp:
@jgoux can u have a look at this :) 💚

jgoux (approved):
LGTM!

PR #5382, opened 2026-05-28
feat(cli): migrate network bans

7ttp:
## TL;DR

Ports `supabase network-bans get` and `supabase network-bans remove` from the Go proxy to native TypeScript 

## What's Introduced

Replaces the `LegacyGoProxy` path for both `network-bans` leaf commands with native handlers
 wired through the existing legacy management-api runtime and instrumentation path 

## Coverage

Added a small integration suite covering this as well


Coly010 (approved):
Thank you! This was great work! 

There were some minor things that appeared in my feedback, so I addressed them in the commit I added. Thanks again!

PR #5388, opened 2026-05-29
chore(ci): gh action hardening

staaldraad:

chore

## Additional context

Applies most of zizmor's hardening suggestions. Brings some action usage up to date (client-id instead of app-id). And does some action pinning.


jgoux (approved):
🛡️ LGTM!

Coly010 (approved):

PR #5389, opened 2026-05-29
ci: notify Slack on beta and stable CLI releases

Coly010:
## What changed

Adds a Slack release notifier to the CLI release pipeline.

- **New `.github/workflows/slack-notify.yml`** — a reusable workflow (`on: workflow_call`) that posts a release message to a Slack **Incoming Webhook** via `curl`, mirroring the pattern used in `supabase-js`. Takes `version` + `channel` inputs and a required `SLACK_RELEASE_WEBHOOK` secret, and renders a Block Kit message with links to the changelog (GitHub release), the commit, and the workflow run.
- **`.github/workflows/release.yml`** — adds a `notify-slack` job that calls the reusable workflow once the release succeeds.

## Why / behaviour

- `needs: [plan, release]` with an `if:` that contains no status function keeps the implicit `success()` gate, so the notification only fires when both `plan` and `release` succeeded.
- The `if:` further restricts to **non-dry-run beta/stable** cuts — alpha and dry runs stay silent.
- Nothing depends on `notify-slack`, so a webhook/Slack failure cannot affect the already-completed release.
- Reuses existing `plan` outputs (`version`, `channel`); `secrets: inherit` forwards the webhook, same as the existing `release` job.

## Reviewer note: required repo configuration

This needs a new repository secret before it can post:

1. Create a Slack **Incoming Webhook** for the target channel.
2. Add it as a repo secret named **`SLACK_RELEASE_WEBHOOK`** (Settings → Secrets and variables → Actions). Must be a secret (not a variable) so `secrets: inherit` can forward it.

A `workflow_dispatch` dry-run will not exercise the Slack job (gated on `dry_run != 'true'` by design); the first real signal is the next merged beta cut from `develop`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -208,3 +208,22 @@ jobs:
       POSTHOG_API_KEY: ${{ secrets.POSTHOG_API_KEY }}
       POSTHOG_ENDPOINT: ${{ secrets.POSTHOG_ENDPOINT }}
       GH_APP_PRIVATE_KEY: ${{ secrets.GH_APP_PRIVATE_KEY }}
+
+  # Posts to the release Slack channel once the pipeline succeeds. Listing
+  # `release` in `needs` without a status function in `if:` keeps the implicit
+  # success() gate, so this only runs when both plan and release succeeded.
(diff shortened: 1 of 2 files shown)


jgoux (approved):
Very nice!

PR #5394, opened 2026-05-29
feat(cli): migrate vanity-subdomains

7ttp:
## TL;DR

Ports `supabase vanity-subdomains get`, `check-availability`, `activate`, and `delete` from the legacy Go proxy to native TypeScript 

## What's Introduced

This replaces the `LegacyGoProxy` path for the `vanity-subdomains` command family with native handlers 
wired through the existing legacy management-api runtime, instrumentation, and JSON error handling path

## Ref:

closes CLI-1294


Coly010 (approved):
Thank you! LGTM! Just added a minor fix around marking project-ref as a safe flag. I removed it, because I don't think we want to potentially leak that data into telemetry.

PR #5396, opened 2026-05-29
fix(cli): drop redundant identify from identity stitch

seanoliver:
## Problem

Both the Go CLI's `StitchLogin` and the TS CLI's `X-Gotrue-Id` interception fire `$create_alias` AND `$identify(distinctID, nil)` back-to-back on a user's first successful stitch. The `$identify` carries no person properties, so `$create_alias` already owns the entire device→user merge in the person graph. The Identify is pure event-volume waste.

This is the follow-up to #5366. That PR gated stitching in ephemeral environments (CI, Docker, npx) to stop the identify spike. This one tightens the path that still runs — persistent developer laptops — by dropping the redundant call.

We verified the redundancy empirically: of 638K `posthog-go` `$identify` events on 2026-05-26, zero carried `$anon_distinct_id` linkage, confirming the merge comes entirely from the alias.

## Changes

- Go: remove `s.analytics.Identify(distinctID, nil)` from `StitchLogin` in `apps/cli-go/internal/telemetry/service.go`.
- TS: remove `analytics.identify(gotrueId)` from the stitch in `apps/cli/src/legacy/auth/legacy-platform-api.layer.ts`. The TS `identify()` may have set default person properties (`cli_version`, `os`, `arch`); accepting that small loss since subsequent `cli_*` capture events carry the same data as event properties.
- Tests updated on both surfaces to assert `Alias` fires and `Identify` does not.

Kept the surfaces 1:1 so the Go/TS telemetry parity holds. `identify` stays on the `Analytics` interface — the `login` command still uses it for `cli_login_completed`, which is a separate path.

Halves the remaining persistent-laptop stitch volume (~15K/day → ~7K/day). No effect on the spike population already handled by #5366.

GROWTH-890


jgoux (approved):
Thank you!

PR #5400, opened 2026-05-30
fix(cli): suppress legacy task spinner when stdout is not a TTY

### LINKED ISSUE #5397: branches list -o json emits spinner ANSI to stdout since v2.102.0, breaking JSON parsing
`supabase branches list -o json` writes a progress-spinner escape sequence to
stdout ahead of the JSON. Parsing the captured output fails.

**Works:** v2.101.0
**Breaks:** v2.102.0

**Environment:** GitHub Actions (Ubuntu), stdout is not a TTY, `CI=true`.

**Reproduce:**
```sh
branches_json="$(supabase branches list --project-ref "$REF" -o json)"
node -e 'JSON.parse(process.argv[1])' "$branches_json"
```

**Expected:** stdout holds only the JSON array.

**Actual:** stdout starts with `1b 5b 3f 32 35 6c` (`ESC [ ? 2 5 l`, hide-cursor)
followed by a gray `│` spinner frame, so the parse throws:
```
SyntaxError: Unexpected token '', "[?25l[90"... is not valid JSON
```

**Likely cause:** v2.102.0 ported `branches` to TypeScript (#5374). The new
renderer writes the spinner to stdout even when stdout is not a TTY. The Go
build wrote clean JSON. Other commands ported in that release probably share
the bug.

**Workaround:** pin `version: 2.101.0`.

zlotnika:
v2.102.0 ported `branches`, `secrets`, `orgs`, `sso`, `ssl-enforcement`, `network-restrictions`, `backups`, etc. to TS. Each ported handler wraps its API call in:

```ts
const fetching = output.format === "text" ? yield* output.task("...") : undefined;
```

`output.format` is the TS `--output-format` flag, not the Go `--output` flag (`goFmt`). Passing Go-style `--output json` leaves `output.format === "text"`, so the `@clack` spinner starts after 200ms and writes cursor-hide (`ESC [?25l`) + a frame to stdout ahead of the JSON — breaking jq in CI. Go pre-2.102.0 emitted clean JSON.

Fix at the source: `textOutputLayer.task` returns a no-op task when `tty.stdoutIsTty` is false. Covers all 28+ ported call sites without per-handler patches; the `jsonOutputLayer` / `streamJsonOutputLayer` task impls already route through stderr / events.

Test asserts `spinner()` is never called when `stdoutIsTty: false`.

Fixes #5397

### DIFF
--- a/apps/cli/src/shared/output/output.layer.ts
+++ b/apps/cli/src/shared/output/output.layer.ts
@@ -181,6 +181,20 @@ export const textOutputLayer = Layer.effect(
           : Effect.sync(() => log.info(JSON.stringify(event))),
       task: (message: string) =>
         Effect.sync(() => {
+          if (!tty.stdoutIsTty) {
+            // Non-TTY stdout (CI, pipe, redirect) — suppress the @clack spinner,
+            // which writes cursor-hide and animated-frame ANSI to stdout and
+            // pollutes machine-parsed output. See supabase/cli#5397.
+            const noop = () => Effect.void;
+            return {
+              message: noop,
+              succeed: noop,
+              fail: noop,
+              info: noop,
+              cancel: noop,
+              clear: noop,
+            };
+          }
           let shown = false;
           let settled = false;
           let currentMessage = message;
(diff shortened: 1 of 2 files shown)


7ttp (commented):
closing in favour of: https://github.com/supabase/cli/pull/5410

> THANKS! for ur efforts towards contributing to Supabase !! 💚

PR #5405, opened 2026-05-31
fix(cli): telemetry json parse crash

### LINKED ISSUE #5395: SyntaxError: JSON Parse error: Unrecognized token ''
# Bug report

- [ ] I confirm this is a bug with Supabase, not with my own application. 
- [ ] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Can't start local supabase using the CLI `supabase start`

## To Reproduce

```
PS C:\Users\{USER}\Github\[PROJECT]> scoop install supabase
WARN  'supabase' (2.101.0) is already installed.
Use 'scoop update supabase' to install a new version.

PS C:\Users\{USER}\Github\[PROJECT]> scoop update supabase 
supabase: 2.101.0 (latest version)
Latest versions for all apps are installed! For more information try 'scoop status'

PS C:\Users\{USER}\Github\[PROJECT]> supabase start
175 | ${Rm(_.cause,"  ")}
176 | }`:_.stack).join(`
177 | `),Rm=(w,_)=>{let R=w.stack.split(`
178 | `),D=`${_}[cause]: ${R[0]}`;for(let J=1,Y=R.length;J<Y;J++)D+=`
179 | ${_}${R[J]}`;if(w.cause)D+=` {
180 | ${_}}`;return D},kQ=`~effect/Fiber/${FB}`,RS0={_A:_0,_E:_0},DS0={id:0},x5=()=>globalThis[YY];class dT{constructor(w,_=!0){this[kQ]=RS0,this.setContext(w),this.id=++DS0.id,this.currentOpCount=0,this.currentLoopCount=0,this.interruptible=_,this._stack=[],this._observers=[],this._exit=void 0,this._children=void 0,this._interruptedCause=void 0,this._yielded=void 0,this.runtimeMetrics?.recordFiberStart(this.context)}[kQ];id;interruptible;currentOpCount;currentLoopCount;_stack;_observers;_exit;_currentExit;_children;_interruptedCause;_yielded;context;currentScheduler;currentTracerContext;currentSpan;currentLogLevel;minimumLogLevel;currentStackFrame;runtimeMetrics;maxOpsBeforeYield;currentPreventYield;_dispatcher=void 0;get currentDispatcher(){return this._dispatcher??=this.currentScheduler.makeDispatcher()}getRef(w){return RT(this.context,w)}addObserver(w){if(this._exit)return w(this._exit),y1;return this._observers.push(w),()=>{let _=this._observers.indexOf(w);if(_>=0)this._observers.splice(_,1)}}interruptUnsafe(w | ... truncated 

SyntaxError: JSON Parse error: Unrecognized token ''
      at ~effect/Effect/successCont (B:/~BUN/root/supabase.exe:180:7863)
      at runLoop (B:/~BUN/root/supabase.exe:180:2053)
      at evaluate (B:/~BUN/root/supabase.exe:180:1441)
      at <anonymous> (B:/~BUN/root/supabase.exe:180:5706)
      at <anonymous> (B:/~BUN/root/supabase.exe:189:12261)
      at <anonymous> (node:fs:225:13)

Bun v1.3.13 (Windows x64)
```

## System information

- OS: Windows
- Browser (if applies) [e.g. chrome, safari]
- Version of supabase-js: [e.g. 6.0.2]
- Version of Node.js: [e.g. 10.10.0]

7ttp:
fixes `supabase start` crashing with `JSON Parse error: Unexpected EOF` from a malformed `telemetry.json` by 
recovering and regenerating unreadable telemetry state 

- closes https://github.com/supabase/cli/issues/5395


apps/cli/src/shared/telemetry/consent.ts
```diff
@@ -4,6 +4,14 @@ import type { ConsentState, TelemetryConfig } from "./types.ts";
 
 export const getConfigDir = CliConfig.useSync((cliConfig) => cliConfig.supabaseHome);
 
+function parseTelemetryConfig(content: string): TelemetryConfig | null {
+  try {
+    return JSON.parse(content) as TelemetryConfig;
+  } catch {
```
7ttp, line 10: failing here bricks commands like `supabase start` on corrupted local telemetry state,
 so returning `null` lets the existing first run path regenerate `telemetry.json` :D

Coly010 (approved):

apps/cli/src/shared/telemetry/consent.ts
jgoux: Thanks again for jumping on this and fixing the immediate crash. I agree with the core behavior here: malformed telemetry state should never crash unrelated CLI commands, and treating it as absent local state is the right outcome.

I opened a follow-up PR to align the implementation a bit more with the Effect style we’re using in the TypeScript CLI: https://github.com/supabase/cli/pull/5412

The main changes are:

- avoid raw `JSON.parse` for local state files and decode through `Effect` `Schema.fromJsonString(...)` instead
- avoid `as TelemetryConfig` casts, since they can make structurally invalid JSON look valid to TypeScript
- avoid `null` internally and use `Option` for absent config, while still outputting `null` at JSON/API boundaries where that is the expected serialized value

So the user-facing fix remains the same, but malformed JSON and valid-but-wrong-shape JSON are both handled safely and silently.

PR #5411, opened 2026-06-01
fix(cli): route TS telemetry to production PostHog

pamelachia:
## Summary

Both TypeScript CLI shells defaulted their PostHog write key to a hard-coded staging-project key, so released binaries sent `cli_command_executed` (and other CLI events) to the staging PostHog project instead of production. As commands were ported from the Go binary to native TypeScript handlers, their telemetry silently moved off production. This injects the PostHog key at build time from the existing `POSTHOG_API_KEY` release secret (the same one the Go binary already consumes via `-ldflags`), so released TS binaries report to production. Outside a release build the key is empty and telemetry no-ops, mirroring the Go binary.

## Changes

- `posthog-config.ts` (legacy shell) and `cli-config.layer.ts` (next shell): the PostHog key default now reads a build-injected `process.env.SUPABASE_CLI_POSTHOG_KEY`, empty when not injected. The runtime `SUPABASE_TELEMETRY_POSTHOG_KEY` override is unchanged.
- `build.ts`: inject the key via `bun build --define` on both the standard and musl compile paths, sourced from `POSTHOG_API_KEY` (already present in the release build env).
- `analytics.layer.ts` and `legacy-analytics.layer.ts`: return a no-op analytics client when the key is empty, so local and source builds emit nothing.
- Updated the cli-config unit test to expect the empty no-op default.

## Linear

- fixes GROWTH-895


jgoux:
Thanks for fixing this! I pushed a small follow-up to keep the TS telemetry path closer to the Go build and a bit more idiomatic for the Effect codebase.

Main changes:
- Inject both `POSTHOG_API_KEY` and `POSTHOG_ENDPOINT` at build time, so the TS shells now mirror the Go binary’s key + endpoint behavior.
- Centralized the Bun `--define` args in `build.ts` to avoid duplicating the PostHog injection logic.
- Moved PostHog config resolution into a shared helper used by both legacy and next telemetry paths.
- Represented the optional PostHog key as `Option.Option<string>` instead of using `""` as a sentinel value for “not configured”.
- Kept runtime env overrides (`SUPABASE_TELEMETRY_POSTHOG_KEY/HOST`) taking precedence over shipped build-time values.
- Added tests for no key, build-injected config, and runtime override behavior.

The important behavior stays the same: outside release builds, telemetry has no PostHog key and silently no-ops.

jgoux (approved):

PR #5453, opened 2026-06-03
chore(ci): update apply-release-notes workflow for authorization

avallete:

Use pure json review author_association to gate the approval to supabase org members.


avallete:
App now can read members, no need for it.

PR #5466, opened 2026-06-04
fix(cli-go): avoid incompatible compose dependency

jgoux:
Targets the Dependabot branch for #5461.

`compose-go/v2@2.11.0` changes `CreateHostPath` from `bool` to `types.OptOut`, but `docker/compose/v2@2.40.3` still expects the old `bool` API and no newer v2 module is available. This caused the Go CLI build and lint jobs to fail while compiling Docker Compose internals.

The start command only used Docker Compose for best-effort image prefetching before the actual Supabase container startup path. This replaces that helper with direct Docker inspect/pull calls through the existing retrying Docker client, keeps missing-image pulls parallel like Compose did, de-dupes repeated image tags, and removes the now-unused Compose dependency tree.

Added a fake-Docker regression test that verifies missing image pulls overlap concurrently.


Coly010 (approved):

apps/cli-go/internal/start/start.go
```diff
@@ -198,21 +194,22 @@ func (cli *RetryClient) ImageInspect(ctx context.Context, refStr string, options
...
+		}
+		if _, err := cli.ImageInspect(ctx, service.Image); err == nil {
+			continue
+		}
+		out, err := cli.ImagePull(ctx, service.Image, image.PullOptions{})
+		if err != nil {
+			continue
+		}
+		_, _ = io.Copy(io.Discard, out)
+		_ = out.Close()
```
avallete: **question**

Will that still pull images in parallels like compose did ? I believe this might hit start time badly if it doesn't.

avallete (changes requested):

apps/cli-go/internal/start/start.go
jgoux: You're right, we want to keep the dep on docker compose just for the "nice output" in parallel.
I have another PR to upgrade both docker and compose so we can keep using them.

PR #5467, opened 2026-06-04
fix(cli): restore Go debug output parity

jgoux:
Restores the Go CLI debug side effects for native TypeScript legacy Management API commands.

The TypeScript path was resolving profiles, credentials, and generated API URLs correctly, but it no longer emitted the debug breadcrumbs that Go printed from profile loading, access-token lookup, root command setup, and HTTP transport logging. That made `supabase --profile supabase projects list --debug` hide which profile file, resolved profile host, token source, and Management API host were used.

This ports those debug side effects into the TS legacy config, credentials, and platform API layers, and moves typed Management API HTTP debug logging to the point where generated requests have the active profile base URL attached. The output order now matches the Go management-command flow, including the repeated token-source line around the root debug banner.


Coly010 (approved):

Coly010:
🔥

PR #5468, opened 2026-06-04
feat(cli): port services

7ttp:
## TL;DR

ports `supabase services` to native ts

## What’s introduced

 adds native TS implementations for services 
moves the local service version rendering and linked version lookup into shared command logic, 
adds integration coverage for text/json output and linked version fallback behavior

## Ref

- closes CLI-1307


Coly010 (approved):
🔥

apps/cli/src/legacy/commands/services/services.handler.ts
```diff
@@ -1,8 +1,112 @@
...
+  const output = yield* Output;
+  const legacyOutput = yield* LegacyOutputFlag;
+  const cliConfig = yield* LegacyCliConfig;
+  const credentials = yield* LegacyCredentials;
+  const linkedProjectCache = yield* LegacyLinkedProjectCache;
+  const telemetryState = yield* LegacyTelemetryState;
+  const fs = yield* FileSystem.FileSystem;
+  const path = yield* Path.Path;
+
+  const projectRefPath = path.join(cliConfig.workdir, "supabase", ".temp", "project-ref");
```
avallete, line 31: **suggestion**(non-blocking): use LegacyProjectRefResolver.resolveOptional here instead of re-reading .temp/project-ref inline. It already does this projectId/ref-file chain via legacyTempPaths() and mirrors Go's flags.LoadProjectRef.

apps/cli/src/shared/services/services.shared.ts
```diff
@@ -0,0 +1,383 @@
...
+
+export interface ServiceFetchConfig {
+  readonly apiUrl: string;
+  readonly projectHost: string;
+  readonly projectRef: string;
+  readonly accessToken?: Redacted.Redacted<string>;
+  readonly userAgent: string;
+  readonly headers?: Readonly<Record<string, string | undefined>>;
+  readonly api?: ApiClient;
+  readonly tenantBaseUrlOverride?: string;
```
avallete, line 87: **note**(non-blocking): tenantBaseUrlOverride bypasses the PROJECT_REF_PATTERN guard. Test-only today, but it sits on an exported type.

PR #5469, opened 2026-06-04
chore(cli-go): migrate Docker SDK to Moby modules

jgoux:
Targets the Dependabot branch for #5461.

`compose-go/v2@2.11.0` changed API shapes used by the existing `github.com/docker/compose/v2` dependency, which made the Go CLI fail while compiling Docker Compose internals. The earlier helper-only fix avoided that mismatch by removing Compose from image prefetching, but it also moved us away from Compose's pull behavior and progress output.

This migrates the Go CLI Docker integration to the Docker v29 module layout instead: direct Docker client/API imports now use `github.com/moby/moby/client` and `github.com/moby/moby/api`, while image prefetching uses `github.com/docker/compose/v5`. That keeps Compose responsible for parallel pulls and progress display while aligning our code with the public Moby modules documented for Docker v29.

The migration also updates Docker option/result types across the Go CLI, replaces removed Docker utility imports with their Moby equivalents or narrow local test helpers, and removes the old pre-v1.42 volume-prune compatibility branch because Docker v29 no longer targets that API range.


jgoux:
As this change is quite large, we don't want to take the risk of breaking users workflows before the port is completed. Closing.

PR #5470, opened 2026-06-04
feat(cli): port bootstrap command to native TypeScript

Coly010:
## What changed

Replaces the Phase-0 Go proxy for `supabase bootstrap` with a native TypeScript implementation in the legacy shell. `bootstrap` is a meta-orchestrator that chains: workdir resolve/prompt → template list/download → blank `init` → ensure-login → `projects create` → `projects api-keys` (backoff) → `link` services → health poll (backoff) → write `.env` → `db push` → start suggestion.

Per the "hoist before you duplicate" policy, the shared orchestration is extracted from the already-ported `login` / `projects create` / `projects api-keys` / `link` handlers into `legacy/shared/`, and those handlers are refactored to delegate (no behavior change — their existing tests still pass):

- `legacy-ensure-login` (browser login flow + post-login telemetry)
- `legacy-project-create-core`
- `legacy-get-api-keys`
- `legacy-link-services-core`
- `legacy-tenant-keys` (`legacyExtractServiceKeys`)
- the login api/crypto layers move to `legacy/shared/` so `bootstrap` can compose them without a cross-command import

## Why

Phase 1+ native port of `bootstrap` (the last Quick-Start command still proxied), unblocking native machine output (`--output-format json|stream-json`) and removing the Go dependency for everything except the migration push.

## Reviewer-relevant context

- **`db push` is delegated to the bundled Go binary (interim).** It is the only non-native step, pending a separate native `db push` port. `LegacyGoProxy.exec` exits the process on a non-zero exit rather than returning a failure, so Go's push backoff cannot be reproduced from the proxy (single attempt). Documented in `SIDE_EFFECTS.md`.
- **Go-parity decisions worth a look:**
  - No `cli_project_linked` telemetry — Go's `bootstrap` calls `link.LinkServices` (services-only), **not** `link.Run`, so it deliberately skips the project-linked event, status check, and `linked-project.json` temp write.
  - API keys are fetched **without** `reveal` (Go's `RunGetApiKeys` uses empty params).
  - Workdir / DB password env vars are read **prefixed** (`SUPABASE_WORKDIR`, `SUPABASE_DB_PASSWORD`) to match Go's `viper.SetEnvPrefix("SUPABASE")`.
  - `Using workdir …` prints only when the resolved workdir differs from the cwd (matches Go's `ChangeWorkDir` guard).
- **Hardening:** template download rejects entries that would escape the target directory; GitHub and health error bodies are sanitized before surfacing.
- `docs/go-cli-porting-status.md` flips `bootstrap` `wrapped` → `ported` with the interim-db-push note.

## Follow-ups (out of scope)

- Native `db push` port will remove the proxy delegation (eliminating the single-attempt and transient `--password`-in-process-table limitations).
- `bootstrap.e2e.test.ts` deferred — integration coverage is comprehensive; e2e needs replayed API fixtures plus handling of the real Go `db push` subprocess.

### DIFF
--- a/apps/cli/src/legacy/commands/bootstrap/bootstrap.handler.ts
+++ b/apps/cli/src/legacy/commands/bootstrap/bootstrap.handler.ts
@@ -28,6 +28,11 @@ import {
 } from "./bootstrap.errors.ts";
 import { deriveDbConfig } from "./bootstrap.pgconfig.ts";
 import { suggestAppStart } from "./bootstrap.suggest.ts";
+import {
+  LEGACY_BOOTSTRAP_MAX_RETRIES,
+  legacyBootstrapBackoff,
+  legacyBootstrapRetryNotify,
+} from "./bootstrap.retry.ts";
 import { type LegacyStarterTemplate, LegacyTemplateService } from "./bootstrap.templates.ts";
 import type { LegacyBootstrapFlags } from "./bootstrap.command.ts";
 
@@ -39,14 +44,9 @@ const SCRATCH_TEMPLATE: LegacyStarterTemplate = {
   start: "supabase start",
 };
 
-// Go's `utils.NewBackoffPolicy`: exponential backoff, 3s initial interval,
-// `maxRetries = 8` (`internal/utils/retry.go`). Injectable so tests run fast.
-const LEGACY_BOOTSTRAP_MAX_RETRIES = 8;
-const legacyBootstrapBackoff = Schedule.exponential("3 seconds");
-
 export const legacyBootstrap = Effect.fn("legacy.bootstrap")(function* (
   flags: LegacyBootstrapFlags,
-  retrySchedule: typeof legacyBootstrapBackoff = legacyBootstrapBackoff,
+  retrySchedule: Schedule.Schedule<unknown> = legacyBootstrapBackoff,
 ) {
   const output = yield* Output;
   const tty = yield* Tty;
@@ -70,6 +70,10 @@ export const legacyBootstrap = Effect.fn("legacy.bootstrap")(function* (
   // while leaving the surrounding process untouched.
   const originalCwd = process.cwd();
   let createdRef: string | undefined;
+  // Resolved bootstrap workdir, hoisted so the linked-project-cache finalizer writes
+  // beside the other `supabase/.temp/` files instead of `cliConfig.workdir`. Go achieves
+  // this by re-running `flags.LoadConfig` after `ChangeWorkDir` (`bootstrap.go:98-100`).
+  let resolvedWorkdir: string | undefined;
 
   yield* Effect.gen(function* () {
     // A. Resolve workdir (flag -> env -> prompt -> cwd). `bootstrap.go:30-40`.
@@ -88,6 +92,7 @@ export const legacyBootstrap = Effect.fn("legacy.bootstrap")(function* (
     const workdir = path.isAbsolute(workdirInput)
       ? workdirInput
       : path.join(runtimeInfo.cwd, workdirInput);
+    resolvedWorkdir = workdir;
 
     // B. List templates + resolve the starter. `bootstrap.go:38-58` / `cmd:25-58`.
     const samples = yield* templateService.listSamples;
@@ -177,11 +182,13 @@ export const legacyBootstrap = Effect.fn("legacy.bootstrap")(function* (
     createdRef = projectRef.length > 0 ? projectRef : undefined;
 
     // H. Fetch api keys with backoff; each attempt prints "Linking project...".
-    // `bootstrap.go:88-97`.
+    // `bootstrap.go:88-97`. The notify wrapper reproduces Go's `NewErrorCallback`
+    // (`<err>\nRetry (n/8):` after each failed attempt); a fresh counter per block.
+    const apiKeysNotify = legacyBootstrapRetryNotify();
     const keys = yield* Effect.gen(function* () {
       if (isText) yield* output.raw("Linking project...\n", "stderr");
       return yield* legacyGetProjectApiKeys(projectRef);
-    }).pipe(Effect.retry(retry));
+    }).pipe(apiKeysNotify, Effect.retry(retry));
     const { anon } = legacyExtractServiceKeys(keys);
 
     // I. Link services (best-effort, anon key) + mandatory project-ref write.
@@ -197,6 +204,7 @@ export const legacyBootstrap = Effect.fn("legacy.bootstrap")(function* (
     yield* fs.writeFileString(paths.projectRef, projectRef);
 
     // J. Poll health until db is healthy. `bootstrap.go:106-113`.
+    const healthNotify = legacyBootstrapRetryNotify();
     yield* Effect.gen(function* () {
       if (isText) yield* output.raw("Checking project health...\n", "stderr");
       const services = yield* api.v1
@@ -209,7 +217,7 @@ export const legacyBootstrap = Effect.fn("legacy.bootstrap")(function* (
           });
         }
       }
-    }).pipe(Effect.retry(retry));
+    }).pipe(healthNotify, Effect.retry(retry));
 
     // K. Derive db config + write .env (non-fatal). `bootstrap.go:114-121`.
     const dbConfig = deriveDbConfig(projectRef, created.dbPassword, cliConfig.projectHost);
@@ -274,7 +282,9 @@ export const legacyBootstrap = Effect.fn("legacy.bootstrap")(function* (
     ),
     Effect.ensuring(
       Effect.suspend(() =>
-        createdRef === undefined ? Effect.void : linkedProjectCache.cache(createdRef),
+        createdRef === undefined
+          ? Effect.void
+          : linkedProjectCache.cache(createdRef, resolvedWorkdir),
       ),
     ),
     Effect.ensuring(telemetryState.flush),
--- a/apps/cli/src/legacy/commands/bootstrap/bootstrap.handler.ts
+++ b/apps/cli/src/legacy/commands/bootstrap/bootstrap.handler.ts
@@ -1,13 +1,297 @@
-import { Effect, Option } from "effect";
+import { Effect, FileSystem, Option, Path, Schedule } from "effect";
+import * as HttpClientError from "effect/unstable/http/HttpClientError";
+
+import { LegacyPlatformApi } from "../../auth/legacy-platform-api.service.ts";
+import { LegacyCliConfig } from "../../config/legacy-cli-config.service.ts";
+import { LegacyLinkedProjectCache } from "../../telemetry/legacy-linked-project-cache.service.ts";
+import { LegacyTelemetryState } from "../../telemetry/legacy-telemetry-state.service.ts";
+import { LegacyWorkdirFlag, LegacyYesFlag } from "../../../shared/legacy/global-flags.ts";
+import { Output } from "../../../shared/output/output.service.ts";
 import { LegacyGoProxy } from "../../../shared/legacy/go-proxy.service.ts";
+import { RuntimeInfo } from "../../../shared/runtime/runtime-info.service.ts";
+import { Tty } from "../../../shared/runtime/tty.service.ts";
+import { legacyAqua, legacyBold } from "../../shared/legacy-colors.ts";
+import { legacyEnsureLogin } from "../../shared/legacy-ensure-login.ts";
+import { legacyGetProjectApiKeys } from "../../shared/legacy-get-api-keys.ts";
+import { sanitizeLegacyErrorBody } from "../../shared/legacy-http-errors.ts";
+import { legacyLinkServicesCore } from "../../shared/legacy-link-services-core.ts";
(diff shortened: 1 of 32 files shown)


apps/cli/src/legacy/commands/bootstrap/bootstrap.handler.ts
```diff
@@ -1,13 +1,297 @@
...
+  name: "scratch",
+  description: "An empty project from scratch.",
+  url: "",
+  start: "supabase start",
+};
+
+// Go's `utils.NewBackoffPolicy`: exponential backoff, 3s initial interval,
+// `maxRetries = 8` (`internal/utils/retry.go`). Injectable so tests run fast.
+const LEGACY_BOOTSTRAP_MAX_RETRIES = 8;
+const legacyBootstrapBackoff = Schedule.exponential("3 seconds");
```
avallete: **issue**

I think the go version of the policy include two things that seems to differ:

1. It uses a jitter randomization (IDK if that's doable with Effect)
2. It has a max interval of 60s and max elapsed time of 15 mins.

With this tell me if I'm wrong but it seems like this might cause a behavioral drift for the retry over a stuck health check it might wait longer.

Separately, Go's `NewErrorCallback` prints `<err>\nRetry (n/8):` between attempts (debug logger for the first 2 failures, **stderr from the 3rd on**). The retries here reprint `"Linking project..."` / `"Checking project health..."` but not that error/`Retry` notice, so stderr diverges once there are ≥3 failures (or under `--debug`).

This might be acceptable drifts if the behavior is to convoluted to reproduce. Just wanna flag it out.

apps/cli/src/legacy/commands/bootstrap/bootstrap.handler.ts
```diff
@@ -1,13 +1,297 @@
...
+        } catch {
+          /* original cwd vanished — nothing to restore to */
+        }
+      }),
+    ),
+    Effect.ensuring(
+      Effect.suspend(() =>
+        createdRef === undefined ? Effect.void : linkedProjectCache.cache(createdRef),
+      ),
+    ),
```
avallete, line 289: **issue**

`linked-project.json` can be written to the wrong directory.

`linkedProjectCache.cache` resolves its path from `cliConfig.workdir` (see `legacy-linked-project-cache.layer.ts`), and `cliConfig.workdir` is resolved once at layer-build time (`legacy-cli-config.layer.ts:110`): raw `--workdir` flag → `SUPABASE_WORKDIR` env → cwd-walk. But the bootstrap target workdir is this handler's own resolved/absolutized `workdir` which is what `project-ref` (line 197) and `legacyLinkServicesCore` correctly use.

When `--workdir` is absolute they coincide (and every integration test hardcodes `cliConfig.workdir`, so this is masked). But when the workdir comes from the **interactive prompt** (no flag/env), `cliConfig.workdir` is the cwd-walk result (original cwd/ancestor), so `linked-project.json` lands in the wrong place while `project-ref` lands in the right one. Go avoids this by calling `flags.LoadConfig(fsys)` _after_ `ChangeWorkDir`, re-resolving paths relative to the workdir before `PersistentPostRun` caches. Consider feeding the resolved `workdir` into the cache, or documenting the limitation in `SIDE_EFFECTS.md`.

Reproduced with the real `legacyCliConfigLayer` + `legacyLinkedProjectCacheLayer` driving the handler through the prompt path (no flag/env): `project-ref` lands in the prompted workdir `.../myproj/supabase/.temp/`, but `linked-project.json` lands in the cwd-walk parent `.../supabase/.temp/` — a different directory than Go writes to. The existing integration tests miss this because they pass an absolute `--workdir` and a mock `LegacyCliConfig` whose `workdir` is hardcoded to that same dir, so the two paths always coincide in the suite.

apps/cli/src/legacy/commands/bootstrap/bootstrap.templates.ts
```diff
@@ -0,0 +1,229 @@
...
+              const resolvedLocal = path.resolve(localPath);
+              if (
+                resolvedLocal !== resolvedTarget &&
+                !resolvedLocal.startsWith(resolvedTarget + path.sep)
+              ) {
+                return yield* new LegacyBootstrapTemplateDownloadError({
+                  message: `failed to download template: entry escapes target directory: ${entryPath}`,
+                });
+              }
+              downloads.push({ localPath, remoteUrl: entry.download_url ?? "" });
```
avallete: **nitpick**

`download_url` null fallback. GitHub returns `null` `download_url` for files larger than 1 MB (and for submodules). The `?? ""` fallback makes `downloadFile` issue `GET ""` and surface a confusing transport error instead of a clear "unsupported entry" message. Edge case, but worth an explicit guard since Go's `GetContents` handles large files via a different path.

apps/cli/src/legacy/commands/bootstrap/bootstrap.templates.ts
```diff
@@ -0,0 +1,229 @@
...
+              message: `failed to download template: expected a directory listing for ${contentPath}`,
+            });
+          }
+          const listing = payload as ReadonlyArray<GithubContentEntry>;
+          for (const entry of listing) {
+            const entryPath = entry.path ?? "";
+            if (entry.type === "file") {
+              const relative = entryPath.startsWith(root)
+                ? entryPath.slice(root.length)
+                : entryPath;
```
avallete: **nitpick**

Prefix match should be segment-aware. `startsWith(root)` would mis-slice a sibling sharing the prefix (e.g. `root="examples/app"`, entry `"examples/app-2/x"` → `"-2/x"`). The contents API only returns children of the queried directory so it can't happen in practice, but matching on `root + "/"` (or splitting on path segments) would be more obviously correct and defensive.

apps/cli/src/legacy/commands/bootstrap/bootstrap.handler.ts
avallete: Here is the integration test used to reproduce the issue (AI generated):

```ts
// apps/cli/src/legacy/commands/bootstrap/bootstrap.workdir-cache.integration.test.ts
import { existsSync, mkdtempSync, rmSync } from "node:fs";
import { tmpdir } from "node:os";
import { join } from "node:path";

import { BunServices } from "@effect/platform-bun";
import { describe, expect, it } from "@effect/vitest";
import { Effect, Layer, Option, Schedule } from "effect";

import {
  mockAnalytics,
  mockBrowser,
  mockOutput,
  mockRuntimeInfo,
  mockStdin,
  mockTty,
} from "../../../../tests/helpers/mocks.ts";
import {
  type LegacyApiHandler,
  LEGACY_VALID_REF,
  legacyJsonResponse,
  mockLegacyCredentialsTracked,
  mockLegacyLoginApi,
  mockLegacyLoginCrypto,
  mockLegacyPlatformApi,
  mockLegacyTelemetryStateTracked,
} from "../../../../tests/helpers/legacy-mocks.ts";
import { legacyCliConfigLayer } from "../../config/legacy-cli-config.layer.ts";
import { legacyLinkedProjectCacheLayer } from "../../telemetry/legacy-linked-project-cache.layer.ts";
import {
  LegacyOutputFlag,
  LegacyProfileFlag,
  LegacyWorkdirFlag,
  LegacyYesFlag,
} from "../../../shared/legacy/global-flags.ts";
import { LegacyGoProxy } from "../../../shared/legacy/go-proxy.service.ts";
import { LegacyTemplateService } from "./bootstrap.templates.ts";
import { legacyBootstrap } from "./bootstrap.handler.ts";
import type { LegacyBootstrapFlags } from "./bootstrap.command.ts";

const FAST_BACKOFF = Schedule.exponential("1 milli");

const PROJECT = {
  id: LEGACY_VALID_REF,
  ref: LEGACY_VALID_REF,
  organization_id: "org-1",
  organization_slug: "acme",
  name: "alpha",
  region: "us-east-1",
  status: "COMING_UP",
};
const ORGS = [{ id: "org-1", slug: "acme", name: "Acme Inc" }];
const API_KEYS = [{ name: "anon", api_key: "anon-key" }];
const HEALTHY = [{ name: "db", healthy: true, status: "ACTIVE_HEALTHY" }];

describe("legacy bootstrap linked-project cache location", () => {
  it.live(
    "writes linked-project.json into the (prompted) bootstrap workdir, not cliConfig.workdir",
    () => {
      // cwd = parent; the user types a *subdir* at the prompt. No --workdir flag
      // and no SUPABASE_WORKDIR env => the case the rest of the suite never covers.
      const parent = mkdtempSync(join(tmpdir(), "bootstrap-cache-"));
      const subdir = "myproj";
      const bootstrapWorkdir = join(parent, subdir);

      // Token via env => ensure-login is a no-op and the cache has a bearer token.
      const prevToken = process.env["SUPABASE_ACCESS_TOKEN"];
      const prevWorkdir = process.env["SUPABASE_WORKDIR"];
      process.env["SUPABASE_ACCESS_TOKEN"] = "sbp_" + "a".repeat(40);
      delete process.env["SUPABASE_WORKDIR"];

      const out = mockOutput({ format: "text", promptTextResponses: [subdir] });

      const handler: LegacyApiHandler = (request, recorded) => {
        const url = recorded.urlWithParams;
        if (recorded.method === "POST" && /\/v1\/projects(\?|$)/.test(url)) {
          return Effect.succeed(legacyJsonResponse(request, 201, PROJECT));
        }
        if (url.includes("/api-keys")) {
          return Effect.succeed(legacyJsonResponse(request, 200, API_KEYS));
        }
        if (url.includes("/health")) {
          return Effect.succeed(legacyJsonResponse(request, 200, HEALTHY));
        }
        if (url.includes("/v1/organizations")) {
          return Effect.succeed(legacyJsonResponse(request, 200, ORGS));
        }
        // GET /v1/projects/{ref} — read by the linked-project cache.
        if (recorded.method === "GET" && url.includes(`/v1/projects/${LEGACY_VALID_REF}`)) {
          return Effect.succeed(legacyJsonResponse(request, 200, PROJECT));
        }
        return Effect.succeed(legacyJsonResponse(request, 404, {}));
      };
      const api = mockLegacyPlatformApi({ handler });

      const proxyLayer = Layer.succeed(LegacyGoProxy, { exec: () => Effect.void });
      const templateLayer = Layer.succeed(LegacyTemplateService, {
        listSamples: Effect.succeed([]),
        download: () => Effect.void,
      });

      // GlobalFlag services don't cross sibling boundaries in Layer.mergeAll
      // (CLAUDE.md item 5), so provide them explicitly into the real config layer.
      const flagsLayer = Layer.mergeAll(
        Layer.succeed(LegacyProfileFlag, "supabase"),
        Layer.succeed(LegacyWorkdirFlag, Option.none()),
        Layer.succeed(LegacyYesFlag, false),
        Layer.succeed(LegacyOutputFlag, Option.none()),
      );
      const runtime = mockRuntimeInfo({ cwd: parent });
      const credentials = mockLegacyCredentialsTracked();

      const configLayer = legacyCliConfigLayer.pipe(
        Layer.provide(flagsLayer),
        Layer.provide(runtime),
        Layer.provide(BunServices.layer),
      );
      const cacheLayer = legacyLinkedProjectCacheLayer.pipe(
        Layer.provide(configLayer),
        Layer.provide(credentials.layer),
        Layer.provide(api.httpClientLayer),
        Layer.provide(BunServices.layer),
      );

      const layer = Layer.mergeAll(
        BunServices.layer,
        out.layer,
        api.layer,
        api.httpClientLayer,
        configLayer,
        cacheLayer,
        credentials.layer,
        mockTty({ stdinIsTty: true, stdoutIsTty: false }),
        runtime,
        mockLegacyTelemetryStateTracked().layer,
        mockAnalytics().layer,
        templateLayer,
        proxyLayer,
        mockLegacyLoginApi({ gotrueId: "gotrue-user" }).layer,
        mockLegacyLoginCrypto().layer,
        mockBrowser(),
        mockStdin(true),
        flagsLayer,
      );

      const flags: LegacyBootstrapFlags = {
        template: Option.some("scratch"),
        password: Option.some("s3cret"),
      };

      return Effect.gen(function* () {
        yield* legacyBootstrap(flags, FAST_BACKOFF);

        const projectRef = join(bootstrapWorkdir, "supabase", ".temp", "project-ref");
        const cacheInWorkdir = join(bootstrapWorkdir, "supabase", ".temp", "linked-project.json");
        const cacheInParent = join(parent, "supabase", ".temp", "linked-project.json");

        // project-ref already goes to the right place...
        expect(existsSync(projectRef)).toBe(true);
        // ...so linked-project.json must land beside it (Go writes both into workdir).
        expect(existsSync(cacheInWorkdir)).toBe(true);
        expect(existsSync(cacheInParent)).toBe(false);
      }).pipe(
        Effect.provide(layer),
        Effect.ensuring(
          Effect.sync(() => {
            if (prevToken !== undefined) process.env["SUPABASE_ACCESS_TOKEN"] = prevToken;
            else delete process.env["SUPABASE_ACCESS_TOKEN"];
            if (prevWorkdir !== undefined) process.env["SUPABASE_WORKDIR"] = prevWorkdir;
            rmSync(parent, { recursive: true, force: true });
          }),
        ),
      );
    },
  );
});
```

avallete (approved):
LGTM 👍

Issue #5476, opened 2026-06-04
db diff misses reloption-only changes on existing
  views


nirav-gondaliya:
### Discovered while testing supabase/cli#3973's fix

The PR for supabase/cli#3973 restores `WITH (...)` on **newly created** views. A related but distinct gap remains: when a view already exists and only its reloptions change, `db diff` emits nothing.

### Repro

1. Declare a view with `with (security_invoker=true)`, run `supabase db diff -f initial` and apply the migration.
2. Change the schema file to `with (security_invoker=false)` (or any other reloption swap).
3. Run `supabase db diff -f update_view`.

### Expected

The diff captures the reloption change, e.g.:

```sql
ALTER VIEW "public"."user_details" SET (security_invoker=false);
```

### Actual

```
No schema changes found
```

The old value (`true`) silently persists in the database. For `security_invoker` specifically, that means a view can remain `SECURITY DEFINER` after the user thinks they've moved it to `SECURITY INVOKER` — same risky failure mode as supabase/cli#3973, just on updates instead of creations.

### Root cause

The underlying diff engines used by `db diff` (djrobstep/migra, @pgkit/migra) do not read `pg_class.reloptions` when comparing views, so reloption-only differences are invisible to them. The supabase/cli#3973 fix paper-overs this for the create path by re-attaching reloptions from the live target catalog into emitted `CREATE VIEW` statements; there is no `CREATE VIEW` statement to patch in the update case.

### Possible directions

* Extend the post-processor to also diff source-vs-target reloptions and emit `ALTER VIEW ... SET (...)` / `RESET (...)` for reloption-only changes.
* Or, accept this as one more reason to move declarative-schema workflows over to pg-delta (per commit 6c19b38), and document the migra limitation in the meantime.

avallete:
This is natively handled by pg-delta as mentionned here: https://github.com/supabase/cli/issues/3973#issuecomment-4630009899

The way forward is to move from `migra` to `pg-delta` for diffing. Closing this one.

PR #5478, opened 2026-06-04
fix(cli): preserve view reloptions in db diff output

### LINKED ISSUE #3973: Declarative Database Schema - Views do NOT apply 'with(security_invoker=true)'
# Bug report

## Describe the bug

Using a declarative database schema to create a view with a 'security_invoker=true' permission setting, the security is not carried over into the migration file. Which makes using declarative schema's less useful and leaves Supabase environment open to permission / security issues.

## To Reproduce

1. Create file './schemas/view.sql'
2. Add the following:

-- Create a private schema
CREATE SCHEMA "private" AUTHORIZATION service_role;

GRANT USAGE ON SCHEMA "private" TO service_role;
GRANT USAGE ON SCHEMA "private" TO authenticated;

GRANT ALL ON ALL TABLES IN SCHEMA "private" TO authenticated, service_role;
GRANT ALL ON ALL ROUTINES IN SCHEMA "private" TO authenticated, service_role;
GRANT ALL ON ALL SEQUENCES IN SCHEMA "private" TO authenticated, service_role;

ALTER DEFAULT PRIVILEGES FOR ROLE postgres IN SCHEMA "private" GRANT ALL ON TABLES TO authenticated, service_role;
ALTER DEFAULT PRIVILEGES FOR ROLE postgres IN SCHEMA "private" GRANT ALL ON ROUTINES TO authenticated, service_role;
ALTER DEFAULT PRIVILEGES FOR ROLE postgres IN SCHEMA "private" GRANT ALL ON SEQUENCES TO authenticated, service_role;

ALTER ROLE authenticator SET pgrst.db_schemas TO 'public, graphql, private';

-- Create a private table
create table "private"."users" (
    "id" uuid,
    "created_at" timestamp with time zone not null default now(),
    "created_by" uuid not null default auth.uid(),
    "updated_at" timestamp with time zone,
    "updated_by" uuid,
    "deleted_at" timestamp with time zone,
    "deleted_by" uuid,
    "is_deleted" boolean not null default false,
    "version" smallint not null default 1,
    "name" text,
    "detail" jsonb
);

COMMENT ON TABLE "private"."users" IS 'User profile table.';

-- Create view
create or replace view "public"."user_details" with(security_invoker=true) 
as
  select
    users.id,
    users.name as name
  from
    "private".users;

3. Generate the migration sql using:

supabase db diff -f initial_setup

4. You should now have a migrations file in the "./migrations" folder. Open the file and you see:

create or replace view "public"."user_details" as  SELECT id,
    created_at,
    is_deleted,
    name,
    detail
   FROM private.users
  WHERE (is_deleted = false);

## Expected behavior

The view should be defined as:

create or replace view "public"."user_details" with(security_invoker=true) as  SELECT id,
    created_at,
    is_deleted,
    name,
    detail
   FROM private.users
  WHERE (is_deleted = false);

## System information

- OS: macOS
- Version of supabase-js: 2.33.9
- Version of Node.js: 22.14.0

## Additional context
Declarative schemas don't work for 'schema' definitions or 'permissions' consistently.

If I put the below into the declarative schema file - it does not translate into a migration file correctly. It should copy exactly when a db is reset, and empty. But it condenses all of this into a single 'if' statement.

Something isn't working quite right as a result.

>>
CREATE SCHEMA "private" AUTHORIZATION service_role;

GRANT USAGE ON SCHEMA "private" TO service_role;
GRANT USAGE ON SCHEMA "private" TO authenticated;

GRANT ALL ON ALL TABLES IN SCHEMA "private" TO authenticated, service_role;
GRANT ALL ON ALL ROUTINES IN SCHEMA "private" TO authenticated, service_role;
GRANT ALL ON ALL SEQUENCES IN SCHEMA "private" TO authenticated, service_role;

ALTER DEFAULT PRIVILEGES FOR ROLE postgres IN SCHEMA "private" GRANT ALL ON TABLES TO authenticated, service_role;
ALTER DEFAULT PRIVILEGES FOR ROLE postgres IN SCHEMA "private" GRANT ALL ON ROUTINES TO authenticated, service_role;
ALTER DEFAULT PRIVILEGES FOR ROLE postgres IN SCHEMA "private" GRANT ALL ON SEQUENCES TO authenticated, service_role;

ALTER ROLE authenticator SET pgrst.db_schemas TO 'public, graphql, private';

nirav-gondaliya:
## Summary

`supabase db diff` was silently dropping the `WITH (...)` clause on `CREATE VIEW` statements (`security_invoker`, `security_barrier`, `check_option`, plus reloptions on materialized views). The underlying diff engines (djrobstep/migra, @pgkit/migra) don't read `pg_class.reloptions`, so a declarative view defined as

```sql
create or replace view "public"."user_details"
with (security_invoker=true) as
select id, name from "private"."users";
```
ended up in the generated migration as create or replace view "public"."user_details" as SELECT ...

— silently turning a SECURITY INVOKER view into a SECURITY DEFINER one and bypassing RLS on the underlying table.

## What changed

- New helper apps/cli-go/internal/db/diff/view_options.go queries pg_class.reloptions on the diff target and splices a WITH (...) clause back into any CREATE [OR REPLACE] [MATERIALIZED] VIEW statement that lacks one. Failures (connection / query) are soft — the original diff is returned with a stderr warning so the diff is never lost.
- DiffDatabase calls the helper after the underlying diff engine returns, so the fix flows to every DiffFunc implementation.
- Removed the matching limitation from apps/cli-go/docs/supabase/db/diff.md.

## Scope

This PR closes the new-view case from #3973. A related gap surfaced during testing — migra doesn't detect reloption-only changes on existing views, so flipping security_invoker=true → false on an already-applied view produces No schema changes found. That is a separate migra limitation (the patcher only runs when there's a CREATE VIEW to patch) and is tracked separately at #5476.

## Why the fix is engine-agnostic

The patcher's regex matches view "x"."y"\s+as — bare as. Engines that already emit with (...) between the name and as (e.g. pg-schema-diff) do not match the regex and pass through unchanged.

Closes #3973

### DIFF
--- a/apps/cli-go/docs/supabase/db/diff.md
+++ b/apps/cli-go/docs/supabase/db/diff.md
@@ -12,4 +12,3 @@ While the diff command is able to capture most schema changes, there are cases w
 
 - Changes to publication
 - Changes to storage buckets
-- Views with `security_invoker` attributes
--- a/apps/cli-go/internal/db/diff/diff.go
+++ b/apps/cli-go/internal/db/diff/diff.go
@@ -211,7 +211,13 @@ func DiffDatabase(ctx context.Context, schema []string, config pgconn.Config, w 
 	} else {
 		fmt.Fprintln(w, "Diffing schemas...")
 	}
-	return differ(ctx, shadowConfig, config, schema, options...)
+	out, err := differ(ctx, shadowConfig, config, schema, options...)
+	if err != nil {
+		return out, err
+	}
+	// Restore view reloptions that the underlying diff engine dropped
+	// (e.g. WITH (security_invoker=true)). See issue #3973.
+	return applyViewReloptionsFromTarget(ctx, out, config, options...), nil
 }
 
 func migrateBaseDatabase(ctx context.Context, config pgconn.Config, migrations []string, fsys afero.Fs, options ...func(*pgx.ConnConfig)) error {
--- a/apps/cli-go/internal/db/diff/view_options.go
+++ b/apps/cli-go/internal/db/diff/view_options.go
@@ -1,0 +1,117 @@
+package diff
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"regexp"
+	"strings"
+
+	"github.com/jackc/pgconn"
+	"github.com/jackc/pgx/v4"
+	"github.com/supabase/cli/internal/utils"
+	"github.com/supabase/cli/pkg/pgxv5"
+)
+
+// viewDefinitionPattern matches a CREATE [OR REPLACE] [MATERIALIZED] VIEW
+// "schema"."name" AS prefix. Capture groups: 1 = schema, 2 = view name, 3 =
+// the trailing "AS" token, whose start offset is used to splice a WITH (...)
+// clause in front of it without touching the rest of the statement.
+var viewDefinitionPattern = regexp.MustCompile(`(?i)create\s+(?:or\s+replace\s+)?(?:materialized\s+)?view\s+(?:if\s+not\s+exists\s+)?"([^"]+)"\."([^"]+)"\s+(as\b)`)
+
+// SELECT_VIEW_RELOPTIONS reads reloptions for every view and materialized view
+// in the target database. djrobstep/migra, pg-schema-diff and @pgkit/migra
+// emit CREATE VIEW statements without the WITH (...) clause, so the CLI
+// reattaches it from this query after the diff is produced.
+//
+// See https://github.com/supabase/cli/issues/3973.
+const SELECT_VIEW_RELOPTIONS = `SELECT n.nspname AS nspname,
+       c.relname AS relname,
+       c.reloptions AS reloptions
+  FROM pg_class c
+  JOIN pg_namespace n ON c.relnamespace = n.oid
+ WHERE c.relkind IN ('v','m')
+   AND c.reloptions IS NOT NULL
+   AND array_length(c.reloptions, 1) > 0`
+
+type viewKey struct {
+	schema string
+	name   string
+}
+
+type viewReloptionsRow struct {
+	Nspname    string   `db:"nspname"`
+	Relname    string   `db:"relname"`
+	Reloptions []string `db:"reloptions"`
+}
+
+// PatchViewReloptions rewrites CREATE [OR REPLACE] [MATERIALIZED] VIEW
+// statements in sql to include a WITH (...) clause for any view present in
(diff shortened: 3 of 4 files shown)


avallete:
Hey there !

Thank's for your contribution. Moving forward we want to deprecate diff based on `migra`/`pgschema` in favor of diff based on `pg-delta` which already cover this case: https://github.com/supabase/cli/issues/3973#issuecomment-4630009899

Could you give it a try ? This code path will be removed bit by bit as users migrate and `pg-delta` mature.

nirav-gondaliya:
Hey @avallete , thanks for the pointer, **confirmed** `pg-delta` **handles this end-to-end**. I diffed two DBs where the target had:
  
- A regular view with `WITH (security_invoker=true, security_barrier=true, check_option=local)`
- A materialized view with `WITH (autovacuum_enabled=false)`

`supabase db diff --from <src> --to <tgt>` (which routes through `DiffPgDeltaRef`) emitted both `CREATE VIEW … WITH (…) AS …` and `CREATE MATERIALIZED VIEW … WITH (…) AS … WITH DATA` with every reloption preserved. So the catalog-level diff in pg-delta solves the original #3973 footgun cleanly — no SQL-regen patch needed on that path.

One open question before I close: pg-delta is still flagged experimental and `--use-migra` is the default. Until pg-delta is the default (or graduates from experimental), users on the stock `supabase db diff` will still silently lose `security_invoker` and fall through to `SECURITY DEFINER`, which is a real security regression rather than a cosmetic one. Would you want a minimal stopgap on the migra path for the security-relevant reloptions (`security_invoker`, `security_barrier`) until pg-delta is the default? Happy to either:
  
1. Trim this PR to just those two options on the migra/pgschema paths as a stopgap, or
2. Close it and rely entirely on pg-delta + the `--use-pg-delta` / `[experimental.pgdelta] enabled = true` docs.

Either works for me — your call on which fits the deprecation timeline better.

avallete:
Hey there ! Thanks for testing it out !

We plan to make delta the new default very soon, this issue/limitation with migra has been around for a long time. I think we can rely on pg-delta being pushed along as the new default.

nirav-gondaliya:
Sounds good — closing this in favor of pg-delta then. Thanks @avallete for the quick turnaround let me know if I can help with specific issues.

Closes without merge — fix is already covered by `pg-delta` (catalog-level diff preserves view reloptions natively). Users on the current default can opt in via `--use-pg-delta` or `[experimental.pgdelta] enabled = true` in `config.toml` until it becomes the default.

PR #5480, opened 2026-06-04
fix(cli): missing saml.id in sso list 

### LINKED ISSUE #5475: sso list fails with SchemaError in v2.102.0
### Confirmations
- [x] I confirm I have searched the Docs, GitHub Discussions, and Discord.

### Describe the bug
`supabase sso list` fails to parse the Management API response and exits with:

```
failed to list sso providers: SchemaError(Missing key
  at ["items"][0]["saml"]["id"])
```

The command never prints the provider list. The same project's providers are
returned as valid JSON by the underlying Management API endpoint
(`GET /v1/projects/{ref}/config/auth/sso/providers`), so the response is well-formed —
the CLI is rejecting it during deserialization because it treats `items[].saml.id`
as a required field.

### Steps to reproduce
1. Have a project with at least one existing SAML SSO provider.
2. Run: `npx supabase@latest sso list --project-ref <ref>`
3. Observe the SchemaError above.

For comparison, the raw API returns fine:
```bash
curl -s https://api.supabase.com/v1/projects/<ref>/config/auth/sso/providers \
  -H "Authorization: Bearer $SUPABASE_ACCESS_TOKEN" | jq .
```

### Version bisection
- ✅ Works: `2.100.0`, `2.101.0`
- ❌ Broken: `2.102.0`, `2.103.0`, `2.104.0`

The regression was introduced in **2.102.0**. Possibly related to the SSO command /
API response decoding being ported to native TypeScript with a schema that requires
`saml.id`, which the list response does not always include.

### Expected behavior
`sso list` parses the response and prints the providers, as it does on 2.101.0.

### Sample response shape (redacted)
`items[0]` from the raw API looks like this — note `saml.id` is <present/absent — FILL IN>:
```json
{
  "items": [
    {
      "id": "<uuid>",
      "saml": {
        "...": "... (redacted — include whether an `id` key exists here)"
      },
      "domains": [ { "domain": "<redacted>" } ]
    }
  ]
}
```

### Workaround
Pin to the last working version: `npx supabase@2.101.0 sso list --project-ref <ref>`.

### System information
- OS: macOS 26.5 (arm64)
- CLI: 2.102.0+ via `npx`
- Node: v24.10.0

7ttp:
## TL;DR

fixes `supabase sso list` failing when the Management API omits `items[].saml.id`

This regression was introduced in the native ts sso port, which decoded the list response more strictly than the API returns in practice
basically this makes `sso list` handle that response shape safely while keeping the existing behaviour..

## ref:
- closes https://github.com/supabase/cli/issues/5475

### DIFF
--- a/apps/cli/src/legacy/commands/sso/list/list.handler.ts
+++ b/apps/cli/src/legacy/commands/sso/list/list.handler.ts
@@ -1,4 +1,4 @@
-import type { SupabaseApiError } from "@supabase/api/effect";
+import { operationDefinitions } from "@supabase/api/effect";
 import { Effect, Option } from "effect";
 
 import { LegacyPlatformApi } from "../../../auth/legacy-platform-api.service.ts";
@@ -11,7 +11,7 @@ import {
   encodeToml,
   encodeYaml,
 } from "../../../shared/legacy-go-output.encoders.ts";
-import { mapLegacyHttpError } from "../../../shared/legacy-http-errors.ts";
+import { sanitizeLegacyErrorBody } from "../../../shared/legacy-http-errors.ts";
 import { LegacyLinkedProjectCache } from "../../../telemetry/legacy-linked-project-cache.service.ts";
 import { LegacyTelemetryState } from "../../../telemetry/legacy-telemetry-state.service.ts";
 import { legacySuggestUpgrade } from "../../../shared/legacy-upgrade-suggest.ts";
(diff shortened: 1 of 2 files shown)


apps/cli/src/legacy/commands/sso/list/list.handler.ts
```diff
@@ -11,7 +11,7 @@ import {
   encodeToml,
   encodeYaml,
 } from "../../../shared/legacy-go-output.encoders.ts";
-import { mapLegacyHttpError } from "../../../shared/legacy-http-errors.ts";
+import { sanitizeLegacyErrorBody } from "../../../shared/legacy-http-errors.ts";
```
7ttp, line 14: superseded / overlapped with https://github.com/supabase/cli/pull/5479 
 which seems like the better approach for this regression

PR #5491, opened 2026-06-05
fix(cli): Exclude partition children from pg-delta diffs

### LINKED ISSUE #5445: declarative sync drops partitions created by pg_partman
**Describe the bug**
When running `supabase db schema declarative sync` the migrations include drop statements for all partitions created by pg_partman.

**To Reproduce**
1. Create the following test-table:
```sql
CREATE EXTENSION pg_partman WITH SCHEMA extensions;

CREATE TABLE public.part_test (
  id uuid DEFAULT gen_random_uuid() NOT NULL,
  ts timestamp with time zone DEFAULT now() NOT NULL,
  PRIMARY KEY (ts, id)
)
PARTITION BY RANGE (ts);
```
2. Run `supabase db declarative sync` to set up declarative schemas.
3. Start partitioning with pg_partman:
```sql
SELECT extensions.create_parent(
  p_parent_table := 'public.part_test',
  p_control := 'ts',
  p_type := 'range',
  p_interval := '7 days',
  p_premake := 7
);
```
4. Run `supabase db schema declarative sync` to generate a new migration.
5. The migration includes drop tables for the partitions:
```
Generated migration SQL:
DROP TABLE public.part_test_default;
DROP TABLE public.part_test_p20260415;
DROP TABLE public.part_test_p20260422;
DROP TABLE public.part_test_p20260429;
DROP TABLE public.part_test_p20260506;
DROP TABLE public.part_test_p20260513;
DROP TABLE public.part_test_p20260520;
DROP TABLE public.part_test_p20260527;
DROP TABLE public.part_test_p20260603;
DROP TABLE public.part_test_p20260610;
DROP TABLE public.part_test_p20260617;
DROP TABLE public.part_test_p20260624;
DROP TABLE public.part_test_p20260701;
DROP TABLE public.part_test_p20260708;
DROP TABLE public.part_test_p20260715;
DROP TABLE public.part_test_p20260722;
```

**Expected behavior**
Partitions created by pg_partman should not be considered when creating statements.
Ideally I would like to be able to define a filter that get passed along to pg_delta.
Maybe as a command flag or even better in config.toml.

```bash
supabase db schema declarative sync --filter '{"not":{"table/is_partition":true}}'
```

**System information**
 - Version of OS: Ubuntu 24.04 (WSL)
 - Version of CLI: v2.104.0
 - Version of Docker: v29.4.2
 - Versions of services:
   - supabase/postgres: 17.6.1.084
   - supabase/gotrue: v2.188.1
   - postgrest/postgrest: v14.4
   - supabase/realtime: v2.102.1
   - supabase/storage-api: v1.43.3
   - supabase/edge-runtime: v1.74.0
   - supabase/studio: 2026.05.25-sha-65c570e
   - supabase/postgres-meta: v0.96.6
   - supabase/logflare: 1.42.0
   - supabase/supavisor: 2.9.5

**Potential workaround**
I could put the partitioning in a seed file and do the following:
```bash
supabase db reset --no-seed # reset database without seeding
supabase db schema declarative sync --no-apply # generate migrations without partman tables
supabase db reset # apply the new migration and seed
```

This would work fine as a workaround for now.
On prod i would just start the partitioning once manually.

mittal-parth:

Bug fix

Declarative sync compares migrations-side and declarative-side catalogs. When runtime partition children exist on the source (for example from pg_partman) but declarative files only define the parent table, pg-delta proposed DROP TABLE statements for every child partition.

## What is the new behavior?

This change adds a default FilterDSL rule into the embedded pg-delta scripts to ignore tables with table/is_partition: true. Parent partitioned tables remain in scope; operational child partitions are left out of diff and declarative export plans.

## Additional context

Closes #5445

### DIFF
--- a/apps/cli-go/internal/db/diff/pgdelta_test.go
+++ b/apps/cli-go/internal/db/diff/pgdelta_test.go
@@ -2,10 +2,31 @@ package diff
 
 import (
 	"runtime"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
+
+const partitionFilterSnippet = `"table/is_partition": true`
+
+func TestPgDeltaTemplatesExcludePartitionChildren(t *testing.T) {
+	t.Run("pgdelta.ts applies default partition filter", func(t *testing.T) {
+		assert.Contains(t, pgDeltaScript, partitionFilterSnippet)
+		assert.Contains(t, pgDeltaScript, "partitionFilter")
+	})
+
+	t.Run("pgdelta_declarative_export.ts applies default partition filter", func(t *testing.T) {
+		assert.Contains(t, pgDeltaDeclarativeExportScript, partitionFilterSnippet)
+		assert.Contains(t, pgDeltaDeclarativeExportScript, "partitionFilter")
+	})
+
+	t.Run("partition filter is always composed into supabase.filter", func(t *testing.T) {
+		for _, script := range []string{pgDeltaScript, pgDeltaDeclarativeExportScript} {
+			assert.True(t, strings.Contains(script, "filterParts = [supabase.filter!, partitionFilter]"))
+		}
+	})
+}
 
 func TestContainerRef(t *testing.T) {
 	t.Run("passes empty string through", func(t *testing.T) {
--- a/apps/cli-go/internal/db/diff/templates/pgdelta.ts
+++ b/apps/cli-go/internal/db/diff/templates/pgdelta.ts
@@ -19,17 +19,21 @@ async function resolveInput(ref: string | undefined) {
 const source = Deno.env.get("SOURCE");
 const target = Deno.env.get("TARGET");
 
+// Runtime partition children (e.g. pg_partman) are operational state, not declarative schema.
+const partitionFilter = { not: { "table/is_partition": true } };
+
 const includedSchemas = Deno.env.get("INCLUDED_SCHEMAS");
+const filterParts = [supabase.filter!, partitionFilter];
 if (includedSchemas) {
   const schemas = includedSchemas.split(",");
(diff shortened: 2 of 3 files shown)


avallete:
Thanks for digging into this @mittal-parth — the diagnosis of where the spurious `DROP`s come from is exactly right. We don't want to land the `relispartition`-based filter though, because it's too coarse and introduces a worse regression than the bug it fixes:

`{ not: { "table/is_partition": true } }` is applied to **every** change (it filters `Change` objects, symmetrically), so it doesn't just suppress the unwanted child `DROP`s — it also drops the **CREATE** and **ALTER** changes for partitions. That means a user who manages partitions declaratively (`CREATE TABLE foo_2024 PARTITION OF foo …`) would have those partitions silently never created, and a genuinely-intended partition drop would be silently suppressed. `relispartition` can't tell an operational pg_partman child apart from a user-declared partition, so any rule keyed on it will misfire one way or the other.

We also can't lean on `pg_depend` here: the table extractor already filters anything with a `pg_depend` edge to `pg_extension`, and pg_partman's children survive that — which proves they carry no extension dependency. The only authoritative signal for "this is an operational partman child" is pg_partman's own `part_config` table (a child is managed iff its `pg_inherits` parent is registered there).

So the correct fix lives in pg-delta's Supabase integration as a small pg_partman-aware handler (read `part_config` at extract time, exclude only the children of managed parents), not as a catalog filter in these templates. We're tracking that as part of the extensions-diffing work and will make sure #5445 is covered there. I'm going to close this PR in favor of that approach, but thank you — the repro and root-cause here directly shaped the design. 🙏

mittal-parth:
Got it, thanks @avallete!

PR #5493, opened 2026-06-05
fix(cli): auto-retry db dump/pull via the IPv4 pooler on IPv6-only networks

avallete:
Closes [CLI-1593](https://linear.app/supabase/issue/CLI-1593/improve-db-dump-ipv6-error-guidance)

## What

`supabase db dump` and `db pull` run `pg_dump` inside a Docker container. Supabase direct database hosts (`db.<ref>.supabase.co:5432`) are **IPv6-only** unless the IPv4 add-on is enabled, so on environments without working IPv6 in the container (very common on Docker Desktop for macOS) the operation failed with an opaque `error running container: exit 1`.

This PR makes that path **self-healing**: when a remote dump/pull fails because the direct host is unreachable over IPv6, the CLI transparently resolves the project's **IPv4 transaction pooler**, warns the user, and retries once. If no pooler is available it falls back to an actionable error message pointing at `--db-url`.

## Why

The host running the CLI often *does* have IPv6 (so the pre-flight dial succeeds and the direct config is selected), but the `pg_dump` **container** does not — so the failure only surfaces deep inside the container as a libpq/getaddrinfo error, hidden behind the generic container exit code. Users were left stuck with no hint, even though a working IPv4 pooler existed for their project.

## Behavior

```mermaid
flowchart TD
    A["db dump / db pull (remote)"] --> B["Run pg_dump in Docker container<br/>(tee stderr for classification)"]
    B --> C{Succeeded?}
    C -->|yes| OK["Write dump ✓"]
    C -->|no| D{"stderr is an<br/>IPv6 connectivity error?"}
    D -->|no| SUG["Classify error → actionable suggestion"]
    D -->|yes| E{"Host is a direct<br/>db.&lt;ref&gt;.supabase.co?"}
    E -->|no| SUG
    E -->|yes| F{"IPv4 pooler<br/>config resolvable?"}
    F -->|no| SUG2["Suggest --db-url with the<br/>transaction pooler URL"]
    F -->|yes| G["Warn user · reset output ·<br/>retry once via IPv4 pooler"]
    G --> H{Retry succeeded?}
    H -->|yes| OK2["Write dump ✓<br/>(transparent recovery)"]
    H -->|no| SUG3["Classify retry error → suggestion"]
```

Happy-path auto-recovery (linked project, host has IPv6, container does not):

```mermaid
sequenceDiagram
    actor U as User
    participant CLI as supabase db dump
    participant C as pg_dump container
    participant API as link cache / Management API
    U->>CLI: db dump (linked → direct host)
    CLI->>C: pg_dump → db.ref.supabase.co:5432 (IPv6)
    C-->>CLI: error: "No address associated with hostname"<br/>/ "Network is unreachable" (no IPv6 in container)
    Note over CLI: classify captured stderr → IPv6 connectivity error
    CLI->>API: resolve IPv4 transaction pooler + login role
    API-->>CLI: pooler config (port 5432)
    CLI-->>U: ⚠ Warning: retrying via the IPv4 connection pooler
    CLI->>C: pg_dump → aws-0-…pooler.supabase.com (IPv4)
    C-->>CLI: dump output
    CLI-->>U: dump written ✓
```

## How

- **`internal/db/dump/pooler_fallback.go` — `RunWithPoolerFallback`** wraps the Docker-backed `pg_dump` operations. It runs the closure with an stderr-capturing exec; on failure it classifies the captured stderr and, if it's an IPv6 error against a direct host with a resolvable pooler, warns, resets the output, and retries once via the pooler. `resetOutput` rewinds the destination between attempts (`bytes.Buffer.Reset`, file `Truncate`+`Seek`, stdout ignored) so a partial first attempt isn't left behind. `--dry-run` skips the wrapper entirely.
- **`internal/db/dump/dump.go` + `internal/db/pull/pull.go`** route their remote dump paths through `RunWithPoolerFallback` (dump data/role/schema; pull's experimental role+schema dump and `dumpRemoteSchema`).
- **`internal/utils/flags/db_url.go` — `ResolvePoolerConfigForFallback`** returns an authenticated IPv4 transaction-pooler config: it prefers the pooler URL persisted at `supabase link` time, otherwise fetches it from the Management API, forces the transaction port, and authenticates via `SUPABASE_DB_PASSWORD` or a temporary login role. It's injected through a package variable so tests can stub the network call.
- **`internal/utils/connect.go`** — detection (`isIPv6ConnectivityError`) covers `Address family for hostname not supported`, `No address associated with hostname`, `Network is unreachable`, and (gated on an IPv6 literal so genuine project-not-found / tenant errors keep their own hint) `No route to host` / `Cannot assign requested address`. The IPv6-literal regex matches both Go's bracketed `[…]` and libpq's parenthesised `(…)` forms. `ProjectRefFromDirectDbHost`, `WarnIPv6PoolerFallback`, and the existing `SetConnectSuggestion` / `SuggestIPv6Pooler` provide ref extraction, the retry warning, and the non-recoverable suggestion.

## Non-recoverable fallback (message only)

When auto-retry isn't possible (not an IPv6 error, not a direct host, or no pooler), the command still fails — but with guidance instead of a bare exit code:

```
Your network does not support IPv6, which is required for direct connections to the database.
Retry through the IPv4 transaction pooler by passing it to --db-url "postgres://postgres.<ref>:[YOUR-PASSWORD]@aws-0-<region>.pooler.supabase.com:6543/postgres"
```

## Tests

- `dump_test.go`: auto-retry succeeds via the pooler (asserts warning, output truncation/rewrite, no leftover suggestion); IPv6 failure with no pooler still surfaces the suggestion; `Cannot assign requested address` classification.
- `connect_test.go`: detection matrix incl. the new signatures and `ProjectRefFromDirectDbHost`; `SuggestIPv6Pooler` enrichment.
- `db_url_test.go`: `ResolvePoolerConfigForFallback` (persisted-URL vs Management API resolution).

## Notes

- These `db` commands are still proxied to the bundled Go binary, so the fix lives in `apps/cli-go`.
- The auto-retry only triggers for **direct Supabase hosts** — explicit `--db-url`/`--local` targets are never silently rerouted.

https://claude.ai/code/session_01UaPk7dGPmiCqoKJHyV7SLz


jgoux (commented):
I found two behavior gaps worth tightening before this lands. The targeted and full `apps/cli-go` Go test suites pass locally, so these are about uncovered command behavior rather than current test failures.

apps/cli-go/internal/db/pull/pull.go
```diff
@@ -151,7 +154,9 @@ func dumpRemoteSchema(ctx context.Context, path string, config pgconn.Config, fs
 		return errors.Errorf("failed to open dump file: %w", err)
 	}
 	defer f.Close()
-	return migration.DumpSchema(ctx, config, f, dump.DockerExec)
+	return dump.RunWithPoolerFallback(ctx, config, f, false, func(ctx context.Context, config pgconn.Config, out io.Writer, exec migration.ExecFunc) error {
```
jgoux, line 165: This wraps the initial `pg_dump` pass, but the normal `db pull` flow still later calls `diffRemoteSchema` with the original direct config (`run` calls it after `dumpRemoteSchema`, and non-initial pulls call it directly). That diff path ultimately runs inside Docker/Edge Runtime with the direct host as `TARGET`, so the same “host has IPv6 but the container does not” failure can still surface after this fallback succeeds. I think the retry needs to cover the diff pass too, or `db pull` will only be self-healing for part of the workflow.

apps/cli-go/internal/db/dump/dump.go
```diff
@@ -36,15 +34,25 @@ func Run(ctx context.Context, path string, config pgconn.Config, dataOnly, roleO
 	if utils.IsLocalDatabase(config) {
 		db = "local"
 	}
-	if dataOnly {
-		fmt.Fprintf(os.Stderr, "Dumping data from %s database...\n", db)
-		return migration.DumpData(ctx, config, outStream, exec, opts...)
-	} else if roleOnly {
-		fmt.Fprintf(os.Stderr, "Dumping roles from %s database...\n", db)
-		return migration.DumpRole(ctx, config, outStream, exec, opts...)
+	return RunWithPoolerFallback(ctx, config, outStream, dryRun, func(ctx context.Context, config pgconn.Config, out io.Writer, exec migration.ExecFunc) error {
```
jgoux, line 37: Because `dump.Run` always enables the fallback and `RunWithPoolerFallback` only checks whether `config.Host` looks like `db.<ref>.supabase.co`, this also applies to explicit `--db-url` inputs. In that case a user-provided direct URL can be silently retried with a different pooler config, and if no persisted pooler URL exists the resolver reaches the Management API via `GetSupabase()`, which can fatally exit when the user is not logged in. That seems to break the supported no-login `--db-url` flow and contradicts the PR note that explicit targets are never silently rerouted.

avallete: Good catch — fixed in 50f9be2. The auto-retry is now gated to the linked path: `ParseDatabaseConfig` sets `PoolerFallbackEligible = connType == linked`, and the new `PoolerFallbackConfig` helper refuses to reroute unless that's true. So explicit `--db-url` (and `--local`) are never silently rerouted, and the no-login `--db-url` flow can't reach `GetSupabase()` / fatally exit anymore — it falls through to the existing classification (`SetConnectSuggestion` + the token-guarded `SuggestIPv6Pooler`).

---

apps/cli-go/internal/db/pull/pull.go
avallete: Right, the diff pass was uncovered — fixed in 50f9be2. Both `diffRemoteSchema` (migra/pg-delta) and `pullDeclarativePgDelta` now retry through the IPv4 pooler via the shared `dump.PoolerFallbackConfig` helper when the direct host fails over IPv6. The diff/declarative paths run through `RunEdgeRuntimeScript`/migra, which already embed the container stderr in the returned error, so `IsIPv6ConnectivityError` classifies it directly (no extra stderr capture needed). So an initial `db pull` self-heals across both the dump and the diff pass, and the regular (non-initial) pull self-heals on the diff pass too.

One trade-off worth flagging: in the initial *legacy* pull, the dump and the diff each detect/resolve the pooler independently, so you'd see the fallback warning (and a temp login-role resolution) twice. I kept it per-call for simplicity/correctness rather than threading the resolved config through `run()`; happy to thread it if you'd prefer to dedupe.

---

apps/cli-go/internal/db/dump/pooler_fallback.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+		return nil
+	}
+	// Could not auto-recover: classify the failure into an actionable suggestion.
+	utils.SetConnectSuggestion(connErr)
+	if utils.IsIPv6ConnectivityError(connErr) {
+		// Enrich the hint with the project's actual transaction pooler URL so the
+		// user gets a copy-pasteable --db-url.
+		utils.SuggestIPv6Pooler(ctx, config.Host)
+	}
```
avallete, line 64: **issue**

This could leak the password to a terminal, we don't want that.

avallete: Agreed — fixed in 288f5cc. `SuggestIPv6Pooler` was rendering the Management API's `connection_string` verbatim, which can carry a real password. `ipv6PoolerSuggestion` now masks the userinfo password with the `[YOUR-PASSWORD]` placeholder before printing (via `maskPoolerPassword`), so the credential is never echoed to the terminal while the hint stays copy-pasteable. Added a test that feeds a real password through the API mock and asserts it's absent from the suggestion.

---

jgoux (approved):

PR #5500, opened 2026-06-07
fix(diff): include view reloptions changes

### LINKED ISSUE #5476: db diff misses reloption-only changes on existing

pjpjq:
## Summary

- Compare source and target view `pg_class.reloptions` for non-pg-delta `db diff`.
- Generate `ALTER VIEW ... SET/RESET (...)` for reloption-only changes on existing views.
- Add regression coverage for regular views, materialized views, schema filtering, and the connected diff path.

Fixes #5476.

## Tests

- `cd apps/cli-go && go test ./internal/db/diff ./internal/db/pull ./cmd`

### DIFF
--- a/apps/cli-go/internal/db/diff/diff.go
+++ b/apps/cli-go/internal/db/diff/diff.go
@@ -257,6 +257,9 @@ func DiffDatabase(ctx context.Context, schema []string, config pgconn.Config, w 
 	if err != nil {
 		return DatabaseDiff{}, err
 	}
+	if !usePgDelta {
+		output = appendViewReloptionDiff(ctx, output, shadowConfig, config, schema, options...)
+	}
 	return DatabaseDiff{SQL: output}, nil
 }
 
--- a/apps/cli-go/internal/db/diff/view_reloptions.go
+++ b/apps/cli-go/internal/db/diff/view_reloptions.go
@@ -1,0 +1,198 @@
+package diff
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"sort"
+	"strings"
+
+	"github.com/jackc/pgconn"
+	"github.com/jackc/pgx/v4"
+	"github.com/supabase/cli/internal/utils"
+	"github.com/supabase/cli/pkg/pgxv5"
+)
+
+const SELECT_VIEW_RELOPTIONS = `SELECT n.nspname AS nspname,
+       c.relname AS relname,
+       c.relkind::text AS relkind,
+       COALESCE(c.reloptions, ARRAY[]::text[]) AS reloptions
+  FROM pg_class c
+  JOIN pg_namespace n ON c.relnamespace = n.oid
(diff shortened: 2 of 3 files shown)


SahilArate:
Nice approach the post-processor pattern is the right call here given migra's blind spot on pg_class.reloptions.
A few thoughts that might help before merge:
1. No-op guard: if both source and target have reloptions = NULL, the diff should be a clean no-op. Worth adding a test case to confirm there's no spurious SET () emitted.
2. Dropped view edge case: if a view exists in target but not source (i.e. it's being dropped), the post-processor shouldn't emit an ALTER VIEW SET for it — the drop statement from migra already handles it. A guard on view existence in both catalogs would be defensive here.
3. Multiple reloptions at once: if security_invoker and another reloption both change in the same diff, the generated SQL should batch them: ALTER VIEW foo SET (security_invoker=false, check_option=cascaded) rather than two separate statements.
4. RESET vs SET: when a reloption is removed entirely (not just changed), RESET is the correct SQL. Worth verifying the test covers the full remove case, not just value swap.
Happy to help test or iterate on any of these!

avallete:
Hi there !

Thank's for contributing, but given the amount of changes it requires to support this option for a deprecated diff engine I don't think we'll want to get this merged.

The new recommended approach is to use `pg-delta` diff engine which handle this kind of things natively.

PR #5507, opened 2026-06-08
fix: skip unconfigured vector bucket seed for linked projects

### LINKED ISSUE #5506: seed buckets calls Vector Buckets API and fails with FeatureNotEnabled when storage.vector is not configured
## Bug report

`supabase seed buckets --linked --yes` fails on a linked hosted project when Vector Buckets are not enabled for the tenant, even though the project only declares regular Storage buckets and does not use Vector Buckets.

This started after upgrading/pinning the CLI to `2.105.0`. It appears related to Vector Buckets support added in #5230.

## Environment

- Supabase CLI: `2.105.0`
- Command: `supabase seed buckets --linked --yes --debug`
- Project type: hosted Supabase project
- Vector Buckets: not used / not enabled for tenant
- Regular Storage buckets are declared in `supabase/config.toml`

## Config

The project had regular Storage bucket config only:

```toml
[storage]
enabled = true
file_size_limit = "50MiB"

[storage.buckets.test]
public = false
objects_path = "./storage/test"
```

No `[storage.vector]` section was configured.

## Actual behavior

The CLI updates the regular Storage bucket, then calls the Vector Buckets API and fails with `FeatureNotEnabled`:

```text
Supabase CLI 2.105.0
HTTP GET https://<project-ref>.supabase.co/storage/v1/bucket
Updating Storage bucket: test
HTTP PUT https://<project-ref>.supabase.co/storage/v1/bucket/test
Updating vector buckets...
HTTP POST https://<project-ref>.supabase.co/storage/v1/vector/ListVectorBuckets
Error status 403: {"error":"FeatureNotEnabled","statusCode":"403","message":"feature not enabled for this tenant"}
```

## Expected behavior

If the project does not configure or use Vector Buckets, `supabase seed buckets` should not call Vector Buckets endpoints. It should only seed regular Storage buckets.

Alternatively, if Vector Buckets require explicit tenant enablement, the CLI should treat missing `[storage.vector]` as disabled for hosted projects.

## Workaround

Adding an explicit disabled vector config avoids the Vector Buckets API call and allows `seed buckets` to succeed:

```toml
[storage.vector]
enabled = false
max_buckets = 10
max_indexes = 5
```

After adding this section, the same command succeeds and only regular Storage bucket/object endpoints are called.

## Notes

The CLI source appears to default `Storage.VectorBuckets.Enabled` to true when config is optional or not explicitly set, which makes `seed buckets` enter the Vector Buckets path even for projects that only use regular Storage buckets.

addniner:
## Summary
- skip vector bucket seeding for linked projects when no vector buckets are configured
- keep local vector bucket seeding and configured remote vector buckets unchanged

Fixes #5506

## Testing
- go test ./internal/seed/buckets

### DIFF
--- a/apps/cli-go/internal/seed/buckets/buckets.go
+++ b/apps/cli-go/internal/seed/buckets/buckets.go
@@ -11,9 +11,11 @@ import (
 )
 
 func Run(ctx context.Context, projectRef string, interactive bool, fsys afero.Fs) error {
+	shouldSeedVectorBuckets := utils.Config.Storage.VectorBuckets.Enabled &&
+		(len(projectRef) == 0 || len(utils.Config.Storage.VectorBuckets.Buckets) > 0)
 	if len(projectRef) == 0 &&
 		len(utils.Config.Storage.Buckets) == 0 &&
-		!utils.Config.Storage.VectorBuckets.Enabled {
+		!shouldSeedVectorBuckets {
 		return nil
 	}
 	api, err := client.NewStorageAPI(ctx, projectRef)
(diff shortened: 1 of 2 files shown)


jgoux:
Sorry just saw your PR now. It was fixed [here](https://github.com/supabase/cli/pull/5508).

PR #5509, opened 2026-06-08
fix(functions): add apikey compatibility header

kallebysantos:

Bug fix

Currently, the API proxy is overwriting the `Authorization` header when forwarding to `/functions`

## What is the new behavior?

Uses a custom `sb-api-key` header to handle the minted jwt

## Additional context

Towards [SDK-1153](https://linear.app/supabase/issue/SDK-1153/cli-use-a-custom-header-for-minted-jwts)

### DIFF
--- a/apps/cli-go/internal/functions/serve/templates/main.ts
+++ b/apps/cli-go/internal/functions/serve/templates/main.ts
@@ -111,7 +111,10 @@ function getAuthToken(req: Request) {
   const authHeader = req.headers.get("authorization");
   const sbApiKeyCompatibilityToken = req.headers.get("sb-api-key")
 
-  if (!authHeader && !sbApiKeyCompatibilityToken) {
+  // NOTE:(kallebysantos) Kong on legacy CLI stack pass it down as 'Bearer Token' format
+  const cleanSbApiKeyCompatibilityToken = sbApiKeyCompatibilityToken.replace('Bearer', '').trim()
+
+  if (!authHeader && !cleanSbApiKeyCompatibilityToken) {
     throw new Error("Missing authorization header");
   }
 
@@ -120,7 +123,7 @@ function getAuthToken(req: Request) {
   // - Original bearer is not present or is ApiKey
   const bearerToken = extractBearerToken(authHeader ?? '')
   const token = (!bearerToken || bearerToken.startsWith('sb_'))
-    ? sbApiKeyCompatibilityToken
+    ? cleanSbApiKeyCompatibilityToken
     : bearerToken
 
   if (!token) {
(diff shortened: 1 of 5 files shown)


apps/cli-go/internal/start/templates/kong.yml
```diff
@@ -182,10 +182,10 @@ services:
         config:
           add:
             headers:
-              - "Authorization: {{ .BearerToken }}"
+              - "sb-api-key: {{ .BearerToken }}"
```
avallete, line 185: **remark**(not-blocking):

this mints on every /functions request regardless of verify_jwt. Harmless now that it goes to sb-api-key and is stripped before the worker (and only read when verify_jwt is on), but it's wider than necessary could scope it later. (Underlying expression is in  https://github.com/supabase/cli/blob/develop/apps/cli-go/internal/start/start.go#L451-L463 )

apps/cli-go/internal/functions/serve/templates/main.ts
```diff
@@ -96,15 +96,40 @@ const functionsConfig: Record<string, FunctionConfig> = (() => {
...
+  return token
+}
+
 function getAuthToken(req: Request) {
   const authHeader = req.headers.get("authorization");
-  if (!authHeader) {
+  const sbApiKeyCompatibilityToken = req.headers.get("sb-api-key")
+
+  // NOTE:(kallebysantos) Kong on legacy CLI stack pass it down as 'Bearer Token' format
+  const cleanSbApiKeyCompatibilityToken = sbApiKeyCompatibilityToken.replace('Bearer', '').trim()
```
avallete: **nitpick**

`sbApiKeyCompatibilityToken` is `string | null` (headers.get returns null when absent), so .replace(...) throws if sb-api-key is missing. Masked today only because Kong always sets the header on this route but a request reaching the runtime without it (direct-to-runtime, tests) would 500. A `??`  guard would match the ingress behavior.

avallete (approved):
With the latest commit this looks like it resolves the original bug for the CLI stackAuthorization is preserved, the mint moves to sb-api-key, and it's now stripped before the user worker (prepareUserRequest → worker.fetch(userReq)). Two small nits inline.

apps/cli-go/internal/functions/serve/templates/main.ts
kallebysantos: Yha! I noticed it after pushing the last commit...
I'm already fixing it, and I also figured out that [edge-runtime-main.ts](https://github.com/supabase/cli/blob/develop/packages/stack/src/services/edge-runtime-main.ts) needs to be updated as well 😅

kallebysantos: Hey @avallete, I did already `null` guard the mentioned line as well updated the `edge-runtime-main.ts` to match the new feature.

> [!WARNING]  
> I noticed that `edge-runtime-main.ts` doesn't support the Asymmetric JWT 
> The following PRs should be replicated for the TS version as well:
> - https://github.com/supabase/cli/pull/4721
> - https://github.com/supabase/cli/pull/4985

avallete: Tracked here: supabase/cli#5560

PR #5513, opened 2026-06-08
chore(cli): track output_format and fix is_agent CI parity

pamelachia:
## Summary

Adds an `output_format` property to the `cli_command_executed` telemetry event and fixes a Go/TS divergence in how `is_agent` is computed. Together these make the JSON-vs-human output choice measurable and ensure both CLI shells agree on what counts as an agent. This is the measurement-and-parity groundwork for auto-switching agents to JSON output (next subissue under GROWTH-806).

## Changes

- **Record the resolved output format on every `cli_command_executed`.** Added `output_format` to the Go catalog (`events.go`) and its TS mirror (`event-catalog.ts`), and emit it from the Go capture and both TS instrumentation paths. `db query` resolves its format into a command-local flag, so it now mirrors the resolved value onto the global that telemetry reads, keeping the value accurate for the highest-volume agent command. The property does not exist today, so current JSON usage is unmeasurable without it.
- **Fix `is_agent` Go/TS parity.** The TS legacy analytics layer folded CI into `is_agent` (`is_agent = aiTool || isCi`), tagging CI environments as agents while the Go binary does not. `is_agent` now reflects coding-agent detection only, matching Go on both shells. `is_ci` continues to be reported as its own separate property.

## Notes for reviewers

- `output_format` values are shell-native (Go: `pretty|json|yaml|toml|env`; TS: `text|json|stream-json`), disambiguated by `$lib`; normalize in analysis.
- Deliberately out of scope: aligning the CI env-var lists across the two shells. `runtime.isCi` gates behavior in `legacy-platform-api.layer.ts`, so changing those lists would alter non-interactive behavior for some CI environments. That belongs in a separate change, not this telemetry-parity fix.
- The three instrumentation/platform-api unit tests now provide a mock `Output` because the instrumentation reads `output.format`.

## Linear

- fixes GROWTH-912
- parent GROWTH-806

### DIFF
--- a/apps/cli-go/cmd/db.go
+++ b/apps/cli-go/cmd/db.go
@@ -308,6 +308,9 @@ without the envelope.`,
 					outputFormat = "table"
 				}
 			}
+			// db query resolves --output into a command-local flag, so mirror the
+			// resolved value onto the global that telemetry's output_format reads.
+			utils.OutputFormat.Value = outputFormat
 			if flag := cmd.Flags().Lookup("linked"); flag != nil && flag.Changed {
 				return query.RunLinked(cmd.Context(), sql, flags.ProjectRef, outputFormat, agentMode, os.Stdout)
 			}
--- a/apps/cli-go/cmd/root.go
+++ b/apps/cli-go/cmd/root.go
@@ -175,8 +175,9 @@ func Execute() {
 		if service := telemetry.FromContext(executedCmd.Context()); service != nil {
 			ensureProjectGroupsCached(executedCmd.Context(), service)
 			_ = service.Capture(executedCmd.Context(), telemetry.EventCommandExecuted, map[string]any{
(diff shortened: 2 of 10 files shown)


jgoux:
Thanks Pam. I pushed a small follow-up to fix the legacy TS telemetry path.

The issue was that several legacy commands emit Go-style machine output via `--output json|yaml|toml|env`, but they intentionally keep the TS `Output` service in text mode so raw output stays byte-compatible with Go. That meant reading `output.format` would report `text` even when the command actually emitted JSON/YAML/TOML/ENV.

The follow-up now resolves `output_format` from the legacy `--output` flag when it is a Go machine format, and falls back to the TS output format otherwise. I kept `--output pretty` as a fallback case because it defers to the active TS output mode, e.g. `--output pretty --output-format json`.

jgoux (approved):

PR #5553, opened 2026-06-11
fix(cli): avoid auth for local typegen and services

jgoux:
Fixes CLI-1619.

This removes eager Management API client construction from command paths that have tokenless behavior:

- `gen types --local` and `--db-url` now use a lean runtime and lazily construct the platform API client only for linked/project-id generation.
- `services` now uses a lean runtime so it can always print local service versions, while keeping the linked-version lookup optional when a token is available.
- Command-wiring regression coverage exercises tokenless `gen types --local` and `services` invocations through the actual CLI command layers.

### DIFF
--- a/apps/cli/src/legacy/commands/gen/types/types.handler.ts
+++ b/apps/cli/src/legacy/commands/gen/types/types.handler.ts
@@ -3,11 +3,12 @@ import { ChildProcess, ChildProcessSpawner } from "effect/unstable/process";
 import { Effect, FileSystem, Option, Path, Stdio, Stream } from "effect";
 import { LegacyDebugFlag, LegacyNetworkIdFlag } from "../../../../shared/legacy/global-flags.ts";
 import { Output } from "../../../../shared/output/output.service.ts";
-import { LegacyPlatformApi } from "../../../auth/legacy-platform-api.service.ts";
 import { LegacyCliConfig } from "../../../config/legacy-cli-config.service.ts";
 import { LegacyProjectNotLinkedError } from "../../../config/legacy-project-ref.errors.ts";
-import { PROJECT_NOT_LINKED_MESSAGE } from "../../../config/legacy-project-ref.service.ts";
-import { LegacyProjectRefResolver } from "../../../config/legacy-project-ref.service.ts";
+import {
+  LegacyProjectRefResolver,
+  PROJECT_NOT_LINKED_MESSAGE,
+} from "../../../config/legacy-project-ref.service.ts";
 import { mapLegacyHttpError } from "../../../shared/legacy-http-errors.ts";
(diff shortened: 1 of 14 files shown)


apps/cli/src/legacy/commands/gen/types/types.handler.ts
```diff
@@ -225,6 +241,64 @@ export const legacyGenTypes = Effect.fn("legacy.gen.types")(function* (flags: Le
...
+      if (Option.isSome(fileValue)) {
+        return yield* assertValidProjectRef(fileValue.value);
+      }
+      if (tty.stdinIsTty && output.interactive) {
+        return yield* promptForProjectRef().pipe(Effect.flatMap(assertValidProjectRef));
+      }
+      return yield* Effect.fail(
+        new LegacyProjectNotLinkedError({ message: PROJECT_NOT_LINKED_MESSAGE }),
+      );
+    });
```
avallete: **question**

Isn't that a duplicate of similar code somewhere else ? Especially the `resolveProjectRef` seems like something that is used in other places.

Coly010 (approved):

avallete (approved):

apps/cli/src/legacy/commands/gen/types/types.handler.ts
jgoux: Yes, this was duplicated. I moved the ownership back to the shared `LegacyProjectRefResolver` by adding a shared `LegacyPlatformApiFactory`, so the resolver can still create the platform API lazily only when it needs to prompt. `gen types` now calls `projectRef.resolve(...)` instead of carrying its own `resolveProjectRef` copy.

apps/cli/src/legacy/commands/gen/types/types.layers.ts
```diff
@@ -0,0 +1,97 @@
...
+      const telemetryRuntime = yield* TelemetryRuntime;
+
+      return LegacyGenTypesPlatformApi.of({
+        make: legacyMakePlatformApi.pipe(
+          Effect.provideService(Analytics, analytics),
+          Effect.provideService(LegacyCliConfig, cliConfigService),
+          Effect.provideService(LegacyCredentials, credentialsService),
+          Effect.provideService(LegacyDebugLogger, debugLogger),
+          Effect.provideService(FileSystem.FileSystem, fs),
+          Effect.provideService(HttpClient.HttpClient, http),
```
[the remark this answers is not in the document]
jgoux: Fixed. The `gen types` lazy platform API factory now provides `FetchHttpClient.layer` directly when it builds typed Management API clients, matching `legacyManagementApiRuntimeLayer`. The debug-wrapped `legacyHttpClientLayer` is still exposed for the other services that need it, but typed API calls no longer receive a pre-wrapped client.

PR #5560, opened 2026-06-12
feat(functions): add hybrid JWT verification to TS edge runtime

avallete:
Port the asymmetric JWT support from the Go CLI (supabase/cli#4721, #4985) to
the TypeScript edge runtime. The runtime now verifies ES256/RS256 tokens
against a JWKS, falling back to the legacy symmetric secret for HS256, and
exposes the JWKS to user workers via `SUPABASE_JWKS`.

## How this differs from the Go implementation

This is **not** a 1:1 port. The Go CLI verifies tokens with a JWT library,
but the edge runtime's auth gate lives in the main worker script that the CLI
injects as source into the runtime — it can't rely on bundled npm
dependencies. So verification here is implemented by hand directly on
**WebCrypto (`crypto.subtle`)**, with no `jose`/JWT library:

- `HS256` → `crypto.subtle.verify` with the raw symmetric secret (legacy path).
- `ES256` / `RS256` → `crypto.subtle.importKey("jwk", …)` + `verify`, with
  per-algorithm import/verify params (`ECDSA`/`P-256`, `RSASSA-PKCS1-v1_5`).

Tokens are also checked for `exp`/`nbf` (with a small clock-skew window) before
the signature, matching the Go/Docker runtime, which reject expired tokens by
default.

## JWKS resolution

Key resolution is hybrid and ordered: the JWKS injected by the CLI
(`config.jwks`) is checked first, then the auth service's
`/auth/v1/.well-known/jwks.json` endpoint. Candidate keys are filtered by `kty`
and `kid` (keys without a `kid` stay eligible so single-key sets that omit it
still verify). Remote JWKS responses are cached per-URL; failed and empty
responses are not cached.

Note that the CLI-injected `config.jwks` is currently **symmetric-only** (an
`oct` key derived from `jwtSecret`), so it never matches an ES256/RS256 token
today — that local check is future-proofing for when the local stack grows
asymmetric signing keys, and real asymmetric verification currently resolves
through the well-known endpoint. Sourcing `SUPABASE_JWKS` from configured
signing keys / third-party providers is tracked as a follow-up (it first
requires wiring asymmetric signing into the local auth service).

`SUPABASE_JWKS` is also forwarded into the user worker environment so function
code can read the same key set the gate uses.

https://claude.ai/code/session_01Vn3KQfhzbP2X3QNqs36sKR


packages/stack/src/services/edge-runtime-main.ts
```diff
@@ -21,43 +31,60 @@ async function loadConfig() {
...
+    return null;
+  }
+}
+
+async function isValidAsymmetricJwt(jose: Jose, jwks: string, jwksUrl: string, jwt: string) {
+  try {
+    // Prefer the JWKS injected by the CLI; fall back to fetching it from the
+    // local auth service's well-known endpoint for keys minted elsewhere.
+    const keySet = createLocalJwks(jose, jwks) ?? jose.createRemoteJWKSet(new URL(jwksUrl));
+    await jose.jwtVerify(jwt, keySet);
```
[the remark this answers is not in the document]
avallete: Addressed in the WebCrypto rewrite (`f66460e`). The original `jose` path returned a successful local resolver and only failed inside `jwtVerify`, so it never reached the remote JWKS. The current `verifyHybridJwt` checks the local JWKS via `verifyWithJwks`, which returns `false` when no candidate key matches, and only then falls through to fetch `/auth/v1/.well-known/jwks.json`. So an ES256/RS256 token absent from the injected oct-only JWKS now correctly tries the remote endpoint.

---

packages/stack/tests/createStack.e2e.test.ts
```diff
@@ -94,6 +96,33 @@ describe("createStack e2e", () => {
     expect(await res.text()).toBe("later");
   });
 
+  test("enforces hybrid JWT verification on Edge Functions", { timeout: 20_000 }, async () => {
+    writeFunction(projectDir, "secure", "secure");
+    // verify_jwt defaults to true, so this turns verification back on.
+    await stack.reloadFunctions();
```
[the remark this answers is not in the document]
avallete: Addressed. The test now explicitly re-enables verification before asserting auth failures: it calls `await stack.reloadFunctions({ noVerifyJwt: false })` (with a comment noting that `reloadFunctions()` with no opts would keep the original `noVerifyJwt: true` config). The missing/forged-token requests then correctly return 401, and the hybrid verifier is exercised by the valid HS256 token returning 200.

---

packages/stack/src/services/edge-runtime-main.ts
```diff
@@ -21,43 +34,67 @@ async function loadConfig() {
...
+  if (localJwks) {
+    try {
+      await jose.jwtVerify(jwt, localJwks);
+      return true;
+    } catch {
+      // No matching/valid local key — try the remote JWKS below.
+    }
+  }
+  try {
+    await jose.jwtVerify(jwt, jose.createRemoteJWKSet(new URL(jwksUrl)));
```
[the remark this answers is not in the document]
avallete: Addressed in `f66460e`. Remote JWKS are now cached per-URL in a module-level `remoteJwksCache = new Map<string, Promise<Jwks>>()`. `fetchRemoteJwks` returns the cached promise when present and only deletes the entry on fetch failure, so repeated asymmetric invocations reuse the fetched key set instead of hitting `/auth/v1/.well-known/jwks.json` on every request.

---

packages/stack/src/services/edge-runtime.ts
```diff
@@ -27,8 +27,13 @@ interface DockerEdgeRuntimeOptions extends EdgeRuntimeOptions {
 const bootstrapFileName = "index.ts";
 const bootstrapMountDir = "/workspace";
 const bootstrapSourcePath = new URL("./edge-runtime-main.ts", import.meta.url);
+// jose is imported here (rather than in edge-runtime-main.ts) so the bun
+// workspace toolchain never has to resolve the Deno-only `jsr:` specifier; the
+// runtime script consumes it as an ambient `jose` global. See edge-runtime-main.ts.
+const joseImport = 'import * as jose from "jsr:@panva/jose@6";\n';
```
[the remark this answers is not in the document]
avallete: Addressed in `f66460e`. The `jsr:` import of the verifier was removed entirely — verification is now self-contained on top of WebCrypto (`crypto.subtle`), with no external/JOSE dependency. The injected edge-runtime main worker has no remote imports (`grep` for `deno.land|esm.sh|https://|jsr:|npm:` returns nothing), so a default local stack — including the `noVerifyJwt` / no-functions case — no longer depends on JSR being reachable or cached at startup.

---

packages/stack/src/functions.ts
```diff
@@ -198,6 +200,7 @@ export const resolveFunctionsRuntimeConfig = Effect.fnUntraced(function* (
     publishableKey: stackConfig.publishableKey,
     secretKey: stackConfig.secretKey,
     jwtSecret: stackConfig.jwtSecret,
+    jwks: generateJwks(stackConfig.jwtSecret),
```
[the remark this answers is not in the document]
avallete: Leaving this one open as a known limitation rather than fixing it in this PR, because the local stack doesn't yet support asymmetric signing locally:

- The local auth service is configured with the symmetric secret only — `StackBuilder` passes `version`, `port`, `siteUrl`, `jwtExpiry`, and `externalUrl` to the auth service, but not `auth.signing_keys_path` or `auth.third_party`. So local GoTrue mints HS256 tokens signed with `jwtSecret`, and `generateJwks(stackConfig.jwtSecret)` is the correct local key set for them.
- Since no asymmetric key is wired into the local issuer, putting `signing_keys_path`/`third_party` keys into `SUPABASE_JWKS` wouldn't match any locally-minted token today.
- Externally-minted asymmetric tokens are still handled by the remote `/auth/v1/.well-known/jwks.json` fallback for keys the local auth service is aware of.

Properly sourcing `SUPABASE_JWKS` from the configured signing keys / third-party providers requires first plumbing those into the local auth service so it actually signs with them — that's a larger, separate change. Happy to file a follow-up issue if you'd like to track it.

---

packages/stack/src/services/edge-runtime-main.ts
```diff
@@ -27,37 +27,140 @@ function base64UrlToBytes(value: string) {
...
+  let decodedHeader: { alg?: string; kid?: string };
+  try {
+    decodedHeader = JSON.parse(new TextDecoder().decode(base64UrlToBytes(header!)));
+  } catch (error) {
+    console.error("Failed to decode JWT header", error);
+    return false;
+  }
+
+  try {
+    if (decodedHeader.alg === "HS256") {
```
avallete: **No `exp`/`nbf` validation — behavioral divergence from the Go runtime.** `verifyHybridJwt` only checks signatures, so an expired token with a valid signature is accepted on every path. The Go/Docker runtime this ports used `jose.jwtVerify` / `golang-jwt`, both of which reject expired tokens by default — so a function gated locally behaves differently from the same function deployed.

The payload is already at hand (`parts[1]`); decoding it and checking `exp`/`nbf` with a small clock skew is a few lines:

```ts
const claims = JSON.parse(new TextDecoder().decode(base64UrlToBytes(payload!)));
const now = Math.floor(Date.now() / 1000);
const skew = 60;
if (typeof claims.exp === "number" && claims.exp + skew < now) return false;
if (typeof claims.nbf === "number" && claims.nbf - skew > now) return false;
```

---

packages/stack/src/services/edge-runtime-main.ts
```diff
@@ -27,37 +27,140 @@ function base64UrlToBytes(value: string) {
...
+function fetchRemoteJwks(jwksUrl: string): Promise<Jwks> {
+  const cached = remoteJwksCache.get(jwksUrl);
+  if (cached) return cached;
+  const pending = fetch(jwksUrl)
+    .then((res) => res.json())
+    .then(toJwks);
+  pending.catch(() => remoteJwksCache.delete(jwksUrl));
+  remoteJwksCache.set(jwksUrl, pending);
+  return pending;
+}
```
avallete, line 137: **A useless JWKS response can be cached permanently.** The response isn't checked with `res.ok`. If the gateway/auth service answers 502/404 with a JSON body during startup, `res.json()` succeeds, `toJwks` returns `{ keys: [] }`, and that empty set is cached forever — asymmetric tokens then 401 until the runtime restarts. Throwing on `!res.ok` lets the existing eviction path kick in:

```ts
const pending = fetch(jwksUrl)
  .then((res) => {
    if (!res.ok) throw new Error(`JWKS fetch failed: ${res.status}`);
    return res.json();
  })
  .then(toJwks);
```

Consider also not caching an empty key set. Relatedly, the cache has no TTL and no refetch-on-unknown-`kid`, so rotated keys are never picked up without a restart; standard JWKS client behavior is a rate-limited refetch when a token's `kid` misses the cached set. Lower priority for local dev, but cheap to add.

---

packages/stack/src/services/edge-runtime-main.ts
```diff
@@ -27,37 +27,140 @@ function base64UrlToBytes(value: string) {
...
+      return await verifyLegacyHs256(config.jwtSecret, signingInput, signature!);
+    }
+    if (decodedHeader.alg === "ES256" || decodedHeader.alg === "RS256") {
+      const { alg, kid } = decodedHeader;
+      // Prefer the JWKS injected by the CLI; fall back to the auth service's
+      // well-known endpoint for keys minted elsewhere.
+      const localJwks = parseLocalJwks(config.jwks);
+      if (await verifyWithJwks(alg, kid, signingInput, signature!, localJwks)) {
+        return true;
+      }
```
avallete: **The "local JWKS first" branch is dead code for asymmetric tokens today.** `generateJwks` only emits an `oct` key derived from the symmetric secret, and `verifyWithJwks` filters candidates by `kty === "EC" | "RSA"` — so the injected `config.jwks` can never verify an ES256/RS256 token. All real asymmetric verification goes through the remote well-known endpoint.

That's fine as future-proofing for when the stack config grows asymmetric signing keys, but the PR description ("the JWKS injected by the CLI is tried first") oversells what's wired up, and it means this branch is untestable as shipped. Worth a code comment noting the local JWKS is currently symmetric-only.

---

packages/stack/tests/createStack.e2e.test.ts
```diff
@@ -94,6 +96,34 @@ describe("createStack e2e", () => {
     expect(await res.text()).toBe("later");
   });
 
+  test("enforces hybrid JWT verification on Edge Functions", { timeout: 20_000 }, async () => {
+    writeFunction(projectDir, "secure", "secure");
+    // The stack was created with noVerifyJwt; re-enable verification explicitly
+    // (reloadFunctions() with no opts would keep the original config).
+    await stack.reloadFunctions({ noVerifyJwt: false });
```
avallete, line 103: Two things here:

1. **ES256/RS256 paths have zero test coverage.** This test only exercises HS256 (missing header / forged secret / valid token / apikey compat) — the entire asymmetric machinery (`kty`/`kid` filtering, WebCrypto import params, remote fallback, cache) is unverified, and it's the headline feature of the PR. Per the testing pyramid this belongs in unit tests rather than more e2e: `edge-runtime-main.ts` is import-safe outside Deno (the `typeof Deno !== "undefined"` guard), so exporting `verifyHybridJwt`/`verifyWithJwks` and unit-testing them under Bun would cover this cheaply — WebCrypto is available, so a test can generate a P-256 key pair, mint a token, and assert accept/reject.

2. **This mutates shared suite state without restoring it.** `reloadFunctions({ noVerifyJwt: false })` flips verification on for the whole stack. No later test currently calls a function, so it passes today, but anyone appending a function test after this one will hit surprise 401s. Either restore `noVerifyJwt: true` at the end of the test or leave a comment that subsequent tests must not assume open functions.

---

packages/stack/src/services/edge-runtime-main.ts
```diff
@@ -27,37 +27,140 @@ function base64UrlToBytes(value: string) {
...
+  );
+  for (const jwk of candidates) {
+    try {
+      const key = await crypto.subtle.importKey("jwk", jwk, spec.importParams, false, ["verify"]);
+      if (await crypto.subtle.verify(spec.verifyParams, key, sig, data)) return true;
+    } catch (error) {
+      console.error("Asymmetric JWK verification failed", error);
+    }
+  }
+  return false;
```
avallete, line 98: Minor, take or leave:

- `kid` matching is strict: if the token carries a `kid`, JWKS keys without one are excluded from candidates. Defensible, but some issuers omit `kid` on single-key sets — the Go-side behavior is more lenient (try all type-matching keys).
- `console.error` fires per candidate key per request; with a multi-key JWKS and a bad token this gets noisy. Consider logging once per token.

---

avallete (commented):
Reviewed the hybrid JWT verification port. The verifier architecture is sound: hand-rolling on WebCrypto is the right call for an injected script with no bundled deps, algorithm confusion is handled correctly (`alg` allow-list, HS256 never touches JWKS keys, asymmetric candidates filtered by `kty`), and the failed-fetch cache eviction is done properly.

Two things I'd address before merge, detailed inline:

1. **Missing `exp`/`nbf` validation** — expired tokens with valid signatures are accepted, diverging from the Go/Docker runtime this ports (jose/golang-jwt validate expiry by default).
2. **`fetchRemoteJwks` can permanently cache an empty key set** when the auth service responds non-2xx with a JSON body (no `res.ok` check).

Also inline: the local-JWKS branch is currently dead code for asymmetric tokens (the injected JWKS is `oct`-only), the ES256/RS256 paths have no test coverage (suggest unit tests on exported helpers rather than more e2e), and the new e2e test flips `noVerifyJwt` on the shared stack without restoring it.

---

packages/stack/src/services/edge-runtime-main.ts
avallete: Fixed in `c83bda6`. Added `areClaimsValid(payload)`, called in `verifyHybridJwt` before signature checks on every path, rejecting expired (`exp`) and not-yet-valid (`nbf`) tokens with a 60s clock-skew window — matching the Go/Docker runtime. Covered by unit tests for an expired-but-validly-signed token, an `nbf`-in-future token, and the skew boundary.

---

packages/stack/src/services/edge-runtime-main.ts
avallete: Fixed in `c83bda6`. `fetchRemoteJwks` now throws on `!res.ok` (so a 502/404-with-JSON-body no longer parses into an empty set), and the existing `.catch` eviction removes the cache entry. I also explicitly evict an empty key set after parsing, so a successful-but-empty response is retried on the next request rather than cached forever.

On the TTL / refetch-on-unknown-`kid` point: I left that out for now since it's lower priority for local dev, and the empty-set eviction already covers the "keys not published yet at startup" case. Happy to add a rate-limited `kid`-miss refetch in a follow-up if you'd like it.

---

packages/stack/src/services/edge-runtime-main.ts
avallete: Addressed in `c83bda6`. Added a comment on the local-JWKS branch noting that `config.jwks` is currently symmetric-only (`oct` derived from `jwtSecret`), so it never matches an ES256/RS256 token today — the check is future-proofing for when the stack config grows asymmetric signing keys, and real asymmetric verification resolves via the well-known endpoint. I also corrected the PR description so it no longer implies the injected JWKS verifies asymmetric tokens. The branch is now exercised by unit tests that inject an asymmetric JWK into `config.jwks` directly.

---

packages/stack/tests/createStack.e2e.test.ts
avallete: Both addressed in `c83bda6`.

1. Added `src/services/edge-runtime-main.unit.test.ts` (18 cases). It exports and unit-tests `verifyHybridJwt`/`verifyWithJwks`/`areClaimsValid` under Bun, generating real P-256 and RSA key pairs via WebCrypto to mint ES256/RS256 tokens and asserting: local-JWKS accept, wrong-key reject, `kid` mismatch reject, keyless-entry match, remote well-known fallback (mocked `fetch`), plus HS256 and the claim paths. The exports are inert for the runtime since the script is loaded as raw text (`with { type: "text" }`).
2. The e2e test now restores `noVerifyJwt: true` on the shared stack at the end of the hybrid test, with a comment explaining why, so later tests don't inherit JWT enforcement.

---

packages/stack/src/services/edge-runtime-main.ts
avallete: Both taken, in `c83bda6`. `verifyWithJwks` now keeps keys without a `kid` eligible even when the token carries one (`kid === undefined || key.kid === undefined || key.kid === kid`), so single-key sets that omit `kid` still verify — matching the more lenient Go behavior, with a unit test for the keyless-entry case. And I dropped the per-candidate `console.error` (failures across candidates are expected during matching), so a bad token against a multi-key JWKS no longer spams the logs.

---

packages/stack/src/services/edge-runtime-main.ts
```diff
@@ -27,37 +27,179 @@ function base64UrlToBytes(value: string) {
...
+  }
+}
+
+// Cache the well-known JWKS per URL so asymmetric verification does not refetch
+// `/auth/v1/.well-known/jwks.json` on every request.
+const remoteJwksCache = new Map<string, Promise<Jwks>>();
+
+function fetchRemoteJwks(jwksUrl: string): Promise<Jwks> {
+  const cached = remoteJwksCache.get(jwksUrl);
+  if (cached) return cached;
```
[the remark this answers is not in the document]
avallete: Addressed in `80e5620`, though via a TTL rather than refetch-on-miss. A refetch-on-miss wouldn't cover the more important case (a *revoked* key stays in the cached set, so a revoked token keeps matching and never misses). Instead, `remoteJwksCache` entries now carry a `cachedAt` timestamp and expire after a 5-minute TTL, so both rotated-in and revoked keys are picked up within one TTL window without a restart. `res.ok` failures and empty sets still aren't cached. Covered by a unit test that asserts a stale cached set is reused within the TTL and refetched (picking up the rotated key) after it elapses.

---

packages/stack/src/services/edge-runtime-main.ts
```diff
@@ -27,37 +27,179 @@ function base64UrlToBytes(value: string) {
...
+  let claims: { exp?: unknown; nbf?: unknown };
+  try {
+    claims = JSON.parse(new TextDecoder().decode(base64UrlToBytes(payload)));
+  } catch {
+    return false;
+  }
+  const now = Math.floor(Date.now() / 1000);
+  const skew = 60;
+  if (typeof claims.exp === "number" && claims.exp + skew < now) return false;
+  if (typeof claims.nbf === "number" && claims.nbf - skew > now) return false;
```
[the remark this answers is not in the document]
avallete: Fixed in `80e5620`. `areClaimsValid` now rejects `exp`/`nbf` when present but not a number (`typeof exp !== "number"` → reject), so a token carrying string temporal claims like `"exp":"1"` fails the gate instead of slipping through. Unit-tested with string `exp` and `nbf` values.

---

packages/stack/src/services/edge-runtime-main.ts
```diff
@@ -27,37 +27,179 @@ function base64UrlToBytes(value: string) {
...
+export function areClaimsValid(payload: string): boolean {
+  let claims: { exp?: unknown; nbf?: unknown };
+  try {
+    claims = JSON.parse(new TextDecoder().decode(base64UrlToBytes(payload)));
+  } catch {
+    return false;
+  }
+  const now = Math.floor(Date.now() / 1000);
+  const skew = 60;
+  if (typeof claims.exp === "number" && claims.exp + skew < now) return false;
```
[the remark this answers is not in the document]
avallete: Good catch — fixed in `80e5620`. `areClaimsValid` now returns `false` when the decoded payload isn't a non-null object (`typeof claims !== "object" || claims === null`), so a `null` (or bare-string) payload is rejected with the normal 401 path instead of dereferencing `null` and throwing into a 500. Covered by a unit test minting a validly-signed token with a `null` payload (asserts `false`, no throw) plus direct `areClaimsValid` cases.

---

packages/stack/src/services/edge-runtime-main.ts
```diff
@@ -27,37 +27,179 @@ function base64UrlToBytes(value: string) {
...
+    return { keys: [] };
+  }
+}
+
+// Cache the well-known JWKS per URL so asymmetric verification does not refetch
+// `/auth/v1/.well-known/jwks.json` on every request.
+const remoteJwksCache = new Map<string, Promise<Jwks>>();
+
+function fetchRemoteJwks(jwksUrl: string): Promise<Jwks> {
+  const cached = remoteJwksCache.get(jwksUrl);
```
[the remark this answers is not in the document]
avallete: Fixed in `80e5620`, essentially as suggested. `remoteJwksCache` entries now carry a `cachedAt` timestamp and a 5-minute TTL (`REMOTE_JWKS_TTL_MS`); the lookup returns the cached set only while `Date.now() - cachedAt < TTL`, otherwise it refetches. So a rotated/revoked key stops verifying within one TTL window instead of for the process lifetime. (Implementation stores `{ cachedAt, jwks }` and also keeps the existing `res.ok` guard and empty-set eviction.) Covered by a unit test asserting the stale set is reused within the TTL and refetched after it expires.

---

packages/stack/src/services/edge-runtime-main.ts
```diff
@@ -27,37 +27,189 @@ function base64UrlToBytes(value: string) {
...
+  let claims: unknown;
+  try {
+    claims = JSON.parse(new TextDecoder().decode(base64UrlToBytes(payload)));
+  } catch {
+    return false;
+  }
+  // A valid JWT payload must be a JSON object; reject anything else (e.g. `null`
+  // or a bare string) rather than dereferencing it and throwing past the gate.
+  if (typeof claims !== "object" || claims === null) return false;
+  const { exp, nbf } = claims as { exp?: unknown; nbf?: unknown };
```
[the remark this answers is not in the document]
avallete: Fixed in `a211129`. `areClaimsValid` now also rejects arrays (`Array.isArray(claims)`) alongside the `null`/non-object guard, so an array payload like `[]` fails the gate instead of skipping the temporal-claim checks. Added an array case to the non-object payload unit test.

---

packages/stack/src/services/edge-runtime-main.ts
```diff
@@ -108,6 +257,7 @@ async function serveFunction(req: Request, config: any, functionName: string, fu
     SUPABASE_DB_URL: config.dbUrl,
     SUPABASE_PUBLISHABLE_KEYS: JSON.stringify({ default: config.publishableKey }),
     SUPABASE_SECRET_KEYS: JSON.stringify({ default: config.secretKey }),
+    SUPABASE_JWKS: config.jwks,
```
kallebysantos, line 253: As mentioned above, it seems the `config.jwks` is not fully implemented... maybe we should expose the `remoteJwks` instead?

And then add a TODO comment on this line explaining the situation? 
Or better, just remove the `localJwks` — No premature "future work" preparation with broken "dead code"

This `SUPABASE_JWKS` env is consumed by Server SDK. If the values here are not correct, the lib may deny `auth: "user"` requests

kallebysantos (approved):

packages/stack/src/services/edge-runtime-main.ts
avallete: Went with your "better" suggestion in `d239180` — removed the dead `localJwks` path rather than future-proofing broken code:

- Dropped the local-JWKS-first branch (and `parseLocalJwks`) from `verifyHybridJwt`. Since the injected JWKS is symmetric-only (`oct`), it could never verify an ES256/RS256 token anyway, so asymmetric tokens now go straight to the auth service's well-known JWKS.
- Kept `SUPABASE_JWKS: config.jwks` rather than switching to `remoteJwks`, because for the default local issuer that's the **correct** value: GoTrue signs HS256 with the symmetric secret and doesn't publish it at `/.well-known/jwks.json`, so exposing the (empty/asymmetric-only) remote set would actually break the SDK's HS256 `auth: "user"` verification. The `oct` JWKS matches what the Go CLI's `Auth.ResolveJWKS` emits as its fallback when no signing keys are configured.
- Added a TODO on that line: the remaining gap is resolving `auth.signing_keys_path` / `auth.third_party` into `SUPABASE_JWKS` (like `ResolveJWKS` does) so the SDK works for asymmetric-configured projects. That requires plumbing those keys into the local stack and is tracked as the separate open thread above.

So no broken dead code, and the env value is correct for the default local case. Reworked the asymmetric unit tests to exercise the remote path accordingly.

---

kallebysantos (approved):

packages/stack/src/functions.ts
jgoux, line 203: This still always builds `SUPABASE_JWKS` from the legacy `jwtSecret`. The Go implementation this PR is porting from resolves third-party JWKS and configured signing keys first, then falls back to the secret only when no signing key path is configured. In this TS stack, `auth.signing_keys_path` / `auth.third_party` are loaded by config but never carried into `ResolvedStackConfig`, the auth service is started with only `GOTRUE_JWT_SECRET`, and the edge runtime verifies ES256/RS256 only against the local auth service JWKS endpoint. So projects configured with local asymmetric signing keys or third-party auth still won’t have those keys in `SUPABASE_JWKS`, and the runtime gate can reject the ES256/RS256 tokens this change is meant to support. Please either port the JWKS resolution into `packages/stack` or narrow this PR so it does not claim full hybrid/asymmetric local parity yet.

packages/stack/tests/createStack.e2e.test.ts
```diff
@@ -94,6 +96,38 @@ describe("createStack e2e", () => {
     expect(await res.text()).toBe("later");
   });
 
+  test("enforces hybrid JWT verification on Edge Functions", { timeout: 20_000 }, async () => {
```
jgoux, line 99: This e2e name says “hybrid JWT verification,” but the scenario only covers missing credentials, forged HS256, valid HS256, and the `apikey` compatibility path. The ES256/RS256 branch is only covered in unit tests with a mocked JWKS fetch, so we still do not have runtime proof that the actual edge-runtime + gateway + auth JWKS path works. Please add an asymmetric runtime smoke test or rename/scope this e2e to the HS256 legacy path it actually exercises.

packages/stack/tests/createStack.e2e.test.ts
```diff
@@ -94,6 +96,38 @@ describe("createStack e2e", () => {
...
+    writeFunction(projectDir, "secure", "secure");
+    // The stack was created with noVerifyJwt; re-enable verification explicitly
+    // (reloadFunctions() with no opts would keep the original config).
+    await stack.reloadFunctions({ noVerifyJwt: false });
```
jgoux, line 103: After this flips the shared stack to `noVerifyJwt: false`, any assertion failure before the restore at the end leaves JWT verification enabled for the remaining tests in this file. Please wrap the assertions in `try`/`finally` and restore `noVerifyJwt: true` in the `finally` block so later tests do not inherit hidden state after a failure.

jgoux (changes requested):
Thanks for the solid cleanup through the review rounds. I verified the final head locally:

- `pnpm --filter @supabase/stack test:core`
- `pnpm --filter @supabase/stack check:all`
- `pnpm --filter @supabase/stack exec bun --bun vitest run --project e2e tests/createStack.e2e.test.ts`

I found one remaining correctness gap that I think should be addressed before merging: the PR claims hybrid/asymmetric support for the TS edge runtime, but the TS stack still only resolves an HS256 secret-derived JWKS and does not wire `auth.signing_keys_path` / `auth.third_party` into either the auth service or the function `SUPABASE_JWKS` env. That means the new verifier works for mocked remote JWKS and the legacy HS256 path, but still does not provide the Go-side parity for configured local asymmetric keys/third-party providers.

Two lower-severity follow-ups I noticed:

- The added e2e is valuable, but it only exercises HS256 plus the `apikey` compatibility path; ES256/RS256 remain unit-only with mocked `fetch`. A runtime-level asymmetric smoke test would make this much stronger.
- The shared e2e stack should restore `noVerifyJwt` in a `finally`, so a failed assertion does not leak JWT enforcement into later tests.

packages/stack/src/functions.ts
avallete: I think is has already been discussed here: [https://github.com/supabase/cli/pull/5560#discussion_r3404081531](<https://github.com/supabase/cli/pull/5560#discussion_r3404081531>)<br><br>Or I'm I missing something ?

jgoux: Indeed!

avallete:
After discussion with @kallebysantos I'll turn this back into a draft and let him take over. Implementation of the auth for edge functions might see some changes soon.

PR #5563, opened 2026-06-12
chore(cli): organize lazy platform API factory

jgoux:
## Summary

Extracts the lazy Management API client factory into auth-owned service and layer modules so command runtimes can share the same wiring instead of rebuilding it locally.

The lean `gen types` runtime now consumes the shared lazy factory, while eager Management API runtimes adapt their already-built `LegacyPlatformApi` into the factory shape for project-ref resolution. This keeps tokenless command paths lazy without duplicating platform API construction in commands that intentionally authenticate up front.

### DIFF
--- a/apps/cli/src/legacy/auth/legacy-platform-api-factory.layer.ts
+++ b/apps/cli/src/legacy/auth/legacy-platform-api-factory.layer.ts
@@ -1,25 +1,12 @@
 import { FetchHttpClient } from "effect/unstable/http";
-import { Effect, FileSystem, Layer, Path } from "effect";
-import type * as HttpClient from "effect/unstable/http/HttpClient";
+import { Effect, Layer } from "effect";
 
-import { Analytics } from "../../shared/telemetry/analytics.service.ts";
-import { TelemetryRuntime } from "../../shared/telemetry/runtime.service.ts";
-import { LegacyCliConfig } from "../config/legacy-cli-config.service.ts";
-import { LegacyDebugLogger } from "../shared/legacy-debug-logger.service.ts";
-import { LegacyCredentials } from "./legacy-credentials.service.ts";
 import { legacyMakePlatformApi } from "./legacy-platform-api.layer.ts";
 import { LegacyPlatformApi } from "./legacy-platform-api.service.ts";
 import { LegacyPlatformApiFactory } from "./legacy-platform-api-factory.service.ts";
 
 type LegacyPlatformApiDeps =
-  | Analytics
-  | LegacyCliConfig
-  | LegacyCredentials
-  | LegacyDebugLogger
-  | FileSystem.FileSystem
-  | HttpClient.HttpClient
-  | Path.Path
-  | TelemetryRuntime;
+  typeof legacyMakePlatformApi extends Effect.Effect<infer _A, infer _E, infer R> ? R : never;
 
 /**
  * Captures the surrounding Management API context without resolving an access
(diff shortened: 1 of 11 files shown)


apps/cli/src/legacy/auth/legacy-platform-api-factory.layer.ts
```diff
@@ -0,0 +1,53 @@
...
+ * the single typed-API debug wrapper.
+ */
+export const legacyPlatformApiFactoryLayer = Layer.effect(
+  LegacyPlatformApiFactory,
+  Effect.gen(function* () {
+    const context = yield* Effect.context<LegacyPlatformApiDeps>();
+    return LegacyPlatformApiFactory.of({
+      make: legacyMakePlatformApi.pipe(Effect.provideContext(context)),
+    });
+  }),
```
avallete, line 25: Non-blocking: `make` is rebuilt on every call — each invocation re-resolves the access token and re-emits the `--debug` banner lines from `legacyMakePlatformApi`. This matches the previous inline factory, so no regression, but since both `legacyProjectRefLayer` and the `gen types` handler can call `make` within a single command run, an `Effect.cached` on `make` would be a cheap follow-up if the duplicate token resolution / debug output ever becomes noticeable.

---

apps/cli/src/legacy/auth/legacy-platform-api-factory.layer.ts
```diff
@@ -0,0 +1,53 @@
...
+
+type LegacyPlatformApiDeps =
+  | Analytics
+  | LegacyCliConfig
+  | LegacyCredentials
+  | LegacyDebugLogger
+  | FileSystem.FileSystem
+  | HttpClient.HttpClient
+  | Path.Path
+  | TelemetryRuntime;
```
avallete: Non-blocking: `LegacyPlatformApiDeps` is a manually maintained mirror of `legacyMakePlatformApi`'s requirements. The compiler catches omissions (a new dependency not listed here breaks the `R = never` constraint on `make` at the `provideContext` call — nice), but it can silently accumulate stale entries if a dependency is ever removed. Fine as is; just flagging the asymmetry.

---

avallete (approved):
Clean dedup of the three inline factory builds into auth-owned layers. Verified locally on the head commit: `types:check`, `lint:check`, `fmt:check`, `knip:check` all pass, and the full unit + integration suite passes (195 files, 2,157 tests).

Highlights:
- `Effect.context` + `Effect.provideContext` replaces the seven manual `Effect.provideService` calls, so a new `legacyMakePlatformApi` dependency now fails compilation in exactly one place instead of silently missing a provide.
- Factory error channel narrowed from `unknown` to the concrete union, which matches what `legacyMakePlatformApi` can actually fail with.
- Dropping the outer `FetchHttpClient` provide and the extra services from the `gen types` runtime union is safe — `legacyHttpClientLayer` and the factory layer both self-provide fetch, and the handler only yields the services that remain in the union.
- `--debug` HTTP logging parity is preserved since the typed client owns its own debug wrapper via `transformClient`.

Two non-blocking inline notes on memoization and the manually mirrored deps union.

---

apps/cli/src/legacy/auth/legacy-platform-api-factory.layer.ts
jgoux: Handled in `bdf8bea6`: `legacyPlatformApiFactoryLayer` now builds `make` with `Effect.cached`, so the first Management API use in a command run constructs the client and later calls reuse that result.

apps/cli/src/legacy/auth/legacy-platform-api-factory.layer.ts
jgoux: Handled in `bdf8bea6`: `LegacyPlatformApiDeps` is now derived from `typeof legacyMakePlatformApi` instead of manually mirroring its requirements, so removed dependencies stop appearing in the factory layer automatically.

PR #5569, opened 2026-06-12
fix(cli): handle blank project password prompt

jgoux:
Fixes the interactive password prompt path used by `projects create` when the user leaves the database password blank.

`@clack/prompts` can return no string value for that blank password entry. The output adapter was calling `.trim()` directly, which crashed before the project creation flow could generate the fallback password. The adapter now normalizes non-string prompt results to an empty string so existing command-level fallback behavior can run.

### DIFF
--- a/apps/cli/src/shared/output/output.layer.ts
+++ b/apps/cli/src/shared/output/output.layer.ts
@@ -305,7 +305,7 @@ export const textOutputLayer = Layer.effect(
             cancel("Operation cancelled.");
             return yield* Effect.interrupt;
           }
-          return value.trim();
+          return typeof value === "string" ? value.trim() : "";
         }),
       promptConfirm: (message: string, opts?: { defaultValue?: boolean }) =>
         Effect.gen(function* () {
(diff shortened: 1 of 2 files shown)


apps/cli/src/shared/output/output.layer.ts
```diff
@@ -305,7 +305,7 @@ export const textOutputLayer = Layer.effect(
             cancel("Operation cancelled.");
             return yield* Effect.interrupt;
           }
-          return value.trim();
+          return typeof value === "string" ? value.trim() : "";
```
avallete, line 308: Shouldn't that have been caught by type checking ? If the value can be undefined, we should've got an "need an optional operator" warning, or something. I think the root cause of the issue is deeper.

jgoux: Verified this. Type checking does not catch it because `@clack/prompts` declares `password()` as `Promise<string | symbol>`, but an empty password submit can resolve `undefined` at runtime.

The deeper issue appears to be in Clack: `TextPrompt` normalizes a missing value to `""` on finalize, while `PasswordPrompt` only sets its value from `userInput` and does not perform the same empty-submit normalization. So this guard is defensive against the dependency's runtime behavior, not against a type-visible state.

I opened an upstream issue with the reproduction and source-level notes here: https://github.com/bombshell-dev/clack/issues/560

Coly010 (approved):

PR #5593, opened 2026-06-17
fix(cli): accept sparse SSO provider responses

### LINKED ISSUE #5589: supabase sso list/show/update fail with SchemaError on 2.106.0 (regression from 2.101.0)
# `supabase sso list/show/update` fail with `SchemaError` on 2.106.0 (regression from 2.101.0)

## Summary

Starting in CLI **2.106.0**, every `supabase sso` read/write command fails with a
deserialization `SchemaError` against a hosted project that has existing SAML SSO
providers. The same commands work correctly on **2.101.0** against the *same*
project and the *same* providers (no provider data changed between the two runs),
so this is a client-side response-parsing regression rather than a server-side or
data issue.

## Affected commands

- `supabase sso list`
- `supabase sso show <provider-id>`
- `supabase sso update <provider-id> ...` (fails because it performs a GET first)

`supabase sso list` against a project with **zero** providers succeeds
(`{"providers":[]}`); the failure only occurs once at least one SAML provider exists.

## Errors

`supabase sso list`:
```
failed to list sso providers: SchemaError(Missing key
  at ["items"][0]["saml"]["attribute_mapping"]["keys"])
```

`supabase sso show <provider-id>` and `supabase sso update <provider-id> ...`:
```
failed to get sso provider: SchemaError(Missing key
  at ["saml"]["id"])
```

The output format does not matter (`--output-format json` and `--experimental`
produce the same error), which indicates the failure is in response
deserialization, before rendering.

## Likely cause

The Management API returns SAML providers with `attribute_mapping` as an **empty
object** (`{}`) for providers created without a custom attribute mapping. 2.106.0's
schema appears to require `attribute_mapping.keys` to be present whenever
`attribute_mapping` is present, so an empty `{}` (no `keys`) fails to deserialize.
A similar strictness around `saml.id` breaks `show`/`update`.

For reference, the GoTrue admin endpoint
(`GET /auth/v1/admin/sso/providers`) returns `"attribute_mapping": {}` for such
providers, and that endpoint continues to work normally.

## Steps to reproduce

1. Have a hosted project with at least one SAML SSO provider created without a
   custom attribute mapping (so its `attribute_mapping` is `{}`).
2. `supabase sso list --project-ref <project-ref>` on **2.106.0** → `SchemaError`.
3. Downgrade to **2.101.0** and run the identical command → succeeds, lists all
   providers correctly.

## Expected behavior

`sso list/show/update` deserialize providers whose `attribute_mapping` is an empty
object (`{}`) — or whose `attribute_mapping`/`saml.id` shape matches what the
Management API currently returns — and succeed, as 2.101.0 does.

## Environment

- Supabase CLI: **2.106.0** (broken), **2.101.0** (works)
- OS: macOS (Apple Silicon), installed via Homebrew tap
- Project plan: Pro (SAML SSO enabled)

## Workaround

Pin the CLI to 2.101.0 until a fixed release is available.

pjpjq:
## What changed

- Relax the generated SSO provider response models so nested SAML IDs and `attribute_mapping.keys` can be omitted.
- Keep request payload models strict, while allowing list/show/update response parsing to match the Management API shape.
- Skip rendering empty SAML attribute mappings in pretty output.
- Add regression coverage for `sso list`, `sso show`, and `sso update` with empty `attribute_mapping` objects.

## Why

Hosted projects can return SAML providers with `attribute_mapping: {}` and without a nested `saml.id`. The CLI's generated response types treated those fields as required, causing schema decoding to fail before the commands could render or update providers.

Fixes #5589.

### DIFF
--- a/apps/cli-go/api/overlay.yaml
+++ b/apps/cli-go/api/overlay.yaml
@@ -29,6 +29,18 @@ actions:
 - target: $.components.schemas.*.properties.items.items.properties.saml.properties.name_id_format.enum
   description: Converts list of enum with duplicated type to a string
   remove: true
+- target: $.components.schemas.*.properties.saml.required[?(@ == "id")]
+  description: Allows provider responses to omit nested SAML IDs
+  remove: true
+- target: $.components.schemas.ListProvidersResponse.properties.items.items.properties.saml.required[?(@ == "id")]
+  description: Allows provider list responses to omit nested SAML IDs
+  remove: true
+- target: $.components.schemas.*.properties.saml.properties.attribute_mapping.required[?(@ == "keys")]
+  description: Allows provider responses to return an empty SAML attribute mapping object
+  remove: true
+- target: $.components.schemas.ListProvidersResponse.properties.items.items.properties.saml.properties.attribute_mapping.required[?(@ == "keys")]
+  description: Allows provider list responses to return an empty SAML attribute mapping object
+  remove: true
 - target: $.components.schemas.*.properties.connectionString
   description: Removes deprecated field that conflicts with naming convention
   remove: true
--- a/apps/cli-go/internal/sso/get/get_test.go
+++ b/apps/cli-go/internal/sso/get/get_test.go
@@ -67,6 +67,38 @@ func TestSSOProvidersShowCommand(t *testing.T) {
 		assert.Empty(t, apitest.ListUnmatchedRequests())
 	})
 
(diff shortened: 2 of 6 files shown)


jgoux:
Hey @pjpjq, we fully ported the SSO commands to TypeScript so I [fixed it on TypeScript's side](https://github.com/supabase/cli/pull/5594) instead.
Thanks for your contribution anyway! 🙏

PR #5607, opened 2026-06-17
fix(cli): reconcile hybrid stitch+stamp identity with shared LegacyIdentityStitch service

seanoliver:
## Problem

The #5366 gate stopped the ephemeral-env `$identify` spike, but at the cost of attribution: in CI, Docker, and `npx supabase`, `cli_*` events stay orphaned on throwaway device IDs and never link to the authenticated user. GROWTH-891 (#5559) fixes that with a hybrid stitch+stamp model.

While #5559 was in review, #5579 (db lint/advisors port) landed on develop and independently extracted the legacy identity stitch into a shared `LegacyIdentityStitch` service — one per-command `stitchAttempted` guard so the advisor transports alias at most once. That's the architecture we want, but it's a port of the pre-891 behavior: it only stamps when it aliases (persistent, first login), so it doesn't restore CI/Docker/npx attribution; it sets `stitchAttempted` after the file-read yield; and it reads the `runtime.distinctId` field that 891 replaced with a mutable identity slot. Merging #5559 on top as-is would silently drop the attribution feature and reintroduce the race.

This PR reconciles the two: keep #5579's shared-service architecture, fold the hybrid stitch+stamp behavior into it. Supersedes #5559.

## Changes

- **The shared `LegacyIdentityStitch` now stamps everywhere.** On the first authenticated response the user UUID is stamped into `runtime.identity` in every runtime, so captures in CI/Docker/npx carry the real user. The `$create_alias` (pre-login history merge) and the `telemetry.json` write still only happen on a persistent machine.
- **Hardening preserved:** `stitchAttempted` is set before the first yield (no double-stitch race); when an identity already exists we stamp without aliasing (never merge two person graphs); alias fires at most once across all transports sharing the service.
- **`stitchedDistinctId()` returns `runtime.identity.current()`** so the post-run `cli_command_executed` is attributed to the real user in every runtime, including steady state.
- `legacy-analytics.layer.ts` resolves `distinctId` from the identity slot while keeping develop's already-keyed `groups` map.
- Stitch behavior tests live in `legacy-identity-stitch.integration.test.ts` (CI-stamp-no-alias, stale-identity-stamp-no-alias, concurrent-alias-once); the platform-api layer test keeps develop's service-mocked wiring. A few command test runtimes still using the removed `distinctId` field were updated to `makeTelemetryIdentity`.
- Brings the Go + next-TS 891 changes (logout identity reset + device-id rotation, the redundant `$identify` removal, ADR 0013) along through the merge.

## Testing

Typecheck clean, full unit suite (1318) green, and the affected integration suites (identity-stitch, platform-api, login, logout, advisors, lint, services, gen/types, issue, linked-project-cache) pass under bun. Also ran an independent Codex review focused on the spike-regression risk — it confirmed no alias in ephemeral runtimes, alias-at-most-once across transports, the pre-yield race guard, and the no-cross-graph-merge invariant, with no findings.

GROWTH-891


seanoliver:
Heads up on CI: the only red check here is **Codegen**, and it's not from this PR — this branch touches zero codegen/API files.

`go generate` pulls the live staging spec from `api.supabase.green/api/v1-yaml`, and that spec currently has a schema oapi-codegen v2.4.1 can't handle:

`ProjectUpgradeEligibilityResponse.warnings[]` uses a `discriminator` (`propertyName: type`) with inline `oneOf` branches and **no `mapping`** → "discriminator: not all schemas were mapped". The three warning types (`pg_graphql_introspection_change`, `ltree_reindex_required`, `operator_estimator_gate`) look like a recent platform addition, so this is breaking Codegen on all current CLI PRs.

Fix is upstream: add the discriminator `mapping` (or switch the `oneOf` branches to `$ref`s) in the API spec, or adjust the cli's `types.cfg.yaml` to tolerate it. Everything else on this PR is green (unit + integration, all 3 e2e shards, lint, code quality).

jgoux (approved):

PR #5618, opened 2026-06-18
fix(cli): merge matching [remotes.*] block on config push

Coly010:
## What changed

`config push` regressed in v2.106.0 (the native-TS port): when a `[remotes.<name>]` block in `config.toml` targeted the project ref, the command aborted with

> cannot push config: a [remotes.*] block targets project ***, which config push does not yet support.

The Go CLI (v2.105.0) instead merges that remote's subtree over the base config and pushes it. The port had punted on Go's `mergeRemoteConfig`.

This ports the merge faithfully and removes the abort.

## Why this location

The merge is owned by `@supabase/config`, mirroring Go doing it in `pkg/config`. `loadProjectConfig` / `loadProjectConfigFile` now accept an optional `{ projectRef }`. When set, after `env()` interpolation and **before** schema decode, the matching `[remotes.<name>]` raw subtree is deep-merged over the base document (objects recurse; arrays and scalars replace wholesale — viper's `v.Set` semantics), `db.seed.enabled` is forced `false` when the remote omits it, the `remotes` key is stripped, and the merged document is decoded. Doing it on the raw document (not the decoded config) is essential: the decoded remote section carries full schema defaults that would otherwise clobber every field the block doesn't override.

The merge is gated on `projectRef`, so every other `loadProjectConfig` caller is unaffected.

## Notable details for reviewers

- New `DuplicateRemoteProjectIdError` (exported from `@supabase/config`) raised when two remotes share the target `project_id`, carrying Go's verbatim message `duplicate project_id for [remotes.<b>] and [remotes.<a>]`.
- `LoadedProjectConfig` gains optional `document` (merged, post-interpolation raw doc) and `appliedRemote` fields.
- The push handler prints `Loading config override: [remotes.<name>]` to stderr (Go parity) when a remote applies, and now derives optional pointer-section presence (`db.ssl_enforcement`, `storage.image_transformation`, `storage.s3_protocol`, auth subsections) from the merged document instead of re-reading the file — so sections introduced by the remote are detected. Dead code removed (`matchesRemoteProjectRef`, `resolveRemoteByProjectRef`, `LegacyConfigPushUnsupportedRemoteError`).
- `functions deploy` is consolidated onto the same shared merge, deleting its divergent partial copy (`configForProjectRef` / `mergeFunctionConfigByPresence`, which only handled `functions.*` and `edge_runtime.deno_version`). Verified behavior-preserving since deploy reads only those fields. This also corrects deploy's duplicate-`project_id` message to match Go (both remote names bracketed).

Closes CLI-1808

### DIFF
--- a/apps/cli/src/legacy/commands/config/push/SIDE_EFFECTS.md
+++ b/apps/cli/src/legacy/commands/config/push/SIDE_EFFECTS.md
@@ -64,7 +64,7 @@ when its local gate is off.
 | ---- | ------------------------------------------------------------------------------------- |
 | `0`  | success, **including** declining a confirmation prompt (Go returns nil and continues) |
 | `1`  | malformed `config.toml`                                                               |
-| `1`  | a `[remotes.*]` block targets the project ref (unsupported — see Known Gaps)          |
+| `1`  | two `[remotes.*]` blocks declare the same `project_id` as the target ref              |
 | `1`  | list-addons failure (network or non-200)                                              |
 | `1`  | any per-service read/update failure (network or unexpected status)                    |
 
@@ -72,7 +72,9 @@ when its local gate is off.
 
 ### `--output-format text` (Go CLI compatible)
 
-All diagnostics on **stderr**, no stdout. `Pushing config to project: <ref>`, then
+All diagnostics on **stderr**, no stdout. When a `[remotes.<name>]` block matches the
+target ref, `Loading config override: [remotes.<name>]` prints first. Then
+`Pushing config to project: <ref>`, then
 per service either `Remote <X> config is up to date.` or
 `Updating <X> service with config: <unified diff>`; experimental prints
 `Enabling webhooks for project: <ref>`. Confirmations render `<title> [Y/n] `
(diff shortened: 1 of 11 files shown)


jgoux:
I found one parity issue worth fixing before merge:

`packages/config/src/io.ts` currently checks duplicate remote `project_id`s only among remotes whose `project_id` matches the requested `projectRef`:

```ts
const matches = Object.entries(remotes)
  .filter(([, remote]) => isObject(remote) && remote["project_id"] === projectRef)
  .map(([name]) => name);
```

The Go loader builds its duplicate map across all `[remotes.*]` blocks before applying the matching override (`apps/cli-go/pkg/config/config.go:503-514`). That means a config with two non-target remotes sharing the same `project_id`, or two remotes omitting `project_id`, is accepted here but rejected by Go.

Since this PR is explicitly restoring Go parity for remote config loading, `applyRemoteOverride` should scan all remotes for duplicate IDs first, then apply the matching remote.

Coly010:
Good catch — fixed in 65ab5daa.

`applyRemoteOverride` now builds the `project_id -> [remotes.<name>]` map across **all** `[remotes.*]` blocks and fails on the first duplicate before applying the matching override, mirroring Go's `loadFromFile` (`apps/cli-go/pkg/config/config.go:503-518`). A missing `project_id` reads as `""` (matching `viper.GetString`), so two remotes that both omit it now collide on the empty key just like Go.

Added two regression tests: duplicate IDs among non-target remotes, and two remotes omitting `project_id`.

jgoux (approved):

PR #5644, opened 2026-06-22
fix(cli): keep test db local connections plaintext

jgoux:
Fixes CLI-1818.

`supabase test db` now forces plaintext Postgres connections for local and Unix-socket targets by passing `ssl: false` through the node-postgres connection path. This matches the previous Go behavior, where local connections always cleared TLS config, and prevents ambient environment such as `PGSSLMODE=require` from forcing TLS against the local database.

While following the workspace integration suite, this also canonicalizes discovered function deploy import paths before containment checks and aligns Docker bind assertions with real host paths. That keeps source upload and Docker bind collection consistent on macOS temp paths such as `/var` vs `/private/var`.


apps/cli/src/shared/functions/deploy.ts
```diff
@@ -1072,6 +1084,17 @@ async function buildDockerBinds(
...
+      return;
+    }
+    await walkLocalImportMapTargetImports(
+      importMap,
+      target,
+      importMapAllowedRoots,
+      projectRoot,
+      appendImportMapBind,
+      async () => {},
+    );
```
avallete, line 1097: **question**

I wonder why this is needed to fix the connection plaintext issue ?

Maybe a comment could be useful here on the why:

1. Why return if the target is a directory ?
2. Why we walkLocalImporMapTargetIMports otherwise ? 

Knowing that the function itself is already declared with:

```
  if ((await stat(pathname)).isDirectory()) {
    return;
  }
  await walkImportPaths(importMap, pathname, allowedRoots, displayRoot, onFile, onWarning);
```

Would that make this logic recursive or something ? (we call the `walkImportPaths` above, then it will be called here as well).

jgoux: Good question. This is not needed for the `test db` plaintext connection fix itself.

I included it because after changing the CLI workspace I ran the full `supabase:test:integration` target, and the functions deploy integration file failed. The failures exposed two issues in that path:

1. The source upload path already walks imports of local import-map targets, so if an import map points at `lib.ts` and `lib.ts` imports `helper.ts`, both files are uploaded.

2. The Docker bind path only mounted the direct import-map target. That meant `lib.ts` was mounted, but `helper.ts` could be missing when bundling with Docker. The added `walkLocalImportMapTargetImports` makes Docker bind collection match source upload collection.

The directory guard is there because import-map targets may also be directories. Directory targets are already handled by mounting the directory itself, and walking it as a single source file would be wrong. For file targets, we walk their imports to include transitive local dependencies.

The earlier `walkImportPaths` call is for imports reachable from the function entrypoint inside the project root. This later walk is specifically for local files referenced by the import map itself, including targets outside the project root that are allowed because the import map lives outside the project root.

avallete (approved):

PR #5653, opened 2026-06-22
fix: apply import map during GitHub branch Edge Function bundling

### LINKED ISSUE #47000: Deploying functions from a branch tracked by the Github fail on path resolution
# Bug report

## Describe the bug

I am trying to setup deployment from a persistent branch using the "Branching" feature and the Github integration.

Deploys work fine from my machine with `supabase functions deploy`, as they do if I run my own CI/CD. 

I had high hopes with the github integration for branches, but for some reason, at bundling time, the import path resolution doesn't work :

Image

## To Reproduce

not sure... But include relative paths or @paths in supabase edge function definitions and try to deploy using the github integration.

## Expected behavior

The functions are bundled and deployed

Tanusree-samanta:
Closes supabase/supabase#47000

## What changed
- `ShouldUseDenoJsonDiscovery` now checks that the import map file
  actually exists before skipping the `--import-map` flag
- Added test coverage for missing file, existing file, and non-deno
  import map cases

## Why
The GitHub branch deploy pipeline runs in a different working
directory than local `supabase functions deploy`, so Deno's
auto-discovery of `deno.json` silently failed, breaking `@backend/`
style path aliases. This mirrors the explicit resolution already
used locally.

## How to test
- Deploy an Edge Function with path aliases via GitHub branching
- Verify `@backend/` aliases resolve correctly

### DIFF
--- a/apps/cli-go/pkg/function/bundle.go
+++ b/apps/cli-go/pkg/function/bundle.go
@@ -91,7 +91,20 @@ func (b *nativeBundler) Bundle(ctx context.Context, slug, entrypoint, importMap 
 }
 
 func ShouldUseDenoJsonDiscovery(entrypoint, importMap string) bool {
-	return isDeno(filepath.Base(importMap)) && filepath.Dir(importMap) == filepath.Dir(entrypoint)
+	if !isDeno(filepath.Base(importMap)) {
+		return false
+	}
+	if filepath.Dir(importMap) != filepath.Dir(entrypoint) {
+		return false
+	}
+	// Only rely on Deno auto-discovery if the import map file
+	// actually exists at the expected path. In GitHub branch
+	// deploy pipelines, the working directory may differ from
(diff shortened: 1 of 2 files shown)


Tanusree-samanta:
I've opened a PR that addresses this: [link to your PR]

Root cause: `ShouldUseDenoJsonDiscovery` in `pkg/function/bundle.go`
assumed Deno would auto-discover `deno.json` whenever it sat in the
same directory as the entrypoint, so it skipped passing `--import-map`
explicitly. That assumption holds for local `supabase functions deploy`,
but breaks in the GitHub branch deploy pipeline where the working
directory differs, causing `@backend/` style path aliases

jgoux:
Thanks for your contribution. 🙏 

This issue has to be handled in our branching service codebase rather than in the CLI.
I opened a PR in our branching service in order to address it.

PR #5660, opened 2026-06-23
feat(cli): add registry fallbacks for Docker image pulls

jgoux:
## Summary

- When no registry override is set, image pulls now try **ECR → GHCR → Docker Hub** in sequence instead of failing on the first error
- Rate-limit errors (429, `toomanyrequests`) from one registry automatically fall through to the next candidate rather than bubbling to the user
- An explicit `SUPABASE_INTERNAL_IMAGE_REGISTRY` override still restricts pulls to the single configured registry, preserving existing mirror behaviour

## Changes

**Go (`apps/cli-go`)**
- `GetRegistryImageUrls(imageName)` — returns the ordered fallback candidate list
- `GetRegistryAuthForImage(imageTag)` — per-registry auth lookup backed by a `sync.Map` (replaces the single `sync.Once` that only cached one registry's credentials)
- `DockerResolveImageIfNotCached` — tries each candidate for a local inspect hit, then for a pull, and returns the resolved URL; callers use it to set `config.Image` directly
- `HasRegistryOverride()` helper for branching between single-registry and multi-fallback paths

**TypeScript (`apps/cli`)**
- `legacyGetRegistryImageUrlCandidates(imageName)` — mirrors the same ECR/GHCR/Docker Hub logic in `legacy-docker-registry.ts`
- `legacy-docker-run.layer.ts` — `resolveImage` tries candidates with per-attempt retries on transient patterns; `withResolvedImage` injects the result into `LegacyDockerRunOpts` before `runCapture`, `runStream`, and `runInteractive`


apps/cli/src/legacy/shared/legacy-docker-run.layer.ts
```diff
@@ -47,14 +62,125 @@ export const legacyDockerRunLayer: Layer.Layer<
...
+            if (delay === undefined) {
+              break;
+            }
+            yield* Effect.sleep(`${delay} millis`);
+          }
+        }
+
+        return yield* Effect.fail(
+          new LegacyDockerRunError({
+            message: `failed to pull docker image from all registries: ${failures.join("; ")}`,
```
avallete, line 172: **"Docker not installed" no longer surfaces the install hint.** `withResolvedImage` now runs *before* the actual `docker run` spawn in `run`/`runCapture`/`runStream`. When the `docker` binary is missing or the daemon is down, `hasLocalImage` swallows the spawn failure (`Effect.catch(() => false)`) and `pullImage` maps the spawn error to `new Error("spawn")`, so every candidate "fails" and the user lands here with:

```
failed to pull docker image from all registries: public.ecr.aws/supabase/...: attempt 1: Error: spawn; ghcr.io/supabase/...; supabase/...
```

instead of the friendly `SUGGEST_DOCKER_INSTALL` ("Docker Desktop is a prerequisite…") message that the real `run`/`runCapture` spawn paths deliberately emit. A first-time user without Docker now gets a confusing, repeated `Error: spawn` string with no install guidance. Consider detecting the spawn failure here and surfacing `SUGGEST_DOCKER_INSTALL`.

---

apps/cli/src/legacy/shared/legacy-docker-run.layer.ts
```diff
@@ -47,14 +62,125 @@ export const legacyDockerRunLayer: Layer.Layer<
...
+        Effect.catch(() => Effect.succeed(false)),
+      );
+
+    const pullImage = (
+      image: string,
+    ): Effect.Effect<{ readonly exitCode: number; readonly stderr: string }, Error> =>
+      Effect.gen(function* () {
+        const command = ChildProcess.make("docker", ["pull", image], {
+          stdin: "inherit",
+          stdout: "pipe",
```
avallete, line 77: **Pull progress is now hidden, so a first run appears to hang.** `pullImage` pipes and buffers stdout/stderr and only surfaces them on failure. Previously these methods didn't pull explicitly — `docker run` auto-pulled and streamed progress to the inherited terminal (for `run`) or the piped/teed stderr; the Go CLI streams pull progress via `jsonmessage.DisplayJSONMessagesToStream`. Now the explicit pre-pull is silent.

Scenario: `supabase test db` (pg_prove) or `db dump` on a machine that hasn't cached the image shows no output for the entire multi-hundred-MB pull, looking frozen, then suddenly proceeds — a regression from the inline progress users see today. Consider teeing pull progress to stderr (at least for the interactive `run` path).

---

apps/cli-go/internal/utils/docker.go
```diff
@@ -157,32 +157,56 @@ func CliProjectFilter(projectId string) filters.Args {
...
+func getRegistryAuth(registry string) string {
+	// Docker stores Hub credentials under the legacy index server, which
+	// resolves to the `index.docker.io` hostname; a bare `docker.io` lookup
+	// never matches them. Normalising here also routes images Docker treats as
+	// Hub references (a registry override or repository segment without a dot,
+	// colon, or `localhost`) to the right credentials.
+	if registry == "docker.io" {
+		registry = "index.docker.io"
+	}
+	if auth, ok := registryAuth.Load(registry); ok {
```
avallete: **`sync.Map` load-then-store drops the `sync.Once` single-load guarantee.** The comment below claims caching prevents re-reading the Docker config and "re-print[ing] the warning". That holds for sequential reuse, but this is a non-atomic `Load` → `loadRegistryAuth` → `Store`. Under the concurrent image pulls `supabase start` issues (via `WaitAll`/errgroup), two goroutines requesting the same registry can both miss and both call `loadRegistryAuth`, double-reading the config file and double-printing `Failed to load registry credentials`. Low severity, but it contradicts the stated invariant — `LoadOrStore` with a memoized/once-per-key loader would restore single-load semantics.

---

apps/cli-go/internal/utils/docker.go
```diff
@@ -236,27 +309,43 @@ func DockerImagePullWithRetry(ctx context.Context, image string, retries int) er
...
-	return DockerImagePullWithRetry(ctx, imageUrl, 2)
+	var pullErrors []string
+	for _, imageUrl := range imageUrls {
+		if err := DockerImagePullWithRetry(ctx, imageUrl, 2); err == nil {
+			return imageUrl, nil
+		} else {
+			pullErrors = append(pullErrors, fmt.Sprintf("%s: %v", imageUrl, err))
+		}
+	}
+	return "", errors.Errorf("failed to pull docker image from all registries: %s", strings.Join(pullErrors, "; "))
```
avallete: **Aggregated pull error breaks the `IsErrConnectionFailed` unwrap chain.** This collapses the per-registry errors into a string with `%v`/`%s` and a non-wrapping `errors.Errorf` (no `%w`). `DockerStart` relies on `client.IsErrConnectionFailed(err)` to set the `suggestDockerInstall` hint — that unwrap returns `false` for any connection failure that surfaces from the pull loop, since the cause chain is flattened to text. In practice a dead daemon is usually caught earlier at `ImageInspect` (line 322, which does wrap with `%w`), so this is an edge case, but if a connection failure first surfaces during the pull loop the install suggestion is silently lost. Consider wrapping the first/representative error with `%w`.

---

avallete (commented):
Reviewed the registry-fallback changes. The core ECR → GHCR → Docker Hub fallback logic, candidate construction, dedupe, override short-circuit, and the `index.docker.io` auth normalization all look correct and match between the Go and TS implementations, with good test coverage. No correctness bugs in the fallback algorithm itself.

A few quality/parity issues to consider (inline), all in the surrounding plumbing rather than the fallback logic:

1. **TS** — the new pre-pull path bypasses the `SUGGEST_DOCKER_INSTALL` hint when Docker is missing/down, replacing it with a verbose generic "failed to pull from all registries" message.
2. **TS** — image-pull progress is now buffered and hidden, so first-run pulls of large images appear to hang.
3. **Go** — `getRegistryAuth`'s `sync.Map` load-then-store loses the old `sync.Once` single-load guarantee under concurrent pulls.
4. **Go** — the aggregated pull error flattens causes with `%v`, breaking `client.IsErrConnectionFailed` unwrapping in `DockerStart` (edge case, since inspect usually catches a dead daemon first).

Minor (not inline): `apps/cli/src/legacy/commands/test/db/db.handler.ts:175` passes an already-rewritten `legacyGetRegistryImageUrl(...)` into `docker.run`, which `resolveImage` then re-parses — idempotent, but the rewrite now lives in two places.

---

apps/cli/src/legacy/shared/legacy-docker-run.layer.ts
jgoux: Confirmed. With the pre-pull now running before the `docker run` spawn, a missing/unusable runtime surfaced the aggregate pull error instead of the install hint. `resolveImage` now short-circuits on a spawn failure — an effect failure, as opposed to a non-zero exit which returns a value — and fails with `SUGGEST_DOCKER_INSTALL` rather than trying (and reporting) every registry candidate. Fixed in 96bdcef0.

apps/cli/src/legacy/shared/legacy-docker-run.layer.ts
jgoux: Fixed in 96bdcef0. `pullImage` now tees both stdout and stderr to the parent terminal in real time (mirroring Go's `jsonmessage.DisplayJSONMessagesToStream`), keeping the buffered copies only for retry classification and error reporting. Progress is written to stderr so it never corrupts the captured `db dump` stdout, and each stream is decoded separately to avoid splitting a multi-byte sequence across interleaved chunks.

apps/cli-go/internal/utils/docker.go
jgoux: Valid — the `Load` → `loadRegistryAuth` → `Store` was non-atomic, so concurrent pulls for the same registry could both miss and both load. Switched to `LoadOrStore` of a `*registryAuthEntry` that carries a per-key `sync.Once`, so `loadRegistryAuth` runs exactly once per registry even under the concurrent pulls `supabase start` issues. Fixed in 96bdcef0.

apps/cli-go/internal/utils/docker.go
jgoux: Fixed in 96bdcef0. The aggregate now wraps a representative error with `%w` (preferring one that `client.IsErrConnectionFailed` already matches), so the connection-failed type stays in the unwrap chain and `DockerStart` still sets the install hint. Confirmed go-errors `Errorf` wraps via stdlib `fmt.Errorf`, so the `errors.As` walk in `IsErrConnectionFailed` traverses it.

avallete (approved):

PR #5667, opened 2026-06-23
feat(cli): port supabase storage ls/cp/mv/rm to native TypeScript

Coly010:
## What changed

Replaces the Phase 0 `LegacyGoProxy` shims for `storage ls/cp/mv/rm` with native Effect implementations that talk to the Storage gateway directly (`apps/cli-go/internal/storage` as the reference), and adds TS-only `--output-format json|stream-json` on top of byte-faithful Go text output.

- **`ls`** — pagination + recursive BFS over buckets/objects.
- **`cp`** — 4-way scheme branch (download/upload, single/recursive), dst-key resolution, bucket auto-create on `Bucket not found`, content-type sniff/refine, `--cache-control`/`--content-type`/`--jobs`.
- **`mv`** — single move with recursive per-object BFS fallback on `not_found`.
- **`rm`** — group-by-bucket, confirm prompt, chunked deletes (1000), recursive directory + bucket removal.

## Shared-infra promotion (hoist-before-duplicate)

The Storage gateway, credential derivation, bucket-config helpers, content-type sniffing, the gateway runtime layer, and the yes/no prompt are promoted out of `seed buckets` into `legacy/shared/`. `seed buckets` is refactored to consume them in the same change (its existing suite covers the regression).

## Reviewer notes

- **`--linked`/`--local` are per-leaf, not group scoped globals.** Effect CLI requires global-flag names to be unique tree-wide and `seed` already owns `linked`/`local` with opposite Go defaults, so two groups can't both declare them (caught by the bundled-binary smoke test). Cost vs Go's persistent flags: the flags must follow the subcommand token (`storage ls --local`, not `storage --local ls`) — the same shape the `db` family uses. Documented in `storage.flags.ts` + each `SIDE_EFFECTS.md`.
- **Parity over hardening:** a multi-perspective review flagged a path-traversal write primitive in recursive download (remote object names → local paths). Go's `filepath.Join` has no boundary guard, so adding one would diverge from the strict 1:1 port; left identical to Go (noted for a future, explicit hardening).
- Porting tracker flipped to `ported`; the bogus `--copy-metadata` cp note removed (no such Go flag).

## Related

Closes CLI-1321


jgoux (commented):
Reviewed the storage `ls/cp/mv/rm` port end-to-end against the Go reference (parity, shared-infra hoist, Effect/TS correctness, security, tests). Overall this is strong, faithful work — pagination, BFS ordering, the cp 4-way scheme matrix, dst-key resolution, bucket auto-create, mv recursive fallback, and all success/error strings match Go, and the `seed buckets` hoist is behavior-preserving. No behavior-breaking parity divergence found.

Leaving inline notes on two things worth addressing before merge — (1) the path-traversal write primitive in recursive download, and (2) a test-coverage asymmetry where cp/mv/rm lack the linked-path, HTTP-error, and `stream-json` tests that `ls` already has — plus a couple of minor clarity nits. Everything else verified correct.

apps/cli/src/legacy/commands/storage/cp/cp.handler.ts
```diff
@@ -1,18 +1,485 @@
...
+    const iterError = yield* legacyIterateStoragePathsAll(
+      gateway,
+      output,
+      remotePath,
+      (objectPath) =>
+        Effect.gen(function* () {
+          const relPath = objectPath.startsWith(remotePath)
+            ? objectPath.slice(remotePath.length)
+            : objectPath;
+          const dstPath = path.join(localPath, relPath);
```
jgoux, line 236: **Security — path traversal in recursive download.** `objectPath` (and therefore `relPath`) is server-controlled, and `path.join` normalizes `..`, so a malicious/compromised bucket returning a name like `../../../etc/...` resolves *outside* `localPath`. `fs.open(dstPath, { flag: "w" })` below is truncating (not `O_EXCL`) and `makeDirIfNotExist(path.dirname(dstPath))` even creates the parent dirs — so this is a write/overwrite primitive. (`downloadSingle` is fine; its path is user-supplied.)

I understand the parity argument (Go's `filepath.Join` has no guard either), and the blast radius is gated (`--experimental` + `cp -r` + remote→local + hostile endpoint), so this isn't high-severity. But re-introducing a traversal primitive in a fresh, security-reviewable implementation is worth hardening: after computing `dstPath`, assert containment (resolve both, require `resolved.startsWith(localPathResolved + sep)`) and fail with `LegacyStorageFileError` otherwise. At minimum, document the accepted risk in `cp/SIDE_EFFECTS.md`, which currently says nothing about traversal.

apps/cli/src/legacy/commands/storage/cp/cp.handler.ts
```diff
@@ -1,18 +1,485 @@
...
+  const cliConfig = yield* LegacyCliConfig;
+  const telemetryState = yield* LegacyTelemetryState;
+  const linkedProjectCache = yield* LegacyLinkedProjectCache;
+  const resolver = yield* LegacyProjectRefResolver;
+  const fs = yield* FileSystem.FileSystem;
+  const path = yield* Path.Path;
+  const runtimeInfo = yield* RuntimeInfo;
+
+  const jobsFlag = Option.getOrElse(flags.jobs, () => 1);
+  const jobs = jobsFlag < 1 ? 1 : jobsFlag;
```
jgoux, line 74: Nice — this clamp actually *fixes* a latent Go deadlock: `NewJobQueue(0)` makes an unbuffered channel with the priming loop running zero times, so the first `Put` blocks forever. Worth a one-line comment noting the `< 1 → 1` coercion is an intentional deviation from Go (which accepts `--jobs 0`) so a future reader doesn't "restore parity" by removing it.

apps/cli/src/legacy/commands/storage/cp/cp.handler.ts
```diff
@@ -1,18 +1,485 @@
...
+  root: string,
+  dir: string,
+  out: Array<UploadFile>,
+): Effect.Effect<void, PlatformError> =>
+  Effect.gen(function* () {
+    const names = [...(yield* fs.readDirectory(dir))].sort();
+    for (const name of names) {
+      const abs = path.join(dir, name);
+      // afero.Walk uses Lstat (no-follow); a symlink is not regular → skipped.
+      const isSymlink = yield* fs.readLink(abs).pipe(
```
jgoux, line 477: Minor: this treats *any* `readLink` failure as "not a symlink." For a regular file the expected `EINVAL` is correct, but it also swallows genuine IO/permission errors that Go's `Lstat`-based `afero.Walk` would surface. Low impact, but slightly less faithful than Go — consider only treating `EINVAL` as not-a-symlink and propagating other errno values.

apps/cli/src/legacy/commands/storage/cp/cp.integration.test.ts
```diff
@@ -0,0 +1,431 @@
...
+
+function prefixOf(body: unknown): string {
+  return typeof body === "object" &&
+    body !== null &&
+    typeof (body as { prefix?: unknown }).prefix === "string"
+    ? (body as { prefix: string }).prefix
+    : "";
+}
+
+describe("legacy storage cp", () => {
```
jgoux, line 48: Coverage gap vs. `ls.integration.test.ts`: cp is missing symmetrical tests that `ls` already has —
- **linked-path routing** (`local: false`): the most common real-world flow; drives API-key resolution + project-ref loading. Untested here.
- **HTTP error propagation** (non-200 from the gateway).
- **`--output-format stream-json`** (only `json` is tested).
- **`--jobs`**: never exercised (clamping at `--jobs 0`, parallel dispatch).

Adding a linked-path test mirroring `ls` would close the biggest gap and gives a natural hook for the error/stream-json cases in the same pass.

Also minor: a couple of test names leak implementation detail (`…(O_EXCL)`, `…without x-upsert`) — CLAUDE.md asks for scenario/user-intent names (e.g. "refuses to overwrite an existing local file").

apps/cli/src/legacy/commands/storage/mv/mv.integration.test.ts
```diff
@@ -0,0 +1,238 @@
...
+  return {
+    src: opts.src,
+    dst: opts.dst,
+    recursive: opts.recursive ?? false,
+    linked: true,
+    local: opts.local ?? true,
+  };
+}
+
+describe("legacy storage mv", () => {
```
jgoux, line 30: Coverage gap vs. `ls.integration.test.ts`: mv is missing the **linked-path** (`local: false`), **HTTP error propagation**, and **`--output-format stream-json`** tests that `ls` has. Two mv-specific additions worth including:
- the recursive `not_found` fallback is tested only with `local: true` — exercise it on the linked path too.
- the recursive fallback emits `{ message: "", moved }` in json mode, but no test asserts the `moved` count (json test only covers the direct-move `{ message }` path).

apps/cli/src/legacy/commands/storage/rm/rm.integration.test.ts
```diff
@@ -0,0 +1,421 @@
...
+
+function prefixCount(body: unknown): number {
+  return typeof body === "object" &&
+    body !== null &&
+    Array.isArray((body as { prefixes?: unknown }).prefixes)
+    ? (body as { prefixes: unknown[] }).prefixes.length
+    : -1;
+}
+
+describe("legacy storage rm", () => {
```
jgoux, line 25: Coverage gap vs. `ls.integration.test.ts`: rm is missing the **linked-path** (`local: false`), **HTTP error propagation**, and **`--output-format stream-json`** tests that `ls` has. Also: "tolerates a missing bucket on delete" covers a 404 on the *bucket* DELETE, but the *object* DELETE always returns success in the mock — object-deletion failure is uncovered at the handler level.

apps/cli/src/legacy/commands/storage/cp/cp.handler.ts
Coly010: Good catch — your read of the mechanics is correct.

I've taken the "at minimum" option (documenting the accepted risk) rather than adding the containment check, and the reason is parity. Go's `DownloadStorageObjectAll` builds the destination as `filepath.Join(localPath, filepath.FromSlash(relPath))` with `relPath` coming straight from the server-returned object name and **no** containment guard (`apps/cli-go/internal/storage/cp/cp.go:72-73`), then `MkdirIfNotExistFS(filepath.Dir(dstPath))` (`:80`) and opens `O_WRONLY|O_CREATE|O_TRUNC` (`:84`). The object names are never sanitized upstream either (`internal/storage/ls/ls.go`). A hard `resolved.startsWith(localPath + sep)` check that fails the download would make the TS port reject something the Go CLI happily writes — a stdout/exit-code divergence the legacy shell's strict 1:1 contract doesn't allow.

So I've documented it in `cp/SIDE_EFFECTS.md` (which was silent on traversal): `path.join` normalizes `..` exactly like Go's `filepath.Join`, parent dirs are `mkdir -p`'d and files open `O_TRUNC` (a write/overwrite primitive), it matches Go exactly and is intentionally unguarded, and the blast radius is gated behind `--experimental` + `cp -r` + remote→local + a hostile endpoint. `downloadSingle` is noted as unaffected (user-supplied path, `O_EXCL` `wx`). Commit 938bc372.

If we decide the hardening is worth doing for real, I'd rather treat it as a deliberate, separately-tracked divergence from Go than fold it into this strict port — happy to open a follow-up if you'd like to push for it.

apps/cli/src/legacy/commands/storage/cp/cp.handler.ts
Coly010: Accepted — your reading of the Go side is exactly right. `--jobs` is a `uint` in Go (`apps/cli-go/cmd/storage.go`), so `--jobs 0` is accepted and flows into `NewJobQueue(0)`, whose unbuffered channel + zero-run priming loop deadlocks the first `Put` (`apps/cli-go/pkg/queue/queue.go`). I've added a comment above the clamp pointing at `queue.go` and flagging that the `< 1 → 1` coercion is an intentional deviation that should not be "restored to parity" by removing it. Commit 5db60ed3.

apps/cli/src/legacy/commands/storage/cp/cp.integration.test.ts
Coly010: Agreed — accepted. The asymmetry was real: the linked branch (`cp.handler.ts:79`) and the `--jobs` clamp were untested, and only the `json` arm of the machine-format result was exercised. Added to `cp.integration.test.ts` (commit ae539bb7):

- a linked-path upload test mirroring `ls` (asserts the request hits `https://<ref>.supabase.co`, plus the telemetry-flush + linked-cache writes);
- a 503-propagation test on the upload route;
- a `stream-json` result test;
- `--jobs 0` (clamp) and `--jobs 2` (parallel download) cases.

Also renamed the implementation-leaking names per `apps/cli/CLAUDE.md`: "downloads a single remote object to a new local file", "refuses to overwrite an existing local file on a single download", and "uploads a single local file with a sniffed content-type" (the x-upsert / content-type checks stay as assertions). I kept the `errors.Join parity` name — that parenthetical documents the specific Go behavior the test pins (`cp.go:96`) rather than leaking a flag.

apps/cli/src/legacy/commands/storage/mv/mv.integration.test.ts
Coly010: Accepted — both mv-specific points mapped to real gaps. Added to `mv.integration.test.ts` (commit 3a33a96a):

- a linked-path direct-move test (host + the telemetry/linked-cache `Effect.ensuring` wiring at `mv.handler.ts:96-100`);
- a 503-propagation test with `--recursive` set, proving only a `not_found` body triggers the fallback (`mv.handler.ts:71-77`);
- a linked-path recursive `not_found` fallback test;
- an assertion on the `{ message: "", moved }` result (`mv.handler.ts:90-93`), exercised in `stream-json` mode — which also closes the stream-json gap.

apps/cli/src/legacy/commands/storage/rm/rm.integration.test.ts
Coly010: Accepted. Added to `rm.integration.test.ts` (commit cd6ba61c):

- object-DELETE failure — a **500** (not 404), so it escapes the `Bucket not found` tolerance branch and propagates through `deleteObjects` (`rm.handler.ts:154`), asserting `Error status 500` + failure;
- a 503 on the `-r` bucket listing (`rm.handler.ts:98`), asserting failure with no DELETE attempted;
- a linked-path test (host + telemetry/linked-cache at `rm.handler.ts:137-139`);
- a `stream-json` result test.

Left the existing "tolerates a missing bucket on delete" test as-is — it covers the distinct 404-on-*bucket*-DELETE branch, so the new case is the object-DELETE failure you flagged.

jgoux (approved):

Issue #5556, opened 2026-06-11
Getting SIGKILL error


tlherysr:
**Describe the bug**
I cannot run `supabase` command on my Mac. Getting SIGKILL signal.

**To Reproduce**
1. Install supabase with this command: `brew install supabase`
2. Run `supabase --help` after installation.
3. See error

**Expected behavior**
Supabase cli should print the help text.

**Screenshots**
<img width="1348" height="547" alt="Image" src="https://github.com/user-attachments/assets/40203878-dbca-4bc4-8dfc-f7dfe635a57f" />

**System information**
Rerun the failing command with `--create-ticket` flag.
 - Ticket ID: I cannot even run the cli tool so cannot run it with --create-ticket flag.
 - Version of OS: MacOS 27 Developer Beta
 - Version of CLI: 2.105.0
 - Version of Docker: Docker version 29.5.3
 - Versions of services: I cannot run supabase command.

veilofsecurity:
Same issue. Tried latest beta and no change.

avallete:
Hi! Thanks for the report. Since you're on the macOS 27 Developer Beta (and this doesn't reproduce on stable macOS 26.x), we suspect the new OS is killing the binary at launch most likely due to stricter code-signature enforcement or the Rosetta 2 removal in Golden Gate. Could you run the following commands and paste the output so we can confirm which one it is?

**1. Check the binary architecture vs your machine:**

```bash
file $(which supabase)
uname -m
```

**2. Check how the binary is signed:**

```bash
codesign -dvv $(which supabase)
```

**3. Get the exact kill reason from the system log** (run `supabase --help` first, then immediately):

```bash
log show --last 2m --predicate 'sender == "AppleMobileFileIntegrity" OR process == "supabase"' --info
```

If macOS generated a crash report, it would also help to attach the most recent `supabase-*.ips` file from `~/Library/Logs/DiagnosticReports/`.

**In the meantime, two workarounds you can try:**

- If step 1 shows `x86_64` (Intel binary on Apple Silicon): macOS 27 uninstalls Rosetta 2 on upgrade. Either reinstall it with `softwareupdate --install-rosetta`, or better, reinstall the CLI through an ARM-native Homebrew (`/opt/homebrew`). Make sure you're on Homebrew 6.0.0+ (`brew --version`), which is the first version with macOS 27 support.
- If step 2 shows `linker-signed` / no TeamIdentifier: try re-signing the binary in place:

  ```bash
[...]

PR #5675, opened 2026-06-24
ci(cli): add macOS code signing with rcodesign

avallete:
Implement Phase 1 of macOS code signing to fix the SIGKILL issue on macOS 26+ (CLI-1621). The Bun SFE and Go sidecar binaries are now signed with a full ad-hoc signature during the build pipeline, replacing the degenerate linker-signed signature that AMFI rejects.

## Changes

- **Build pipeline signing**: Added `signDarwinBinaries()` and `resolveSignMode()` to `apps/cli/scripts/build.ts` to sign macOS binaries (`supabase` and `supabase-go`) with `rcodesign` between compilation and archiving. This ensures all distribution channels (npm, Homebrew, GitHub Releases) ship the signed bytes.

- **CI integration**: Updated `.github/workflows/build-cli-artifacts.yml` to install `rcodesign` v0.29.0 (pinned with sha256), set `SUPABASE_CLI_REQUIRE_SIGNING=1` to enforce signing in release builds, and verify signatures post-build using `rcodesign print-signature-info`.

- **Smoke test verification**: Extended `apps/cli/tests/smoke-test-macos.ts` with native signature verification via new `verifyMacSignature()` helper in `apps/cli/tests/helpers/macos-signature.ts`. On macOS runners, this checks the signature is valid, carries the correct identifier (`com.supabase.cli` / `com.supabase.cli-go`), and is no longer linker-signed.

- **Documentation**: Added ADR 0013 documenting the decision, rationale, and Phase 2 roadmap (Developer ID + notarization). Updated `release-process.md` and `binary-distribution.md` to describe the signing step and its role in the release pipeline.

## Implementation details

- **No Apple credentials required for Phase 1**: Full ad-hoc signatures are self-contained and do not require an Apple Developer ID. This fixes the SIGKILL without blocking on account provisioning.

- **Linux-only signing**: `rcodesign` runs on the existing Linux build runner, avoiding a macOS job and pipeline split. Verification happens on macOS smoke-test runners.

- **Graceful degradation**: Local builds without `rcodesign` warn and skip signing (unless `SUPABASE_CLI_REQUIRE_SIGNING=1` is set), allowing contributors to build unsigned binaries for testing.

- **Identifiers**: `com.supabase.cli` for the Bun SFE, `com.supabase.cli-go` for the Go sidecar, configured in `MACOS_IDENTIFIERS` map.

See ADR 0013 for full context, Phase 2 roadmap (Developer ID + notarization), and related follow-ups (Windows Authenticode, Linux cosign).

Related #5556 

https://claude.ai/code/session_01GnLjngbm48rMYVwn9Guduc


Coly010 (approved):
Awesome work 🎉 few things that claude highlighted just, nothing blocking though.

The CI signature check can let a mis-signed sidecar through. In the "Verify macOS signatures" step, grep -q 'identifier: com.supabase.cli' is a substring match, so it matches com.supabase.cli-go too. So if supabase-go ever got signed with com.supabase.cli by mistake, this would happily pass it — which kinda defeats the point of separate identifiers. Could we pick the expected id per binary and anchor the grep (com.supabase.cli$ vs com.supabase.cli-go$)?

signDarwinBinaries is reaching back into the module-level shell. The binary list is gated on if (shell === "legacy") reading the script-level const instead of anything passed in. Works fine today, but it's a little easy to miss, and it'll quietly need a tweak if the next shell ever gets a second sidecar. I'd just pass shell (or the resolved binary list) in as an arg so the function stands on its own.

The identifiers live in three places. MACOS_IDENTIFIERS in build.ts, EXPECTED_IDENTIFIERS in the test helper, and the hard-coded string in the workflow yaml — and nothing keeps them in sync. If one drifts we'd sign with one id and verify against another and only catch it on the macOS legs. Pulling them into one shared module that build.ts and the helper both import (and ideally pointing the CI check at the same thing) means one place to change — and it basically fixes the first point for free.

None of this blocks the actual fix though — the sign-on-Linux / verify-on-macOS approach looks solid.

PR #5681, opened 2026-06-24
fix(cli): wait for all image pulls before starting containers

### LINKED ISSUE #5068: "supabase start" doesn't wait for pulls to finish
# Bug report

- [X] I confirm this is a bug with Supabase, not with my own application. 
- [ ] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

This seems to be in the last few versions.

## To Reproduce

1. Delete all docker images on Windows
2. `npx supbaase start`
3. Watch it try to pull, then it starts before pull is finished... this problem has existed for last few versions

## Expected behavior

All Pulls should finish before supabase starts

## System information

- OS: [e.g. macOS, Windows] - Windows
- Browser (if applies) [e.g. chrome, safari] - n/a
- Version of supabase-js: [e.g. 6.0.2] - 2.89.1, 2.103.0

Image

- Version of Node.js: [e.g. 10.10.0] v24.12.0

## Additional context

GIF is almost a minute long, it does not freeze. Wait for it to finish.

Coly010:
## What changed

`supabase start` could start containers before their Docker images had finished downloading.

The command ran two uncoordinated image-acquisition paths:

1. A best-effort concurrent pre-pull (`pullImagesUsingCompose`) using docker-compose's `Pull` with `PullOptions{IgnoreFailures: true}`. It only targets the primary registry and, by design, **silently swallows per-image pull failures** (the `IgnoreFailures` flag is the hook that lets the registry fallback recover).
2. An authoritative lazy per-container pull inside `utils.DockerStart` → `DockerResolveImageIfNotCached` (multi-registry fallback: ECR → GHCR → Docker Hub).

So any image the concurrent pre-pull failed to cache — a transient registry/network/rate-limit hiccup, common on a fresh machine pulling 10+ images at once — was pulled **later**, during the `Starting database… / Starting containers…` phase. That is the "start doesn't wait for pulls" behaviour from the issue. The pre-pull was added in #4394, matching the reporter's "last few versions" regression window.

## The fix

Add `ensureImagesCached`, a completeness pass that runs immediately after the best-effort pre-pull and before any container starts. It resolves every project image through the **same** multi-registry fallback resolver `DockerStart` already uses (`DockerResolveImageIfNotCached`), fanned out concurrently via the existing `utils.WaitAll` primitive.

After it returns, every required image is guaranteed present in the local cache, so the per-container `DockerStart` calls become pure cache hits and never pull mid-start. On the happy path it is just N cheap image inspects; an image that genuinely cannot be pulled from any registry now fails the start cleanly **before** any container is created, instead of limping into a half-pulled start. The compose pre-pull (and its `IgnoreFailures`) is kept as the fast concurrent progress UI — it is simply no longer relied on for completeness.

## TypeScript port

The native-TS port already pulls in a preparation phase that is awaited before startup and fails hard on pull errors, so it does not have this bug. This PR adds regression guards locking that contract in:

- `Stack.unit.test.ts`: `stack.start()` aborts and starts zero containers when a docker pull fails.
- `prefetch.unit.test.ts`: preparation fails with `DockerPullError` when the whole registry fallback chain fails.

Fixes #5068

### DIFF
--- a/apps/cli-go/internal/start/start.go
+++ b/apps/cli-go/internal/start/start.go
@@ -244,7 +244,14 @@ func ensureImagesCached(ctx context.Context, project types.Project) error {
 		_, err := utils.DockerResolveImageIfNotCached(ctx, image)
 		return err
 	})
-	return errors.Join(result...)
+	// Set the install hint once, sequentially, after the concurrent resolve
+	// finishes, rather than from inside DockerResolveImageIfNotCached where the
+	// WaitAll goroutines would race on the CmdSuggestion global.
+	if err := errors.Join(result...); err != nil {
+		utils.SuggestDockerInstallIfConnectionFailed(err)
+		return err
+	}
+	return nil
 }
 
 func run(ctx context.Context, fsys afero.Fs, excludedContainers []string, dbConfig pgconn.Config, ignoreHealthCheck bool, options ...func(*pgx.ConnConfig)) error {
--- a/apps/cli-go/internal/start/start.go
+++ b/apps/cli-go/internal/start/start.go
@@ -215,6 +215,38 @@ func pullImagesUsingCompose(ctx context.Context, project types.Project) error {
 	return service.Pull(ctx, &project, api.PullOptions{IgnoreFailures: true})
 }
 
+// ensureImagesCached guarantees every image required by the project is present in
+// the local Docker cache before any container is started. The compose pre-pull is
+// best effort (PullOptions.IgnoreFailures) and only targets the primary registry,
+// so any image it skips would otherwise be pulled lazily by DockerStart while
+// containers are already starting. Resolving each image here, using the same
+// multi-registry fallback as DockerStart, keeps all pulls ahead of container start.
+//
+// project.Services[*].Image is the registry-normalized URL produced by GetServices
+// (via GetRegistryImageUrl), which DockerResolveImageIfNotCached expands back into
+// the same candidate set DockerStart later resolves against. On a warm cache this
+// is one ImageInspect per image (no pull, no output); that cost is intentional and
+// bounded by the fixed service count.
+func ensureImagesCached(ctx context.Context, project types.Project) error {
+	seen := make(map[string]struct{}, len(project.Services))
+	var images []string
+	for _, service := range project.Services {
+		if service.Image == "" {
+			continue
+		}
+		if _, ok := seen[service.Image]; ok {
+			continue
+		}
+		seen[service.Image] = struct{}{}
+		images = append(images, service.Image)
+	}
+	result := utils.WaitAll(images, func(image string) error {
+		_, err := utils.DockerResolveImageIfNotCached(ctx, image)
+		return err
+	})
+	return errors.Join(result...)
+}
+
 func run(ctx context.Context, fsys afero.Fs, excludedContainers []string, dbConfig pgconn.Config, ignoreHealthCheck bool, options ...func(*pgx.ConnConfig)) error {
 	excluded := make(map[string]bool)
 	for _, name := range excludedContainers {
@@ -236,6 +268,11 @@ func run(ctx context.Context, fsys afero.Fs, excludedContainers []string, dbConf
 		Services: utils.GetServices().Filter(notExcluded),
 	}
 	if err := pullImagesUsingCompose(ctx, project); err != nil {
+		return err
+	}
+	// Pull any images the best-effort compose pre-pull skipped, so that all pulls
+	// finish before containers start (https://github.com/supabase/cli/issues/5068).
+	if err := ensureImagesCached(ctx, project); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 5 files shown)


apps/cli-go/internal/utils/docker.go
```diff
@@ -329,6 +329,12 @@ func DockerResolveImageIfNotCached(ctx context.Context, imageName string) (strin
 		if _, err := Docker.ImageInspect(ctx, imageUrl); err == nil {
 			return imageUrl, nil
 		} else if !errdefs.IsNotFound(err) {
+			// Surface the install hint here too: callers like ensureImagesCached
+			// resolve images before DockerStart runs, so this is the first place a
+			// "Docker not running" failure is seen.
+			if client.IsErrConnectionFailed(err) {
+				CmdSuggestion = suggestDockerInstall
```
jgoux: **Potential `-race` data race (low severity, benign value).**

`ensureImagesCached` now calls `DockerResolveImageIfNotCached` concurrently across `WaitAll` goroutines, and this PR adds a write to the unsynchronized package-level global `CmdSuggestion` right here. If the Docker daemon is down, every goroutine hits this branch and writes concurrently — a data race per the Go memory model. It's behaviorally harmless (all goroutines write the identical constant), but `go test -race` can flag it, and the previous sequential call path didn't have it.

Options: guard with `sync.Once`/mutex, set the suggestion in the caller after aggregating errors, or accept it with a one-line comment acknowledging the benign race. Since it only fires when Docker is already down, accepting is defensible — just worth a conscious decision so it doesn't surface as a confusing race-detector failure later.

apps/cli-go/internal/start/start.go
```diff
@@ -215,6 +215,38 @@ func pullImagesUsingCompose(ctx context.Context, project types.Project) error {
...
+		if service.Image == "" {
+			continue
+		}
+		if _, ok := seen[service.Image]; ok {
+			continue
+		}
+		seen[service.Image] = struct{}{}
+		images = append(images, service.Image)
+	}
+	result := utils.WaitAll(images, func(image string) error {
```
jgoux, line 243: **Minor (non-blocking): no early cancellation on first failure.**

`WaitAll` waits for every goroutine regardless of failures, and `DockerImagePullWithRetry` retries with exponential backoff (~4s, ~8s sleeps). So if image A fails fast (e.g. 503) while image B is mid-retry, the user waits out B's full retry schedule before the aggregated error surfaces. This matches existing `WaitAll` semantics elsewhere, so it's not a regression — but a `ctx`-cancel-on-first-error would be a nice future refinement for faster failure feedback.

jgoux (commented):
## Review summary

Well-constructed, correct fix. The diagnosis is accurate and the fix reuses existing primitives (`WaitAll`, `DockerResolveImageIfNotCached`, `GetServices`) rather than reinventing them. I verified the load-bearing assumption: `GetRegistryImageUrls` only uses the last path segment to rebuild the candidate set, so feeding it an already-normalized `service.Image` is idempotent and produces exactly the set `DockerStart` later resolves against. Coverage also matches — `project.Services` is the same `notExcluded`-filtered set that gets started. Test coverage (Go `gock.IsDone()` invariant check + TS regression guards) is strong.

One real-but-benign concern (`CmdSuggestion` concurrent write) and one minor UX note inline. Both non-blocking; approving in spirit pending a conscious decision on the first.

apps/cli-go/internal/utils/docker.go
Coly010: Good catch — fixed in a244947c. You're right that this was a genuine write/write `-race` on the package-global `CmdSuggestion`: `WaitAll` fans out one goroutine per image (`internal/utils/docker.go:79`) and `ensureImagesCached` is the only concurrent caller, so the new write raced when the daemon was down. It was also redundant with `DockerStart`, which already set the identical hint at the caller after the resolve returns.

I went with your "set it in the caller" option, since it matches the existing `DockerStart` pattern:

- Dropped the write from `DockerResolveImageIfNotCached`.
- Extracted `utils.SuggestDockerInstallIfConnectionFailed(err)` so the connection-failed → hint mapping stays co-located in `utils` and the raw string isn't leaked across packages.
- Call it once on the joined error in `ensureImagesCached`; `DockerStart` now routes through the same helper too.

Detection still works on the aggregated error because `IsErrConnectionFailed` is `errors.As`, which traverses both the `%w` chain and `errors.Join`'s `Unwrap() []error`. Same user-facing hint, no race — `go test -race ./internal/start/... ./internal/utils/...` passes.

apps/cli-go/internal/start/start.go
Coly010: Agreed this is a fair future refinement, but I'm leaving it out of this PR. As you note, it isn't a regression: `WaitAll` (`internal/utils/docker.go:79`) has always waited for every goroutine and joined the errors, and all three call sites depend on that contract identically — this one, `DockerRemoveAll` (`docker.go:111`), and `restartServices` (`internal/db/reset/reset.go:229`, which even comments that it intentionally waits out excluded services).

Adding ctx-cancel-on-first-error means changing the shared `WaitAll` primitive (or forking a variant), so it's better as a separate, consistent codebase-wide change than a one-off bolted onto this targeted fix. Leaving this thread open as a tracked follow-up rather than resolving it.

jgoux (approved):

PR #5683, opened 2026-06-24
refactor(cli): harden macOS signature checks per #5675 review

avallete:
Follow-up to the merged macOS code-signing change (#5675), addressing the three non-blocking review points from @Coly010.

## Changes

1. **Exact identifier match instead of substring.** The previous checks used `includes("identifier: com.supabase.cli")`, which also matches `com.supabase.cli-go` — so a sidecar accidentally signed with the SFE's identifier would have passed. Both the build-time verification in `build.ts` and the macOS smoke-test helper now extract and compare the **whole** identifier value, so each binary is verified against exactly its own identifier.

2. **`signDarwinBinaries` no longer reaches into the module-level `shell`.** It now takes `shell` as a parameter and resolves its binary list via `darwinBinariesForShell(shell)`, so the function stands on its own and the legacy/next split lives in one place.

3. **Single source of truth for identifiers.** New `apps/cli/scripts/macos-signing.ts` exports `MACOS_IDENTIFIERS` plus `macIdentifierFor()` / `darwinBinariesForShell()` helpers, imported by both `build.ts` (signing) and the smoke-test helper (verification). The third copy — the hardcoded, substring-matching `Verify macOS signatures` step in `build-cli-artifacts.yml` — is **removed**: `build.ts` already verifies each signature against the shared source during the build and throws on mismatch (failing the job), so the separate step was redundant and was the remaining drift/substring risk. If you'd prefer to keep an explicit standalone CI verification step, I can re-add one that imports the shared module instead of hardcoding the identifier — let me know.

The sign-on-Linux / verify-on-macOS approach is unchanged.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GnLjngbm48rMYVwn9Guduc)_


jgoux:
@codex review

jgoux (approved):

PR #5686, opened 2026-06-24
feat(cli): Push auth email template HTML from content_path during config push

### LINKED ISSUE #5684: CaC: Support Deploying Auth Email Templates Specified in config.toml to Supabase Cloud projects
### Existing issues

- [x] I have searched the existing issues.

### Affected area

Auth

### Problem to solve

CLI has great support for deploying many features, even supports using email templates locally. Extending the capability of supabase config push to deploy email templates would continue to support solid deployment workflows. We're aiming to put as much as possible in our repo, so that it can be version controlled, accessed by agents, and reviewed through PRs.

### Proposed solution

Add support for deploying auth email templates via supabase CLI

### Alternatives considered

_No response_

### Additional context

_No response_

mittal-parth:

Enhancements

After the native config push port (#5489), only template subjects were pushed to the cloud API - content_path was parsed but the HTML files were never read, so mailer_templates_*_content was never sent.

## What is the new behavior?

This change loads the template and notification HTML from disk before the auth diff/PATCH step (templates from project root, notifications from supabase/) and includes the content in the auth update body.

## Additional context

Closes #5684


jgoux (approved):

PR #5105, opened 2026-04-20
feat(db): support .sql.gz seed files with automatic decompression

Basavaraj8143:
## Summary
This PR adds support for gzipped seed SQL files (`.sql.gz`) during database seeding.

### What’s included
- Auto-detect `.sql.gz` seed files by extension.
- Decompress gzipped SQL before parsing/executing.
- Reuse the same path for seed hashing so change detection works for compressed seeds.
- Add tests for:
  - pending gzipped seeds
  - seeding from gzipped seed files
  - seeding globals from gzipped files
- Update docs/config comments to mention `.sql.gz` support.

## Why
Large seed files are common in DB-first workflows. Compressed seed files reduce repo/storage size while keeping `db reset`/seed automation intact.

## Backward compatibility
- Existing `.sql` behavior is unchanged.
- `.sql.gz` is additive and auto-detected from extension.

PR #5692, opened 2026-06-25
feat(db): support gzipped SQL seeds

jgoux:
Adds support for configured seed files ending in `.sql.gz` during `db reset`.

The implementation decompresses gzip input before parsing and executing seed SQL, and hashes the decompressed SQL content for seed history so equivalent recompressions do not mark seeds dirty. Config templates, generated schema text, and reset docs now mention `.sql.gz` support.

This is an alternative implementation to #5105 in the current Go-backed `db reset` path. The current TypeScript command layer still delegates `db reset` to legacy Go; if native TypeScript seed loading lands separately, it should mirror the same decompressed-content hashing behavior.

### DIFF
--- a/apps/cli-go/docs/supabase/db/reset.md
+++ b/apps/cli-go/docs/supabase/db/reset.md
@@ -4,6 +4,6 @@ Resets the local database to a clean state.
 
 Requires the local development stack to be started by running `supabase start`.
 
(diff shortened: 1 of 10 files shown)


jgoux:
I'll come back to this later.

Issue #5735, opened 2026-06-30
Edge functions deployment failure in Supabase CLI new version


pAkalpa:
### Affected area

Edge Functions

### Supabase CLI version

2.108.0

### Operating system

macOS 26.5.2

### Installation method

Other

### Command

```shell
supabase functions deploy --no-verify-jwt
```

### Actual output

```shell
WARN: config section [inbucket] is deprecated. Please use [local_smtp] instead.
Bundling Function: platform
An error occurred in Effect.tryPromise
Try rerunning the command with --debug to troubleshoot the error.
```

### Expected behavior

With supabase CLI version `2.107.0` `supabase functions deploy --no-verify-jwt` works fine but version `2.108.0` throws an error when used command `supabase functions deploy --no-verify-jwt`

<img width="1349" height="825" alt="Image" src="https://github.com/user-attachments/assets/24cc574a-3b63-47a3-88fd-956b5c5a17d4" />

### Steps to reproduce

1. Install Deno deps.
2. Run command `supabase functions deploy --no-verify-jwt` to deploy edge functions.
3. Getting `An error occured in Effect.tryPromise` message.

### Docker and service versions

```shell
➜ docker --version       
Docker version 29.6.1, build 8900f1d330

➜ docker ps | rg supabase
13b8a255cc52   public.ecr.aws/supabase/postgres:17.6.1.121             "sh -c '\ncat <<'EOF'…"   About an hour ago   Up About an hour (healthy)   0.0.0.0:54322->5432/tcp, [::]:54322->5432/tcp                                                            supabase_db_medirai-dermatron-supabase
e35722448db2   public.ecr.aws/supabase/studio:2026.06.22-sha-2207d7f   "docker-entrypoint.s…"    3 hours ago         Up 3 hours (healthy)         0.0.0.0:54323->3000/tcp, [::]:54323->3000/tcp                                                            supabase_studio_medirai-dermatron-supabase
e621a3f9458c   public.ecr.aws/supabase/postgres-meta:v0.96.6           "docker-entrypoint.s…"    3 hours ago         Up 3 hours (healthy)         8080/tcp                                                                                                 supabase_pg_meta_medirai-dermatron-supabase
835bc20a2862   public.ecr.aws/supabase/edge-runtime:v1.74.2            "sh -c 'cat <<'EOF' …"    3 hours ago         Up 3 hours                   8081/tcp                                                                                                 supabase_edge_runtime_medirai-dermatron-supabase
[...]

Issue #4550, opened 2025-11-29
deno monorepo support does not support using dependencies in workspace packages


alita-moore:
I am getting the following error
```
Relative import path "axios" not prefixed with / or ./ or ../ and not in import map from "file:///Users/alitamoore/workspaces/monorepo/libs/core/dist/src/errors/core.d.ts"
```
when running an edge function locally. The dependency is imported in a workspace package into the function. That worksapce package has dependencies. The workspace package is defined using package.json. This workflow is supported by deno, however, supabase does not interpret the package.json of the workspace package.

Yet the package.json has the necessary dependencies defined, but the system seems to require that it be a deno.json. To work around it I need to manually insert the dependencies in the deno.json even though I'm just duplicating the dependencies between package.json and deno.json.

rdforte:
I am also experiencing the same issue. I am running a monorepo using Deno workspaces and I have a set of schemas in the `schema` directory that I would like to share across the `frontend` and my supabase edge function `supabase/functions/api`. The Deno linter seems to pick up the workspace just fine. It's only when I run `supabase start && supabase functions serve api` and hit the function on my local that I encounter this error.

Here is my current folder structure. I have omitted some of the folders and files for brevity

```
.
├── deno.json
├── deno.lock
├── frontend
├── schema
├── supabase
   ├── config.toml
   ├── functions
      └── api
          ├── deno.json
          ├── index.ts
```

any help on this would be much appreciated. I'm also happy to assist anyway I can.

alita-moore:
can this be re-opened please

alita-moore:
❤️

7ttp:
had a look and is reproducible...

curious if this is expected to be supported here for serve/deploy? 
thoughts? @avallete @jgoux

avallete:
The single-package case from #4549 (a function dir with its own `package.json` and no import map) is supported now. This issue is the harder, distinct one a function with a `deno.json` importing a **local workspace package whose deps live in that package's own** `package.json` and it's **not supported yet**.

The reason is a deliberate constraint in how we bundle today, not an oversight:

* `deno.json` and `package.json` are treated as **mutually exclusive** resolution modes. If an import map / `deno.json` is present, we set `DENO_NO_PACKAGE_JSON=1` (and `noNpm` on serve), so `package.json` is ignored, including a workspace member's. This mutual exclusion exists on purpose because the "both present" combination caused npm-resolution bugs.
[...]

PR #5755, opened 2026-07-01
fix(cli): function deploy paths

### LINKED ISSUE #3467: Functions which import from parent directory fail to deploy using api flag
**Describe the bug**
https://github.com/supabase/cli/issues/2533#issuecomment-2801993444

**To Reproduce**

1. Create `supabase/functions/slug/index.ts`
```typescript
import { handler } from "../../../../common/index.ts";

console.log("Hello from Functions!");

Deno.serve(handler);
```
2. Create `../common/index.ts`
```typescript
export const handler = async (req) => {
  const { name } = await req.json()
  const data = {
    message: `Hello ${name}!`,
  }

  return new Response(
    JSON.stringify(data),
    { headers: { "Content-Type": "application/json" } },
  )
}
```
3. Run `supabase functions deploy --use-api` and observe error
```bash
Uploading asset (update): supabase/functions/slug/index.ts
failed to read file: open ../common/index.ts: invalid argument
exit status 1
```

**Expected behavior**

**Screenshots**

**System information**
Rerun the failing command with `--create-ticket` flag.
 - Ticket ID: [e.g. ab1ac733e31e4f928a4d7c8402543712]
 - Version of OS: [e.g. Ubuntu 22.04]
 - Version of CLI: [e.g. v1.60.0]
 - Version of Docker: [e.g. v25.0.3]
 - Versions of services: [output from `supabase services` command]

**Additional context**
If applicable, add any other context about the problem here.
 - Version of supabase-js [e.g. v2.22.0]
 - Version of Node.js [e.g. v16.20.0]

7ttp:
## TL;DR

`functions deploy` was excluding monorepo imports outside `supabase/` 

this fixes it by using the nearest git root as the source boundary, following existing pattern 

## ref:

- Closes https://github.com/supabase/cli/issues/3467
- Related https://github.com/supabase/cli/issues/5735
- Related https://github.com/supabase/cli/issues/4550


apps/cli/src/shared/functions/deploy.ts
```diff
@@ -325,6 +325,29 @@ async function realpathIfExists(pathname: string) {
   }
 }
 
+async function findGitRoot(startPath: string) {
```
jgoux: Non-blocking: there are already a couple of similar `.git` root walkers in the CLI. Since source-root semantics now affect deploy and Docker bind behavior, it may be worth extracting this into a shared helper so `.git` file vs directory handling stays consistent across project state, signing-key, and functions code.

apps/cli/src/shared/functions/deploy.ts
```diff
@@ -887,31 +910,30 @@ async function resolveImportMapAllowedRoots(projectRoot: string, importMapPath:
...
-    const relativePath = toApiRelativePath(cwd, pathname);
+    const relativePath = toApiRelativePath(sourceRoot, pathname);
     await Effect.runPromise(outputRaw(`Uploading asset (${config.slug}): ${relativePath}\n`));
     form.append("file", new File([contents], relativePath));
   };
 
   const uploadAsset = async (pathname: string, contents: Uint8Array) => {
     const realPathname = await realpath(pathname);
-    if (!isContainedPath(realProjectRoot, realPathname)) {
+    if (!isContainedPath(realSourceRoot, realPathname)) {
```
jgoux, line 918: Non-blocking: now that the containment root is `sourceRoot` (nearest git root when present), the user-facing messages in this block still say `project root`. Could we rename them to `source root` or `git/source root` so monorepo users understand the actual boundary?

apps/cli/src/shared/functions/deploy.ts
```diff
@@ -1098,9 +1121,10 @@ export async function buildDockerBinds(
   const hostFunctionsDir = resolve(functionsDir);
   const hostOutputDir = resolve(outputDir);
   const projectRoot = resolve(functionsDir, "..", "..");
-  const realProjectRoot = await realpath(projectRoot);
+  const sourceRoot = await resolveFunctionsSourceRoot(projectRoot);
```
jgoux, line 1106: Non-blocking: `buildDockerBinds` is also used by `functions serve`, so this source-root expansion changes serve behavior as well as deploy behavior. Could we add a serve integration case for a repo-root workspace import, or call out explicitly if the serve behavior change is intentional but covered elsewhere?

jgoux (commented):
Overall this looks like the right scoped fix for the repo-root import problem in #3467. I left a few non-blocking comments around wording, shared helper drift, and serve coverage because the source boundary now affects more than just API deploy.

apps/cli/src/shared/functions/deploy.ts
7ttp: pulled this into a small shared helper and reused it across
 deploy, project state, and signing-key so the `.git` root handling stays consistent..

apps/cli/src/shared/functions/deploy.ts
7ttp: renamed 💚

apps/cli/src/shared/functions/deploy.ts
7ttp: added a serve integration case for that path as well...

Coly010 (approved):

PR #5761, opened 2026-07-02
chore(dev/onboarding): set up `mise`-based workflow for initial codebase setup

kanadgupta:

Adds a `mise.toml` along with enhancements to our `CONTRIBUTING.md` so first-time contributors have a quick and easy way to install all the correct versions of all of the required software (Node, Bun, Go, `golangci-lint`, pnpm) to get tests running.

I tried to minimize the amount of additional configuration, utilizing existing files (e.g., `.nvmrc`, `go.mod`, `package.json`) as the respective versioning sources of truth where applicable.

I also chose not to commit [a `mise.lock` lockfile](<https://mise.en.dev/dev-tools/mise-lock.html>) for simplicity's sake but I'm happy to add it if folks want to lean into `mise` further in this repo.

If someone clones the repo for the first time, they'll need to manually download all of the executables (Node, Bun, Go, golangci-lint) and look in a few different places to determine which versions to use.

## What is the new behavior?

Now, users can install `mise` and run `mise install` and it'll install the correct versions of all the required software:

https://github.com/user-attachments/assets/4cad1a5c-417c-403a-9ffb-57c88979902b


.github/actions/setup/action.yml
```diff
@@ -20,7 +20,7 @@ runs:
   steps:
     - name: Resolve Bun version
       shell: bash
-      run: echo "BUN_VERSION=1.3.13" >> "$GITHUB_ENV"
+      run: echo "BUN_VERSION=$(cat .bun-version)" >> "$GITHUB_ENV"
```
avallete: **praise**

That's nice.

mise.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[hooks]
+# installs the package manager (e.g., pnpm) for the project, if it is not already installed
+postinstall = "corepack enable"
+
+[monorepo]
+config_roots = [
+  "apps/*",
+  "packages/*",
+]
```
avallete: **remark**

This enables mise's experimental monorepo tasks feature, but the repo defines no mise tasks and there are no nested `mise.toml` for now. Also I wonder if there SHOULD be some ? I will defer to @Coly010 here but I think it might step on the foot of `Nx` to have `mise` and `nx` for tasks orchestration within projects ?

mise.toml
```diff
@@ -0,0 +1,20 @@
+monorepo_root = true
+
+[hooks]
+# installs the package manager (e.g., pnpm) for the project, if it is not already installed
+postinstall = "corepack enable"
```
avallete: **nitpick**
corepack is being unbundled from Node distributions going forward, so this hook will break on a future Node bump. Not blocking, but worth a comment noting it, or consider managing pnpm as a mise tool `(pnpm = "11.4.0")` instead of relying on corepack.

CONTRIBUTING.md
```diff
@@ -4,6 +4,50 @@ Bun monorepo for exploring the next generation of the Supabase CLI and local dev
...
+# macOS / Linux
+curl https://mise.run | sh
+
+# macOS via Homebrew
+brew install mise
+```
+
+See the [`mise` installation docs](https://mise.jdx.dev/getting-started.html) for other package managers (apt, dnf, cargo, npm, Windows, …).
+
+`mise` needs to hook into your shell so it can inject the right tool versions into your `PATH` as you move between directories. Follow the `mise activate` instructions [in this section](https://mise.en.dev/getting-started.html#activate-mise) to add the activation line for your shell to its startup file.
```
avallete: **nitpick**

Should we normalize to have the same domain for both links ?

```suggestion
`mise` needs to hook into your shell so it can inject the right tool versions into your `PATH` as you move between directories. Follow the `mise activate` instructions [in this section](https://mise.jdx.dev/getting-started.html#activate-mise) to add the activation line for your shell to its startup file.
```

avallete (commented):
Few minor comments and nitpicks, leaving approval to @Coly010 given my comment about NX/monorepo.

Thanks for this PR !


Also, the PR title I think should be `chore:` as it's only adding dev tools for contributors, nothing user facing in a release.

CONTRIBUTING.md
kanadgupta: Oh weird I missed that, thanks!

mise.toml
kanadgupta: Ah, didn't realize they're unbundling the two. I had an approach where I was specifying the `pnpm` version before, I'll revert to that!

mise.toml
kanadgupta: I'll remove this! For some odd reason I thought this was required to not throw warnings but evidently not. I imagine the task orchestration in `mise` pales in comparison to `nx`.

kanadgupta: https://github.com/supabase/cli/pull/5761/commits/16598eaa33ae1b83ff6a3dfdab4cc8f47f1486a3

mise.toml
kanadgupta: https://github.com/supabase/cli/pull/5761/commits/68862e363e45cdc26b174c70f942616cbde5d1aa

CONTRIBUTING.md
kanadgupta: https://github.com/supabase/cli/pull/5761/commits/b5d90ecb51245d388ffad9939f9f025db5598275

mise.toml
```diff
@@ -0,0 +1,20 @@
...
+  "packages/*",
+]
+
+[settings]
+# uses .bun-version and .nvmrc to determine the bun and node versions to use
+idiomatic_version_file_enable_tools = ["bun", "node"]
+
+[tools]
+# extracts the go version from `apps/cli-go/go.mod`
+go = "{{ exec(command=`awk '/^go / {print $2}' apps/cli-go/go.mod`) }}"
```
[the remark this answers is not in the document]
kanadgupta: @supabase/cli how much do we accommodate non-Unix developers in this repo? `mise` supports [OS-specific tool installation](https://mise.en.dev/dev-tools/#os-specific-tools) so I'm happy to split this out to be more OS-specific if needed.

CONTRIBUTING.md
```diff
@@ -4,6 +4,50 @@ Bun monorepo for exploring the next generation of the Supabase CLI and local dev
...
+See the [`mise` installation docs](https://mise.jdx.dev/getting-started.html) for other package managers (apt, dnf, cargo, npm, Windows, …).
+
+`mise` needs to hook into your shell so it can inject the right tool versions into your `PATH` as you move between directories. Follow the `mise activate` instructions [in this section](https://mise.jdx.dev/getting-started.html#activate-mise) to add the activation line for your shell to its startup file.
+
+#### Installing the pinned tool versions
+
+From the repo root:
+
+```sh
+mise install
```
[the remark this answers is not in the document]
kanadgupta: Nearly everyone running this will be doing so in an interactive TTY so I don't think we need to account for this

mise.toml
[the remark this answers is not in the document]
kanadgupta: Fixed in https://github.com/supabase/cli/commit/68862e363e45cdc26b174c70f942616cbde5d1aa and https://github.com/supabase/cli/pull/5761/commits/16598eaa33ae1b83ff6a3dfdab4cc8f47f1486a3

mise.toml
```diff
@@ -0,0 +1,20 @@
...
+golangci-lint = "2"
```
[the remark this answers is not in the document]
kanadgupta: Fixed in 47511ef255ae5dad9454c1a032873363ecbd16a9

mise.toml
```diff
@@ -0,0 +1,14 @@
...
+# extracts the pnpm version from `package.json` (in the root)
+pnpm = """{{ exec(command=`jq -r '.packageManager | split("@")[1]' package.json`) }}"""
```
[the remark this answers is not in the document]
kanadgupta: Fixed in 4471f6bd6891ca7bde1ee5aa350ab45b0b6c7d4f

mise.toml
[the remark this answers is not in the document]
jgoux: as it's more of an internal tooling matter, I would wait for a Windows user to complain before actually adding it. 😅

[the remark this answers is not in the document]
avallete: The best thing humans can do in a LLM/Agentic world is to say: "No" 😂

jgoux (approved):

mise.toml
```diff
@@ -0,0 +1,6 @@
+[tools]
+go = "1.25.11"
+"go:github.com/golangci/golangci-lint/v2/cmd/golangci-lint" = "2.1.6"
+
+[settings]
+idiomatic_version_file_enable_tools = ["node", "bun", "pnpm"]
```
kanadgupta, line 6: 🤦 did not realize that this option supported `pnpm` — nice catch!

.github/actions/setup/action.yml
```diff
@@ -9,56 +9,78 @@ inputs:
...
+          bun: readFileSync(".bun-version", "utf8").trim(),
+          mise: readFileSync("mise.toml", "utf8").trim(),
+          node: packageJson.devEngines?.runtime,
+          pnpm: packageJson.packageManager,
+        };
+        const digest = createHash("sha256")
+          .update(JSON.stringify(payload))
+          .digest("hex");
+        console.log(`hash=${digest}`);
+        JS
```
kanadgupta, line 40: I have an idea for simplifying this so we don't need to manually grab all of the source files to construct the cache key — see https://github.com/supabase/cli/pull/5778. Still needs testing!

kanadgupta (commented):
Excited to use `mise` in CI! I have an idea for simplifying the caching setup in https://github.com/supabase/cli/pull/5778 — will see if it actually does the trick.

PR #5764, opened 2026-07-02
fix(cli): CSV-split sso and postgres-config StringSlice flags

Coly010:
## Current Behavior

Go registers `sso add/update --domains/--add-domains/--remove-domains` and `postgres-config update/delete --config` as pflag `StringSliceVar`, which CSV-splits each occurrence (via `encoding/csv`) and accumulates across repeats. The legacy TS port only supported flag repetition (`Flag.atLeast(0)`), so a single comma-separated value like `--domains a.com,b.com` produced one malformed string (`"a.com,b.com"`) instead of two domains — sent to the API as-is, or (for `postgres-config`) failing the `key=value` split validation.

Fixes #CLI-1853

## Expected Behavior

`--domains a.com,b.com` (and `--config foo=bar,baz=qux`) now CSV-split per-occurrence and accumulate across repeats, matching Go's `pflag.StringSliceVar`/`encoding/csv` semantics exactly — including quoted-comma handling (`'"a,b",c'` → `["a,b", "c"]`) and Go's `csv.Reader` malformed-input error messages.

The CSV-parsing algorithm already existed for `--schema` flags (`gen types`, `db lint`, `db dump`, `db pull`, `db diff`, `db schema {generate,sync}`) in `legacy-schema-flags.ts`. This extracts the generic algorithm into a new shared `legacy-string-slice-flag.ts` module (`legacyParseStringSliceFlag`) so the four newly-fixed flags reuse it instead of duplicating it; `legacy-schema-flags.ts` now delegates to the shared module while keeping its schema-specific CSV re-encoder (`legacySchemaToCsvField`) untouched.

Out of scope: the ticket's "bonus" item (aligning `sso update`'s domains mutex error message with cobra's group-error template) is deferred — it depends on the separate functions-mutex-template fix (CLI-1860), which hasn't landed yet.

CLOSES CLI-1853

### DIFF
--- a/apps/cli/src/legacy/commands/postgres-config/delete/delete.command.ts
+++ b/apps/cli/src/legacy/commands/postgres-config/delete/delete.command.ts
@@ -3,18 +3,25 @@ import type * as CliCommand from "effect/unstable/cli/Command";
 
 import { withJsonErrorHandling } from "../../../../shared/output/json-error-handling.ts";
 import { legacyManagementApiRuntimeLayer } from "../../../shared/legacy-management-api-runtime.layer.ts";
+import { legacyParseStringSliceFlag } from "../../../shared/legacy-string-slice-flag.ts";
 import { withLegacyCommandInstrumentation } from "../../../telemetry/legacy-command-instrumentation.ts";
 import { legacyPostgresConfigDelete } from "./delete.handler.ts";
+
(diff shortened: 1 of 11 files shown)


Coly010:
@codex review

avallete (approved):

PR #5768, opened 2026-07-02
fix(cli): run --experimental gate before storage's mutex check

Coly010:
## Current Behavior

Go's cobra runs `PersistentPreRunE` (which includes the `--experimental` gate, `apps/cli-go/cmd/root.go:91-96`) before `ValidateFlagGroups()` (mutual-exclusivity checks, `cobra@v1.10.2/command.go:985,1010`). The TS port's `storage ls/cp/mv/rm` had this backwards — checking `--linked`/`--local` exclusivity before the `--experimental` gate — so `supabase storage ls --linked --local` (without `--experimental`) surfaced the mutex error in TS instead of Go's "must set the --experimental flag" error.

Fixes CLI-1855

## Expected Behavior

All 4 storage leaves now run the experimental gate first, matching Go's precedence exactly (verified against the built binary). Also corrects the inline comments in those files and the shared `legacy-experimental-gate.ts` doc, which claimed Go gates experimental "after flag-group validation" — this was backwards per cobra's actual execution order.

An existing e2e test (`storage.e2e.test.ts`) asserted the old (wrong) precedence for `--linked --local` without `--experimental`; fixed it to add `--experimental` so it keeps testing the mutex message it originally intended, at the real subprocess boundary. New ordering coverage for all 4 leaves lives in a new integration suite (`storage.experimental-gate.integration.test.ts`) rather than duplicating it in e2e, per this repo's e2e scope policy.

Out of scope: the same gate-vs-mutex-check ordering bug also exists in `db schema declarative generate`/`sync` (verified against Go source), unrelated command family — filing as a separate follow-up rather than expanding this PR.


Coly010:
@codex review

Coly010:
@codex review

Coly010:
@codex review

avallete (approved):

PR #5783, opened 2026-07-04
fix(migration): improve gap detection error message

### LINKED ISSUE #45124: Branching: preview `migrate` step DIES on multi-migration gap; `supabase db push --include-all` applies the same gap cleanly
## Summary

When a Supabase preview branch is created (either via GitHub integration on a PR or via Management API / MCP), it spawns with a **subset** of the parent project's migration history applied to the tracking table, then the preview-branching `migrate` step DIES trying to apply the gap. The same gap is applied without error by `supabase db push --include-all --linked` from the CLI against the same preview branch.

This generates spurious "⚠️ Service health check failed" comments on PR previews even though all services are healthy and our migrations are valid.

## Reproduction

Tested against project ref `byvleunmmqjfroztnqgy` (parent) on 2026-04-22.

1. Parent project has 25 migrations tracked in `supabase_migrations.schema_migrations`, earliest `20260417000000` (post-consolidation baseline), latest `20260422021351`. All migrations are committed and confirmed in sync via `supabase migration list --linked`.

2. Created a branch via Management API:
   ```bash
   curl -X POST -H "Authorization: Bearer $SUPABASE_TOKEN" \
     "https://api.supabase.com/v1/projects/byvleunmmqjfroztnqgy/branches" \
     -d '{"branch_name":"health-probe-investigation"}'
   # → branch ref: cfwtrulhimxjapemejyd
   ```

3. Polled `/v1/projects/{preview_ref}/actions`:
   ```
   clone:     EXITED   ✅
   pull:      EXITED   ✅
   health:    EXITED   ✅
   configure: EXITED   ✅
   migrate:   DEAD     ❌
   seed:      CREATED  (skipped)
   deploy:    CREATED  (skipped)
   ```

4. Queried preview's tracking table:
   ```sql
   SELECT COUNT(*), MIN(version), MAX(version) FROM supabase_migrations.schema_migrations;
   -- 9, 20260417000000, 20260419010459
   ```
   Preview was tracking only 9 of the parent's 25 migrations — missing the latest 16.

5. Verified missing-migration artifacts were absent: tables/enums introduced by the 16 missing migrations did NOT exist on the preview, and a column dropped by one of them was still present. Preview's data state was consistent with its tracking — i.e., the gap migrations had not been partially applied.

6. Ran the same migrations via local CLI:
   ```bash
   supabase link --project-ref cfwtrulhimxjapemejyd
   supabase db push --include-all --linked
   # → Applied all 16 missing migrations cleanly
   # → "Finished supabase db push" with no errors
   ```

7. Re-queried preview's tracking table: now `25` rows. Schema artifacts now match parent.

## Logs endpoint returns ResourceNotFoundException

```bash
curl -H "Authorization: Bearer $SUPABASE_TOKEN" \
  "https://api.supabase.com/v1/projects/cfwtrulhimxjapemejyd/actions/cfwtrulhimxjapemejyd/logs"

# → {"message":"failed to stream logs: operation error CloudWatch Logs: GetLogEvents,
#     https response error StatusCode: 400, ResourceNotFoundException:
#     The specified log stream does not exist."}
```

This makes the failure undiagnosable from the dashboard for affected customers — the only way to find out what happened is to bypass and retry manually as I did.

## Service health is fine

The `/v1/projects/{preview_ref}/health?services=...` endpoint returned `healthy: true, status: ACTIVE_HEALTHY` for all 6 services (auth, db, pooler, realtime, rest, storage) at the time `migrate: DEAD` was reported. The bot comment's "Service health check failed" text is therefore misleading — the `health` step did pass. The actual failure is in the `migrate` step.

## Hypothesis

The preview-branching `migrate` agent appears to NOT use `--include-all` (or equivalent), so it bails on out-of-order timestamps that arise during normal multi-developer workflows. Out-of-order timestamps are common when concurrent branches generate migrations independently — Supabase's own docs even warn about this in the `db push` CLI context, but the branching agent doesn't seem to handle it the same way.

## Customer-side impact

PRs containing more than a few migrations beyond the preview snapshot point reliably trigger this. We've seen it on multiple PRs (#1480, #1521, plus the live repro above). Reviewers learn to ignore the ⚠️ which is exactly the wrong outcome — it trains the team to ignore yellow-flag warnings generally.

## Workaround

Add a GitHub Actions job that detects preview branch creation, fetches the preview ref via Management API, and runs `supabase db push --include-all --linked` to repair the migrate state. This is what we're implementing, but it's load-bearing tooling that customers shouldn't need.

## Asks

1. Make the preview-branching `migrate` step use the same migration apply logic as `supabase db push --include-all --linked`.
2. Fix the CloudWatch logs endpoint so `migrate: DEAD` runs surface their actual error to customers.
3. Update the bot comment text — when `migrate` fails, say "Migration apply failed" rather than "Service health check failed". The current text misdirects diagnosis.

Happy to provide more repro info, branch credentials for inspection (within a debug window), or anything else useful.

lakshitsoni26:
**Fixes supabase/supabase#45124**

**Background**
When pushing migrations to a remote preview branch, users encountered a failure when there is a multi-migration gap. The underlying error scenario is `"Found local migration files to be inserted before the last migration on remote database."`

This function currently signals the resolution hint (`--include-all`) via a global variable (`utils.CmdSuggestion`) read only by the CLI's own root command. If the hint is consumed differently downstream — e.g., by a non-interactive caller reading the error value — it would be lost. This PR removes that dependency.

**Potential Cloud Fix**
Our theory is that returning a wrapped error containing the hint natively from `GetPendingMigrations`—instead of sidestepping it into a global variable—will likely resolve the issue for preview branch pushes. If the overarching cloud orchestrator surfaces the raw error text — whether by capturing CLI stdout/stderr or reading a returned error object — this ensures the `--include-all` hint travels with it intact.

**Changes**
- Removed global `utils.CmdSuggestion` mutation for missing remote migrations.
- Updated `GetPendingMigrations` to return `fmt.Errorf("%w\n%s", err, suggestIgnoreFlag(diff))`.
- Updated `TestPendingMigrations` to explicitly assert `assert.ErrorContains(t, err, "--include-all")`.

> [!WARNING]
> **E2E Suite Verification Required**
> The local `test:core` suite could not be run locally in this environment due to missing Docker/OrbStack daemon connectivity (`connection refused`). This fix has been fully verified via unit tests in `internal/migration/up` only.
> **Requesting CI/maintainer confirmation that the e2e suite passes cleanly and exercises this path correctly.**

### DIFF
--- a/apps/cli-go/internal/migration/up/up.go
+++ b/apps/cli-go/internal/migration/up/up.go
@@ -47,7 +47,7 @@ func GetPendingMigrations(ctx context.Context, includeAll bool, conn *pgx.Conn, 
 			pending := localMigrations[len(remoteMigrations)+len(diff):]
 			return append(diff, pending...), nil
 		}
-		utils.CmdSuggestion = suggestIgnoreFlag(diff)
+		return diff, fmt.Errorf("%w\n%s", err, suggestIgnoreFlag(diff))
 	}
 	return diff, err
 }
--- a/apps/cli-go/internal/migration/up/up_test.go
+++ b/apps/cli-go/internal/migration/up/up_test.go
@@ -86,6 +86,7 @@ func TestPendingMigrations(t *testing.T) {
 		_, err := GetPendingMigrations(context.Background(), false, conn.MockClient(t), fsys)
 		// Check error
 		assert.ErrorIs(t, err, migration.ErrMissingRemote)
+		assert.ErrorContains(t, err, "--include-all")
 	})
 }
 


jgoux:
Hey @lakshitsoni26, the Golang CLI is in the process of being migrated to TypeScript and this part is already ported.
I'm closing this PR as we don't intend to evolve the Golang part anymore.

PR #5792, opened 2026-07-06
test: trim low-value unit coverage

jgoux:
Trims unit tests that only asserted exported symbols, object shape, or redundant constant properties, while keeping package export contracts that guard real behavior.

Removes low-value literal default-version assertions from the stack unit tests without adding new CI surface area.

Strengthens JWT generation coverage by checking the decoded header and payload plus the HMAC signature instead of only checking that the token has three segments.


jgoux:
Addressed: removed the added stack check target and updated the PR description. The PR now keeps the cleanup focused on test coverage without adding CI surface area.

Coly010 (approved):

PR #5797, opened 2026-07-06
fix(cli): functions deploy --jobs validation + bundler routing (Go parity)

Coly010:

Bug fix (Go parity).

`supabase functions deploy --jobs` diverges from the Go CLI in two related ways:

1. The `--jobs` guard only errored when local bundling (Docker/legacy-bundle) was active, so `--use-docker=false --jobs 2` (no `--use-api`) silently passed in TS while Go rejects it. The error text also differed from Go's wording.
2. `useLocalBundler` decided Docker-vs-API routing based on whether `--use-api` was *typed* on the command line (`explicitUseApi`), not its *resolved* value — so `--use-api=false` alone silently forced the API bundler instead of falling through to Docker, contrary to Go's `if useApi { useDocker = false }` (which only fires when the resolved value is true).

Go's only guard (`apps/cli-go/cmd/functions.go:79-82`) is:
```go
if useApi {
    useDocker = false
} else if maxJobs > 1 {
    return errors.New("--jobs must be used together with --use-api")
}
```

## What is the new behavior?

- `--jobs > 1` is now rejected whenever `--use-api`'s resolved value is `false`, regardless of `--use-docker`/`--legacy-bundle`, with the exact Go error text: `--jobs must be used together with --use-api`.
- Bundler routing (`useLocalBundler`) now keys off the same resolved `flags.useApi` value, so `--use-api=false` alone falls through to Docker/legacy-bundle instead of silently switching to the API path.
- Both fixes land in the shared `deployFunctions()` used by both the `legacy` and `next` shells, with regression coverage added in both.

Fixes CLI-1861.

### DIFF
--- a/apps/cli/src/legacy/commands/functions/deploy/deploy.e2e.test.ts
+++ b/apps/cli/src/legacy/commands/functions/deploy/deploy.e2e.test.ts
@@ -56,11 +56,35 @@ describe("supabase functions deploy (legacy) — argument validation", () => {
       },
     );
     expect(exitCode).not.toBe(0);
-    // The Go CLI phrases this as either "must be used together with --use-api"
-    // or "cannot be used with local bundling" depending on version — both mean
-    // --jobs is rejected without server-side (--use-api) bundling.
-    expect(stderr).toMatch(/--jobs\b.*(--use-api|local bundling)/i);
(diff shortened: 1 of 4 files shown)


Coly010:
@codex review

avallete (approved):

PR #5798, opened 2026-07-06
fix(cli): telemetry enable/disable fire cli_command_executed on pre-toggle consent

Coly010:

Bug fix (Go parity).

Go's `cli_command_executed` telemetry event is gated on a consent snapshot read once at process start (`PersistentPreRunE`, `apps/cli-go/cmd/root.go:131-138`), before a command's own handler runs, and fired after the handler completes (`:171-181`) based on that snapshot — not on any value the handler itself just wrote. This applies uniformly to every command, including `telemetry enable`/`disable` themselves.

The TS legacy port instead unconditionally suppressed this event for both `telemetry enable` and `telemetry disable` via `analytics: false`, so neither command ever fired it, regardless of prior telemetry state.

## What is the new behavior?

`analytics: false` is removed from both commands. The existing `legacyAnalyticsLayer`/`telemetryRuntimeLayer` machinery already reads consent from disk once, at Effect layer-construction time — before either handler mutates `telemetry.json` — so this reproduces Go's snapshot semantics without any new bespoke logic:

- Running `disable` while telemetry is enabled fires the event one last time (using the pre-toggle, still-enabled snapshot); running it while already disabled stays silent.
- Running `enable` while telemetry is already enabled (a redundant/idempotent invocation) fires the event; the common case (enabling from disabled) stays silent.

This is intentionally broader than the original report's "disable only" framing — tracing Go's actual gate (`internal/telemetry/service.go`'s `canSend()`) showed it's symmetric and state-based, not command-based, so scoping the fix to `disable` alone would have left a real (if narrow) parity gap for the redundant-`enable` case.

Also documents a previously-unmentioned side effect: when the event fires, the command now waits (bounded, ~5s) for a PostHog flush attempt before exiting, since the analytics client's shutdown is a finalizer around the whole command run.

Closes CLI-1868

### DIFF
--- a/apps/cli/src/legacy/commands/telemetry/disable/SIDE_EFFECTS.md
+++ b/apps/cli/src/legacy/commands/telemetry/disable/SIDE_EFFECTS.md
@@ -17,7 +17,8 @@ instead of `~/.supabase/telemetry.json`.
 
 ## API Routes
 
-`disable` is fully local. No network calls are made.
+None called directly. `cli_command_executed` may be sent to PostHog — see
+Telemetry Events Fired below.
 
 ## Environment Variables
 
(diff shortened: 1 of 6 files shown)


Coly010:
@codex review

avallete (approved):

PR #5800, opened 2026-07-06
docs(cli): disclose --high-availability as TS-only on projects create

Coly010:

Docs update (Go parity disclosure) + telemetry-list correctness fix.

`--high-availability` on `projects create` has no Go CLI equivalent — Go's `cmd/projects.go` never registers such a flag, and its `RunE` closure never sets `HighAvailability` on the create request body even though the underlying Management API field exists. Unlike `--reveal` on `projects api-keys` (which explicitly documents itself as a TS-only addition, in a code comment, `SIDE_EFFECTS.md`, and `docs/go-cli-porting-status.md`'s "Flag divergences" list), `--high-availability` was undisclosed everywhere.

It was also listed in `safeFlags: ["org-id", "high-availability"]`, implying Go-parity telemetry-safety that doesn't exist (Go's `markFlagTelemetrySafe` only covers `org-id`). Since it's a boolean flag, this had no actual runtime effect — boolean flag values are always logged verbatim by the instrumentation regardless of `safeFlags` membership — but the array wrongly implied otherwise.

The linked ticket (CLI-1870) offered two valid resolutions: remove the flag for strict parity, or disclose it as TS-only. This PR disclosure, since `--high-availability` is a real, working, user-facing capability that Go's CLI simply never got around to exposing (not a Go bug being ported forward) — removing it would regress anyone currently relying on it.

## What is the new behavior?

`--high-availability` is now disclosed as TS-only in a code comment, `SIDE_EFFECTS.md`, and `docs/go-cli-porting-status.md`'s "Flag divergences" list, matching the `--reveal` precedent exactly. Removed from `safeFlags` (now just `["org-id"]`) to match Go's actual telemetry-safe set.

Fixes CLI-1870.


Coly010:
@codex review

Coly010:
@codex review

Coly010:
@codex review

avallete (approved):

PR #5803, opened 2026-07-06
fix(cli): mark gen bearer-jwt's --role as required (Go parity)

Coly010:

Bug fix (Go-parity divergence).

Go marks `--role` as a required flag on `gen bearer-jwt` (`cmd/gen.go:175`, `MarkFlagRequired("role")`). The TS port declared it optional, so a missing `--role` previously wasn't rejected by TS at all — the command would proceed to spawn the real Go binary via the Phase-0 proxy, which enforced the requirement itself and printed its own clean, single-line cobra error.

Fixes CLI-1859.

## What is the new behavior?

TS's own argument parser now rejects a missing `--role` before ever invoking the Go subprocess, matching Go's declared intent. This is a real behavioral improvement (fails faster, no longer silently depends on the subprocess for validation) but it comes with a caveat worth calling out explicitly: enforcing this at TS parse time newly exposes a pre-existing bug in the vendored `effect` CLI library for this command — a required-flag/choice parse failure dumps the full help doc to stdout and double-prints the error (once from the library, once from this repo's Go-parity error renderer), instead of Go's single clean stderr line. This bug already affects other shipped commands with required flags (verified independently on `sso add`), so it's not something this PR introduces — but this PR is what makes it newly reachable for `gen bearer-jwt` specifically, since `--role` previously never hit TS's own parse-error path at all. Filed CLI-1901 to track the systemic fix, and documented the interim divergence in `SIDE_EFFECTS.md`.

Reviewed via a 4-way parallel pass (architect/engineer/security/DX). Security and engineer approved outright. Architect approved after a suggested follow-up (auditing other Go `MarkFlagRequired` call sites) was independently re-verified and found to already be correctly handled elsewhere in the codebase — no follow-up needed there. DX flagged the CLI-1901 issue above as a blocking concern before this PR could be considered complete; resolved by verifying it independently, filing CLI-1901, and disclosing the interim behavior rather than shipping over it silently.


Coly010:
@codex review

Coly010:
Thanks for the review — both P1 findings are real, and together they change the conclusion for this PR.

**Missing-role parity expectations:** confirmed. Ran `apps/cli-e2e`'s `parity: gen bearer-jwt` test after building both binaries — it fails: stdout, stderr, and filesystem all diverge. Go still prints one clean stderr line and writes `telemetry.json` via its own `PersistentPostRun`. TS now rejects before ever invoking the Go proxy, hitting a pre-existing effect-cli library bug (dumps the full help doc to stdout, double-prints the error) and never runs the proxy at all, so `telemetry.json` never gets written.

Went further and re-checked the original premise directly against the unfixed code: `--help` text mode never renders required/optional for flags at all, and `--help` JSON mode's `required` field is `true` for every non-boolean flag regardless of `Flag.optional` — `--role` already showed `"required": true` before this change too, due to a separate bug in the same library module. So there was never an observable "--help incorrectly shows optional" defect to begin with.

For a pure Phase-0 proxy command, delegating entirely to the Go binary is what guarantees output parity. Duplicating Go's flag validation in TS's own (currently broken) error renderer trades a real, tested parity guarantee for a fix to a defect that doesn't actually exist. Reverted the change (`9addfc89`) and am closing this PR. Filed the underlying library bug as CLI-1901 with this exact repro as a concrete regression case — worth revisiting `gen bearer-jwt --role` once that's fixed.

**CliOutput layer on the parser-error test:** also worth noting — this became moot since the test is reverted, but for the record: the test passed cleanly across three independent runs (mine, and two reviewing agents', including one that reverted the source and confirmed the test correctly failed without the fix), which wouldn't be possible if `Cause.squash` were seeing a missing-service defect instead of the real `MissingOption` error. Good catch to flag regardless — just didn't end up mattering here.

PR #5811, opened 2026-07-07
docs(cli): fix fabricated SUPABASE_API_URL and bare PROJECT_ID env-var references

Coly010:

Docs fix — Legacy port SIDE_EFFECTS.md accuracy.

Two systemic env-var documentation errors across `src/legacy/**/SIDE_EFFECTS.md`, found in the parity audit:

- **~36 files** document `SUPABASE_API_URL` as a real override for the Management API base URL. No such env var exists in Go: `SetEnvPrefix("SUPABASE")` + `AutomaticEnv()` never binds `API_URL`, there's no `os.Getenv("SUPABASE_API_URL")` anywhere, and the `Profile.api_url` field loads on a separate, unbound `viper.New()` instance (`internal/utils/profile.go:99`). The real override is `SUPABASE_PROFILE` (built-in name or a path to a YAML profile file).
- **~18 files** document bare `PROJECT_ID` as if it were itself a settable env var. It's Go's internal viper key name — the real, user-facing env var is `SUPABASE_PROJECT_ID` (`viper.GetString("PROJECT_ID")` under the global `SetEnvPrefix("SUPABASE")` singleton, `internal/utils/flags/project_ref.go:62`).

Both claims were independently verified against the real Go source and, for the viper-specific mechanics, a compiled test program against the exact pinned `github.com/spf13/viper v1.21.0`.

## What is the new behavior?

- Removed the fabricated `SUPABASE_API_URL` rows; where a file didn't already document `SUPABASE_PROFILE`, added a real row for it (including the persisted `~/.supabase/profile` fallback step in the resolution chain, matching `legacy-cli-config.layer.ts`).
- Corrected every bare `PROJECT_ID` mention (Environment Variables tables, Files Read "When" columns, and prose) to `SUPABASE_PROJECT_ID`, converging on the dominant convention already used correctly across ~15 sibling files.
- Left `db/advisors/SIDE_EFFECTS.md` untouched — it already correctly documents that `SUPABASE_API_URL` is not honored.
- Filed follow-ups for two related-but-out-of-scope findings surfaced during review: CLI-1904 (global choice-flag telemetry gap, unrelated) and CLI-1905 (bare `DB_PASSWORD` has the same bug class as this PR's F2, in a set of files this PR doesn't touch).

Docs-only change; no `.ts` files touched. `pnpm check:all` passes.


Coly010:
@codex review

jgoux (approved):

PR #5813, opened 2026-07-07
docs(cli): correct inaccurate SIDE_EFFECTS.md and AGENTS.md claims

Coly010:

Docs fix — Legacy port SIDE_EFFECTS.md / AGENTS.md accuracy (F3-F10, F12 from the parity audit).

11 `SIDE_EFFECTS.md` files and `AGENTS.md` contained factual errors where the code was already correct but the documentation was wrong — wrong file paths, a fabricated stdout message, an inverted claim about whether a command writes the linked-project cache, a missing json/stream-json output envelope, a fictitious subprocess name, a backwards project-ref claim, and a fabricated telemetry-identity call. Each claim was independently re-verified against `apps/cli-go/` (and, for the `completion` exit-code claim, against the actual compiled Go binary and the actual TS CLI) before editing.

## What is the new behavior?

- **F3**: `db/branch/{create,delete,list,switch}` — corrected the current-branch file path to `supabase/.branches/_current_branch`, matching `internal/utils/misc.go:99`.
- **F4**: `db/remote/commit` — removed the false claim it prints `Finished supabase db pull.` (that belongs only to `db pull`).
- **F5**: `db/lint` — corrected: it DOES write the linked-project cache on `--linked`, matching Go's `ensureProjectGroupsCached`.
- **F6**: `db/schema/declarative/generate` — corrected: it has a real json/stream-json success envelope.
- **F7**: `db/schema/declarative/sync` — replaced a fictitious `migration up --local` subprocess with the real `db reset --local` recovery-path subprocess.
- **F8**: `completion` — the doc's exit-code claim was backwards. Go exits `0` on both bare `completion` and an unrecognized shell subcommand (verified against the compiled binary); the legacy TS shell currently exits `1` for both — a genuine, systemic exit-code bug in the shared CLI harness, not `completion`-specific (it reproduces on any bare/unrecognized group-command invocation, e.g. `branches`). Rather than fixing the code here or describing the bug as intended, the doc now states Go's true behavior and flags the current TS divergence, filed separately as CLI-1906.
- **F9**: `config/push` + `AGENTS.md` — corrected the linked-project-cache path (was a fabricated `~/.supabase/<hash>/...`, real path is `<workdir>/supabase/.temp/linked-project.json`) and fixed an adjacent conflation where the project-ref-fallback file was wrongly attributed to the same cache file, when it's actually a separate file (`<workdir>/supabase/.temp/project-ref`).
- **F10**: `branches/update` — corrected: the upgrade-suggest call uses the branch's own resolved ref, not the parent `--project-ref`.
- **F12**: `AGENTS.md` telemetry table — removed a fabricated `analytics.identify(gotrueId)` claim from the `login` row; Go's `StitchLogin` only aliases.

A second, wider pre-existing doc bug was found during F9's verification (the same fabricated linked-project-cache path is copy-pasted across 29 other `SIDE_EFFECTS.md` files not touched here) — filed separately as CLI-1907 rather than expanding this diff.

Docs-only change; no `.ts`/`.go` files touched. `pnpm check:all` passes.

### DIFF
--- a/apps/cli/AGENTS.md
+++ b/apps/cli/AGENTS.md
@@ -90,14 +90,14 @@ Always check `src/shared/` before writing new infrastructure. Do not duplicate w
 
 Also check the following `legacy/` infrastructure before writing equivalent helpers from scratch:
 
-| Path                                                    | What it provides                                                                                                                                                                            |
-| ------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `legacy/config/legacy-cli-config.layer.ts`              | `LegacyCliConfig` — resolves `SUPABASE_PROFILE` (built-in name **or** YAML file path), `--workdir`, `--experimental`, project-id from `supabase/config.toml`                                |
-| `legacy/config/legacy-project-ref.layer.ts`             | `LegacyProjectRefResolver` — `--project-ref` flag → env → linked-project.json → config fallback chain; matches Go's resolver order                                                          |
-| `legacy/telemetry/legacy-telemetry-state.layer.ts`      | `LegacyTelemetryState.flush` — writes `~/.supabase/telemetry.json`, runs in every command's `Effect.ensuring`                                                                               |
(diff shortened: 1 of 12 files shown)


Coly010:
@codex review

jgoux (approved):

PR #5827, opened 2026-07-08
fix(cli): exit 0 for bare group commands to match Go cobra parity

Coly010:
## What changed

Invoking a legacy-shell group command that has subcommands but is itself not runnable — e.g. `supabase branches`, `supabase completion` — with no subcommand and no `--help` exited **1** in the TS legacy shell. Go's cobra CLI exits **0** for the identical invocation: a non-`Runnable()` command with no `RunE` internally returns `flag.ErrHelp`, which cobra's `ExecuteC()` maps to "print help, return nil error". Only explicit `--help` got exit 0 before this fix; the bare/missing-subcommand form did not, even though the printed help text was identical in both cases.

`CliError.ShowHelp` (in `effect/unstable/cli`) already declares the correct exit code via Effect's own `Runtime.errorExitCode` marker (`0` for a clean `ShowHelp` with no errors, `1` otherwise) — `apps/cli/src/shared/cli/run.ts` now delegates to `Runtime.getErrorExitCode(Cause.squash(cause))` instead of hand-rolling `ShowHelp` classification, which is simpler and tracks the library's own source of truth for any future `CliError` additions.

Also added an integration test driving the real `legacyBranchesCommand` through `Command.runWith` (confirming the actual `ShowHelp` cause shape), and an e2e test asserting the real compiled-binary exit code, since this bug is specifically about the real OS process exit code.

## Known related (not fixed here)

A typo'd subcommand under a group (e.g. `supabase branches bogus`) still exits 1 where Go cobra exits 0 for a non-root command — this is pre-existing on `develop`, not a regression from this change, and out of scope for this fix. Filed as a follow-up.

Fixes CLI-1906

### DIFF
--- a/apps/cli/src/shared/cli/run.e2e.test.ts
+++ b/apps/cli/src/shared/cli/run.e2e.test.ts
@@ -1,0 +1,24 @@
+import { describe, expect, test } from "vitest";
+import { runSupabase } from "../../../tests/helpers/cli.ts";
+
+/**
+ * CLI-1906: the real bug here is the actual OS process exit code —
+ * `ProcessControl.exit` calls real `process.exit(code)`, so only a genuine
+ * subprocess run proves the shipped binary's exit code changed. Everything
+ * else about this fix (`exitCodeForFailure`'s classification) is covered by
+ * `run.unit.test.ts` and `run.integration.test.ts`; this is the one minimal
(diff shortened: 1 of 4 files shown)


Coly010:
@codex review

jgoux (approved):

PR #5845, opened 2026-07-09
feat(cli): scaffold fetch-style edge functions

jgoux:
## Summary

Updates the next CLI Edge Function scaffold to use the Web Fetch-style `export default { fetch(...) }` authoring model instead of `Deno.serve(...)`.

New functions now get a function-local `package.json` with no default dependencies, and no generated `deno.json`. This makes the default template dependency-free while establishing the package-manifest boundary for users who add dependencies under `supabase/functions/<function-name>` and import them by bare specifier from function code.

## Notes

Existing `deno.json` manifest detection remains in shared config so existing projects and explicit import-map flows keep working.


avallete (commented):
Is that breaking change ? Is that still something that we wanna review (4 weeks old)

jgoux:
@avallete with Workers coming I don't think it's worth pursuing for now.

PR #5855, opened 2026-07-10
fix(cli): exempt global choice flags from telemetry redaction (Go isEnumFlag parity)

Coly010:
## What changed

Go's `isEnumFlag` (`apps/cli-go/cmd/root_analytics.go:110-116`) unconditionally reports the value of any `*utils.EnumFlag`-backed pflag verbatim in `cli_command_executed` telemetry — no per-flag annotation needed. `--output`, `--dns-resolver`, and `--agent` are all registered as `*utils.EnumFlag` on `rootCmd.PersistentFlags()` (`cmd/root.go:330,331,333`), so Go always sends their real value.

The TS port's `withLegacyCommandInstrumentation` already auto-detects `Flag.choice` flags declared in a command's own `config` (CLI-1866), and already resolves global/persistent flag values as a fallback when a command doesn't declare that CLI name locally (CLI-1896). But CLI-1896 deliberately scoped out the 3 global *choice* flags — nothing taught the global-fallback path that a resolved value might itself be a safe enum, so `--output`/`--dns-resolver`/`--agent` always fell through to `"<redacted>"`.

This adds `GLOBAL_CHOICE_FLAG_NAMES` (`legacy-command-instrumentation.ts`), derived from `LEGACY_GLOBAL_FLAGS` by reusing the existing `getChoiceFlagNames` helper (so the choice-detection predicate has exactly one home), and consults it only on the global-fallback path — i.e. only when the invoking command's own `flags` record doesn't already declare that CLI name. A command that registers its own differently-typed local flag under the same CLI name (`db diff`'s local string `output: Flag.string("output")`, `cmd/db.go:622`) still shadows the global and stays redacted, matching Go's `isEnumFlag` type-asserting that command's own non-enum flag object instead of root's persistent `EnumFlag`. Verified this is the real wiring, not a hypothetical: `db diff` passes `output` in its own `flags` record with no matching `config` entry, so `isFromHandler` is structurally `true` for it and the new global set is never consulted.

Also verified Go's separate `output_format` telemetry property (`PropOutputFormat`) is independent of the generic `flags.output` entry — both already fire in Go too, so no double-counting concern from un-redacting `flags.output`.

Fixes CLI-1904.

## Why

Found during CLI-1866's review as a pre-existing gap in the same problem class, tracked separately since it was out of scope for that PR's diff, then explicitly deferred again by CLI-1896 ("the 3 global choice flags ... remain redacted — that gap is the already-tracked CLI-1904, not this ticket").

## Reviewer-relevant context

- All 4 `review-changes` agents (architect, engineer, security, DX) reviewed this diff and approved with no blocking findings. Two converging nits were fixed in a follow-up commit: deduped the choice-detection predicate (`GLOBAL_CHOICE_FLAG_NAMES` now derives from `getChoiceFlagNames` instead of re-implementing it) and clarified the two `AGENTS.md` telemetry bullets.
- One deliberately-deferred judgement call from engineer-reviewer: no command's own `*.integration.test.ts` today exercises `withLegacyCommandInstrumentation`'s PostHog-capture behavior for *any* flag (that coverage is entirely the job of `legacy-command-instrumentation.unit.test.ts`'s own scenario suite) — this is a pre-existing, repo-wide pattern, not something this PR narrows or regresses. Adding integration-tier telemetry assertions to `db diff`/`db query` specifically would set a new cross-cutting test precedent well beyond this fix's scope, so left as a possible future hardening rather than folded in here.


Coly010:
@codex review

avallete (approved):

PR #5856, opened 2026-07-10
fix(cli): match Go path.Match character-class negation in legacy seed globbing

Coly010:
## Current Behavior

`legacyMatchPattern` (the glob matcher backing `[db.seed].sql_paths` and `db reset --version`) treated a leading `!` inside a `[...]` bracket class the same as `^` — shell/fnmatch-style negation. Go's `path.Match` (which `apps/cli-go`'s `fs.Glob`/`afero.Glob` actually use for this) only negates on `^`; `!` is an ordinary literal class member. So a pattern like `[!a].sql` matched `b.sql` in the TS port where the Go CLI would not, meaning `db push --include-seed` / `db reset` could read, hash, and execute seed SQL outside the set the user's Go-compatible config intended.

Verified directly against the real Go stdlib source (`$GOROOT/src/path/match.go`) available on this machine, and against `apps/cli-go/pkg/config/config.go` (`fs.Glob`) and `apps/cli-go/internal/migration/repair/repair.go` (`GetMigrationFile` / `afero.Glob`) — both go through the same `path.Match` grammar.

## New Behavior

Rather than patching the bug in place, this fixes it by deleting the duplicate: `apps/cli/src/legacy/shared/legacy-path-match.ts` (`legacyPathMatch`) is already a byte-faithful, already-tested port of Go's `path.Match` — introduced earlier for the sibling `db new`/migration seed pipeline (`legacy-seed.ts`) — with correct `^`-only negation, escapes, and malformed-pattern (`ErrBadPattern`) detection. `legacyMatchPattern` and its `matchClass`/`match` helpers are removed, and both call sites are rewired onto `legacyPathMatch`:

- `legacy-seed-ops.ts`'s own seed glob (`globOne`)
- `reset.handler.ts`'s `--version` migration-file glob (behavior-identical here — `v` is validated as digits-only before the glob ever runs, so no bracket syntax is reachable on this path)

A follow-up commit closes a related gap all four `review-changes` reviewers independently flagged: `legacyGlobSeedFiles` was discarding `legacyPathMatch`'s `badPattern` signal, so a malformed pattern (e.g. an unterminated `[` class) fell through to the generic `no files matched pattern` warning instead of Go's `failed to glob files: syntax error in pattern` (`fs.Glob`'s up-front `Match(pattern, "")` validation). The sibling pipeline (`legacy-seed.ts:resolveSeedFiles`) already did this correctly — mirrored it here now that both pipelines share `legacyPathMatch`.

**Behavior change worth calling out for release notes:** any existing `[db.seed].sql_paths` config relying on `[!x]` as shell-style negation will now select a different (Go-correct) set of seed files with no error — e.g. `[!0-9]` now means "the literal characters `!`, `0`-`9`", not "not a digit". Use `[^x]` for negation, matching Go/the real Go CLI. This is a correctness fix (the legacy shell's contract is strict Go parity), but it is a silent behavior change for anyone who had unknowingly depended on the old, non-Go-compatible negation.

## Related Issue(s)

Fixes https://linear.app/supabase/issue/CLI-1880

## Notes for reviewers

`review-changes` (architect/engineer/security/DX, run against the first commit) all returned APPROVE/SHIP IT. Judgement calls deliberately left open rather than expanded into this PR:

- The two seed pipelines (`legacy-seed-ops.ts` and `legacy-seed.ts`) still duplicate the outer `fs.Glob`-style directory-walk/dedup logic (both now delegate the leaf matcher to `legacyPathMatch`, but the surrounding glob-resolution code is still two separate ports). Flagged by the architect reviewer as a good candidate for a follow-up consolidation, out of scope for this fix.
- `sql_paths`' glob dialect (Go `path.Match` semantics, `^`-only negation, no POSIX classes) isn't documented anywhere user-facing (`packages/config/src/db.ts`). Flagged by the DX reviewer as worth a docs follow-up, not blocking.

### DIFF
--- a/apps/cli/src/legacy/commands/db/reset/reset.handler.ts
+++ b/apps/cli/src/legacy/commands/db/reset/reset.handler.ts
@@ -30,11 +30,8 @@ import { LegacyTelemetryState } from "../../../telemetry/legacy-telemetry-state.
 import { legacyDropUserSchemas } from "../shared/legacy-drop-schemas.ts";
 import { LegacyDbBootstrapSeam } from "../shared/legacy-db-bootstrap.seam.service.ts";
 import { legacyListLocalMigrations } from "../shared/legacy-pgdelta.cache.ts";
-import {
-  legacyGetPendingSeeds,
-  legacyMatchPattern,
-  legacySeedData,
-} from "../shared/legacy-seed-ops.ts";
+import { legacyGetPendingSeeds, legacySeedData } from "../shared/legacy-seed-ops.ts";
+import { legacyPathMatch } from "../../../shared/legacy-path-match.ts";
 import { legacyUpsertVaultSecrets } from "../../../shared/legacy-vault.ts";
 import { legacySeedBucketsRun } from "../../../shared/legacy-seed-buckets.ts";
 import type { LegacyDbResetFlags } from "./reset.command.ts";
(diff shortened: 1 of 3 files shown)


Coly010:
@codex review

avallete (approved):

PR #5880, opened 2026-07-15
fix(cli): stop telemetry hangs and errors on blocked networks

pamelachia:
On networks that block or blackhole `eu.posthog.com` (corporate proxies, CI and Docker egress policies, air-gapped machines), telemetry breaks the CLI itself: stable v2.109.1 pays a 5s trailing hang and exits 1 with an UnknownError payload after every command, long-running commands print raw SDK stack traces to stderr, and the Go binary's shutdown wait is unbounded. The only workaround was `supabase telemetry disable`, which permanently removes those users from CLI analytics. This PR makes telemetry strictly fire-and-forget: a failed delivery never changes a command's output, exit code, or runtime beyond a 2s cap.

**Changed:**
- **Exit 1 after successful commands**: the TS shutdown finalizer was `Effect.promise(...).pipe(Effect.ignore)`, but `Effect.promise` turns the shutdown-deadline rejection into a defect that `Effect.ignore` cannot swallow, so a telemetry timeout failed the whole command. The rejection is now caught on the promise itself.
- **Unbounded and stacking exit delays**: one shared 2s budget (`EXIT_DELAY_CAP_MS`) bounds each request and the entire `_shutdown` drain on the TS side, and the Go client gets the same 2s `ShutdownTimeout` (the SDK default waits indefinitely, including retries). The previous 5s shutdown deadline left room for two sequential 2s request timeouts (~4s exits).
- **Work outliving the deadline** (review round 3): the SDK's shutdown deadline only stops the wait, it does not cancel work. The in-flight request kept running past scope release and the drain then started the queued request, keeping the runtime alive ~4s instead of 2s. A lifecycle `AbortController` now composes into every request's signal and fires when the deadline hits, so no request survives or starts after the scope closes. Covered by a unit regression asserting zero active requests one tick after release (red on the previous code) and an e2e test that runs the compiled binary against a TCP blackhole and asserts wall-clock process exit, empty stderr, and that telemetry actually reached the wire.
- **Stderr noise**: Go routes SDK logs to the `--debug` logger; TS reports delivery failures as delivered inside the wrapped fetch, because posthog-node has no logger hook, only hardcoded `console.error` calls, so an infallible transport is the one supported way to get no retries, no output, and no stall.

**Note:** the exit bounds are caps, not floors. On healthy networks both shells return as soon as the flush completes (`flushAt: 1` keeps the send overlapping command runtime), and fully detaching the send instead would race process exit and silently drop events on healthy networks. The accepted tradeoff is that deliveries slower than 2s are dropped; telemetry is best-effort, observable under `--debug` on the Go side only.

## Linear
- fixes GROWTH-1001


apps/cli/src/shared/telemetry/posthog-client.ts
```diff
@@ -0,0 +1,44 @@
...
+// batches are retried with multi-second delays that stall process shutdown.
+// Reporting every attempt as delivered keeps telemetry fire-and-forget — a
+// blocked or offline network must never surface errors or delays for a
+// command that succeeded.
+export function fireAndForgetFetch(
+  fetchImpl: (url: string, init?: RequestInit) => Promise<Response> = globalThis.fetch,
+): NonNullable<PostHogOptions["fetch"]> {
+  return async (url, options) => {
+    try {
+      const response = await fetchImpl(url, options);
```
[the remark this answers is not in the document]
pamelachia: Fixed in 77ce8398: set requestTimeout: 2000 on the client, so a blackholed connection holds exit for at most ~2s (verified: capture plus shutdown completes in 2014ms against a non-routable address, zero stderr bytes). Kept the send awaited rather than fully detached: floating the promise would race process exit and silently drop events on healthy networks, which is the common case.

apps/cli-go/internal/telemetry/client.go
```diff
@@ -38,7 +43,14 @@ func NewClient(apiKey string, endpoint string, baseProperties map[string]any, fa
 			return posthog.NewWithConfig(apiKey, config)
 		}
 	}
-	config := posthog.Config{}
+	// Without a positive ShutdownTimeout, Close blocks indefinitely when the
+	// endpoint is unreachable, hanging every command on networks that block
+	// PostHog. The default logger prints delivery failures to stderr even for
+	// commands that succeeded, so route them to the --debug logger instead.
+	config := posthog.Config{
+		ShutdownTimeout: shutdownTimeout,
```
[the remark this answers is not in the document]
pamelachia: Fixed in c367a3ed: ShutdownTimeout reduced from 5s to 2s (verified: Close() returns in 2.001s against a blackholed endpoint). It is a cap, not a floor: on healthy networks Close returns as soon as the flush completes. A fully nonblocking shutdown would drop the event on every short-lived command since the final flush happens at Close, so a tight bound is the intended trade.

avallete:
@codex review

apps/cli/src/shared/telemetry/posthog-client.ts
```diff
@@ -0,0 +1,36 @@
...
+          host,
+          flushAt: 1,
+          flushInterval: 0,
+          requestTimeout: 2_000,
+          fetch: fireAndForgetFetch,
+        }),
+    ),
+    // Catch on the promise: Effect.promise turns rejections into defects,
+    // which escape Effect.ignore and fail the command.
+    (client) => Effect.promise(() => client._shutdown(5_000).catch(() => undefined)),
```
avallete: **issue** one event can be in flight while a later exit event remains queued. Shutdown then performs two sequential 2-second requests (eg: `shutdownMs: 3987` with two captures against the pinned SDK). Apply the deadline to the entire shutdown and add a two-event blackhole test.

The existing port-9 test rejects immediately and misses this case

pamelachia: Fixed in 878576ba. `_shutdown` now gets the same 2s budget as `requestTimeout` (one shared `EXIT_DELAY_CAP_MS`): in the pinned SDK, `shutdownTimeoutMs` is raced against the entire drain (in-flight request plus the queued flush loop), so the deadline covers the whole shutdown instead of stacking per request.

Added the two-event blackhole test: the fetch stub never resolves but honors the SDK's abort signal, and the second capture happens only after the first request is confirmed in flight (two synchronous captures batch into a single request, so that ordering is load-bearing). It asserts total shutdown under 3s: red at ~4.0s on the previous code (matches your 3987ms), green at ~2.0s now.

The deadline rejection stays silent: its logger call in @posthog/core is debug-gated and the existing catch on the promise swallows the rejection.

apps/cli/src/shared/telemetry/posthog-client.ts
```diff
@@ -0,0 +1,38 @@
...
+          host,
+          flushAt: 1,
+          flushInterval: 0,
+          requestTimeout: EXIT_DELAY_CAP_MS,
+          fetch: fireAndForgetFetch,
+        }),
+    ),
+    // Catch on the promise: Effect.promise turns rejections into defects,
+    // which escape Effect.ignore and fail the command.
+    (client) => Effect.promise(() => client._shutdown(EXIT_DELAY_CAP_MS).catch(() => undefined)),
```
avallete: **issue:** the 2s shutdown deadline does not currently bound process exit.

`posthog-node` rejects `_shutdown(2_000)` while allowing pending promises to continue. With one request in flight and another event queued, the second fetch starts as shutdown times out and keeps Bun alive for its own request timeout.

I reproduced this against a local TCP blackhole:

- One event: scope 2.01s, process 2.49s
- Two events: scope 2.01s, process 4.35s

The regression test only measures `Effect.scoped`, so it passes before the process actually exits.

Could we give the scoped client a lifecycle `AbortController`, combine it with each request’s signal, and abort outstanding fetches when shutdown reaches its deadline? A small subprocess test measuring actual Bun exit would protect the intended ~2s cap.

Roughly:
```
const resource = Effect.sync(() => {
  const shutdown = new AbortController();
  const client = new PostHog(apiKey, {
    // ...
    fetch: (url, options) =>
      fireAndForgetFetch(url, {
        ...options,
        signal: options.signal
          ? AbortSignal.any([options.signal, shutdown.signal])
          : shutdown.signal,
      }),
  });
  return { client, shutdown };
});

Effect.acquireRelease(
  resource,
  ({ client, shutdown }) =>
    Effect.promise(async () => {
      try {
        await client._shutdown(EXIT_DELAY_CAP_MS);
      } catch {
        // Telemetry must not fail the command.
      } finally {
        shutdown.abort();
      }
    }),
).pipe(Effect.map(({ client }) => client));
```
The regression should also assert that no fetch remains active after scope release or preferably use a small subprocess test that verifies Bun exits within the cap.

avallete (commented):
The Go-side timeout and debug-log routing in this PR look good and should stay. I don’t think we need to additionally coordinate one shared shutdown budget across TS parents and Go children: that would broaden the change into cross-runtime telemetry ownership for a transitional sidecar. Let’s keep the blocking scope focused on ensuring the TS client’s 2s shutdown cap also bounds actual Bun process exit.

apps/cli/src/shared/telemetry/posthog-client.ts
pamelachia: Fixed in 5fbf9cd9 (tests hardened in 35894b93 and a7e032dd), using the shape you proposed: a lifecycle AbortController composed into each request's signal via AbortSignal.any, fired in the release finalizer's `finally`, so the deadline now cancels outstanding requests instead of abandoning them. I reproduced your numbers before changing anything: two events against a local TCP blackhole gave scope 2.03s / process exit 4.05s on the previous code, and 2.02s process exit with the abort in place.

Two notes from verifying:

- The shipped binary was already exit-bounded: `run.ts` calls `process.exit` right after scope close, which killed the leaked work. Your measurement is still the right one, since any consumer that does not hard-exit (your harness, tests, library-style use of the wrapper) pays the full overrun, and a release finalizer that leaves requests running, and even starts new ones, is broken regardless of what masks it.
- Tests, per your suggestion: the in-process regression now also asserts zero active requests one settle tick after scope release (red on the previous code). That is the discriminating check, because the compiled binary's explicit exit masks the leak at process level. For the boundary you measured, a new e2e spawns the compiled binary against a TCP blackhole and asserts wall-clock process exit under 4.5s, empty stderr, and at least one connection received on the wire, so it cannot pass vacuously with telemetry disabled (the e2e harness disables telemetry for every spawned process by default).

Also re-verified the SDK assumptions after merging develop in: on posthog-node 5.41.0 / @posthog/core 1.43.1, `_shutdown`'s new `hasTimedOut` flag stops the drain loop only after the current flush settles, so the queued-request start you observed was still live on the new version too.

pamelachia:
also if this doesn't work, and you have a preferred method @avallete that isn't how i did it, can i close this PR and I'll create a linear issue for CLI team to figure this out?

avallete (approved):
All good, thank's for all the adjustments you made through the review loops @pamelachia !

Issue #5890, opened 2026-07-16
Supabase CLI 2.109.1 on Windows can resolve different Management API identities from different process contexts on the same computer and Windows user account.


pifoub:
### Affected area

Auth

### Supabase CLI version

v2.109.1

### Operating system

Windows 11

### Installation method

Other

### Command

```shell
Installation method: Scoop. not tied to command
```

### Actual output

```shell
## Description

Supabase CLI 2.109.1 on Windows can resolve different Management API identities from different process contexts on the same computer and Windows user account.

The persistent Windows generic credential is:

- Target: `Supabase CLI:supabase`
- Username/account: `supabase`
- Type: `CRED_TYPE_GENERIC`
- Persistence: `CRED_PERSIST_LOCAL_MACHINE`

Repeated `supabase logout` commands exited successfully, but the credential remained. Its Windows `LastWritten` timestamp did not change, proving it was not deleted and recreated.

## Environment

- Windows 11
- Supabase CLI 2.109.1
- Installed through Scoop
- PowerShell 7.6.3
- Hosted Supabase projects under two different Supabase identities
- Commands run both interactively and through a desktop application's non-interactive process runner

## Observed behavior

Sanitized CLI traces identify two distinct authenticated Supabase users through the Management API response identity.

A representative sequence:

1. Interactive `supabase login` authenticates Identity B.
2. Interactive `supabase link` succeeds against Identity B's project.
3. Interactive `supabase projects list` returns Identity B's projects.
4. Minutes later, a non-interactive `supabase projects list` launched through another process context returns Identity A's projects.
5. No login, startup task, environment-variable change, or credential write occurs between steps 3 and 4.

This occurred repeatedly.

The behavior presents as if a deleted or replaced credential is being restored, but trace evidence instead shows different Windows process/security contexts resolving different authenticated identities.

## Logout behavior

Multiple `supabase logout` commands returned exit status 0, while the original Windows credential remained readable.

The CLI implementation treats keyring-unavailable/not-found as a successful not-logged-in outcome, so exit status 0 does not establish that the credential was deleted.

## Profile-file behavior

Deleting:

    C:\Users\<user>\.supabase\profile

does not prevent credential loading. Debug output shows:

    NotFound: FileSystem.readFile (...\.supabase\profile)
    Using access token for profile: supabase
    Using profile: supabase (supabase.co)

The CLI silently falls back to the built-in `supabase` profile and continues reading the Windows credential.

## Expected behavior

[...]

PR #5891, opened 2026-07-16
fix(cli): redact API keys

7ttp:
## TL;DR

fixes project keys being written to CLI traces, matching the Go cli's behavior...

## Problem

The Go-cli use to send keys through the `apikey` header:

[https://github.com/supabase/cli/blob/2cb2d6e83a3f7d65ebdab905c35b8f8257e685dc/apps/cli-go/pkg/fetcher/gateway.go#L22-L33](<https://github.com/supabase/cli/blob/2cb2d6e83a3f7d65ebdab905c35b8f8257e685dc/apps/cli-go/pkg/fetcher/gateway.go#L22-L33>)

Its debug transport only logs the request method and URL:

[https://github.com/supabase/cli/blob/2cb2d6e83a3f7d65ebdab905c35b8f8257e685dc/apps/cli-go/internal/debug/http.go#L14-L16](<https://github.com/supabase/cli/blob/2cb2d6e83a3f7d65ebdab905c35b8f8257e685dc/apps/cli-go/internal/debug/http.go#L14-L16>)

Effect tracing captures request headers, but `apikey` is not included in its default redaction list...<br>this allowed values to be written to local traces...

## sol

exclude `apikey` header attributes at the tracing boundary....<br>requests and all other telemetry remain unchanged, while NDJSON and debug output no longer contain project keys....

## ref:

* reported in: supabase/cli#5890

### DIFF
--- a/apps/cli/src/shared/telemetry/tracing.layer.ts
+++ b/apps/cli/src/shared/telemetry/tracing.layer.ts
@@ -76,6 +76,7 @@ class ExportableSpan implements Tracer.Span {
   }
 
   attribute(key: string, value: unknown): void {
+    if (key.endsWith(".header.apikey")) return;
     this.attributes.set(key, value);
   }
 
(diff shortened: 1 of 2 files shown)


7ttp:
@codex review

avallete (approved):

PR #5902, opened 2026-07-18
feat(cli): allow suppressing update notifier via SUPABASE_NO_UPDATE_NOTIFIER

### LINKED ISSUE #5853: Support suppressing "A new version of Supabase CLI is available" message
### Existing issues

- [x] I have searched the existing issues.

### Affected area

Other

### Problem to solve

I'm using supabase from Nix. I want to suppress the message suggesting upgrading because I'm intentionally using the version pinned in Nix.

### Proposed solution

Add an environment variable to avoid checking for upgrade

### Alternatives considered

_No response_

### Additional context

_No response_

GMartinho:
## What

Adds a `SUPABASE_NO_UPDATE_NOTIFIER` environment variable that suppresses the "A new version of Supabase CLI is available" upgrade notifier. When set to a truthy value (`1`, `true`, …), the CLI skips both the GitHub release check and the stderr suggestion.

## Why

Closes #5853. Users who pin the CLI version through a package manager such as Nix are intentionally on a fixed version and have no way to opt out of the periodic upgrade check and its message. This gives them a standard, npm-style env var (`NO_UPDATE_NOTIFIER`, prefixed with `SUPABASE_` to match the CLI's other environment variables) to turn it off.

## Notes

- The variable is read directly from the environment (`os.Getenv` + `strconv.ParseBool`) rather than through viper, because the upgrade check in `Execute()` also runs for `--help` and `--version`, which bypass the `cobra.OnInitialize` hook that configures viper's automatic env binding. Reading it directly makes suppression work consistently across every invocation. Boolean semantics match `viper.GetBool` (used by `SUPABASE_DEBUG` etc.).
- When disabled, the network call to GitHub is skipped entirely, not just the message — so pinned installs make no release request at all.
- Behavior unchanged when the variable is unset or falsy (`0`/`false`).

### DIFF
--- a/apps/cli-go/cmd/root.go
+++ b/apps/cli-go/cmd/root.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"os"
 	"os/signal"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


7ttp (commented):
closing in favour of: https://github.com/supabase/cli/pull/6143

> Thanks! for ur efforts towards contributing to Supabase!! 💚

PR #5944, opened 2026-07-24
fix(cli): render SMS/MFA config templates instead of emitting the raw text/template escape

Coly010:
## What changed

`supabase init` (and the `bootstrap` scratch-project path — both go through the shared `renderProjectConfigTemplate` in `apps/cli/src/shared/init/project-init.templates.ts`) wrote two broken lines into the generated `supabase/config.toml`:

```toml
template = "Your code is {{ `{{ .Code }}` }}"
```

for both `[auth.sms]` and `[auth.mfa.phone]`. The Go CLI's template source contains that `text/template` self-escape, but Go renders the scaffold through `text/template` (`config.Eject`, `pkg/config/config.go:555,572`), which resolves the backtick raw-string action to the literal GoTrue template. The TS renderer copied the escape verbatim and runs no template engine, so the raw escape landed in users' configs — a functionally broken SMS/MFA OTP template (GoTrue would send the literal text instead of substituting the code).

The fix pre-renders the two lines in the TS template constant:

```toml
template = "Your code is {{ .Code }}"
```

Verified byte-identical against the Go CLI's actual `config.Eject` output (dumped via a throwaway Go test harness): these two lines were the only divergence, and after the fix the full generated file matches Go's eject byte-for-byte, including the trailing newline.

## Test changes

The existing byte-parity test compared the TS render against the **raw Go template source**, so both sides carried the unresolved escape and the test passed while the shipped output was broken. It now emulates Go's eject-time rendering (`resolveGoTemplateEscapes`) before comparing, so it would have caught this bug. Also added:

- a dedicated regression test asserting the two generated `template =` lines are the rendered GoTrue templates, not raw Go escapes;
- a guard test that fails loudly if the Go scaffold ever gains a `{{ ... }}` construct the suite does not model, so this class of parity drift cannot silently return (raised by review).

## Deliberately out of scope (from review)

- **Configs already generated with the broken lines are not auto-migrated.** Users who ran `init`/`bootstrap` with an affected build keep the raw escape in their `config.toml`; the manual fix is replacing the two `template =` lines under `[auth.sms]` and `[auth.mfa.phone]` with `"Your code is {{ .Code }}"`. A detect-and-warn on config load (particularly on the `config push` path, which can carry the broken template to hosted GoTrue) was suggested in review as a possible fast-follow, but auto-editing or warning about user-owned config is a scope/UX decision beyond this parity fix.

Fixes CLI-1971

Linear: https://linear.app/supabase/issue/CLI-1971/initbootstrap-write-broken-smsmfa-template-lines-into-configtoml

### DIFF
--- a/apps/cli/src/shared/init/project-init.templates.ts
+++ b/apps/cli/src/shared/init/project-init.templates.ts
@@ -260,7 +260,7 @@ enable_signup = false
 # If enabled, users need to confirm their phone number before signing in.
 enable_confirmations = false
 # Template for sending OTP to users
-template = "Your code is {{ \`{{ .Code }}\` }}"
+template = "Your code is {{ .Code }}"
 # Controls the minimum amount of time that must pass before sending another sms otp.
 max_frequency = "5s"
 
@@ -308,7 +308,7 @@ verify_enabled = false
 enroll_enabled = false
 verify_enabled = false
 otp_length = 6
-template = "Your code is {{ \`{{ .Code }}\` }}"
+template = "Your code is {{ .Code }}"
 max_frequency = "5s"
 
 # Configure MFA via WebAuthn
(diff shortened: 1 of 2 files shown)


Coly010:
@codex review

kanadgupta (approved):

PR #5946, opened 2026-07-24
fix(cli): suppress the --debug hint when a confirmation prompt is declined

Coly010:
## What changed

When a confirmation prompt is declined, the Go CLI prints only a red `context canceled` on stderr and exits 1 — `recoverAndExit` (`apps/cli-go/cmd/root.go:287-303`) explicitly skips the `SuggestDebugFlag` fallback for `errors.Is(err, context.Canceled)`. The TS text `Output.fail` renderer appended the fallback hint for every suggestion-less error, including cancellations, so every prompting command (`logout`, `db push` (3 prompts), `db reset`, `migration fetch`/`repair`/`down`, `functions deploy --prune`, `gen signing-key`, `secrets unset`, `projects delete`, `branches create`, `bootstrap`) printed:

```
context canceled
Try rerunning the command with --debug to troubleshoot the error.   ← spurious
```

This PR mirrors Go's guard at the render boundary:

- New shared sentinel `CONTEXT_CANCELED_MESSAGE` in `apps/cli/src/shared/output/errors.ts` — the byte-for-byte render of Go's `context.Canceled`.
- The text-layer `fail` (`apps/cli/src/shared/output/output.layer.ts`) skips the debug-hint fallback when the normalized message equals the sentinel. The explicit-suggestion branch stays first, matching Go printing a pre-set `utils.CmdSuggestion` even for canceled errors.
- All 14 cancellation construction sites (12 files: the legacy decline paths above plus `shared/functions/deploy.ts`) now reference the constant instead of an inline literal, so the sentinel is single-sourced like Go's.
- `LEGACY_LOGOUT_CANCELLED_MESSAGE` is folded into the shared constant, and the documented intent in `logout.errors.ts` (which already described the Go behavior this PR implements) now matches reality.

## Why the fix is shared rather than legacy-scoped

The failure-rendering seam (`shared/cli/run.ts` `handledProgram` → `Output.fail`) is shell-agnostic, and `next/` genuinely shares a cancellation producer: `shared/functions/deploy.ts` (used by `next functions deploy --prune`) fails with the same `context canceled` message on a declined prune prompt. Suppressing a troubleshooting hint for a user-declined prompt is the correct behavior in both shells, so the fix lives in the shared layer instead of forking the renderer per shell. Aside from that, `next/` handlers treat declines gracefully (no `context canceled` error), so next-shell text rendering is otherwise unchanged. This is deliberate and reviewed against the CLI-1546 invariant (no spinner/stream routing was touched).

## Tests

- Unit (`output.layer.unit.test.ts`): byte assertions that the sentinel message renders as exactly the red `context canceled` line with no hint (fails without the fix), and that an explicit caller suggestion still prints for a cancellation (Go ordering).
- Declined-prompt stderr-byte e2e per affected family, each asserting exit code 1, last stderr line exactly `context canceled`, and no debug-hint line: `logout` (new), `migration fetch` (extended), `db reset` (new file — Docker-free, the destructive prompt fires before any connection is dialed), `gen signing-key` (extended).
- The existing non-canceled guards stay: a generic error still gets the hint (unit + `run.e2e.test.ts` asserts stderr ends with the hint for an unrecognized flag).
- `stripAnsi` hoisted to `tests/helpers/cli.ts` for the files this PR touches.

## Deliberately-open judgement calls

- **Exact-match vs `errors.Is`:** the TS check is message equality, narrower than Go's chain-walking `errors.Is`. A future producer surfacing a *wrapped* cancellation (`"...: context canceled"`) through `Output.fail` would keep the hint where Go suppresses it. No such producer exists today (mid-flight Ctrl-C takes the interrupt/exit-130 path and never reaches `Output.fail`); the invariant is documented on the constant. Reviewers (architect/engineer/parity passes) agreed a tag-registry or normalized-error flag isn't warranted for a single Go sentinel.
- **Pre-existing divergences noted during the parity audit, out of scope here:** (a) Ctrl-C *at* a prompt → TS clack `Operation cancelled.`/exit 130 vs Go `context canceled`/exit 1; (b) the hint's `--debug` detection uses `process.argv` vs Go's `viper.GetBool("DEBUG")` (which also honors `SUPABASE_DEBUG`) — affects only the non-canceled branch.

Fixes CLI-1973

https://linear.app/supabase/issue/CLI-1973/declined-confirmation-prompts-print-a-spurious-debug-hint-context

### DIFF
--- a/apps/cli/src/shared/output/errors.ts
+++ b/apps/cli/src/shared/output/errors.ts
@@ -1,5 +1,31 @@
 import { Data } from "effect";
 
+/**
+ * Byte-for-byte render of Go's `context.Canceled` sentinel.
+ *
+ * Every declined confirmation prompt in the Go CLI surfaces as a bare
+ * `context.Canceled` (e.g. `errors.New(context.Canceled)` in
+ * `apps/cli-go/internal/logout/logout.go:19`), and `recoverAndExit`
+ * (`apps/cli-go/cmd/root.go:287-303`) deliberately skips the
+ * `SuggestDebugFlag` hint for it — declining a prompt is a user decision,
+ * not an error worth troubleshooting. Handlers that port those decline
+ * paths construct their cancellation errors with this exact message, and
+ * the text `Output.fail` renderer keys on it to suppress the `--debug`
+ * hint, mirroring Go's `!errors.Is(err, context.Canceled)` guard.
+ *
+ * Two invariants of that renderer check:
+ * - The value must stay trim-invariant: it round-trips through
+ *   `normalizeCliError`'s trimming `readString` before reaching the
+ *   renderer's equality check (`shared/output/normalize-error.ts`).
+ * - The check is exact-match, narrower than Go's chain-walking
+ *   `errors.Is`: a future producer surfacing a WRAPPED cancellation
+ *   (`"...: context canceled"`) through `Output.fail` would keep the hint
+ *   where Go suppresses it — no such producer exists today (mid-flight
+ *   Ctrl-C takes the interrupt/exit-130 path and never reaches
+ *   `Output.fail`), but widen the check if one ever appears.
+ */
+export const CONTEXT_CANCELED_MESSAGE = "context canceled";
+
 export class NonInteractiveError extends Data.TaggedError("NonInteractiveError")<{
   readonly detail: string;
   readonly suggestion: string;
--- a/apps/cli/tests/helpers/cli.ts
+++ b/apps/cli/tests/helpers/cli.ts
@@ -13,9 +13,7 @@ import {
   registerTempStackProject,
 } from "./stack-e2e-cleanup.ts";
 
-/** Strip ANSI SGR escape sequences (colors/styles) from captured CLI output. */
-// eslint-disable-next-line no-control-regex
-export const stripAnsi = (text: string): string => text.replace(/\x1b\[[0-9;]*m/gu, "");
+export { stripAnsi } from "./ansi.ts";
 
 const BINARY_EXT = process.platform === "win32" ? ".exe" : "";
 const SHIM_PATH = fileURLToPath(new URL("../../dist/supabase.js", import.meta.url));
--- a/apps/cli/tests/helpers/cli.ts
+++ b/apps/cli/tests/helpers/cli.ts
@@ -13,6 +13,10 @@ import {
   registerTempStackProject,
 } from "./stack-e2e-cleanup.ts";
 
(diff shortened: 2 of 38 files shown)


Coly010:
@codex review

apps/cli/tests/helpers/cli.ts
```diff
@@ -13,6 +13,10 @@ import {
   registerTempStackProject,
 } from "./stack-e2e-cleanup.ts";
 
+/** Strip ANSI SGR escape sequences (colors/styles) from captured CLI output. */
+// eslint-disable-next-line no-control-regex
+export const stripAnsi = (text: string): string => text.replace(/\x1b\[[0-9;]*m/gu, "");
```
kanadgupta: (nit) could we use this in the other places we have this helper function defined?

Image

apps/cli/src/shared/output/errors.ts
```diff
@@ -1,5 +1,31 @@
 import { Data } from "effect";
 
+/**
+ * Byte-for-byte render of Go's `context.Canceled` sentinel.
+ *
+ * Every declined confirmation prompt in the Go CLI surfaces as a bare
+ * `context.Canceled` (e.g. `errors.New(context.Canceled)` in
+ * `apps/cli-go/internal/logout/logout.go:19`), and `recoverAndExit`
+ * (`apps/cli-go/cmd/root.go:287-303`) deliberately skips the
```
kanadgupta, line 9: (nit and more of a broader codebase convention question) does it make sense to include GitHub permalinks when referencing `cli-go`? Since that code is likely being removed at some point?

kanadgupta (approved):
Two non-blocking comments otherwise LGTM

apps/cli/src/shared/output/errors.ts
Coly010: it'll likely get removed when we remove the go, but for now while we still perform parity audits, its worth having

apps/cli/tests/helpers/cli.ts
Coly010: honestly yes it probably should be

Coly010: Done in 3651e9a2 — all 16 local copies now use the shared helper. It lives in `tests/helpers/ansi.ts` (re-exported from `cli.ts`) so unit/integration tests don't pull in `cli.ts`'s subprocess/e2e machinery just for ANSI stripping. One copy (`agent-output.e2e.test.ts`) stripped full CSI sequences rather than SGR-only, so the shared helper adopted that superset.

PR #5968, opened 2026-07-28
fix(cli): print Connecting line and relative paths in db pull output (CLI-1978)

Coly010:
## What changed

Two stdout/stderr parity divergences in the native `db pull` paths, both empirically confirmed against the Go binary:

1. **Missing `Connecting to remote/local database...` stderr line.** Go's `pull.Run` goes through `utils.ConnectByConfig`, which prints the line to stderr before dialing (`apps/cli-go/internal/utils/connect.go:344-348`), local vs remote keyed off `utils.IsLocalDatabase`. The TS handler called `connection.connect` directly and never printed it (`db push` and the other db/migration commands already print it manually — pull was the outlier). The line is now emitted right before dialing, same stream and ordering as Go. The delegated `--experimental` branch deliberately skips it: the Go child's own `ConnectByConfig` prints the line, so the parent printing too would double it. The delegate decision is hoisted into `delegatesExperimentalPull` and shared by both the Connecting-line guard and the delegate branch.

2. **`Schema written to` / `Declarative schema written to` printed absolute paths; Go prints workdir-relative ones.** Go chdirs into the workdir and joins relative constants (`pull.go:76` via `new.GetMigrationPath` → `supabase/migrations/<ts>_<name>.sql`; `pull.go:119` prints `utils.GetDeclarativeDir()` verbatim — the config's `declarative_schema_path` or the relative `supabase/database` default). The TS lines now render the workdir-relative path at the print sites only. `writtenMigrations` keeps absolute paths for file I/O, and the machine-output (`--output-format json`) fields `schemaWritten`/`schemaFiles` deliberately stay absolute — that machine contract is unchanged.

Tests: the weak `.toContain("Schema written to")` assertions are tightened to exact relative-path lines (with an absolute-workdir leak guard), plus new assertions for the Connecting line: remote and local wording, exact bytes and stderr ordering (before `Dumping schema…`/`Creating shadow…`/`Preparing declarative…`), presence in machine mode, and absence on both `--experimental` delegate forms (flag and `SUPABASE_EXPERIMENTAL` env). The tightened suite fails 10 tests against the unfixed handler.

## Review notes (deliberately out of scope, same bug class)

- `pull.debug.ts` / `legacy-debug-bundle.ts` print the pg-delta debug-bundle path absolute where Go prints the relative `supabase/.temp/pgdelta/debug/<id>` (`pgdelta_pull_debug.go:46-48`) — distinct lines from this issue's scope.
- `db schema declarative generate` (`generate.handler.ts:275`) prints the same `Declarative schema written to` sentence with an absolute path; Go prints the relative `GetDeclarativeDir()` there too (`declarative.go:156`).
- The delegated `--experimental` parent still dials before delegating (mirroring Go's early connectivity check); a parent-side connect failure on that path surfaces without a Connecting line. Pre-existing behavior, now documented in a comment.

Fixes CLI-1978

https://linear.app/supabase/issue/CLI-1978/db-pull-output-parity-missing-connecting-to-database-line-absolute

### DIFF
--- a/apps/cli/src/legacy/commands/db/pull/pull.handler.ts
+++ b/apps/cli/src/legacy/commands/db/pull/pull.handler.ts
@@ -323,9 +323,34 @@ export const legacyDbPull = Effect.fn("legacy.db.pull")(function* (flags: Legacy
         yield* proxy.exec(rebuildDelegateArgs(flags), { env });
       });
 
+    // viper resolves `EXPERIMENTAL` from *either* the global `--experimental`
+    // pflag or `SUPABASE_EXPERIMENTAL` (`cmd/root.go:318-320,327,334`), so honor
+    // both forms; the legacy root only forwards `--experimental` to Go proxy
+    // argv, never into env. Resolved before connecting so the Connecting line
+    // below knows whether this run delegates to the Go child. Declarative mode
+    // never delegates (Go checks `usePgDelta` before `EXPERIMENTAL`,
+    // `pull.go:47-50`).
+    const delegatesExperimentalPull =
+      !useDeclarative &&
+      (experimental || legacyParseBoolEnv(toml.envLookup("SUPABASE_EXPERIMENTAL")));
+
     // Connectivity check (Go's `ConnectByConfig` at the top of `pull.Run`).
     yield* Effect.scoped(
       Effect.gen(function* () {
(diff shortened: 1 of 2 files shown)


Coly010:
@codex review

kanadgupta (approved):

PR #5969, opened 2026-07-28
fix(cli): use Go's bare invalid-version error in db reset (CLI-1979)

Coly010:
## What changed

`supabase db reset --version <non-integer>` in the TS legacy shell emitted `failed to parse <v>: invalid version number`. The Go CLI's `reset.Run` returns the **bare** `repair.ErrInvalidVersion` = `invalid version number` (`apps/cli-go/internal/db/reset/reset.go:35-36`); the `failed to parse <v>:` wrapper belongs to `migration repair` only (`apps/cli-go/internal/migration/repair/repair.go:29`). This drops the wrapper on the reset path only:

- `reset.handler.ts`: emit the bare `invalid version number` (exit code and `--debug` hint unchanged).
- `reset.errors.ts`: JSDoc now cites the correct Go source (`reset.go:35-36`) instead of the repair-path lines that led to the original conflation.
- `reset.integration.test.ts`: the previous `toContain("invalid version number")` assertion passed for both the wrapped and bare messages (so it could not catch this divergence) — it now pins the exact tag + message.
- `repair.integration.test.ts`: guard assertion that `migration repair` keeps its `failed to parse <v>:` wrapper (correct Go parity there), so a future unification can't regress either side.

`db reset --last` validation and messages are untouched (already at parity). `SIDE_EFFECTS.md` for `db reset` already documented the bare message, so code and doc now agree.

## Why

Strict 1:1 Go parity for the legacy shell. Empirically confirmed on both binaries: Go prints `invalid version number`, TS printed `failed to parse abc: invalid version number` (both exit 1). Audit ref: `apps/cli/docs/go-parity-audit-2026-07-24.md` §3.2 (R1).

## Deliberately left open

Reviewers noted a pre-existing, adjacent divergence: the reset path's `INTEGER_PATTERN` accepts >int64 numeric versions (e.g. 20 digits) that Go's `strconv.Atoi` rejects with `invalid version number`; TS instead falls through to the glob `file does not exist` error. `migration repair` already solves this via the shared `legacyParseMigrationVersion` helper. Out of CLI-1979's scope (wrapper text only) — candidate for a follow-up issue.

Fixes CLI-1979

https://linear.app/supabase/issue/CLI-1979/db-reset-version-non-integer-error-text-diverges-from-go

### DIFF
--- a/apps/cli/src/legacy/commands/db/reset/reset.handler.ts
+++ b/apps/cli/src/legacy/commands/db/reset/reset.handler.ts
@@ -21,6 +21,7 @@ import {
 } from "../../../shared/legacy-db-config.toml-read.ts";
 import { LegacyDbConnection } from "../../../shared/legacy-db-connection.service.ts";
 import { legacyApplyMigrations } from "../../../shared/legacy-migration-apply.ts";
+import { legacyParseMigrationVersion } from "../../../shared/legacy-migration-timestamp.format.ts";
 import { legacyPromptYesNo } from "../../../../shared/legacy/legacy-prompt-yes-no.ts";
 import {
   type LegacyDbConnType,
@@ -48,7 +49,6 @@ import {
   LegacyDbResetVersionFlagsError,
 } from "./reset.errors.ts";
 
-const INTEGER_PATTERN = /^[+-]?\d+$/u;
 const MIGRATE_FILE_PATTERN = /^([0-9]+)_(.*)\.sql$/u;
 
 const applyError = (message: string) => new LegacyDbResetApplyError({ message });
@@ -187,9 +187,14 @@ export const legacyDbReset = Effect.fn("legacy.db.reset")(function* (flags: Lega
 
     // Version / last resolution (Go's reset.Run lines 34-52), filesystem only.
     let resolvedVersion = "";
-    if (Option.isSome(flags.version)) {
+    // Go's `len(version) > 0` guard (reset.go:34) skips validation entirely for an
+    // empty --version, falling through as if no version were given at all.
+    if (Option.isSome(flags.version) && flags.version.value.length > 0) {
       const v = flags.version.value;
-      if (!INTEGER_PATTERN.test(v)) {
+      // Go's `strconv.Atoi` (== `ParseInt(s, 10, 0)`) rejects non-numeric text AND
+      // values outside the int64 range; `legacyParseMigrationVersion` mirrors that
+      // exactly (`migration repair` uses the same helper for its own version parse).
+      if (legacyParseMigrationVersion(v) === undefined) {
         // Go's reset.Run returns the bare repair.ErrInvalidVersion (reset.go:35-36);
         // the `failed to parse <v>:` wrapper belongs to `migration repair` only.
         return yield* Effect.fail(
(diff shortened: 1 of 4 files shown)


Coly010:
@codex review

apps/cli/src/legacy/commands/db/reset/reset.handler.ts
```diff
@@ -190,9 +190,11 @@ export const legacyDbReset = Effect.fn("legacy.db.reset")(function* (flags: Lega
     if (Option.isSome(flags.version)) {
       const v = flags.version.value;
       if (!INTEGER_PATTERN.test(v)) {
+        // Go's reset.Run returns the bare repair.ErrInvalidVersion (reset.go:35-36);
+        // the `failed to parse <v>:` wrapper belongs to `migration repair` only.
         return yield* Effect.fail(
           new LegacyDbResetInvalidVersionError({
-            message: `failed to parse ${v}: invalid version number`,
+            message: "invalid version number",
```
kanadgupta, line 202: The message now byte-matches Go, but the *condition* still diverges from Go's `strconv.Atoi` in two edge cases, so this exact error isn't emitted in exactly the cases Go emits it:

- **int64 overflow** — `Atoi` == `ParseInt(s, 10, 0)`, which rejects values outside the int64 range, so Go's `db reset --version 99999999999999999999` fails with this bare `invalid version number` (`reset.go:35-36`). `INTEGER_PATTERN` (`/^[+-]?\d+$/u`, line 51) accepts it, so the TS path falls through to the glob check and reports `glob supabase/migrations/…: file does not exist` instead.
- **empty version** — Go guards with `len(version) > 0` (`reset.go:34`), so an empty `--version` skips validation entirely and proceeds; here `Option.isSome` + the regex turn it into this error.

There's already a shared exact-`Atoi` mirror for this: `legacyParseMigrationVersion` (`src/legacy/shared/legacy-migration-timestamp.format.ts:64`), which `migration repair` uses for this same validation (`repair.handler.ts:186` — repair's tests even pin the 20-digit overflow case). Replacing the `INTEGER_PATTERN.test(v)` check with `legacyParseMigrationVersion(v) === undefined` (plus a `v.length > 0` guard mirroring `reset.go:34`) would close both gaps and drop the duplicated regex.

Both divergences predate this PR, so a follow-up is fine if you'd rather keep this diff minimal — but since the ticket's goal is byte-parity on this exact error path, worth tracking.

kanadgupta (approved):
Tiny behavioral divergence that Claude discovered (see below) but otherwise LGTM

apps/cli/src/legacy/commands/db/reset/reset.handler.ts
Coly010: Good catch — fixed in e57fc7fb: swapped `INTEGER_PATTERN` for `legacyParseMigrationVersion` (the same helper `migration repair` uses for its Atoi parity) and gated the whole validation block on a non-empty `--version`, so:
- an out-of-int64-range version now hits the bare `invalid version number` error instead of falling through to the glob check
- an empty `--version` now skips validation entirely and falls through to a full reset, matching Go's `len(version) > 0` guard (`reset.go:34`)

Added integration tests for both cases (`rejects an out-of-int64-range --version`, `treats an empty --version like no version at all`) — both fail against the old code.

PR #5973, opened 2026-07-28
fix(cli): always send domains in sso update PUT body to match Go (CLI-1981)

Coly010:
## What changed

`sso update` now **always** sends `domains` in the PUT body, matching the Go CLI.

Go's `internal/sso/update/update.go:82-109` looks gated — `if len(Domains) != 0 { … } else if params.AddDomains != nil || params.RemoveDomains != nil { merge existing → body.Domains }` — but `cmd/sso.go:171-172` declares `--add-domains`/`--remove-domains` with a **non-nil** `[]string{}` default and passes them unconditionally, so `AddDomains != nil` is always true from the CLI. Every Go `sso update` therefore enters the fetch-and-merge branch and sets `body.Domains` (the recomputed existing set, or `[]`). `UpdateProviderBody.Domains` is a non-nil `*[]string` under `json:"domains,omitempty"`, and `omitempty` never omits a non-nil pointer, so even the empty set serializes as `"domains":[]`.

The TS handler gated the merge on `addDomains.length > 0 || removeDomains.length > 0` and omitted `domains` entirely when no domain flags were passed — a different request shape (`sso update <id> --metadata-url X`: Go sends `{"domains":[…existing],"metadata_url":"X"}`, TS sent `{"metadata_url":"X"}`) with potential server-side domain-state drift.

- `update.handler.ts` — the merge branch is now the unconditional `else` of the `--domains` replace branch.
- `mergeDomains` seed filter relaxed from `domain.length > 0` to `typeof domain === "string"` — Go's seed check is nil-ness only (`domain.Domain != nil`, `update.go:89`), so an empty-string domain from the GET response is kept.
- `SIDE_EFFECTS.md` — PUT body documents `domains` as always present.
- Integration tests: PUT-body assertions for all flag combinations, including no-flags → recomputed existing set, no-flags + provider without domains → literal `"domains":[]`, GET response missing `domains` entirely, explicit empty `--domains=` falling into the merge, and empty-string-domain retention. All five fail without the fix.

## Verification level for the Go body shape: **live-captured**

Built the Go binary from `apps/cli-go` and ran it against a local HTTP recorder (`SUPABASE_PROFILE` YAML pointing `api_url` at the recorder). Captured raw PUT bodies:

| Scenario | Go PUT body (verbatim bytes) |
| --- | --- |
| no flags, provider has `old1.com`,`old2.com` | `{"domains":["old1.com","old2.com"]}` |
| no flags, provider has no domains | `{"domains":[]}` |
| `--metadata-url` only (issue repro) | `{"domains":["old1.com","old2.com"],"metadata_url":"https://example.com"}` |
| explicit empty `--domains=` | `{"domains":["old2.com","old1.com"]}` (merge branch; map-iteration order is nondeterministic) |

Go's own `update_test.go` additionally pins `body.Domains` non-nil for the `--domains` and `--add/--remove-domains` paths.

## Review notes (deliberately left open)

- `sso add` has the same gate class (`create.go:60` gates on `!= nil`, TS gates on `.length > 0`), so `sso add --domains=` diverges — that is audit §3.10's separate `sso add` item, intentionally not touched here.
- The GET→PUT read-modify-write window (concurrent domain edits overwritten by the resent snapshot) is inherent to Go's design; parity is the bar.

Fixes CLI-1981
https://linear.app/supabase/issue/CLI-1981/sso-update-omits-domains-from-the-put-body-where-go-always-sends-it

### DIFF
--- a/apps/cli/src/legacy/commands/sso/update/SIDE_EFFECTS.md
+++ b/apps/cli/src/legacy/commands/sso/update/SIDE_EFFECTS.md
@@ -20,13 +20,13 @@
 
 ## API Routes
 
-| Method | Path                                                         | Auth         | Request body                                                                    | Response (used fields)                                             |
-| ------ | ------------------------------------------------------------ | ------------ | ------------------------------------------------------------------------------- | ------------------------------------------------------------------ |
-| `GET`  | `/v1/projects/{ref}/config/auth/sso/providers/{provider_id}` | Bearer token | none                                                                            | `{id, saml?, domains?, created_at?, updated_at?}`                  |
-| `PUT`  | `/v1/projects/{ref}/config/auth/sso/providers/{provider_id}` | Bearer token | `{metadata_xml?, metadata_url?, domains?, attribute_mapping?, name_id_format?}` | `{id, saml?, domains?, ...}` (parsed loosely)                      |
-| `GET`  | `<metadata-url>`                                             | none         | `Accept: application/xml`, 10s timeout                                          | XML body (UTF-8) — validation when `--skip-url-validation` not set |
-| `GET`  | `/v1/projects/{ref}`                                         | Bearer token | none                                                                            | `{organization_slug}` — upgrade-gate side-call on 4xx              |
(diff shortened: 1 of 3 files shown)


Coly010:
@codex review

kanadgupta (approved):

PR #5977, opened 2026-07-28
fix(cli): match Go's init error texts for already-exists and orioledb gate (CLI-1986)

Coly010:
## What changed

`supabase init`'s two error paths on the **legacy shell** now byte-match the Go CLI (verified by building and running `apps/cli-go` locally):

1. **Config already exists** (no `--force`) — was a friendlier TS-only message; now:
   ```
   failed to create config file: open supabase/config.toml: file exists
   Run supabase init --force to overwrite existing config file.
   ```
   This is Go's wrapped `O_EXCL` `*os.PathError` from `utils.InitConfig` (`config.go:243-246`) plus the `CmdSuggestion` from `internal/init/init.go:38-42`. The path is Go's *relative* `utils.ConfigPath`, so it's constant regardless of cwd/`--workdir`.

2. **`--use-orioledb` without `--experimental`** — was `The --use-orioledb flag requires --experimental.`; now cobra's message from Go's `PreRun` `MarkFlagRequired("experimental")` (`cmd/init.go:32-36`):
   ```
   required flag(s) "experimental" not set
   ```
   The second line is Go's generic `recoverAndExit` debug hint, which the legacy text output layer already appends when an error carries no suggestion. This follows the same handler-level convention as `db dump`'s `data-only` gate and `vanity-subdomains`' `desired-subdomain` gate (bare message, no `Error:` prefix — matching Go's `SilenceErrors` + `recoverAndExit` rendering, verified against the built binary).

Also fixed while there (issue scope): the shared gitignore append now always prepends `\n` when `supabase/.gitignore` exists — **even when it's empty** — matching `internal/init/init.go:80-96` (verified against the built Go binary).

## Shell-scoping decision

The shared `initProject` module stays mechanism-only (it already returned `{ created: false }` and left error shaping to each shell). **Error shaping lives at each shell's handler boundary:**

- **Legacy** gets new legacy-scoped error classes in `legacy/commands/init/init.errors.ts` with Go's exact bytes.
- **Next deliberately keeps its friendlier texts**: the graceful `Supabase project already initialized.` no-op and `The --use-orioledb flag requires --experimental.` — now pinned by tests. Its `InitExperimentalRequiredError` moved from `shared/` to `next/commands/init/init.errors.ts` since it's next-only after this change; the now-unused shared `InitAlreadyExistsError` was deleted.

The gitignore newline fix is in shared code and applies to both shells (mechanical scaffolding, same bytes as Go).

## Notes / deliberately left open

- **Windows**: byte parity for the already-exists message now covers Windows too (as of 6d262569) — `init.handler.ts` selects the message on `RuntimeInfo.platform`, rendering Go's `filepath.Join`-built path and `ERROR_FILE_EXISTS` errno text (`failed to create config file: open supabase\config.toml: The file exists.`), pinned by an integration test with a mocked `win32` platform. Documented in `SIDE_EFFECTS.md`.
- Pre-existing, out of scope (flagged during review): `normalize-error.ts`'s `MissingOption` case renders `Error: required flag(s) … not set` *with* an `Error:` prefix, while the built Go binary prints no prefix via `recoverAndExit` — worth a separate parity check; `apps/cli/CLAUDE.md`'s porting checklist quotes a stale debug-hint wording ("get more details" vs the actual "troubleshoot the error"); Go's `failed to read git ignore file` error branch for an unreadable existing `.gitignore` remains unported.

Fixes CLI-1986
Linear: https://linear.app/supabase/issue/CLI-1986/init-error-texts-diverge-from-go-already-exists-use-orioledb-gate


Coly010:
@codex review

apps/cli/src/legacy/commands/init/init.handler.ts
```diff
@@ -43,10 +41,16 @@ export const legacyInit = Effect.fn("legacy.init")(function* (flags: LegacyInitF
...
+    // always `supabase/config.toml` regardless of cwd or `--workdir`. The full
+    // literal is the byte-exact Go output on Linux/macOS (POSIX EEXIST text);
+    // Windows Go prints its own separator and OS errno text and is deliberately
+    // out of scope for this parity fix.
     return yield* Effect.fail(
-      new InitAlreadyExistsError({
-        detail: `Config already exists at ${result.configPath}.`,
-        suggestion: "Run `supabase init --force` to overwrite the existing config.",
+      new LegacyInitConfigExistsError({
+        message: "failed to create config file: open supabase/config.toml: file exists",
```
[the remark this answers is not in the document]
Coly010: Accepted — fixed in 6d262569. You're right that Windows is a published target for this binary (`packages/cli-windows-x64` ships `bin/supabase.exe`), so the POSIX-only literal was a real parity gap. Go's output there genuinely differs: `utils.ConfigPath` is built with `filepath.Join` ([internal/utils/misc.go:82](https://github.com/supabase/cli/blob/2b81bf4ce36e162850c1100c169625547b84a895/apps/cli-go/internal/utils/misc.go#L82)), so the wrapped `*os.PathError` ([internal/utils/config.go:243-246](https://github.com/supabase/cli/blob/2b81bf4ce36e162850c1100c169625547b84a895/apps/cli-go/internal/utils/config.go#L243-L246)) uses a backslash, and the `O_EXCL` open fails with `ERROR_FILE_EXISTS`, which Go's `syscall.Errno.Error()` renders as `The file exists.` — full Windows form: `failed to create config file: open supabase\config.toml: The file exists.` The suggestion line is unchanged since `errors.Is(err, os.ErrExist)` matches on Windows too ([internal/init/init.go:38-42](https://github.com/supabase/cli/blob/2b81bf4ce36e162850c1100c169625547b84a895/apps/cli-go/internal/init/init.go#L38-L42)).

The handler now selects the message on `RuntimeInfo.platform`, and the integration suite covers it with a mocked `win32` platform asserting the Windows string and the composed stderr. Two scope notes: the vitest suites only run on Ubuntu in CI, so coverage is via the mocked platform rather than a Windows runner; and since this PR's POSIX string was established by running the built Go binary, the Windows literal (derived from `filepath.Join` + documented `CREATE_NEW`/`FormatMessage` semantics) is flagged for a one-time confirmation on a real Windows box as follow-up.

Coly010:
@codex review

apps/cli/src/legacy/commands/init/SIDE_EFFECTS.md
```diff
@@ -53,6 +53,20 @@ In interactive mode (`-i`/`--interactive`), may prompt for IDE settings preferen
 
 Success is emitted as raw text even when the legacy shell is invoked with non-text output modes.
 
+When `supabase/config.toml` already exists and `--force` is not set (stderr, byte-matching Go's wrapped `O_EXCL` open error and `CmdSuggestion` on Linux/macOS; on Windows the Go CLI surfaces the OS path separator and errno text instead, which this port does not reproduce):
```
kanadgupta: This parenthetical is stale as of 6d26256 ("render Windows form of init already-exists error to match Go"): the port now *does* reproduce the Windows form — `init.handler.ts` has a `win32` branch emitting `failed to create config file: open supabase\config.toml: The file exists.`, and there's an integration test pinning it. Please update this note (and ideally document the Windows two-line output here too, since this file is the parity contract). The PR description's "byte parity is scoped to Linux/macOS" note is similarly outdated.

kanadgupta (commented):
Claude verified it all works as expected so LGTM. It did flag one small issue (see below):

**One real issue (inline):** `SIDE_EFFECTS.md` still says the Windows form is not reproduced, which the follow-up commit `6d26256` made stale; the PR description's Windows note is similarly outdated.

kanadgupta (approved):

apps/cli/src/legacy/commands/init/SIDE_EFFECTS.md
Coly010: Good catch, fixed in dddd902 — updated the SIDE_EFFECTS.md note (and the PR description) to document both platform variants instead of the stale "Windows isn't reproduced" claim.

PR #5993, opened 2026-07-29
test(cli): make homeDir-dependent tests hermetic

kanadgupta:
Tests that combined `mockRuntimeInfo`'s fixed default `homeDir` (`/tmp/supabase-cli-test-home`) with a real `FileSystem` layer silently picked up stale files left by older test runs or manual CLI invocations — e.g. a leftover `~/.supabase/profile` containing `supabase-staging` broke two `legacy-cli-config` unit tests on a developer machine.

- `mockRuntimeInfo` / `mockTelemetryRuntime` now default to a per-process unique path that is never created on disk, so the stale-state class of failure can't recur.
- Tests that build the real `legacyCliConfigLayer` / `legacyCredentialsLayer` against a real filesystem (the six `*.layers.unit.test.ts` layer-exposure tests, the four experimental-gate tests, and `legacy-cli-config.layer.unit.test.ts`) now pin `homeDir` to a per-test temp dir and isolate ambient env via a new `legacyIsolatedHomeLayer` helper, which also replaces the hand-rolled `TelemetryRuntime` blocks with `mockTelemetryRuntime`.
- The hermeticity rule is documented in the testing policy in `apps/cli/AGENTS.md`.

Verified by recreating a poisoned `/tmp/supabase-cli-test-home` (stale profile + access token) and running the full core suite against it.


kanadgupta:
Test failures are due to an unrelated issue, fix for that is in https://github.com/supabase/cli/pull/5997

avallete (approved):

PR #6004, opened 2026-07-30
test(stack): verify postgres data survives native to docker mode transition

Coly010:

Test only. Adds `packages/stack/tests/postgresDataPersistence.e2e.test.ts`, a new e2e test.

`packages/stack` can run Postgres either natively as a binary process or as a Docker
container, chosen by `StackConfig.mode`. Both paths mount/set `PGDATA` to the same
`dataDir`, and the design intent is that a stack stopped while running natively can be
restarted in Docker mode against the same `dataDir` without losing data. This had never
been tested end-to-end, and there was specific reason for suspicion: the Docker
entrypoint in `packages/stack/src/services/postgres.ts` execs
`postgres -D /etc/postgresql -p ${port}` — a different path than the
`/var/lib/postgresql/data` volume mount target — so it was unverified whether the two
actually resolve to the same data inside the `supabase/postgres` image.

## What is the new behavior?

The new test starts a stack in native mode, writes a marker row via `Bun.SQL` directly
against Postgres, disposes the stack (explicit `dataDir`s are never auto-cleaned, per
`cleanup.ts`), then starts a second stack in Docker mode against the same `dataDir` and
verifies:
- the first stack really ran postgres as a native process (no matching Docker container)
- the second stack really ran postgres as a Docker container
- the marker row written natively is still present and unchanged after the transition

Ran repeatedly (including a sanity check that points the Docker stack at a *different*
fresh `dataDir`, which reproducibly fails with `relation "public.persistence_marker" does
not exist`, confirming the assertion has real detection power): the persistence
assertion passes consistently. The `/etc/postgresql` vs `/var/lib/postgresql/data` path
mismatch does not break persistence in practice — data written natively is correctly
visible after switching to Docker mode against the same `dataDir`.


packages/stack/tests/postgresDataPersistence.e2e.test.ts
```diff
@@ -0,0 +1,202 @@
...
+
+dockerDescribe("postgres native/docker data persistence e2e", () => {
+  let dataDir: string;
+
+  beforeAll(() => {
+    dataDir = mkdtempSync(join(tmpdir(), "supabase-e2e-persist-"));
+  });
+
+  afterAll(() => {
+    rmSync(dataDir, { recursive: true, force: true });
```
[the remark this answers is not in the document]
Coly010: Thanks for flagging this — the fix I pushed a few minutes ago for a CI e2e failure (a different symptom, `EFAULT` from Bun's `rmSync` on Linux, per this repo's ci-triage) touched this exact line: I wrapped the outer `afterAll`'s `rmSync(dataDir, { recursive: true, force: true })` in `try { ... } catch {}`.

That already satisfies your stated "at minimum" bar — it's textually the same pattern already used in `createStack-docker.e2e.test.ts`'s teardown, which bind-mounts a host temp dir into a Docker-mode Postgres container the same way and accepts the same silent-catch fallback for the identical UID-ownership exposure.

On the stronger suggestion (actively restoring ownership or removing via a privileged container step): leaving that out for now. It'd be inconsistent with the existing accepted precedent in the sibling fixture, and it's a local-dev-only disk-litter concern — our e2e jobs run on ephemeral CI runners that are destroyed after each job, so there's no accumulation risk in CI. Given this repo's e2e tests are meant to stay minimal, the extra machinery doesn't earn its complexity here.

Resolving as already addressed by 38becf7d.

packages/stack/tests/postgresDataPersistence.e2e.test.ts
```diff
@@ -0,0 +1,202 @@
...
+
+      apiPort = new URL(stack.url).port;
+      const containerName = dockerContainerNameFor(apiPort);
+
+      try {
+        await stack.start();
+      } catch (startError) {
+        let logs: string;
+        try {
+          logs = execSync(`docker logs ${containerName}`, { encoding: "utf8" });
```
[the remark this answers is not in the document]
Coly010: Accepting this — confirmed end to end. `makePostgresServiceDocker` runs the container with `--rm` (`packages/stack/src/services/postgres.ts:180-181`), so a Postgres crash removes the container before this catch block's `docker logs` call runs, and that call will just report "no such container." `stack.logHistory("postgres")` is a real API already available on `stack` here (`StackHandle.logHistory`), backed by a log buffer fed from the child process's live stdout/stderr as it runs, so it survives the container being removed.

Pushed in f275be29: added a `stack.logHistory("postgres")` capture as the primary diagnostic source, and kept the existing `docker logs` exec as a secondary best-effort capture (still useful for failure modes where the container survives, e.g. a health-check timeout rather than a crash). Thanks for catching this — as written it would have silently produced a useless message on exactly the failure this block exists to explain.

packages/stack/tests/postgresDataPersistence.e2e.test.ts
```diff
@@ -0,0 +1,220 @@
...
+
+    test(
+      "starts postgres successfully by reusing the native dataDir",
+      { timeout: TEST_TIMEOUT_MS },
+      () => {
+        if (startFailureDiagnostics !== undefined) {
+          throw new Error(startFailureDiagnostics);
+        }
+      },
+    );
```
kanadgupta: This placeholder test plus the `startFailureDiagnostics` state and the guard-throws in the two real tests (lines 190-195 and 209-213) reimplement what vitest already does when a `beforeAll` hook throws: the suite's tests are reported as failed with the hook's error, and `afterAll` still runs `dispose()`. Phase 1 of this same file and `createStack-docker.e2e.test.ts` both use the simpler dispose-and-rethrow pattern. Throwing `new Error(startFailureDiagnostics)` at the end of the catch block in `beforeAll` would preserve the full diagnostics while deleting this empty test, the module-level `startFailureDiagnostics` variable, and both guard blocks (~30 lines). As a bonus it removes the misleading "Skipped:" messages, which actually report as test failures, not skips.

packages/stack/tests/helpers/e2e.ts
```diff
@@ -45,6 +45,20 @@ export async function fetchFunctionWhenReady(
...
+ * Convert a `Bun.SQL` query result into a plain array of rows.
+ *
+ * The result of `sql.unsafe()`/a tagged-template query is a decorated `Array`
+ * subclass (`SQLResultArray`) that also carries extra own properties
+ * (`count`, `command`, `lastInsertRowid`, `affectedRows`) alongside the real
+ * rows, which makes it fail `toEqual`/structural-equality assertions against
+ * a plain array literal even when the rows themselves match exactly. Individual
+ * row objects are already plain, so copying the container is enough.
+ */
+export function toPlainRows<T>(result: Iterable<T>): T[] {
```
kanadgupta: `toPlainRows` has exactly one consumer (`queryMarkerRows` in the new test), so landing it in the shared helpers file is speculative — the commit message calls it a "shared helper for Bun.SQL result comparisons" but no other test compares Bun.SQL results this way (`setupTestTable` never reads results, and the other e2e tests go through supabase-js). Inlining `Array.from(result)` at the call site with a one-line comment would keep the shared-helper surface from growing for a single use; move it here if/when a second consumer appears.

kanadgupta (approved):
Test is great! One suggestion re: the `shared` function and some feedback on the placeholder test with the help of Claude. Non-blocking though, LGTM!

packages/stack/tests/postgresDataPersistence.e2e.test.ts
Coly010: Fixed in 5436f3c1: beforeAll now throws with full diagnostics directly, matching Phase 1's dispose-and-rethrow pattern; removed the placeholder test, startFailureDiagnostics, and both guard blocks.

packages/stack/tests/helpers/e2e.ts
Coly010: Fixed in 5436f3c1: inlined Array.from(result) at the one call site and removed toPlainRows from the shared helpers.

PR #4933, opened 2026-03-05
fix(functions): forward NPM_AUTH_TOKEN to deploy bundler container

addniner:
## Summary

The Docker bundler for `functions deploy` forwards `NPM_CONFIG_REGISTRY` from the host environment into the container. When `.npmrc` uses `${NPM_AUTH_TOKEN}` for private registry authentication, the token also needs to be available inside that bundler container.

This PR forwards `NPM_AUTH_TOKEN` from the host environment alongside `NPM_CONFIG_REGISTRY`.

Addresses the CI/CD host environment case in #4927.

## Reproduction

| # | Method | Result | Error |
|---|--------|--------|-------|
| 1 | `supabase functions deploy hello` (`.npmrc` with `${NPM_AUTH_TOKEN}`) | **Fail** | `401 unauthenticated` |
| 2 | `NPM_AUTH_TOKEN=xxx supabase functions deploy hello` | **Fail** | `401 unauthenticated` (env not forwarded to container) |
| 3 | `.npmrc` with hardcoded token | **Pass** | - |
| 4 | `NPM_AUTH_TOKEN=xxx` with patched CLI | **Pass** | - |

Tested with a private package (`@addniner/test-private-pkg`) on GitHub Packages, with Docker cache cleared between attempts.

## Testing

- `go test ./internal/functions/deploy`

Issue #4927, opened 2026-03-04
`functions deploy` fails to read `.npmrc` environment variables for private npm registry auth (workspace `deno.json`)


addniner:
#### Describe the bug

When using a **root-level workspace `deno.json`** at `supabase/functions/deno.json` (instead of per-function `deno.json` files), `supabase functions deploy` fails to resolve the environment variable in `.npmrc`.

The `.npmrc` file uses `${NPM_AUTH_TOKEN}` for private registry authentication. `functions serve` reads this variable correctly (from `supabase/functions/.env`), but `functions deploy` does not — resulting in a registry authentication failure.

#### `.npmrc` configuration

```
@myorg:registry=https://npm.pkg.github.com
//npm.pkg.github.com/:_authToken=${NPM_AUTH_TOKEN}
```

The `NPM_AUTH_TOKEN` is provided via `supabase/functions/.env`, which `functions serve` loads automatically.

#### Project structure

```
supabase/functions/
├── deno.json              # Root workspace config
│   ├── "workspace": ["./*"]
│   └── "imports": {
│         "@my-private-pkg": "npm:@myorg/my-package@^1.0.0",
│         ...
│       }
├── .npmrc                 # Private registry auth
├── .env                   # NPM_AUTH_TOKEN=ghp_xxxxx
├── function-a/
│   ├── index.ts
│   └── deno.json          # Function-specific imports only
├── function-b/
│   └── ...
└── _shared/               # Shared utilities
```

**Why this structure?**
The [official docs recommend per-function `deno.json`](https://supabase.com/docs/guides/functions/dependencies), but this creates a dependency management problem with `_shared/` code — shared utilities would need their npm dependencies duplicated across every function's `deno.json`. A root workspace `deno.json` solves this. This approach is also discussed as practical in [Discussion #33595](https://github.com/orgs/supabase/discussions/33595).

#### Steps to reproduce

1. Set up the project structure above with a private npm package
2. Configure `.npmrc` with `${NPM_AUTH_TOKEN}` environment variable
3. Set the token in `supabase/functions/.env`
4. Run `supabase functions serve` → **Works**, private package resolves correctly
5. Run `supabase functions deploy` → **Fails**, authentication error (`.npmrc` env var not resolved)

#### Expected behavior

`functions deploy` should resolve environment variables in `.npmrc` the same way `functions serve` does.

#### Workaround

[...]

PR #6005, opened 2026-07-30
fix(cli): match Go bundler env and deploy path anchoring (CLI-1985)

Coly010:
Two `functions deploy` divergences from the pinned Go CLI (`apps/cli-go`), resolved per the CLI-1985 ruling (Colum, 2026-07-30: take each point's documented recommended option).

Fixes CLI-1985

### Point 1 — `NPM_AUTH_TOKEN` is no longer forwarded into the Docker bundler (strict parity; **breaking** for private-registry users)

**Decision:** remove the forwarding. The eszip bundler container now receives only `NPM_CONFIG_REGISTRY` from the host, exactly matching Go (`apps/cli-go/internal/functions/deploy/bundle.go:68-70`).

**Evidence:**
- The Go CLI never forwarded `NPM_AUTH_TOKEN` at any point in its history — only `NPM_CONFIG_REGISTRY` (added in `8e17f033`).
- The Go-side PR proposing the token forwarding (#4933, addressing #4927) was **closed unmerged** on 2026-06-22 ("The command is now ported in TypeScript so I'm closing this PR").
- The TS-only forwarding came from #5645, which ported the unmerged #4933. CLI-1985 ruled strict parity over that TS-only addition.

**User-visible change (flagging prominently):** users whose `.npmrc` expands `${NPM_AUTH_TOKEN}` for private npm registries will find `--use-docker` / `--legacy-bundle` deploys failing registry auth again (the pre-#5645 and Go CLI behavior; re-opens the CI/CD-host case of #4927). Workarounds: inline the token in `.npmrc`, or deploy via the default `--use-api` path. Per the strict-parity contract (stderr bytes included), no TS-only warning was added when the variable is set — a DX reviewer requested one and it was rejected on parity grounds; the breaking impact is documented here and in the commit message instead.

**Shared-code caveat (per the ruling):** `dockerNpmEnv` lives in `apps/cli/src/shared/functions/deploy.ts` and serves **both** shells — `next/` (`functions deploy`) and `legacy/`. The removal therefore applies to the next/ shell too. The strict-parity contract only binds the legacy shell, but keeping one code path is the simplest correct design per repo policy, so next/ loses the forwarding as well — stated here explicitly. `functions serve` is unaffected (it has its own env handling, matching Go's serve which loads `supabase/functions/.env`).

### Point 2 — API-deploy upload paths re-anchored at the workdir (align to the pinned oracle; behaviour change)

**Directive:** confirm the intended reference point first, then align or record.

**Evidence found:**
- Upstream Go **never** anchored deploy paths at the git root. The full history of `pkg/function/deploy.go` (pre- and post-monorepo move) shows `toRelPath` anchored at `os.Getwd()` since `29021998` ("convert all paths to relative for deploy", #3403), unchanged since. The Go CLI chdirs to the workdir (`internal/utils/misc.go:238`), so `os.Getwd()` ≡ the workdir.
- The TS git-root anchoring came from #5755 (merged 2026-07-02), a deliberate TS-side monorepo fix closing #3467 (Go hard-fails on imports outside the workdir: `failed to read file: open ../common/index.ts`) — **not** a port of newer upstream Go behavior. There is no newer upstream Go reference to record against.

**Decision (per the ruling's matrix — upstream never did this → align):** uploaded multipart file names and the server-recorded `entrypoint_path` / `import_map_path` / `static_patterns` are now anchored at the workdir with Go's exact `toRelPath` semantics (relative to `os.Getwd()`, forward slashes, `../`-relative when the file lies outside the workdir).

**Scope note:** #5755's import-walk *containment boundary* (which files may be uploaded at all) is intentionally **kept** at the nearest git root. The boundary is a TS-only safeguard with no Go equivalent — Go's walker uploads any reachable import unbounded (and then hard-fails opening `..` paths through `afero.NewIOFS`, which is exactly bug #3467). Reverting the boundary would re-break #3467 and is outside CLI-1985's anchoring scope.

**User-visible change:** in monorepos where the git root is an ancestor of the workdir, redeploys now record `supabase/functions/<slug>/index.ts` (matching what the Go CLI records and the dashboard shows for Go deploys) instead of `apps/myapp/supabase/functions/<slug>/index.ts`. Imports outside the workdir but inside the git root still deploy, uploading with Go-style `../`-relative names — the same name shape Go's `toRelPath` emits, so nothing new is required of the server. Non-monorepo projects (git root == workdir, the common case) are byte-for-byte unchanged.

## What changed

- `apps/cli/src/shared/functions/deploy.ts` — `dockerNpmEnvNames` trimmed to `NPM_CONFIG_REGISTRY`; `deployViaApi` now threads the workdir as the path anchor through `uploadFunctionSource` / `writeSourceDeployForm` / `createSourceMetadata` while the git-root `sourceRoot` remains the containment boundary; ENOENT warn display paths follow the workdir anchor (matching Go's workdir-relative walker paths). Docker bind construction is untouched.
- `apps/cli/src/legacy/commands/functions/deploy/deploy.integration.test.ts` — new regression test: workdir≠git-root monorepo deploy asserts workdir-anchored metadata, `../`-relative upload names, and the Go-parity `Uploading asset` stderr line.
- `apps/cli/src/next/commands/functions/deploy/deploy.integration.test.ts` — the two git-root upload tests updated to the workdir anchoring; the npm env test now asserts `NPM_CONFIG_REGISTRY` is forwarded and `NPM_AUTH_TOKEN` is not.
- `apps/cli/src/legacy/commands/functions/deploy/SIDE_EFFECTS.md` — env table states only `NPM_CONFIG_REGISTRY` is forwarded; new note documents the workdir anchoring and the TS-only git-root boundary.

All four changed/added tests fail against the previous implementation and pass with this change.

### DIFF
--- a/apps/cli/src/shared/functions/deploy.ts
+++ b/apps/cli/src/shared/functions/deploy.ts
@@ -319,6 +319,21 @@ function isContainedInAnyPath(roots: ReadonlyArray<string>, candidate: string) {
   return roots.some((root) => isContainedPath(root, candidate));
 }
 
+/**
+ * Go parity (`apps/cli-go/pkg/function/deploy.go:251-284`, via
+ * `afero.IOFS.Open` → `fs.ValidPath`): `writeForm`'s `addFile` opens every
+ * uploaded path through an `fs.FS`, which rejects any path containing a `..`
+ * element before the read (and thus the upload) happens. A workdir≠git-root
+ * layout can otherwise produce a multipart `File` name like
+ * `../packages/shared/src/index.ts` that escapes the anchor dir — reject it
+ * the same way Go does, before any upload is attempted.
+ */
+function hasParentPathSegment(relativePath: string) {
+  return toSlash(relativePath)
+    .split("/")
+    .some((segment) => segment === "..");
+}
+
 async function realpathIfExists(pathname: string) {
   try {
     return await realpath(resolve(pathname));
@@ -917,6 +932,9 @@ async function writeSourceDeployForm(
     // (`apps/cli-go/pkg/function/deploy.go:94-103`, relative to `os.Getwd()`),
     // NOT at `sourceRoot` — see the CLI-1985 note in `deployViaApi`.
     const relativePath = toApiRelativePath(workdir, pathname);
+    if (hasParentPathSegment(relativePath)) {
+      throw new Error(`failed to read file: open ${relativePath}: invalid argument`);
+    }
     await Effect.runPromise(outputRaw(`Uploading asset (${config.slug}): ${relativePath}\n`));
     form.append("file", new File([contents], relativePath));
   };
--- a/apps/cli/src/shared/functions/deploy.ts
+++ b/apps/cli/src/shared/functions/deploy.ts
@@ -261,7 +261,14 @@ const dockerComposeProjectLabel = "com.docker.compose.project";
  * directory using its OWN workdir rather than the caller's cwd.
  */
 export const dockerWorkdirLabel = "com.supabase.cli.workdir";
-const dockerNpmEnvNames = ["NPM_CONFIG_REGISTRY", "NPM_AUTH_TOKEN"] as const;
+/**
+ * Go parity (`apps/cli-go/internal/functions/deploy/bundle.go:68-70`): the eszip
+ * bundler container receives only `NPM_CONFIG_REGISTRY` from the host
+ * environment. `NPM_AUTH_TOKEN` is deliberately NOT forwarded — the Go-side PR
+ * proposing it (supabase/cli#4933) was closed unmerged, and CLI-1985 ruled
+ * strict parity over the TS-only forwarding that #5645 had added.
+ */
+const dockerNpmEnvNames = ["NPM_CONFIG_REGISTRY"] as const;
 
 export function dockerProjectLabels(projectId: string) {
   return {
@@ -890,6 +897,7 @@ async function resolveImportMapAllowedRoots(projectRoot: string, importMapPath: 
 
 async function writeSourceDeployForm(
   sourceRoot: string,
+  workdir: string,
   config: ResolvedDeployFunctionConfig,
   metadata: SourceDeployMetadata,
   outputRaw: (text: string) => Effect.Effect<void, never>,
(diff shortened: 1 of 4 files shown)


Coly010:
@codex review

apps/cli/src/shared/functions/deploy.ts
```diff
@@ -905,7 +913,10 @@ async function writeSourceDeployForm(
       return;
     }
     uploadedAssets.add(realPathname);
-    const relativePath = toApiRelativePath(sourceRoot, pathname);
+    // Uploaded file names are anchored at the workdir like Go's `toRelPath`
+    // (`apps/cli-go/pkg/function/deploy.go:94-103`, relative to `os.Getwd()`),
+    // NOT at `sourceRoot` — see the CLI-1985 note in `deployViaApi`.
+    const relativePath = toApiRelativePath(workdir, pathname);
```
kanadgupta, line 934: This makes `../packages/shared/src/index.ts`-style multipart `File` names reach the server for the first time from any CLI: Go hard-fails before uploading files outside the workdir (`writeForm`'s `fsys.Open` via `fs.FS` rejects `..` path elements — that's bug #3467, per your own scope note), and the previous TS implementation anchored at the git root, so names never contained `../`. The PR body's "nothing new is required of the server" claim therefore doesn't hold — this exact name shape has plausibly never been received by the platform.

The integration tests mock the API, so CI can't validate that server-side bundling accepts these names and resolves the entrypoint's relative imports to them. Since this is exactly the #5755 monorepo scenario that worked before this PR (with git-root names), a server-side rejection or path normalization would silently regress a working feature.

Could you run one real `--use-api` deploy of the workdir≠git-root shape against a live project (or link server-side evidence that relative-name resolution handles `..`) and note the result in the PR?

kanadgupta (approved):
Claude discovered a potential issue where the server might reject the file name and suggests trying this yourself — I think this would be better suited for the e2e tests I'm hoping to work on soon. Will defer to you on the best course of action here — LGTM otherwise!

apps/cli/src/shared/functions/deploy.ts
Coly010: Rather than validating this only via a live `--use-api` deploy, e1b18f3d adds a Go-parity guard: any bundled file whose relative name would escape the anchor dir (contain a `..` segment) now hard-fails before upload, matching Go's `fs.FS` `..`-rejection in `writeForm`/`addFile` (`pkg/function/deploy.go:251-284`, via `afero.IOFS.Open` → `fs.ValidPath`). This closes the gap directly — that input can no longer reach the server at all, regardless of server-side behavior.

The guard lives in `appendAsset` in `apps/cli/src/shared/functions/deploy.ts`, the single funnel every uploaded multipart `File` passes through (entrypoint, static files, import map, and import-map targets alike), so it catches the escape uniformly no matter how it's reached. I also found the same escape was reachable via an explicit `--import-map ../outside/path.json` flag (not just an auto-discovered monorepo import), and updated the existing deploy integration tests that had asserted the escaping upload succeeded — they now assert the hard-fail instead.

PR #6009, opened 2026-07-30
chore(cli): apply CLI-1989 parity ruling for db push pipeline-incompatible statements

Coly010:

Parity-ruling record + pinned-Go oracle alignment for `db push` / `db reset` / `migration up` pipeline-incompatible statement handling.

Fixes [CLI-1989](https://linear.app/supabase/issue/CLI-1989/parity-ruling-db-push-runs-pipeline-incompatible-statements-outside)

## ⚖ Parity ruling applied (Colum, 2026-07-30)

The parity audit flagged that the TS migration apply (`apps/cli/src/legacy/shared/legacy-migration-apply.ts`) runs pipeline-incompatible statements (`CREATE [UNIQUE] INDEX CONCURRENTLY`, `REINDEX … CONCURRENTLY`, `VACUUM`, `ALTER SYSTEM`, `CLUSTER`) standalone outside the batch transaction, while the pinned Go reference (`apps/cli-go/pkg/migration/file.go`) had no such handling — a migration containing `CREATE INDEX CONCURRENTLY` failed under pinned Go (SQLSTATE 25001) but succeeded under TS.

**Ruling:** confirm the intended upstream reference, record it, and update the pinned `apps/cli-go` to match so future audits don't re-flag this. The TS behaviour is **not** reverted.

### Confirmed provenance

* **Bug:** supabase/cli#5139 — `db reset` fails with SQLSTATE 25001 ("CREATE INDEX CONCURRENTLY cannot be executed within a pipeline").
* **Reference design:** PR [supabase/cli#5156](<https://github.com/supabase/cli/issues/5156>) by @wucm667 — `isPipelineIncompatible` / `trimLeadingSQLComments` + flush-then-run-standalone in Go's `ExecBatch`. **Closed WITHOUT merging** on 2026-06-24: there is no merged upstream Go commit. The closing comment adopted the design directly into the TS port instead (PR-branch commit 29d3fb0e, part of supabase/cli#5671, squash-merged to develop as b48fad60) because the Go path was being retired for the migration commands.

### TS vs. reference semantics

Compared statement-for-statement against the [#5156](<https://github.com/supabase/cli/issues/5156>) diff: the TS classification patterns, comment/BOM trimming, flush-then-standalone execution order, final-batch history insert, and global statement-index accounting in `At statement: N` errors all match. The only deltas are regex-engine trivia (Go RE2 `\s`/`\z` vs JS `\s`/`$`) with no observable effect on valid SQL — no TS behaviour change was needed.

### What this PR does

1. **Pinned Go updated** (`chore(cli-go)` commit): the [#5156](<https://github.com/supabase/cli/issues/5156>) diff applies cleanly and is self-contained (`pkg/migration/file.go` + tests, additive `pkg/pgtest` simple-query mock helpers; no dependency or API churn), so it is applied **verbatim** to `apps/cli-go`. `go test ./migration/...` in `apps/cli-go/pkg` and the full main-module `go test ./...` pass; golangci-lint adds no new findings (the 5 pre-existing gosec findings in `internal/utils/*` are untouched).
2. **Provenance recorded** (`docs(cli)` commit): the provenance note in `legacy-migration-apply.ts`, the `db push` / `db reset` / `migration up` SIDE_EFFECTS.md (including the non-atomic flush-boundary semantics: a mid-file failure leaves earlier batches committed with no history row, so a re-run replays the file from the top — prefer `… IF NOT EXISTS` forms), and the `db push` / `db reset` / `migration up` rows in `docs/go-cli-porting-status.md`.
3. **Classifier test parity** (`test(cli)` commit): the TS `legacyIsPipelineIncompatible` suite claimed to mirror Go's `TestIsPipelineIncompatible` but was missing two of its negatives (string literal, leading-comment-only); those plus BOM / unterminated-comment edge cases are added. 40/40 pass.

### Shipped-sidecar behaviour note

`apps/cli-go` is not just an audit-only reference — it is compiled into the shipped `supabase-go` sidecar (`build:go-sidecar` in `apps/cli/package.json`; `go build -o supabase-go` in CI). Local `db start` and local `db reset` delegate migration apply to that sidecar via the hidden `db __db-bootstrap` seam, which calls `ExecBatch` in `pkg/migration/file.go`. So the pinned-Go update in this PR also extends the supabase/cli#5139 `CREATE INDEX CONCURRENTLY` / pipeline-mode fix to that delegated local path: a migration with `CREATE INDEX CONCURRENTLY` previously failed under local `db start`/`db reset` with SQLSTATE 25001 and now succeeds, matching the already-working remote TS path. This is filed as `chore` because it's realigning the pinned parity oracle rather than introducing new intentional TS behaviour, but it's fair to note it arguably deserves `fix` framing given it's a real shipped runtime behaviour change on the local path, not purely an audit fix. A follow-up `docs(cli)` commit (5a7f833a) adds a one-line note on `ExecBatch`'s definition recording this dual-path (remote TS + local sidecar) reach, and a note on the known `\v`/Unicode-whitespace classification delta between JS `\s` and Go's RE2 `\s` (a residual valid-SQL divergence: PostgreSQL >= 14 treats `\v` as SQL whitespace but Go RE2 `\s` doesn't match it, so e.g. `VACUUM\v(FULL)` classifies as pipeline-incompatible in TS but not under the Go oracle).

Follow-up candidates (not in this PR): an error-message hint when a standalone statement fails mid-file, and a user-facing docs-site note about `CONCURRENTLY` migration best practice.


Coly010:
@codex review

apps/cli-go/pkg/migration/file.go
```diff
@@ -72,23 +77,46 @@ func NewMigrationFromReader(sql io.Reader) (*MigrationFile, error) {
...
-	if len(m.Version) > 0 {
-		if err := m.insertVersionSQL(conn, batch); err != nil {
-			return err
+func isPipelineIncompatible(sql string) bool {
+	upper := strings.ToUpper(trimLeadingSQLComments(sql))
+	return createIndexPattern.MatchString(upper) ||
+		reindexPattern.MatchString(upper) ||
+		vacuumPattern.MatchString(upper) ||
+		alterSystemPattern.MatchString(upper) ||
+		clusterPattern.MatchString(upper)
```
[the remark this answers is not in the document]
Coly010: **Rejected — real gap, but it is a shared design gap in the sanctioned reference (closed Go PR #5156), not a defect in this port.**

This PR's mandate (CLI-1989 ⚖ ruling, see the PR body) is to record provenance and make the pinned `apps/cli-go` match the intended reference **verbatim**. The ported `isPipelineIncompatible` patterns are byte-identical to #5156's diff, which has no `DROP INDEX CONCURRENTLY` pattern — and the sanctioned TS implementation (`legacy-migration-apply.ts`) behaves identically: both classifiers return `false` for `DROP INDEX CONCURRENTLY idx` (verified by running both against the same probe set). Adding the pattern here would desync the port from #5156 and from TS unless both sides changed together, which is new scope beyond this ruling.

Noted as a follow-up candidate (classifier extension applied to Go + TS in lockstep, with `REINDEX`-style tests).

apps/cli-go/pkg/migration/file.go
```diff
@@ -72,23 +77,46 @@ func NewMigrationFromReader(sql io.Reader) (*MigrationFile, error) {
...
+		case strings.HasPrefix(trimmed, "--"):
+			if idx := strings.IndexByte(trimmed, '\n'); idx >= 0 {
+				trimmed = strings.TrimLeft(trimmed[idx+1:], " \t\n\r")
+				continue
+			}
+			return ""
+		case strings.HasPrefix(trimmed, "/*"):
+			if idx := strings.Index(trimmed, "*/"); idx >= 0 {
+				trimmed = strings.TrimLeft(trimmed[idx+2:], " \t\n\r")
+				continue
```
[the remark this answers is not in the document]
Coly010: **Rejected — shared design gap in the sanctioned reference (closed Go PR #5156), not a defect in this port.**

`trimLeadingSQLComments` here is a verbatim transcription of #5156's implementation (first-`*/`-wins, non-nesting), and the sanctioned TS port (`legacyTrimLeadingSqlComments`) transcribes the same logic. Verified empirically: for `/* outer /* inner */ still comment */ CREATE INDEX CONCURRENTLY …` both the ported Go and TS return `false` — identical behaviour, including the misclassification. Fixing nesting only here would desync the pinned oracle from both #5156 and TS; per the CLI-1989 ruling this PR syncs verbatim and records provenance.

Noted as a follow-up candidate alongside the `DROP INDEX CONCURRENTLY` gap (both sides in lockstep).

apps/cli-go/pkg/migration/file.go
```diff
@@ -29,6 +29,11 @@ type MigrationFile struct {
 var (
 	migrateFilePattern = regexp.MustCompile(`^([0-9]+)_(.*)\.sql$`)
 	typeNamePattern    = regexp.MustCompile(`type "([^"]+)" does not exist`)
+	createIndexPattern = regexp.MustCompile(`^CREATE\s+(UNIQUE\s+)?INDEX\s+CONCURRENTLY(\s|\z)`)
+	reindexPattern     = regexp.MustCompile(`^REINDEX(\s|\().*\sCONCURRENTLY(\s|\z)`)
```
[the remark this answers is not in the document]
Coly010: **Rejected — shared design gap in the sanctioned reference (closed Go PR #5156), not a defect in this port; narrower than stated.**

The pattern is byte-identical to #5156's (`^REINDEX(\s|\().*\sCONCURRENTLY(\s|\z)`). Empirically (Go RE2 and JS agree on every probe):
- `REINDEX INDEX\nCONCURRENTLY idx` → **true** in both — the single `\s` matches the newline, so the common multiline form IS classified.
- `REINDEX (VERBOSE)\nINDEX CONCURRENTLY idx` → false in both — only when `.*` must cross the newline does the match fail, identically in the ported Go and the sanctioned TS.

So there is no port defect and no Go↔TS divergence; the residual miss is #5156's own design, out of scope for this provenance/sync PR per the CLI-1989 ruling. Noted as a follow-up candidate with the other two classifier gaps.

apps/cli-go/pkg/migration/file.go
```diff
@@ -72,23 +77,46 @@ func NewMigrationFromReader(sql io.Reader) (*MigrationFile, error) {
...
-	// Batch migration commands, without using statement cache
-	batch := &pgconn.Batch{}
-	for _, line := range m.Statements {
-		batch.ExecParams(line, nil, nil, nil, nil)
-	}
-	// Insert into migration history
-	if len(m.Version) > 0 {
-		if err := m.insertVersionSQL(conn, batch); err != nil {
-			return err
+func isPipelineIncompatible(sql string) bool {
```
kanadgupta, line 80: **Observation (non-blocking): this is a runtime change, not just an oracle sync.** `apps/cli-go` is compiled into the shipped `supabase-go` sidecar (`build:go-sidecar` in `apps/cli/package.json`; `go build -o supabase-go` in CI), and `db start` / local `db reset` delegate migration apply to its `db __db-bootstrap` seam, which calls `ExecBatch` here. So this PR also extends the #5139 fix to the delegated local path — a migration with `CREATE INDEX CONCURRENTLY` previously succeeded on the TS remote path but failed with SQLSTATE 25001 under local `db reset`/`db start`; after this PR both succeed. That's the right convergence, but the PR body frames the Go update purely as audit-facing ("so future audits don't re-flag") — worth a sentence acknowledging the shipped-sidecar behavior change, and a thought on whether `chore` is the right release signal for it.

apps/cli/src/legacy/shared/legacy-migration-apply.ts
```diff
@@ -25,6 +25,16 @@ const BOM_CODE_POINT = 0xfeff;
...
+// Provenance (CLI-1989, parity ruling 2026-07-30): the intended reference for this
+// behaviour is the Go fix proposed for supabase/cli#5139 in PR supabase/cli#5156
+// (`isPipelineIncompatible` / `trimLeadingSQLComments` in `pkg/migration/file.go`).
+// That PR was closed WITHOUT merging — its design was adopted directly into this TS
+// apply instead in PR supabase/cli#5671 (squash-merged to develop as b48fad60; the
+// #5156 closing comment cites the PR-branch commit 29d3fb0e) because the Go path was
+// being retired for the migration commands. The pinned Go oracle (`apps/cli-go`) therefore
+// predated the fix; it now carries the same port of the closed PR (applied alongside
+// this note) so TS-vs-Go parity audits compare like for like.
 const CREATE_INDEX_CONCURRENTLY_PATTERN = /^CREATE\s+(?:UNIQUE\s+)?INDEX\s+CONCURRENTLY(?:\s|$)/u;
```
kanadgupta, line 44: **Minor, for audit-precision:** the PR body's "regex-engine trivia … no observable effect on valid SQL" is *almost* right, but there is one residual valid-SQL divergence: JS `\s` matches `\v` (RE2's `\s` is `[\t\n\f\r ]` and does not), and PostgreSQL ≥14 treats `\v` as SQL whitespace. Verified empirically: `VACUUM\v(FULL)` and `CREATE\vINDEX\vCONCURRENTLY idx` classify **true** here but **false** in the (now-synced) Go oracle, so under Go they'd batch and fail with 25001. Not worth changing behavior over, but since this PR exists to stop audits re-flagging this area, consider adding one line to this provenance note recording the known `\v` (and Unicode-whitespace) delta so the next parity sweep doesn't rediscover it.

kanadgupta (approved):
Claude flagged a couple minor things in the PR description that don't match the nature of the actual changes. Don't have the context to validate them for myself atm but LGTM otherwise!

apps/cli-go/pkg/migration/file.go
Coly010: Verified — apps/cli-go is compiled into the shipped supabase-go sidecar (build:go-sidecar / go build -o supabase-go in CI) and db start/db reset do delegate migration apply through the __db-bootstrap seam into this ExecBatch call (via apply.MigrateAndSeed). Updated the PR description to acknowledge this is a shipped runtime behavior change on the local path, not just audit-facing, and added a one-line comment on ExecBatch in file.go noting the dual-path reach (5a7f833a).

apps/cli/src/legacy/shared/legacy-migration-apply.ts
Coly010: Fixed in 5a7f833a: added a note documenting the known \v/Unicode-whitespace classification delta between JS \s and Go's RE2 \s.

PR #6017, opened 2026-07-31
fix(cli): reload kong after db reset

### LINKED ISSUE #6016: supabase db reset can leave Kong routing to a stale container IP, causing 502 on /auth/v1/*
### Affected area

Local development

### Supabase CLI version

2.110.0

### Operating system

Fedora Linux 44 (kernel 7.1.4-204.fc44.x86_64)

### Installation method

brew

### Command

```shell
supabase db reset
```

### Actual output

```shell
# The reset itself reports success and exits 0:
Resetting local database...
Recreating database...
Initialising schema...
Seeding globals from roles.sql...
Restarting containers...
Finished supabase db reset on branch main.

# But the auth route through Kong is now dead, and stays dead:
$ curl -o /dev/null -w '%{http_code}\n' http://127.0.0.1:54321/auth/v1/health
502

# Any /auth/v1 request fails the same way. A seed script creating the local admin
# through the admin API got the body form:
#   could not create <address>: 502 {
#     "message":"An invalid response was received from the upstream server"
#   }

# Kong is dialling an address the auth container no longer holds
# (auth was on 172.18.0.8 at this point):
$ docker logs supabase_kong_<project>
[error] 1110#0: *4461 connect() failed (111: Connection refused) while connecting to upstream,
  client: 172.18.0.1, server: kong, request: "GET /auth/v1/health HTTP/1.1",
  upstream: "http://172.18.0.4:9999/health", host: "127.0.0.1:54321"

# GoTrue itself is healthy and answers on its real address:
$ docker inspect -f '{{.State.Health.Status}}' supabase_auth_<project>
healthy
$ curl -o /dev/null -w '%{http_code}\n' http://172.18.0.8:9999/health
200

# Restarting Kong is the only thing that clears it:
$ docker restart supabase_kong_<project>
$ curl -o /dev/null -w '%{http_code}\n' http://127.0.0.1:54321/auth/v1/health
200
```

### Expected behavior

supabase db reset should leave the local stack fully routable.

Any container the reset restarts should either keep a stable address, or Kong should be
restarted (or made to re-resolve its upstreams) as part of the same operation, so that
/auth/v1/* keeps working once the command reports success.

Today the command exits 0 and `supabase status` reports every container healthy — because
each container individually is healthy — while requests through the gateway return 502.

### Steps to reproduce

Root cause first, because it explains why the reset-driven repro is intermittent:

Kong addresses its upstreams by container hostname and caches the resolved address for the
life of its process:

    url: http://supabase_auth_<project>:9999/
    url: http://supabase_storage_<project>:5000/

`supabase db reset` recreates the db container and restarts auth, realtime and storage, but
never restarts kong. Docker assigns addresses from a free pool on start, so a restarted
container can return on a different address. auth and storage are stopped and started
together, so whichever starts first takes the lower free address; when that order flips
between resets the two swap, and Kong's cached value is then wrong. It does not recover on
its own — Kong has to be restarted.

Intermittent path (what a user actually hits):

1. `supabase start`, wait for all containers healthy.
2. `curl -o /dev/null -w '%{http_code}\n' http://127.0.0.1:54321/auth/v1/health` → 200
3. Run `supabase db reset`.
4. Repeat 2-3. On the resets where the start order flips, step 2 returns 502 and keeps
   returning 502 until kong is restarted.

Deterministic path (forces the address change directly, no reset needed — substitute your
project suffix for <project>):

1. `supabase start`, wait for all containers healthy.
2. Confirm the route works:
       curl -o /dev/null -w '%{http_code}\n' http://127.0.0.1:54321/auth/v1/health   # 200
3. Make auth and storage exchange addresses:
       docker stop  supabase_auth_<project> supabase_storage_<project>
       docker start supabase_storage_<project>   # takes the address auth just released
       docker start supabase_auth_<project>      # lands on a different one
4. Confirm they swapped:
       docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' supabase_auth_<project>
       docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' supabase_storage_<project>
5. The route is now broken and stays broken:
       curl -o /dev/null -w '%{http_code}\n' http://127.0.0.1:54321/auth/v1/health   # 502
6. `docker restart supabase_kong_<project>` restores it.

Step 5 reproduced on the first attempt. Container names never change in any of this — only
the addresses behind them.

### Crash report ID

_No response_

```shell
$ docker version
Client: Docker Engine - Community
 Version:           29.6.2
 API version:       1.55
Server: Docker Engine - Community
  Version:          29.6.2
  API version:      1.55 (minimum version 1.40)

# `supabase services`, as emitted when piped (local versions only, remote omitted):
supabase/postgres        17.6.1.143
supabase/gotrue          v2.193.0
postgrest/postgrest      v14.15
supabase/realtime        v2.113.4
supabase/storage-api     v1.66.4
supabase/edge-runtime    v1.74.2
supabase/studio          2026.07.13-sha-b5ada96
supabase/postgres-meta   v0.96.6
supabase/logflare        1.47.1
supabase/supavisor       2.9.7
```

### Additional context

Evidence that kong is the container the reset leaves behind
-----------------------------------------------------------

Container identity before and after a single `supabase db reset`:

BEFORE

    /supabase_auth_<p>      id=0f4d55029bc3  started=2026-07-31T01:16:46Z  ip=172.18.0.4
    /supabase_db_<p>        id=99b4e26d6a1f  started=2026-07-31T01:16:30Z  ip=172.18.0.2
    /supabase_kong_<p>      id=c5ccc88fe58c  started=2026-07-31T01:17:23Z  ip=172.18.0.3
    /supabase_realtime_<p>  id=5e135e1136eb  started=2026-07-31T01:16:47Z  ip=172.18.0.6
    /supabase_storage_<p>   id=d5c6a2947293  started=2026-07-31T01:16:46Z  ip=172.18.0.8

AFTER

    /supabase_auth_<p>      id=0f4d55029bc3  started=2026-07-31T01:19:38Z  ip=172.18.0.4
    /supabase_db_<p>        id=4d675b22d679  started=2026-07-31T01:19:22Z  ip=172.18.0.2
    /supabase_kong_<p>      id=c5ccc88fe58c  started=2026-07-31T01:17:23Z  ip=172.18.0.3
    /supabase_realtime_<p>  id=5e135e1136eb  started=2026-07-31T01:19:39Z  ip=172.18.0.6
    /supabase_storage_<p>   id=d5c6a2947293  started=2026-07-31T01:19:38Z  ip=172.18.0.8

db is recreated (container ID changes). auth, realtime and storage are restarted (same ID,
new StartedAt). kong is untouched — its StartedAt is older than the reset that just ran.

On a separate reset the addresses did move, with auth and storage exchanging 172.18.0.4 and
172.18.0.8, which is the state the Kong log in "Actual output" was captured from.

Why this bites tooling in particular
------------------------------------

Anything that talks to GoTrue immediately after a reset — a seed script creating a user via
POST /auth/v1/admin/users, for example — fails with a 502 that looks like a transient startup
race but is not, and will not clear no matter how long it waits.

Possible fixes
--------------

- Restart kong alongside the other containers in `db reset`, so it re-resolves its upstreams.
  Smallest change, and consistent with what the CLI already does for the services it restarts.
- Give the compose services stable addresses so a restart cannot move them.
- Configure Kong's DNS caching so a stale upstream record expires quickly rather than
  persisting for the life of the process.

Relationship to #4535
---------------------

https://github.com/supabase/cli/issues/4535 is closed and reports the same 502 symptom, but
attributes it to Docker Compose appending a folder-name suffix to container names. That
diagnosis does not appear to be correct: the Kong log quoted in that issue is also a stale
IP address, not an unresolved container name, which is the same mechanism described here.

This report reproduces the failure with entirely stable container names, on CLI 2.110.0 —
substantially newer than the 2.62.10 in that issue. The underlying defect therefore still
exists and that thread records the wrong cause.

Note also that `SUPABASE_DB_ONLY=true supabase db reset`, offered as a workaround in #4535,
does not help on 2.110.0: it still prints "Restarting containers..." and takes the same time,
so the containers still move.

7ttp:
## **TL;DR** 
 fixes `/auth/v1/*` (and any gateway route) returning 502 forever after `supabase db reset`, which was happening because the reset restarts auth/storage/
`realtime/pooler` but Kong's nginx keeps dialling their old cached container IPs for the life of its process. Now sorted by running an in-place `kong reload` after the satellite restarts (the same pattern `functions serve` uses), so the gateway re-resolves the moved containers while staying up the whole time 

skipped cleanly, when Kong is absent/stopped, loud failure instead of a silent broken gateway
otherwise. 

## ref: 
- closes https://github.com/supabase/cli/issues/6016

### DIFF
--- a/apps/cli-go/internal/db/reset/reset.go
+++ b/apps/cli-go/internal/db/reset/reset.go
@@ -1,6 +1,7 @@
 package reset
 
 import (
+	"bytes"
 	"context"
 	_ "embed"
 	"fmt"
@@ -287,16 +288,32 @@ func reloadKong(ctx context.Context) error {
 		// Kong may be excluded from the stack.
 		return nil
 	} else if err != nil {
-		return errors.Errorf("failed to inspect kong: %w", err)
+		return suggestKongRecovery(errors.Errorf("failed to inspect kong: %w", err))
 	}
 	if !resp.State.Running {
 		// A stopped gateway has no stale cache to flush.
 		return nil
 	}
-	if _, err := utils.DockerExecOnce(ctx, utils.KongId, nil, []string{"kong", "reload"}); err != nil {
-		return errors.Errorf("failed to reload kong: %w", err)
+	var out bytes.Buffer
+	if err := utils.DockerExecOnceWithStream(ctx, utils.KongId, "", nil, []string{"kong", "reload"}, &out, &out); err != nil {
+		if msg := strings.TrimSpace(out.String()); len(msg) > 0 {
+			return suggestKongRecovery(errors.Errorf("failed to reload kong: %w:\n%s", err, msg))
+		}
+		return suggestKongRecovery(errors.Errorf("failed to reload kong: %w", err))
 	}
 	return nil
+}
+
+// suggestKongRecovery decorates a gateway-left-unconfirmed failure with the
+// advisory next step; the caller-neutral wording also covers branch switch,
+// which shares RestartDatabase.
+func suggestKongRecovery(err error) error {
+	utils.CmdSuggestion = fmt.Sprintf(
+		"Local services restarted, but API routes may return 502 until the gateway reloads.\nTry restarting it with %s, and check %s if the failure persists.",
+		utils.Aqua("docker restart "+utils.KongId),
+		utils.Aqua("docker logs "+utils.KongId),
+	)
+	return err
 }
 
 func resetRemote(ctx context.Context, version string, config pgconn.Config, fsys afero.Fs, options ...func(*pgx.ConnConfig)) error {
--- a/apps/cli-go/internal/db/reset/reset.go
+++ b/apps/cli-go/internal/db/reset/reset.go
@@ -265,11 +265,38 @@ func restartServices(ctx context.Context) error {
 		return nil
 	})
 	// Do not wait for service healthy as those services may be excluded from starting
-	return errors.Join(result...)
+	if err := errors.Join(result...); err != nil {
+		return err
+	}
+	return reloadKong(ctx)
 }
 
 func listServicesToRestart() []string {
 	return []string{utils.StorageId, utils.GotrueId, utils.RealtimeId, utils.PoolerId}
+}
+
+// reloadKong reloads Kong after the restarts above so its nginx re-resolves each
+// upstream container's address. Kong caches resolved addresses for the life of a
+// worker process, and a restarted container can come back on a different one,
+// leaving the gateway returning 502 for that route until Kong restarts. An
+// in-place reload (the `functions serve` pattern) keeps the gateway serving
+// throughout. https://github.com/supabase/cli/issues/6016
+func reloadKong(ctx context.Context) error {
+	resp, err := utils.Docker.ContainerInspect(ctx, utils.KongId)
+	if errdefs.IsNotFound(err) {
+		// Kong may be excluded from the stack.
+		return nil
+	} else if err != nil {
+		return errors.Errorf("failed to inspect kong: %w", err)
+	}
+	if !resp.State.Running {
+		// A stopped gateway has no stale cache to flush.
+		return nil
+	}
+	if _, err := utils.DockerExecOnce(ctx, utils.KongId, nil, []string{"kong", "reload"}); err != nil {
+		return errors.Errorf("failed to reload kong: %w", err)
+	}
+	return nil
 }
 
 func resetRemote(ctx context.Context, version string, config pgconn.Config, fsys afero.Fs, options ...func(*pgx.ConnConfig)) error {
--- a/apps/cli-go/internal/db/reset/reset_test.go
+++ b/apps/cli-go/internal/db/reset/reset_test.go
@@ -73,11 +73,19 @@ func TestResetCommand(t *testing.T) {
 		utils.GotrueId = "test-auth"
 		utils.RealtimeId = "test-realtime"
 		utils.PoolerId = "test-pooler"
+		utils.KongId = "test-kong"
 		for _, container := range listServicesToRestart() {
 			gock.New(utils.Docker.DaemonHost()).
 				Post("/v" + utils.Docker.ClientVersion() + "/containers/" + container + "/restart").
 				Reply(http.StatusOK)
 		}
+		// Kong is not running so the reload is skipped (a successful exec attach is not gock-mockable, see TestExecOnce)
+		gock.New(utils.Docker.DaemonHost()).
+			Get("/v" + utils.Docker.ClientVersion() + "/containers/" + utils.KongId + "/json").
+			Reply(http.StatusOK).
+			JSON(container.InspectResponse{ContainerJSONBase: &container.ContainerJSONBase{
+				State: &container.State{Running: false},
+			}})
 		// Seeds storage
 		gock.New(utils.Docker.DaemonHost()).
 			Get("/v" + utils.Docker.ClientVersion() + "/containers/" + utils.StorageId + "/json").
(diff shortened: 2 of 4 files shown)


apps/cli-go/internal/db/reset/reset.go
```diff
@@ -265,13 +265,40 @@ func restartServices(ctx context.Context) error {
...
+		return nil
+	} else if err != nil {
+		return errors.Errorf("failed to inspect kong: %w", err)
+	}
+	if !resp.State.Running {
+		// A stopped gateway has no stale cache to flush.
+		return nil
+	}
+	if _, err := utils.DockerExecOnce(ctx, utils.KongId, nil, []string{"kong", "reload"}); err != nil {
+		return errors.Errorf("failed to reload kong: %w", err)
```
Coly010: Small thing I want to flag: this makes a kong reload failure hard-fail the whole `db reset`, even though the actual reset (recreate/migrate/seed) already succeeded by this point. `functions serve`'s version of this (`restartEdgeRuntime` in serve.go) just warns and carries on instead of failing. Was going fatal here intentional? If so that's fine, just want to make sure it's a deliberate call rather than drift from the pattern it's based on — a flaky `kong reload` exec would otherwise turn a good reset into a reported failure.

apps/cli-go/internal/db/reset/reset_test.go
```diff
@@ -360,6 +376,132 @@ func TestRestartDatabase(t *testing.T) {
 		assert.Empty(t, apitest.ListUnmatchedRequests())
 	})
 
+	t.Run("skips kong reload when kong is not running", func(t *testing.T) {
```
Coly010, line 379: One gap — none of these new tests actually cover kong reload succeeding (kong running + `kong reload` exits 0), which is the main case this PR is fixing. The comment says a successful exec attach isn't gock-mockable, but `functions/serve/serve_test.go` mocks exactly that (create/start/inspect exec endpoints, `ExitCode: 0`) for the same `kong reload` call — might be worth borrowing that pattern to get real coverage on the happy path here.

Coly010 (changes requested):
Nice fix for the 502-after-reset issue, love it. Left two comments below — nothing major, just want to make sure they're deliberate before this merges.

apps/cli-go/internal/db/reset/reset.go
7ttp: yeah, deliberate 

what we're fixing was `reset` reporting success over a broken gateway, so letting a failed reload slide would just recreate that one level up..?

IMO `reset` mostly runs in scripts where the `exit code` is the only thing checked, and the service restarts a few lines above are already fatal
so this follows suit

PS: *serve* can afford to warn since it's interactive and reloads again on the next file change anyway....

**re** flaky-exec case tho:
rerunning reset is safe/? &  
I'll reword the error so it's clear the database itself was `reset` fine..but kong failed..

apps/cli-go/internal/db/reset/reset_test.go
7ttp: so i tried exactly that pattern first, turns out it can't work:

 the exec attach is a tcp hijack gock can't intercept
 (docker_test.go has a "TODO: mock tcp hijack" for exactly this). 
serve_test's mocks look like they cover it but never have

 run `go test ./internal/functions/serve/... -run TestServeCommand -v`
 u'll see "Warning: failed to reload Kong: ... unable to upgrade to tcp"
 on every run

 the warn-and-continue swallows it, and gock only flags unmatched requests, not unused mocks...

so the happy path is covered the only way it can be rn: 
every mockable branch is tested (absent/stopped kong, inspect failure, reload failure), and the reload-success case was verified on a real docker stack, stale gateway healed 502→200 with kong's StartedAt unchanged....

happy to add a live-docker e2e for that heal if u want it in-tree :) 
(doesn't really benefit, we're planning to remove the go path anyway soon)

Coly010 (approved):

AlbinoGeek:
Man you rock. I reported this yesterday and you already solved it. @7ttp 
I would have, but I read the PR policy and saw I had to wait haha.

7ttp:
> Man you rock. I reported this yesterday and you already solved it. @7ttp I would have, but I read the PR policy and saw I had to wait haha.

haha :) thanks a ton for reporting! 💚
also yeah its currently available in our beta channel
https://github.com/supabase/cli/releases/tag/v2.111.1-beta.2

Please do let us know, if you find anything else

AlbinoGeek:
@7ttp Dumb question, how do I opt-in to the beta channel without having to manually install each version? Is there some documentation on that? (I don't really care the method, I use brew, dnf, uv, bunx, whatever.)

7ttp:
> @7ttp Dumb question, how do I opt-in to the beta channel without having to manually install each version? Is there some documentation on that? (I don't really care the method, I use brew, dnf, uv, bunx, whatever.)

totally not dumb at all :)

- npm/bunx (the reliable one rn): we publish a `beta` dist-tag — `bunx supabase@beta ...` 
resolves the newest beta on every run or maybe js do `npm i -g supabase@beta`
- brew: `brew install supabase/tap/supabase-beta` auto-tracks via `brew upgrade` 
- scoop on windows: `scoop install supabase-beta`, same idea

dnf's the one real gap we attach .rpms to each github release but there's no dnf repo to subscribe to, so nothing to auto-track, bunx is the way on fedora. 

`supabase --version` after installing confirms the channel ya took off :)

PS: it's not undocumented tho/btw the install sec of the repo README lists all three channel variants, 
thereforeee that's ur docs link 😄

PR #6048, opened 2026-08-03
fix(cli): skip unmounted workdir

### LINKED ISSUE #6035: supabase start fails on Podman: db container WorkingDir is set to the host project path (regression in 2.111.0)
### Affected area

Local development

### Supabase CLI version

2.111.0

### Operating system

Fedora 44 Kinoite (Aurora)

### Installation method

npm

### Command

```shell
DOCKER_HOST=unix:///run/user/$(id -u)/podman/podman.sock npx supabase@2.111.0 start
```

### Actual output

```shell
Applying migration 20260612004752_realtime_publication.sql...
Seeding data from supabase/seed.sql...
{"_tag":"Error","error":{"code":"UnknownError","message":"Error: workdir \"/home/<user>/<project>\" does not exist on container d2b69b79cbdd20194983938b340315..."}}
```

### Expected behavior

`supabase start` completes on Podman, as it does on 2.109.1.

Either set the container's `WorkingDir` to a path that exists in the image (2.109.1 used `/`), or mount
the project directory at the path being set as the working directory.

### Steps to reproduce

The behavioural difference reduces to one command, with no Supabase involved:

  $ podman run --rm --workdir /home/<user>/<project> alpine pwd
  Error: workdir "/home/<user>/<project>" does not exist on container 2084202...

  $ podman run --rm --workdir /tmp alpine pwd                     # control: path exists
  /tmp

  $ podman run --rm -v "$PWD:$PWD" --workdir "$PWD" alpine pwd    # control: mounted
  /home/<user>/<project>

Podman refuses a workdir that does not exist inside the container; Docker creates it.

Through the CLI:
  1. export DOCKER_HOST=unix:///run/user/$(id -u)/podman/podman.sock
  2. npx supabase@2.111.0 start   (in a project directory)
  3. Fails at "Seeding data from supabase/seed.sql..." — after all migrations applied.

Comparing the two containers on the SAME Podman:
  2.109.1  db WorkingDir = "/"                        -> exists in every image, works
  2.111.0  db WorkingDir = "/home/<user>/<project>"   -> not mounted, absent, fails

### Crash report ID

_No response_

```shell
Version of CLI:               2.111.0   (2.109.1 works; 2.110.0 also affected)
Version of container runtime: Podman 5.8.3, via the Docker-compatible socket
Operating system:             Fedora 44 (atomic / Universal Blue), kernel 7.0.11
SELinux:                      Enforcing
Node:                         v24.13.0
```

### Additional context

RULED OUT during diagnosis, so the search need not repeat it:

- Not the Postgres image. supabase/postgres:17.6.1.140 and 17.6.1.141 both behave
  identically; both report /etc/postgresql-custom writable under a plain `podman run`.
- Not a path or symlink quirk. `--workdir /definitely-not-here` fails the same way, and
  the real project path works through any name once mounted. Any non-existent path
  fails; any mounted path succeeds.
- Not SELinux. A separate SELinux issue precedes this one on 2.110/2.111 (below);
  relabelling clears it entirely, and then this workdir error appears on its own.

---

A SECOND, SEPARATE ISSUE IN THE SAME VERSIONS (SELinux + the bind-mounted secret).
Reported here because it is on the same upgrade path — a reader hitting one hits both.

2.111 generates the pgsodium key on the host and bind-mounts it read-only:

  bind  <project>/supabase/.temp/start-secrets/supabase_db_<project>/secret-0
     -> /etc/postgresql-custom/pgsodium_root.key   rw=false

On an SELinux-enforcing host that file is labelled user_home_t, which container_t
cannot stat. pgsodium_getkey.sh is:

  if [[ ! -f "${KEY_FILE}" ]]; then
      head -c 32 /dev/urandom | od -A n -t x1 | tr -d ' \n' > "${KEY_FILE}"
  fi
  cat $KEY_FILE

Because the file cannot be stat'd, `! -f` is true, the script tries to CREATE the key,
and the mount is read-only:

  /usr/lib/postgresql/bin/pgsodium_getkey.sh: line 8: /etc/postgresql-custom/pgsodium_root.key: Read-only file system
  od: write error
  ... database system is shut down
  LegacyHealthCheckTimeoutError: supabase_db_<project>: container is not ready: starting

Workaround: chcon -R -t container_file_t <project>/supabase before start.
Proper fix: add :Z (or :z) to that bind mount so Podman relabels it automatically —
this would fix it for every SELinux host, not just this one.

---

A THIRD, MINOR ONE. On 2.110/2.111, `supabase start` fails with

  LegacyStartVolumeCreateError: volume with name supabase_db_<project> already exists

when a volume left behind by 2.109.1's backup-preserving `stop` is present — it tries
to create rather than reuse it. `supabase stop --no-backup` clears it.

7ttp:
## TL;DR

Fixes `supabase start` failing on Podman with: `Error: workdir "/home/<user>/<project>" does not exist on container <id>`

 which was caused by the edge runtime container always being created with `--workdir <project root>` 
a path that only exists inside the container when a bind mounts something at or under it,
 so it's there for a project with functions and absent for one without. Docker quietly creates the missing directory,
Podman rejects the container outright. 
Now sorted by emitting `--workdir` only when a bind actually mounts that path, with tests around both directions...

No behaviour change for anyone already working: 
with ≥1 enabled function the flag is emitted exactly as before, on every runtime. A zero-function project simply stops getting an empty directory nothing ever read, the entrypoint is fully absolute (`--main-service=/root`)..

## refs

- closes supabase/cli#6035 
 the other 2/3 reported problems are already fixed: 
the SELinux relabel on the pgsodium secret bind by supabase/cli#6000 &
the volume already exists rejection by supabase/cli#6037

### DIFF
--- a/apps/cli/src/legacy/commands/functions/serve/serve.integration.test.ts
+++ b/apps/cli/src/legacy/commands/functions/serve/serve.integration.test.ts
@@ -14,6 +14,7 @@ import {
   mockLegacyTelemetryStateTracked,
   useLegacyTempWorkdir,
 } from "../../../../../tests/helpers/legacy-mocks.ts";
+import { toDockerPath } from "../../../../shared/functions/deploy.ts";
 import {
   mockOutput,
   mockProcessControl,
@@ -488,6 +489,9 @@ describe("legacy functions serve integration", () => {
             value.endsWith(":/root/index.ts:ro,Z"),
           ),
         ).toBe(true);
+        expect(extractFlagValues(dockerRun.args, "--workdir")).toEqual([
+          toDockerPath(tempRoot.current),
+        ]);
         expect(dockerRun.args[dockerRun.args.length - 1]).toBe(
           "edge-runtime start --main-service=/root --port=8081 --policy=per_worker\n",
         );
--- a/apps/cli/src/legacy/commands/start/services/edge-runtime.service.integration.test.ts
+++ b/apps/cli/src/legacy/commands/start/services/edge-runtime.service.integration.test.ts
@@ -176,25 +176,68 @@ describe("legacyStartEdgeRuntimeContainer", () => {
       }),
   );
 
-  it.effect(
-    "sets --workdir and --ulimit nofile=65536:65536, matching Go's WorkingDir/Ulimits container.Config",
-    () =>
-      Effect.gen(function* () {
-        const mock = mockDockerSpawner();
-        const out = mockOutput();
+  it.effect("sets --ulimit nofile=65536:65536, matching Go's Ulimits container.Config", () =>
+    Effect.gen(function* () {
+      const mock = mockDockerSpawner();
+      const out = mockOutput();
 
-        yield* legacyStartEdgeRuntimeContainer(baseInput(tempWorkdir.current)).pipe(
-          Effect.provideService(ChildProcessSpawner.ChildProcessSpawner, mock.spawner),
-          Effect.provide(out.layer),
-        );
+      yield* legacyStartEdgeRuntimeContainer(baseInput(tempWorkdir.current)).pipe(
+        Effect.provideService(ChildProcessSpawner.ChildProcessSpawner, mock.spawner),
+        Effect.provide(out.layer),
+      );
 
-        const runCall = mock.runCall!;
-        const workdirIndex = runCall.args.indexOf("--workdir");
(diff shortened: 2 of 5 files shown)


7ttp:
@codex review

Coly010 (approved):

PR #6049, opened 2026-08-03
test(cli): resolve e2e images via the production resolver

7ttp:
## TL;DR

Follow-up to the design note on #6030: `tests/helpers/docker-image.ts` hand-rolled ~130 lines of the candidate/cache-check/retry algorithm that already exists as `legacyMakeDockerImageResolver`, so the two could silently drift. The helper now
drives the production resolver through a real `ChildProcessSpawner`, and the one
piece that couldn't move, #6030's per-candidate budget split, which stops a stalled registry starving 
the ECR → GHCR → Docker Hub fallbacks, is ported into
the resolver as an opt-in `deadline`. One implementation of everything; nothing left to drift....

The deadline is inert in production: both callers pass one argument, every new path is gated on it, and the pre-existing resolver tests pass untouched — the only other production edit reverts #6030's export-widening of the retry
constant. The helper keeps only test policy (memoization, serialized resolves, a docker-CLI probe, a wedged-daemon backstop), with unit coverage it never had...

## Refs

- Follow-up to #6030
 <details>
<summary>basically fixes: (ss)</summary>

image

</details>

### DIFF
--- a/apps/cli/src/legacy/commands/start/lib/image-prepull.unit.test.ts
+++ b/apps/cli/src/legacy/commands/start/lib/image-prepull.unit.test.ts
@@ -125,7 +125,7 @@ describe("legacyEnsureImagesCached", () => {
     }),
   );
 
-  // Every pull attempt fails, so this drives the real LEGACY_DOCKER_PULL_RETRY_DELAYS_MS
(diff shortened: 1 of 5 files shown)


7ttp:
@codex review

Coly010 (approved):

PR #6060, opened 2026-08-04
fix(cli): deprecate db diff --use-pg-schema (CLI-1960)

Coly010:
## What changed

`db diff --use-pg-schema` wraps the **in-process Go library** `github.com/stripe/pg-schema-diff` (`apps/cli-go/internal/db/diff/pgschema.go`) — a direct in-process call (`sql.Open` + `pgschema.Generate`), no container image, no shell-out. There is no TS/WASM binding for this library. Per the M9 "Go removal" milestone's own pre-named decision, this is the **sanctioned keep-in-Go exception** — it is not being ported.

This PR:
- Adds a new TS-only stderr deprecation warning, printed before delegating to Go, additive to Go's own existing "experimental" warning (`cmd/db.go:121`). Fires only when `--use-pg-schema` is actually honored (mirrors Go's own warning gate — no notice on the explicit `--from/--to` path or when the flag has no effect).
- Updates `--help` text for the flag to note the deprecation.
- Updates `SIDE_EFFECTS.md` and `docs/go-cli-porting-status.md` to record the deprecation and the keep-in-Go rationale.
- No removal timeline is promised — removal date is explicitly out of scope for this issue (a follow-up decision via PostHog usage telemetry, per the issue).

## Why (keep-in-Go rationale)

Per `github.com/stripe/pg-schema-diff` being an in-process library with no container/binary boundary and no TS/WASM equivalent, porting isn't feasible within this milestone. Contrast with `--use-pgadmin`, which IS a plain container invocation and is tracked separately as a real port candidate (CLI-1968) — that one is NOT a keep-in-Go case. Full rationale recorded on the Linear issue and in `SIDE_EFFECTS.md`.

Docs are phrased conditionally ("will be the sole remaining Go delegation once `--use-pgadmin`, the `db __shadow`/`db __db-bootstrap` seams, and other in-flight M9 issues are done") rather than claiming present-tense exclusivity, since several other delegations are still live elsewhere in the milestone.

Fixes CLI-1960

## Review notes

Reviewed independently by go-parity-auditor, engineer-reviewer, and architect-reviewer. All three converged on two doc issues (an unresolvable tooling-path reference, and the present-tense "sole remaining Go delegation" overclaim), plus an unowned removal-timeline promise in the warning text — all fixed in the second commit. Behavioral change confirmed additive/stderr-only with the underlying Go delegation untouched.


Coly010:
@codex review

jgoux (approved):

PR #6061, opened 2026-08-04
fix(cli): remove db schema declarative __catalog seam for migrations mode (CLI-1959)

Coly010:
## What changed

Replaces the hidden Go seam `db schema declarative __catalog --mode migrations` — used by `db diff --from/--to migrations` and `db schema declarative sync`'s migrations-catalog diff source — with native TypeScript catalog export, invoking pg-delta directly via the already-shipped `legacyExportCatalogPgDelta` edge-runtime path instead of proxying through the Go binary.

`baseline`/`declarative` catalog modes remain seam-backed (out of scope here — they need native shadow-provisioning primitives that are CLI-1956's territory, a separate in-progress issue). `LegacyCatalogMode` is narrowed to `"baseline" | "declarative"`, making the remaining Go-backed surface exhaustive/compile-checked.

**Bundled parity fix:** the old code routed `db diff --from migrations` through the wrong Go-equivalent function — one shaped for `declarative.go`'s semantics (setup-token-folded cache key, zero-migrations→baseline reuse, a `Creating shadow database...` stderr line) — when Go's actual `db diff` path uses `explicit.go`'s `resolveMigrationsCatalogRef` (bare migration-hash cache key, no stderr line, always consults cache). This meant `db diff --from migrations` could never reuse a cache warmed by `db push`, and printed a spurious progress line. Both are fixed. The generated SQL was unaffected (a zero-migration catalog and the platform baseline are the same snapshot by Go's own logic) — this was a caching-efficiency and stderr-parity bug, not a correctness bug in the diff output.

## Why

Part of the M9 "Go removal" milestone. This is the last piece of `legacy-pgdelta.seam.layer.ts`'s `exportCatalog` surface that doesn't need shadow-provisioning primitives from CLI-1956.

## Review notes

Reviewed independently by go-parity-auditor, engineer-reviewer, and architect-reviewer (all approved after a follow-up commit). Findings addressed:
- Added test coverage for the previously-untested dominant real-world path (`sync` with local migrations present, cache hit/miss, `--no-cache`), including the `Creating shadow database...` stderr-parity assertions — all verified load-bearing via mutation checks.
- Extracted shared logic between the two new cache-key/path-resolution helpers and fixed a parameter-order inconsistency between sibling functions that could have caused a silent, permanent cache miss.
- Exported `LegacyBaselineTomlConfig` (was duplicated as an inline structural type) and corrected the module header, which had gone stale (the file now also owns shadow-provisioning orchestration, not just pure cache-key derivation).

Fixes CLI-1959


Coly010:
@codex review

jgoux (approved):

PR #6063, opened 2026-08-04
fix(cli): port db test to native TypeScript (CLI-1962)

Coly010:
## What changed

Ports `db test` (a hidden, Phase-0 Go-proxy command) and its `test db` alias to native TypeScript. Go registers the same `RunE` function under two cobra commands (`cmd/db.go:422-429` hidden, `cmd/test.go:17-20` visible) with byte-identical flags (`--db-url`, `--linked`, `--local` — default `true` on both). This PR mirrors that structure: hoisted the shared implementation (handler, errors, pg-prove-args, a new command-handler, and runtime layers) into `apps/cli/src/legacy/shared/legacy-test-db.*`, with both `db/test/test.command.ts` and `test/db/db.command.ts` reduced to thin registrations differing only in their `commandPath` (needed for Go-parity telemetry, since `cmd.CommandPath()` differs even though `RunE` is identical).

The underlying pgTAP-enable/pg_prove-container-invocation logic itself was already natively correct from a prior PR (#5522) — this PR is primarily a rewire (removing the `db test` proxy handler), not a re-port. Confirmed via go-parity-auditor: bind-mount/workingDir derivation, network-mode switching, Bitbucket `SecurityOpt` handling, Linux-only `ExtraHosts`, and exit-code propagation (`error running container: exit N`) all already matched Go.

This also resolves the audit's previously-flagged "`--local` default-true modelling caveat" — confirmed the resolved default is `true` on both entry points, matching Go exactly.

**Bonus, undisclosed-until-now parity fix:** `test db`'s help text previously read "Run pgTAP tests on the local or linked database." — Go's actual `Short` (shared via `dbTestCmd.Short`) is "Tests local database with pgTAP". Both entry points now byte-match.

## Why

Part of the M9 "Go removal" milestone. Removes the `LegacyGoProxy` dependency for `db test`/`test db` entirely.

## Review notes

Reviewed independently by go-parity-auditor, engineer-reviewer, and architect-reviewer — all approved after a follow-up commit. The engineer-reviewer caught (via mutation testing) that one new test didn't actually exercise the real command wiring and another assertion was vacuous; both are now fixed and mutation-verified. Also fixed: two stale Go line-number references in carried-over comments, a stale doc pointer in `docs/go-cli-porting-status.md`, and single-sourced the Short/description strings between the two command registrations to prevent future drift.

Fixes CLI-1962


Coly010:
@codex review

jgoux (approved):

PR #6066, opened 2026-08-04
fix(cli): relative declarative dir

7ttp:
## TL;DR 

fixes db schema declarative generate printing the absolute `declarative` dir in
its written-to line and its overwrite prompt which worked in GO cause it chdirs into
the workdir and prints `GetDeclarativeDir()` verbatim,

 diverged in TS cause the handler resolves the dir against the workdir for file I/O and printed that same resolved value
while the tests only asserted a message prefix:
 now sorted cause we switch both renders to the config-verbatim `legacyResolveDeclarativeDir` value through the shared written-line helper exactly like pull and sync already do and is tested locally across
the generate sync and pull suites plus the full workspace....

## ref: 
- extends #5968 
 ( CLI-1978 )


7ttp:
@codex review

jgoux (approved):

PR #6070, opened 2026-08-05
feat(stack): add lazy service lifecycle

jgoux:
Replacement stack 2 of 3, based on #6069.

Stack: #6069 → #6070 → #6071

Adds lazy startup around explicit ownership boundaries:

- gives `process-compose` sole ownership of desired service state and restart behavior
- declares eager startup, activation dependencies, and private companion ownership in one policy
- activates HTTP services at the existing proxy boundary while starting Realtime eagerly
- exposes `Dormant` as an explicit service state instead of parallel boolean bookkeeping
- lets the foreground stack and detached daemon reserve ports until their actual bind boundaries
- makes the daemon own allocation and atomically claim its live-state record

This deliberately omits the custom Realtime WebSocket bridge, managed port-lock protocol, and duplicate lifecycle intent sets from the previous stack.

Supersedes #6042
Supersedes #6043
Supersedes #6044
Supersedes #6045
Supersedes #6046

### DIFF
--- a/apps/cli/src/next/commands/branches/switch/switch.handler.ts
+++ b/apps/cli/src/next/commands/branches/switch/switch.handler.ts
@@ -1,10 +1,4 @@
-import {
-  StateManager,
-  daemonLayer,
-  resolveDaemonConfig,
-  resolveManagedStack,
-  stopDaemon,
-} from "@supabase/stack/effect";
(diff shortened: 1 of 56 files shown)


jgoux:
Consolidated into #6072 to simplify review and merging.

PR #6071, opened 2026-08-05
feat(cli): enable lazy local stack startup

jgoux:
Replacement stack 3 of 3, based on #6070.

Stack: #6069 → #6070 → #6071

Makes lazy startup the policy for CLI-managed stacks while keeping the package default eager:

- maps CLI start, branch switching, and Functions development through the shared lazy stack configuration
- treats a completed lazy startup phase as sufficient to display connection details
- keeps one gated live test for the real daemon, proxy, Docker, and first-request activation boundary
- documents that Realtime remains eager while HTTP services activate through the proxy

The separate test-only commit isolates the legacy SSO integration suite from host keyring state discovered during the full regression run; it does not affect runtime behavior.

Supersedes #6047

### DIFF
--- a/apps/cli/README.md
+++ b/apps/cli/README.md
@@ -123,7 +123,9 @@ Important areas:
 
 The local stack commands use `@supabase/stack` for lifecycle, daemon transport, status, and logs.
 That stack layer now has an explicit preparation phase, so foreground and detached `start` flows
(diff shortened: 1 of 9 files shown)


jgoux:
Consolidated into #6072 to simplify review and merging.

PR #6075, opened 2026-08-05
chore(cli-go): delete internal/start outright (unreachable) (CLI-1966)

Coly010:
## What

Deletes `internal/start` (Go's `supabase start` implementation, 1,509 LOC) outright, rather than only excluding it from the bundled/release `supabase-go` binary via a build tag (the approach this PR originally took).

## Why the revised approach

The original build-tag version kept `apps/cli-go/internal/start/` intact in-tree as "the migration's parity oracle," reasoning that the source might still be useful as a reference even though it doesn't ship. Re-examined under a stricter lens: keeping dead code around as a parity oracle only makes sense if it could ever again be read as a reference for porting something. It can't here — nothing in the TypeScript CLI will ever delegate to it, and nothing else Go-side that's still reachable calls into it either. That's not a parity oracle, it's dead weight, and this milestone's charter is removing Go, not archiving it.

Verified both reachability paths, not just "does the bundled binary need it":

1. **Direct TS→Go delegation** — grepped the TS codebase for anything that shells out to Go's `start` command. Confirmed (per the issue's own claim) that native TS `start` talks to Docker directly and never invokes Go's `start`. The one place a stale comment suggested otherwise (`legacy-pgdelta.seam.service.ts`'s `ensureLocalDatabaseStarted` doc, which said it "starts the stack via the bundled `supabase-go start`") was checked against the actual implementation and against Go's own `cmd/db_schema_declarative.go` — both actually use `supabase-go db start` → `internal/db/start`, a separate, still-live package. The comment was simply wrong; fixed in this PR.
2. **Indirect reachability** — checked every other TS→Go delegation seam still alive in this milestone (per `docs/go-cli-porting-status.md` and Linear CLI-1970's own "expected alive set": `db diff --use-pgadmin`/`--use-pg-schema`, `db pull --experimental`, the `db branch`/`db remote`/`gen keys` proxies, `db test`, the hidden `db __db-bootstrap`/`__shadow`/`__catalog` seams) against a repo-wide `grep -rn "github.com/supabase/cli/internal/start" apps/cli-go --include="*.go"`. The **only** hit anywhere in the module was `cmd/start_full.go`'s own import — now deleted along with it.
3. `cmd/start.go`'s tag-neutral cobra registration never imported `internal/start` — it only duplicated the `excludableContainers()` helper for the flag's help text, confirming the registration is genuinely just flags/cobra wiring.
4. `internal/start`'s own tests (`start_test.go`) only import broadly-shared test infra (`apitest`, `pgtest`, `pkg/config`, etc.) that's used elsewhere too, so deleting them orphans nothing.

## What changed

- Deleted `apps/cli-go/internal/start/` (source, tests, embedded templates) and the now-pointless `cmd/start_full.go`/`cmd/start_bundled.go` split. `cmd/start.go`'s `RunE` is now a single, permanent stub with the same behavior the bundled build's stub already had (same error text, same `--help`/`__complete` surface). `cmd/start_test.go` pins the stub's exact error text and the command/flag surface, since the CI job that used to verify this against a real binary is gone.
- `go mod tidy` drops the entire exclusive dependency tree (docker-compose/v2, buildx, moby/buildkit, k8s client-go + friends, aws-sdk-go-v2 + friends, notary, secret-detector) — each confirmed exclusive to `internal/start` via `go mod why` before removal. `docker/cli` (and its own transitive deps) stay, correctly, because `internal/utils` has its own independent, direct import of `docker/cli/cli/command`.
- Removed `cli-go-ci.yml`'s `Start` job entirely rather than rescoping it — rescoping only made sense when a "full" build with a real `internal/start` implementation still existed to run `./main start` against; once it's deleted, nothing in that job can function. Deleted the now-orphaned `apps/cli-go/e2e-test.sh` + `apps/cli-go/tests/` fixtures it alone used. Added back a lightweight `go list -deps` guard so a future accidental import of the heavy orchestration tree gets caught immediately rather than silently regrowing the binary.
- Removed the now-meaningless `-tags bundled` flag from `apps/cli/scripts/build.ts` and `tools/release/local-release.ts` — there is no remaining `//go:build bundled` constraint anywhere in the module, so both files are now byte-identical to pre-milestone `develop`.
- Fixed a real functional break the deletion surfaced: `shared/functions/serve-main-offline.e2e.test.ts` read `internal/start/templates/kong.yml` straight off disk at test runtime; repointed to the existing TS transcription (`LEGACY_START_KONG_YML_TEMPLATE`), verified byte-identical to the deleted Go template and passing.
- Repointed stale `apps/cli-go/internal/start/...` provenance comments across the TS tree to the last commit where the file existed (`a253ccba25c21356ccd33044c4474aecb77d1ae4`), and removed a `test.todo("parity: start")` in `apps/cli-e2e` that could never be fulfilled once no Go implementation was left to diff TS's behavior against.
- Updated `docs/binary-distribution.md`, `docs/go-cli-porting-status.md`, and `apps/cli/AGENTS.md` to describe `internal/start` as removed rather than excluded from the bundle.

## Measured size impact

Rebuilt the simplified (single-implementation, no tags) binary with the exact release flags for every target and compared against the numbers this PR previously measured for the tagged/bundled build:

| Target | Previous "after" (tagged) | This PR's build (deleted) | Delta |
|---|---|---|---|
| darwin/arm64 | 48,735,282 | 48,735,282 | 0 |
| darwin/amd64 | 51,880,688 | 51,880,688 | 0 |
| linux/arm64 | 47,448,226 | 47,448,226 | 0 |
| linux/amd64 | 50,852,002 | 50,852,002 | 0 |
| windows/amd64 | 51,985,408 | 51,985,408 | 0 |
| windows/arm64 | 47,868,928 | 47,869,440 | +512 |

Byte-for-byte identical (within noise from source-layout changes on windows/arm64). This is expected: deleting the package outright and excluding it via a build tag produce the same compiled bytes — the size win was already fully realized by the previous commit; this change is about deleting genuinely dead source, not further shrinking the binary. The full before/after-this-milestone comparison (97–103 MB → ~47–52 MB per platform, ~50% reduction) is unchanged from the original PR description.

## Review

Independently reviewed by go-parity-auditor, engineer-reviewer, and architect-reviewer against the full revised diff. All three confirmed the reachability analysis holds (re-derived independently, including rebuilding and diffing the binary's `--help`/`__complete`/`completion` output against the pre-deletion build). Convergent and individual findings were fixed: `cmd/start_test.go` now isolates the shared `utils.CmdSuggestion` global and asserts the cobra/flag surface; two stale mirror-workflow comments that referenced the now-removed `Start` job were updated; a stale "validation is kept" claim in `cmd/start.go` was corrected (validation moved to the TS port); write-only package globals in `cmd/start.go` were cleaned up; a docs claim about `apps/cli-go` being unconditionally authoritative was qualified for this one exception; and a new fast-tier unit test (`cli-go-path-references.unit.test.ts`) now guards every `new URL(".../cli-go/...")` literal in the TS tree against exactly the kind of silent breakage this PR found and fixed.

Linear: [CLI-1966](https://linear.app/supabase/issue/CLI-1966/cut-internalstart-from-the-bundled-go-build-42-mb-per-platform)


Coly010:
@codex review

Coly010:
@codex review

avallete (approved):

Coly010:
@codex review

Coly010:
@codex review

PR #6076, opened 2026-08-05
chore(cli): drop Go-deprecated db branch/remote and gen keys proxies (CLI-1964)

Coly010:
## What changed

Removes seven Phase-0 `LegacyGoProxy` wrapper commands from the TS legacy shell entirely, rather than porting or aliasing them:

- `db branch create|delete|list|switch`
- `db remote changes|commit`
- `gen keys`

This includes deleting each command's directory (`.command.ts`/`.handler.ts`/`SIDE_EFFECTS.md`), removing their registrations from `db.command.ts`/`gen.command.ts`, updating `docs/go-cli-porting-status.md`, fixing a stale comment in `start/lib/db-setup.ts` that referenced the deleted `db/branch/*` path, and adjusting one existing test (`shared/cli/hidden-flag.unit.test.ts`) whose assertions drove requests through the now-deleted routing. Added in-process regression coverage confirming the seven commands now surface as `UnknownSubcommand` (`hidden-flag.unit.test.ts` for `db branch`/`db remote`, a new `gen/gen.command.unit.test.ts` for `gen keys`).

Linear: [CLI-1964](https://linear.app/supabase/issue/CLI-1964/drop-go-deprecated-wrapped-commands-db-branch-db-remote-gen-keys)

## Why drop instead of port or alias

Go itself already retires all seven commands, so porting them would anchor Go code for a workflow Go itself is steering users away from:

- `db branch create|delete|list|switch` — Go's `dbBranchCmd` parent is `Hidden: true` and every leaf is `Deprecated`, pointing at `branches create|delete|list` (Management-API preview branches, already natively ported). These are genuinely different product surfaces — the Go `db branch` family clones/renames a **local** Postgres container, which the Management-API branch model has no equivalent for — so no native alias is possible either way.
- `db remote changes|commit` — Go's `dbRemoteCmd` parent is also `Hidden: true`. `changes` is deprecated in favor of `db diff --use-migra --linked` (already native; covers the same linked-diff use case, though Go's `db diff` never registered a `-p/--password` flag the way `db remote` did — a pre-existing Go asymmetry, not something introduced here). `commit` is deprecated in favor of `db pull` (already native, full flag parity including `--password`; the only difference is passing the migration name as a positional argument: `db pull remote_commit`).
- `gen keys` — Go's `genKeysCmd` is `Deprecated: use "gen signing-key" instead.` `gen signing-key` is already natively ported. Note this one has a real functional gap (HS256 `anon`/`service_role` keys + a derived DB password for a Fly-based preview-branch workflow, vs. `gen signing-key`'s unrelated asymmetric JWT signing key) — but the workflow it served (`<ref>-<branch>.fly.dev`) is decommissioned, so the gap is against dead infrastructure, not a live use case.

Note that all seven commands still run against the Go binary today — cobra prints a deprecation warning and then executes a `Deprecated` command rather than refusing it. This PR is a deliberate TS-side decision to not carry the proxy forward, not a claim that Go itself has already removed them. As a result, this is a user-facing behavior change for the legacy shell: `supabase db branch ...`, `supabase db remote ...`, and `supabase gen keys` now fail with an unknown-subcommand error instead of proxying through to Go.

Audit ref: `apps/cli/docs/go-parity-audit-2026-07-24.md` §4 (drop rather than port).


Coly010:
@codex review

Coly010:
Closing — dropping these Go-deprecated proxies (`db branch *`, `db remote *`, `gen keys`) entirely is a breaking change we're not ready to make right now. Reverting to keep-in-Go for these commands instead: the existing `LegacyGoProxy` delegation stays as-is. CLI-1964 is being cancelled; the retained-Go-delegation set is being consolidated into CLI-1970 instead.

PR #6131, opened 2026-08-09
feat(cli): skip vault sync

### LINKED ISSUE #3815: vault: do not push local secrets to remote db
Hi,

`supabase db push` copies local vault secrets to the remote database.
This is not ideal as my project uses different secrets in the local (dev) env than the remote (prod) env to separate API envs.
The CLI does not have a flag to omit updating vault secrets.

Is there a way around this? For now I'll have to manually revert my remote secrets after the push. It is tedious as the project has many secrets.

Another user faced the same issue https://github.com/supabase/cli/issues/3242#issuecomment-2961085948. I created a new issue as the problem is unique enough.

Thanks.

7ttp:
## TL;DR

Adds `--skip-vault` to `supabase db push`, allowing migrations to run without resolving or syncing configured Vault secrets

## What’s introduced?

`supabase db push --skip-vault` skips `[db.vault]` decryption and remote synchronization while preserving migrations, roles, seeds, target selection, and confirmation behavior. 
Existing behavior remains unchanged when the flag is omitted...

## Why was it needed?

Database pushes could overwrite environment-specific remote secrets with local values. 
They could also fail before connecting when Vault decryption keys were intentionally unavailable.

## Ref

- Closes https://github.com/supabase/cli/issues/3815
- Closes CLI-1800


avallete (commented):
Nitpick, but I think a "includeVault" flag, that can be set to false would be more consistent with the other flags we already have (includeSeed/includeRoles/includeAll/…)

Coly010 (approved):

avallete (approved):

PR #6157, opened 2026-08-11
chore(cli): generate docs spec from TS (CLI-2171)

7ttp:
## TL;DR

re-points CLI reference docs generation at the TypeScript command tree, replacing the retired Go generator with `scripts/generate-docs-spec.ts`.

## What's introduced?

- `scripts/generate-docs-spec.ts`, walking the legacy Effect command tree and merging the overlays and examples under `apps/cli/docs` into the reference spec
- docs-local introspection helpers and unit tests
- build time validation of the static tables, so command or flag renames fail loudly

Purely additive, no existing source files modified.

## Why was it needed?

- the reference was generated from the Go command tree, which the Go CLI deprecation deletes
- TS only additions never appeared on the docs site

The output was parity checked against the Go generator before its removal: all 135 command ids, anchors, flags and defaults match, and TS only surfaces like `completion` and `issue` now appear.

## ref:
- stacks on top of: https://github.com/supabase/cli/pull/6155 
- closes: CLI-2171

### DIFF
--- a/apps/cli/scripts/generate-docs-spec.ts
+++ b/apps/cli/scripts/generate-docs-spec.ts
@@ -1,0 +1,38 @@
+/**
+ * Emits the `clispec 001` CLI reference document for the supabase.com docs
+ * site (`supabase/supabase` `apps/docs/spec/cli_v1_commands.yaml`) from the
+ * legacy Effect command tree plus the content under `apps/cli/docs/`
+ * (`supabase/` description overlays, `templates/examples.yaml`).
+ *
+ * Contract (documented in `docs/README.md`): the spec YAML is the ONLY thing
+ * written to stdout, so `bun scripts/generate-docs-spec.ts > cli_v1_commands.yaml`
+ * yields a clean parseable file; diagnostics go to stderr. The spec version
+ * defaults to `latest` and can be overridden with an argument (a leading
+ * `v` is stripped) — the workspace package.json version is a semantic-release
+ * placeholder, so it is never used.
+ */
+import path from "node:path";
+import process from "node:process";
+import { legacyReadDocsContent } from "../src/legacy/docs/legacy-docs-spec.content.ts";
+import {
+  legacyBuildDocsSpec,
+  legacyStringifyDocsSpec,
+} from "../src/legacy/docs/legacy-docs-spec.ts";
+import { legacyRoot } from "../src/legacy/cli/root.ts";
+
+function resolveVersion(): string {
+  const argument = process.argv[2];
+  if (argument === undefined || argument === "") return "latest";
+  return argument.startsWith("v") ? argument.slice(1) : argument;
+}
+
+const content = legacyReadDocsContent(path.resolve(import.meta.dir, "../docs"));
+
+const spec = legacyBuildDocsSpec({
+  root: legacyRoot,
+  version: resolveVersion(),
+  overlays: content.overlays,
+  examples: content.examples,
+});
+
+process.stdout.write(legacyStringifyDocsSpec(spec));
--- a/apps/cli/src/legacy/docs/legacy-docs-spec.content.ts
+++ b/apps/cli/src/legacy/docs/legacy-docs-spec.content.ts
@@ -1,0 +1,88 @@
+import { readdirSync, readFileSync } from "node:fs";
+import path from "node:path";
+import { parse } from "yaml";
+import type { LegacyDocsExample } from "./legacy-docs-spec.ts";
+
+/**
+ * Loads the docs-spec content inputs from `apps/cli/docs/`: the description
+ * overlay markdown under `supabase/` (keyed by docs-relative POSIX path, the
+ * same keys `legacyDocsOverlayPath` produces on every platform) and the
+ * per-command examples from `templates/examples.yaml` (keyed by doc id).
+ * Shared by the generator (`scripts/generate-docs-spec.ts`) and the unit
+ * tests.
+ */
+export interface LegacyDocsContent {
+  readonly overlays: ReadonlyMap<string, string>;
+  readonly examples: Readonly<Record<string, ReadonlyArray<LegacyDocsExample>>>;
+}
+
+export function legacyReadDocsContent(docsDir: string): LegacyDocsContent {
+  const overlays = new Map<string, string>();
+  const walk = (dir: string): void => {
+    for (const entry of readdirSync(dir, { withFileTypes: true })) {
+      const entryPath = path.join(dir, entry.name);
+      if (entry.isDirectory()) walk(entryPath);
+      else if (entry.name.endsWith(".md")) {
+        const key = path.relative(docsDir, entryPath).split(path.sep).join("/");
+        overlays.set(key, readFileSync(entryPath, "utf8"));
+      }
+    }
+  };
+  walk(path.join(docsDir, "supabase"));
+
+  const examplesPath = path.join(docsDir, "templates/examples.yaml");
+  return { overlays, examples: legacyParseExamples(parse(readFileSync(examplesPath, "utf8"))) };
+}
+
+/**
+ * Narrows the parsed `examples.yaml` document to its expected shape — a
+ * mapping of doc id to example entries with optional string fields — failing
+ * with the offending doc id instead of letting a malformed file flow into
+ * the published spec.
+ */
+function legacyParseExamples(
+  parsed: unknown,
+): Readonly<Record<string, ReadonlyArray<LegacyDocsExample>>> {
+  if (typeof parsed !== "object" || parsed === null || Array.isArray(parsed)) {
+    throw new Error("legacy-docs-spec.content.ts: examples.yaml must be a mapping of doc ids.");
+  }
+  const examples: Record<string, ReadonlyArray<LegacyDocsExample>> = {};
+  for (const [docId, entries] of Object.entries(parsed)) {
+    if (!Array.isArray(entries)) {
+      throw new Error(
+        `legacy-docs-spec.content.ts: examples.yaml entry "${docId}" must be a list of examples.`,
+      );
+    }
+    examples[docId] = entries.map((entry, index) => {
+      if (typeof entry !== "object" || entry === null) {
+        throw new Error(
+          `legacy-docs-spec.content.ts: examples.yaml entry "${docId}"[${index}] must be a mapping.`,
+        );
+      }
+      const fields = new Map<string, unknown>(Object.entries(entry));
+      return {
+        ...legacyOptionalString(docId, index, fields, "id"),
+        ...legacyOptionalString(docId, index, fields, "name"),
+        ...legacyOptionalString(docId, index, fields, "code"),
+        ...legacyOptionalString(docId, index, fields, "response"),
+      };
+    });
+  }
+  return examples;
+}
+
+function legacyOptionalString(
+  docId: string,
+  index: number,
+  fields: ReadonlyMap<string, unknown>,
+  field: "id" | "name" | "code" | "response",
+): Partial<Record<"id" | "name" | "code" | "response", string>> {
+  if (!fields.has(field)) return {};
+  const value = fields.get(field);
+  if (typeof value !== "string") {
+    throw new Error(
+      `legacy-docs-spec.content.ts: examples.yaml "${docId}"[${index}].${field} must be a string.`,
+    );
+  }
+  return { [field]: value };
+}
--- a/apps/cli/src/legacy/docs/legacy-docs-spec.content.ts
+++ b/apps/cli/src/legacy/docs/legacy-docs-spec.content.ts
@@ -54,12 +54,19 @@ function legacyParseExamples(
       );
     }
     examples[docId] = entries.map((entry, index) => {
-      if (typeof entry !== "object" || entry === null) {
+      if (typeof entry !== "object" || entry === null || Array.isArray(entry)) {
         throw new Error(
           `legacy-docs-spec.content.ts: examples.yaml entry "${docId}"[${index}] must be a mapping.`,
         );
       }
       const fields = new Map<string, unknown>(Object.entries(entry));
+      for (const key of fields.keys()) {
+        if (key !== "id" && key !== "name" && key !== "code" && key !== "response") {
+          throw new Error(
+            `legacy-docs-spec.content.ts: examples.yaml "${docId}"[${index}] has unknown field "${key}" — allowed fields are id, name, code, response.`,
+          );
+        }
+      }
       return {
         ...legacyOptionalString(docId, index, fields, "id"),
         ...legacyOptionalString(docId, index, fields, "name"),
--- a/apps/cli/src/legacy/docs/legacy-docs-spec.tables.ts
+++ b/apps/cli/src/legacy/docs/legacy-docs-spec.tables.ts
@@ -232,6 +232,33 @@ export const LEGACY_DOCS_CHOICE_OVERRIDES: Readonly<Record<string, ReadonlyArray
 };
 
 /**
+ * Flags excluded from the published reference, keyed `"<doc id> <flag id>"`.
+ * Cobra hid deprecated flags from help and docs; the Effect port keeps
+ * `--include-raw-output` parse-visible for compatibility (see
+ * `commands/domains/SIDE_EFFECTS.md`), so the docs exclusion is restored
+ * here. Validated at build time like the other tables.
+ */
+export const LEGACY_DOCS_EXCLUDED_FLAGS: ReadonlySet<string> = new Set([
+  "supabase-domains-activate include-raw-output",
+  "supabase-domains-create include-raw-output",
+  "supabase-domains-delete include-raw-output",
+  "supabase-domains-get include-raw-output",
+  "supabase-domains-reverify include-raw-output",
+]);
+
+/**
+ * Usage argument overrides, keyed by doc id — for commands whose published
+ * argument rendering cannot be derived from the Effect tree. Both were
+ * hand-written cobra `Use` strings: the parser accepts zero occurrences
+ * (`secrets set --env-file` passes no positionals; `storage rm` validates in
+ * the handler), but the documented shape is required.
+ */
+export const LEGACY_DOCS_ARG_OVERRIDES: Readonly<Record<string, string>> = {
+  "supabase-secrets-set": "<NAME=VALUE> ...",
+  "supabase-storage-rm": "<file> ...",
+};
+
+/**
  * Flags rendered with a Required badge on the docs site, keyed
  * `"<doc id> <flag id>"`. Effect optionality cannot express this — several
  * of these TS flags are intentionally `Flag.optional` at parse time for
--- a/apps/cli/src/legacy/docs/legacy-docs-spec.tables.ts
+++ b/apps/cli/src/legacy/docs/legacy-docs-spec.tables.ts
@@ -1,0 +1,250 @@
+import type { LegacyDocsFlag } from "./legacy-docs-spec.ts";
+
+/**
+ * Static data for the docs spec generator — information the Effect command
+ * tree cannot express structurally:
+ *
+ * - Docs-site section tags: envelope compatibility only (the site's
+ *   navigation comes from `common-cli-sections.json`, not from tags).
+ * - Experimental leaves: each appends the root `--experimental` flag with
+ *   `required: true` to its flag list.
+ * - Deprecated commands: visible in `--help` (with a "(deprecated)" suffix)
+ *   but excluded from the published reference.
+ * - Flag defaults: `Flag.withDefault` captures the value in a closure
+ *   (`Param.ts`: `map(optional(self), Option.getOrElse(...))`), so defaults
+ *   that differ from the type's zero value are recorded here, keyed
+ *   `"<doc id> <flag id>"`.
+ *
+ * The values were frozen from the retired Go generator's output when the
+ * pipeline was re-pointed at the TS tree, and every entry was verified
+ * against it while both generators coexisted.
+ * Flag display types derive purely from the Effect tree — Go-only scalar
+ * typing (uint/duration/time) was deliberately not ported, so the reference
+ * shows the TS types `--help` shows.
+ * `legacyBuildDocsSpec` validates every key here against the walked tree and
+ * fails the build on stale entries.
+ */
+
+export interface LegacyDocsInfoTag {
+  readonly id: string;
+  readonly title: string;
+}
+
+export const LEGACY_DOCS_INFO_TAGS: ReadonlyArray<LegacyDocsInfoTag> = [
+  { id: "quick-start", title: "Quick Start" },
+  { id: "local-dev", title: "Local Development" },
+  { id: "management-api", title: "Management APIs" },
+  { id: "other-commands", title: "Additional Commands" },
+];
+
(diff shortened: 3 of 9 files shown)


7ttp:
@codex review

apps/cli/src/legacy/docs/legacy-docs-spec.tables.ts
```diff
@@ -0,0 +1,207 @@
...
+/**
+ * Flag defaults that differ from the primitive type's zero value, keyed
+ * `"<doc id> <flag id>"` (root global flags use doc id `supabase`). TS-only
+ * flags add entries by hand.
+ */
+export const LEGACY_DOCS_DEFAULT_OVERRIDES: Readonly<Record<string, string>> = {
+  "supabase agent": "auto",
+  "supabase dns-resolver": "native",
+  "supabase output": "pretty",
+  "supabase profile": "supabase",
```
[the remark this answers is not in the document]
7ttp: added the `supabase output-format` default ("text") to the overrides table and asserted it in the unit test — fixed in: https://github.com/7ttp/cli/commit/d6c12b362

apps/cli/scripts/generate-docs-spec.ts
```diff
@@ -0,0 +1,41 @@
...
+
+function resolveVersion(): string {
+  const argument = process.argv[2];
+  if (argument !== undefined && argument !== "") {
+    return argument.startsWith("v") ? argument.slice(1) : argument;
+  }
+  const packageJson = JSON.parse(
+    readFileSync(path.resolve(import.meta.dir, "../package.json"), "utf8"),
+  ) as { version: string };
+  return packageJson.version;
```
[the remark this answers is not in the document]
7ttp: dropped the package.json fallback entirely — it defaults to `latest` like the go generator did (the workspace version is a semantic-release placeholder, so it was never usable) — fixed in: https://github.com/7ttp/cli/commit/d6c12b362

apps/cli/src/legacy/docs/legacy-docs-spec.tables.ts
```diff
@@ -0,0 +1,207 @@
...
+  "supabase-telemetry": ["local-dev"],
+  "supabase-test": ["local-dev"],
+  "supabase-unlink": ["local-dev"],
+  "supabase-vanity-subdomains": ["management-api"],
+};
+
+/** Leaves that require `--experimental`. */
+export const LEGACY_DOCS_EXPERIMENTAL: ReadonlySet<string> = new Set([
+  "supabase-db-schema-declarative-generate",
+  "supabase-db-schema-declarative-sync",
```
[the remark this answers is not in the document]
7ttp: split them into LEGACY_DOCS_EXPERIMENTAL_OPTIONAL — the flag is still documented on both declarative commands but without the required badge, since `[experimental.pgdelta] enabled = true` also passes the gate — fixed in: https://github.com/7ttp/cli/commit/d6c12b362

apps/cli/src/legacy/docs/legacy-docs-spec.ts
```diff
@@ -0,0 +1,450 @@
...
+  const docId = commandPath.join("-");
+  const overrideKey = `${docId} ${single.name}`;
+
+  let acceptedValues: ReadonlyArray<LegacyDocsAcceptedValue> | undefined;
+  let enumType: string | undefined;
+  const choices = legacyChoiceKeysOf(single.primitiveType);
+  if (choices !== undefined) {
+    enumType = legacyDocsEnumTypeString(choices);
+    const type = enumType;
+    acceptedValues = choices.map((value) => ({ id: value, name: value, type }));
```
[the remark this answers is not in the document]
7ttp: added LEGACY_DOCS_CHOICE_OVERRIDES pinning root `-o` to env|pretty|json|toml|yaml (build-time validated like the other tables) — table/csv stay out of the global docs — fixed in: https://github.com/7ttp/cli/commit/d6c12b362

apps/cli/src/legacy/docs/legacy-docs-spec.ts
```diff
@@ -0,0 +1,479 @@
...
+}
+
+function legacyDocsFlagDoc(
+  commandPath: ReadonlyArray<string>,
+  param: Param.AnyFlag,
+): LegacyDocsFlag | undefined {
+  const unwrapped = legacyUnwrapParam(param);
+  if (unwrapped === undefined) return undefined;
+  const { single, isVariadic } = unwrapped;
+  if (single.hidden) return undefined;
```
Coly010, line 226: This filters `single.hidden` only, but the Go docs walk also dropped deprecated flags (cobra `MarkDeprecated` hides them — `apps/cli-go/cmd/domains.go:97`), so regenerating from this branch adds `--include-raw-output` ("(Deprecated) use -o json instead.") to all five `supabase-domains-*` pages; the published v2.98.2 spec omits it. The domains port already documents that Effect CLI can't reproduce cobra's help-hiding (`src/legacy/commands/domains/SIDE_EFFECTS.md`), but the docs generator can restore the exclusion cheaply — e.g. a build-validated `LEGACY_DOCS_EXCLUDED_FLAGS` table like the other overrides.

This also contradicts the header docstring's "flag sets … matched" verification claim, so worth re-running that diff once fixed.

apps/cli/src/legacy/docs/legacy-docs-spec.ts
```diff
@@ -0,0 +1,479 @@
...
+    ...(acceptedValues === undefined ? {} : { accepted_values: acceptedValues }),
+  };
+}
+
+function legacyDocsArgumentUsage(param: Param.AnyArgument): string | undefined {
+  const unwrapped = legacyUnwrapParam(param);
+  if (unwrapped === undefined) return undefined;
+  const { single, isOptional, isVariadic, variadicMin } = unwrapped;
+  if (single.hidden) return undefined;
+  const optional = isOptional || (isVariadic && variadicMin === 0);
```
Coly010, line 273: `isVariadic && variadicMin === 0` renders bare `Argument.variadic()` with optional brackets, so `supabase storage rm` publishes `[file] ...` where Go published `<file> ...`, and `supabase secrets set` publishes `[NAME=VALUE] ...` where Go published `<NAME=VALUE> ...`. Go's usage strings were hand-written `Use` fields, so the structural derivation can't recover them here.

Note a tree-level minimum isn't the right fix for `secrets set` — zero positionals are legitimately valid with `--env-file` (Go still documented `<NAME=VALUE>` regardless). A small docs-side usage override table (in the spirit of the other `LEGACY_DOCS_*` tables) would restore the published strings without touching parse behavior.

apps/cli/src/legacy/docs/legacy-docs-spec.ts
```diff
@@ -0,0 +1,479 @@
...
+      );
+      flags.push(...inherited);
+    }
+
+    let usage: string | undefined;
+    if (isLeaf) {
+      const argUsages = internals.config.arguments
+        .map(legacyDocsArgumentUsage)
+        .filter((part) => part !== undefined);
+      usage = [...commandPath, ...argUsages, "[flags]"].join(" ");
```
Coly010, line 429: Appending `"[flags]"` unconditionally deviates from cobra's `UseLine`, which added it only when the command had its own local (non-inherited) flags — diffed against published v2.98.2 this changes 52 of the 112 usage strings (`supabase backups list` → `supabase backups list [flags]`, `supabase seed buckets`, `supabase inspect db bloat`, …). Arguably the new form is more truthful (those commands do accept inherited flags), but it's an undisclosed deviation for a generator whose docstring claims verified output parity — either replicate the local-flags condition or add this to the "deliberately not reproduced" waiver list.

apps/cli/src/legacy/docs/legacy-docs-spec.ts
```diff
@@ -0,0 +1,479 @@
...
+
+    const flags: Array<LegacyDocsFlag> = [];
+    if (isLeaf) {
+      const own = [
+        ...flagDocsFor(commandPath, internals.config.flags),
+        ...(LEGACY_DOCS_EXTRA_FLAGS[docId] ?? []),
+      ].toSorted((a, b) => legacyDocsCompare(a.id, b.id));
+      own.forEach((flag) => seenFlagKeys.add(`${docId} ${flag.id}`));
+      flags.push(...own);
+      if (LEGACY_DOCS_EXPERIMENTAL.has(docId)) flags.push(experimentalFlag);
```
Coly010: Published flag *order* deviates on ~20 pages: flags that were inherited persistent flags in cobra (`project-ref`, `linked`, `local`) are own config flags in the Effect tree, so they sort into the own-flags block ahead of the injected `--experimental` — e.g. `supabase-vanity-subdomains-get` flips `[experimental, project-ref]` → `[project-ref, experimental]`, `supabase-storage-rm` `[recursive, experimental, linked, local]` → `[linked, local, recursive, experimental]`. The docs site renders the array in order, so these pages visibly reorder. Fine if intentional, but it isn't covered by the docstring's waiver list — worth either matching or waiving explicitly.

apps/cli/src/legacy/docs/legacy-docs-spec.ts
```diff
@@ -0,0 +1,479 @@
...
+  ): void => {
+    const docId = commandPath.join("-");
+    const children = legacyDocsVisibleChildren(command);
+    const internals = legacyCommandInternals(command);
+    const isLeaf = children.length === 0;
+    emittedIds.add(docId);
+
+    const overlay = input.overlays.get(legacyDocsOverlayPath(commandPath));
+    const description =
+      overlay === undefined ? (command.description ?? "") : legacyDocsStripOverlayHeading(overlay);
```
Coly010: Go normalized tabs before emitting (`strings.ReplaceAll(cmd.Long, "\t", "    ")` — `apps/cli-go/docs/main.go:183`); without that, the completion family's descriptions embed literal tabs (`\n\tsource <(supabase completion bash)`) where the published spec has four spaces. The tabs survive the release step's `prettier -w`, and tab-indented snippets may stop rendering as code blocks depending on the docs renderer. A `.replaceAll("\t", "    ")` here restores byte parity.

apps/cli/src/legacy/docs/legacy-docs-spec.ts
```diff
@@ -0,0 +1,479 @@
...
+  };
+}
+
+/**
+ * Serializes the spec for publication. Emitted under YAML 1.1 quoting rules
+ * so scalars like `yes`/`no` stay strings for downstream YAML 1.1 parsers
+ * (PyYAML, Psych, go-yaml v2); the output remains equally valid YAML 1.2.
+ */
+export function legacyStringifyDocsSpec(spec: LegacyDocsSpec): string {
+  return stringify(spec, { indent: 2, lineWidth: 0, version: "1.1" });
```
Coly010: `stringify` defaults to `aliasDuplicateObjects: true`, and since every experimental leaf pushes the same two shared flag objects, the emitted YAML contains anchors/aliases (`&a2` + 16 `*a2`, plus `&a1`/`*a1` for the optional variant) — the Go-generated file has none. Semantically equivalent for parsing consumers (the docstring's waiver rationale), but it makes the published file diff noisily against its predecessor and trips any textual tooling on the spec. `aliasDuplicateObjects: false` restores plain output.

apps/cli/src/legacy/docs/legacy-docs-spec.ts
```diff
@@ -0,0 +1,479 @@
...
+    commandPath: ReadonlyArray<string>,
+    inheritedParams: ReadonlyArray<Param.AnyFlag>,
+  ): void => {
+    const docId = commandPath.join("-");
+    const children = legacyDocsVisibleChildren(command);
+    const internals = legacyCommandInternals(command);
+    const isLeaf = children.length === 0;
+    emittedIds.add(docId);
+
+    const overlay = input.overlays.get(legacyDocsOverlayPath(commandPath));
```
Coly010: Unlike the static tables, the content inputs get no consumption guard: an `examples.yaml` doc id that matches no emitted command, or an overlay whose path maps to none, is silently ignored. After a command rename the tables throw, but the page's examples and long description just vanish (falling back to the terse tree description) with a green build. A post-walk check that every overlay path and examples key was consumed — allowing the 11 deliberately-orphaned deprecated `inspect-db` overlays, e.g. via `skippedExcludedIds` — would extend the build-fails-on-drift guarantee to the content files.

Coly010 (commented):
Reviewed the generator against the retired Go pipeline's actual output (regenerated the spec from this branch and diffed against the published `cli_v1_commands.yaml` v2.98.2). Inline comments fall into two groups:

- **Output changes today**: the deprecated `--include-raw-output` newly published on the domains pages, required variadic args rendered optional, ` [flags]` added to 52 usage strings, flag reordering on ~20 pages, YAML anchors, and literal tabs. The first two look like straight bugs; the rest may be acceptable, but they aren't covered by the docstring's "deliberately not reproduced" list, so worth deciding explicitly either way.
- **Drift guards**: the stale-key validation is one-directional (consolidated comment on the tables file), plus consumption checks for overlays/examples, an examples.yaml validation gap, and negative tests for the guards.

Checked and deliberately *not* flagging: the Bun-only `import.meta.dir` in `scripts/generate-docs-spec.ts` (everything runs the script via bun — `package.json` and `docs/README.md` both pin it), and the leaf-scoped `withGlobalFlags`/`withSharedFlags` gap in the walk (no leaf uses those combinators today).

apps/cli/src/legacy/docs/legacy-docs-spec.tables.ts
```diff
@@ -0,0 +1,250 @@
...
+  "supabase-inspect-db-total-table-sizes",
+  "supabase-inspect-db-unused-indexes",
+]);
+
+/**
+ * Flag defaults that differ from the primitive type's zero value, keyed
+ * `"<doc id> <flag id>"` (root global flags use doc id `supabase`). TS-only
+ * flags add entries by hand.
+ */
+export const LEGACY_DOCS_DEFAULT_OVERRIDES: Readonly<Record<string, string>> = {
```
Coly010, line 136: The stale-key guard makes renames fail the build, but validation is one-directional — nothing catches the reverse drift that the Go generator structurally prevented. In descending order of silence:

- **Defaults**: a changed `Flag.withDefault` (or a new flag with a non-zero default) publishes the old/zero value indefinitely — the override key still resolves. Go read pflag's live `DefValue`. This one has no runtime signal correcting the user.
- **Deprecations**: a newly deprecated command not added to `LEGACY_DOCS_EXCLUDED` stays published as a normal command (cobra auto-dropped anything with `Deprecated` set).
- **Experimental**: a new `legacyRequireExperimental` leaf not added to `LEGACY_DOCS_EXPERIMENTAL` is published without `--experimental` (Go injected it from the same slice the runtime gate read).
- **Required**: a newly required flag not in `LEGACY_DOCS_REQUIRED` gets no badge (Go read the live `MarkFlagRequired` annotation).

`LEGACY_DOCS_TAGS` got a forward guard (the untagged-top-level throw); the other tables didn't. Some of these can't be automated cheaply (defaults aren't recoverable from Effect params; the gate lives in handlers) — so at minimum a "when adding a command/flag, touch these tables" checklist in `docs/README.md` would make the duties discoverable. A unit test asserting `LEGACY_DOCS_EXPERIMENTAL` against the `legacyRequireExperimental` call sites is also cheap, if a bit hacky.

apps/cli/src/legacy/docs/legacy-docs-spec.content.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+  const examples: Record<string, ReadonlyArray<LegacyDocsExample>> = {};
+  for (const [docId, entries] of Object.entries(parsed)) {
+    if (!Array.isArray(entries)) {
+      throw new Error(
+        `legacy-docs-spec.content.ts: examples.yaml entry "${docId}" must be a list of examples.`,
+      );
+    }
+    examples[docId] = entries.map((entry, index) => {
+      if (typeof entry !== "object" || entry === null) {
```
Coly010: This mapping check doesn't reject arrays (unlike the top-level check above), so a mis-indented entry like `- - id: oops` passes as `[{}]`; and since unknown fields are dropped without error, a typo'd key (`titel:`) also collapses to `{}`. Both publish an empty example block instead of failing with the offending doc id, which is what this function's contract promises. Adding `Array.isArray(entry)` here and rejecting unknown fields would close both paths.

apps/cli/src/legacy/docs/legacy-docs-spec.unit.test.ts
```diff
@@ -0,0 +1,230 @@
...
+  });
+
+  it("keeps content that does not start with a heading", () => {
+    expect(legacyDocsStripOverlayHeading("Creates a resource.\nMore.\n")).toBe(
+      "Creates a resource.\nMore.\n",
+    );
+  });
+});
+
+describe("legacyBuildDocsSpec", () => {
```
Coly010, line 81: All of the guard paths this design leans on are only exercised in their no-op path: nothing tests `legacyValidateDocsTables`' stale-key throw, the untagged-top-level-command throw, the missing `--experimental` root-flag throw, or `legacyParseExamples`' malformed-shape rejections. A regression in any of them stays green while silently un-guarding the frozen tables. A handful of negative tests with a synthetic tree / inline fixtures would pin them.

apps/cli/src/legacy/docs/legacy-docs-spec.ts
7ttp: fixed in: https://github.com/supabase/cli/pull/6157/commits/dd532e50cce8d8df7f5ea4ba8b04636301b33183

apps/cli/src/legacy/docs/legacy-docs-spec.ts
7ttp: fixed in: https://github.com/supabase/cli/pull/6157/commits/dd532e50cce8d8df7f5ea4ba8b04636301b33183

apps/cli/src/legacy/docs/legacy-docs-spec.ts
7ttp: fixed in: https://github.com/supabase/cli/pull/6157/commits/dd532e50cce8d8df7f5ea4ba8b04636301b33183

apps/cli/src/legacy/docs/legacy-docs-spec.ts
7ttp: fixed in: https://github.com/supabase/cli/pull/6157/commits/dd532e50cce8d8df7f5ea4ba8b04636301b33183

apps/cli/src/legacy/docs/legacy-docs-spec.ts
7ttp: fixed in: https://github.com/supabase/cli/pull/6157/commits/dd532e50cce8d8df7f5ea4ba8b04636301b33183

apps/cli/src/legacy/docs/legacy-docs-spec.ts
7ttp: fixed in: https://github.com/supabase/cli/pull/6157/commits/dd532e50cce8d8df7f5ea4ba8b04636301b33183

apps/cli/src/legacy/docs/legacy-docs-spec.ts
7ttp: fixed in: https://github.com/supabase/cli/pull/6157/commits/dd532e50cce8d8df7f5ea4ba8b04636301b33183

apps/cli/src/legacy/docs/legacy-docs-spec.tables.ts
7ttp: fixed in: https://github.com/supabase/cli/pull/6157/commits/dd532e50cce8d8df7f5ea4ba8b04636301b33183

apps/cli/src/legacy/docs/legacy-docs-spec.content.ts
7ttp: fixed in: https://github.com/supabase/cli/pull/6157/commits/dd532e50cce8d8df7f5ea4ba8b04636301b33183

apps/cli/src/legacy/docs/legacy-docs-spec.unit.test.ts
7ttp: fixed in: https://github.com/supabase/cli/pull/6157/commits/dd532e50cce8d8df7f5ea4ba8b04636301b33183

7ttp:
(will patch the studio next, once this lands in stable release)

avallete (commented):
## Review

**Overall assessment:** the implementation is sound. I checked out `dd532e5`, ran the unit suite (32/32 pass), ran the generator end-to-end (clean parseable YAML, 144 commands = the 135 Go-parity ids plus the TS-only surfaces), and ran the workspace checks (`types`/`lint`/`fmt`/`knip` all green). I also read the retired Go generator out of git history and confirmed the semantics this port claims to mirror: `info.options` was never populated in Go (correctly omitted here), the root flag walk only ever init'd the help flag (so no `--version` in root flags is correct), and `--experimental` required-ness was injected per-leaf, not on the root list — all matched.

**Codex + reviewer feedback:** all 13 threads are genuinely fixed in the current head, not just marked resolved. Verified live in the emitted spec: `output-format` defaults to `text`; version defaults to `latest` with the `v` prefix stripped (package.json fallback gone); the two declarative-schema commands document `--experimental` without the required badge (`LEGACY_DOCS_EXPERIMENTAL_OPTIONAL`); root `-o` is pinned to `env|pretty|json|toml|yaml` while `db query` gets its own `json|table|csv` flag via `LEGACY_DOCS_EXTRA_FLAGS`; `--include-raw-output` is absent from all five domains pages; `storage rm` / `secrets set` render `<file> ...` / `<NAME=VALUE> ...`; no tabs and no YAML anchors in the output; orphaned overlays/examples throw; the guard paths have negative tests; and the `LEGACY_DOCS_EXPERIMENTAL`-mirrors-runtime-gate test plus the README maintenance checklist cover the reverse-drift concern. The `[flags]` suffix and flag-ordering deviations were waived explicitly in the header docstring rather than replicated, which the review thread offered as an acceptable resolution.

### Findings

**NON-BLOCKING OBSERVATION — `LEGACY_DOCS_EXTRA_FLAGS` validation is weaker than the other tables.** Its keys are validated against `emittedIds`, which includes group commands, but injection only happens on leaves (`legacy-docs-spec.ts:378`). An entry mistakenly keyed to a group id (e.g. `supabase-db` instead of `supabase-db-query`) passes validation and silently never renders — the exact silent-degradation mode the tables were designed to prevent, and one `LEGACY_DOCS_ARG_OVERRIDES` already guards against via `consumedArgOverrideIds`. Smallest fix: track consumption the same way. Low probability today (the table has one entry), fine as a follow-up.

**NON-BLOCKING OBSERVATION — key order of the injected `--experimental` flag.** `{ ...experimental, required: true }` (`legacy-docs-spec.ts:302`) appends `required` after `default_value`, whereas every other required flag (and the Go `FlagDoc` field order) emits `required` before `default_value`. Purely cosmetic in parsed-YAML terms; only worth touching if you want the published file to diff minimally against its predecessor on the 17 experimental leaves.

### Verdict

**Approve with non-blocking comments.**

---

avallete (approved):

PR #6163, opened 2026-08-12
chore(stack): resolve deferred managed-layer hygiene (CLI-2174)

jgoux:
Resolves the deferred hygiene items from the CLI-2106 review cycle (CLI-2174). Stacked on #6152.

## Items 2–5 (original hygiene set)

- **Shared atomic claim primitive** (items 2+3): the duplicated temp-write → `link()` → `EEXIST` protocol in `managed/identity.ts` and `StateManager.ts` is extracted into `managed/atomic-claim.ts` as a plain-async primitive; Effect callers wrap it. Each call site keeps its own race semantics (identity re-reads the winning marker, StateManager fails with `StateClaimError`).
- **Hardlink-less filesystem fallback** (item 3): when `link()` refuses with `EPERM`/`ENOTSUP` (exFAT, FAT32, some network mounts), the claim falls back to an exclusive `writeFile(..., { flag: "wx" })`, which still settles the race. The link-first rationale and the harmlessness of a SIGKILL-stranded `.tmp.<id>` sibling are documented on the primitive.
- **Transaction reentrancy guard** (item 4): `runTransaction` in `managed/sqlite.ts` now refuses a nested call before issuing `BEGIN` — and without issuing `ROLLBACK` — so an inner attempt can never discard the outer transaction's writes. Tracked per handle via a `WeakSet`.
- **Coverage-guard comment parsing** (item 5): `error-actionability-coverage.unit.test.ts` strips comments (string-literal-aware) before scanning for error-class definitions, so a comment containing `class X extends Error` no longer trips the guard.

## Items 7–12 (Effect-refactor verification follow-ups)

- **Explicit closed flag** (item 7): the facade's post-close guard branches on a `closed` flag set by `close()`/`asyncDispose` instead of sniffing "disposed" out of rejection text, so a callback's own rejection is reported as itself.
- **Original failure preserved through claim release** (item 8): a new `releasingClaimOnFailure` combinator on the catch paths of `updateStackRecord`/`deleteStack` discards anything the release raises (interruption included) and re-raises the original cause; the interrupt re-raise stays scoped to the recording site.
- **Publication-poll runtime guard** (item 9): the poll's final answer is checked at runtime; a bounded schedule ending on `None` dies with a clear message instead of silently yielding `undefined`.
- **No fabricated report entries from interrupts** (item 10): the liveness-probe, runtime-inspection, and state-reclamation absorption points re-raise interrupt-only causes instead of recording them.
- **Claim-before-mask contract gap documented** (item 11): the provision mask now carries a comment pinning the constraint (sync adapters have no suspension point; the mask must extend over row creation before async repositories become real), cross-referencing the same shape in `deleteStack`.
- **`migrateSchema` reuses `runTransaction`** (item 12): migration no longer hand-rolls `BEGIN`/`COMMIT`; it runs on a freshly opened handle before any repository transaction exists, so the new reentrancy guard is not tripped.

## Items resolved without code

- **Item 1 (synchronous cold-start WAL retry)**: already fixed on the CLI-2106 branch — WAL conversion now retries via an Effect `Schedule.exponential` instead of a blocking `Atomics.wait`.
- **Item 6 (`credentials.*` fixture ownership)**: no existing issue owns credential-reference resolution; proposed CLI-2114 (first CLI slice wiring the managed layer) as the owner on Linear.

### DIFF
--- a/apps/cli/src/shared/telemetry/error-actionability-coverage.unit.test.ts
+++ b/apps/cli/src/shared/telemetry/error-actionability-coverage.unit.test.ts
@@ -40,6 +40,69 @@ import {
 const ERROR_DEFINITION_PATTERN =
   /TaggedError\(\s*"([A-Za-z0-9_]+)"|class\s+[A-Za-z0-9_]+\s+extends\s+[A-Za-z0-9_.]*Error\(\s*"([A-Za-z0-9_]+)"|class\s+([A-Za-z0-9_]+)\s+extends\s+Error\b/gs;
 
+// Removes `//` line comments and `/* */` block comments from a source string
+// so the regex-based scan below never mistakes a comment merely mentioning
+// `class X extends Error` (or a TaggedError example) for a real definition.
+// String and template literals are treated as opaque runs — a `//` or `/*`
+// inside `"https://..."` (or a backtick template) must survive untouched.
+// Stripped comment bytes are replaced with spaces (newlines are preserved)
+// so line numbers and any line-based logic elsewhere stay unaffected. This
+// is a compact scanner, not a full tokenizer: it does not special-case regex
+// literals or `${...}` interpolation inside template literals.
+function stripComments(source: string): string {
+  let out = "";
+  let i = 0;
+  const n = source.length;
+  while (i < n) {
+    const c = source[i];
+    const next = source[i + 1];
+
+    if (c === '"' || c === "'" || c === "`") {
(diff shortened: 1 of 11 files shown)


jgoux:
Superseded by #6166 — GitHub closed this PR when the head branch was renamed to the Linear-prescribed name (`julien/cli-2174-...`). Same commits, same review state.

PR #6166, opened 2026-08-12
chore(stack): resolve deferred managed-layer hygiene (CLI-2174)

jgoux:
Resolves the deferred hygiene items from the CLI-2106 review cycle (CLI-2174). Follow-up to #6152 (originally stacked on it; rebased onto `develop` after its squash-merge). Supersedes #6163, which GitHub closed during a head-branch rename to the Linear-prescribed name.

## Items 2–5 (original hygiene set)

- **Shared atomic claim primitive** (items 2+3): the duplicated temp-write → `link()` → `EEXIST` protocol in `managed/identity.ts` and `StateManager.ts` is extracted into `managed/atomic-claim.ts` as a plain-async primitive; Effect callers wrap it. Each call site keeps its own race semantics (identity re-reads the winning marker, StateManager fails with `StateClaimError`).
- **Hardlink-less filesystem fallback** (item 3): when `link()` refuses with `EPERM`/`ENOTSUP` (exFAT, FAT32, some network mounts), the claim falls back to an exclusive `writeFile(..., { flag: "wx" })`, which still settles the race. The link-first rationale and the harmlessness of a SIGKILL-stranded `.tmp.<id>` sibling are documented on the primitive.
- **Transaction reentrancy guard** (item 4): `runTransaction` in `managed/sqlite.ts` now refuses a nested call before issuing `BEGIN` — and without issuing `ROLLBACK` — so an inner attempt can never discard the outer transaction's writes. Tracked per handle via a `WeakSet`.
- **Coverage-guard comment parsing** (item 5): `error-actionability-coverage.unit.test.ts` strips comments and masks string/template-literal contents (scanning interpolation expressions as code) before looking for error-class definitions, so commented or quoted `class X extends Error` text no longer trips the guard.

## Items 7–12 (Effect-refactor verification follow-ups)

- **Explicit closed flag** (item 7): the facade's post-close guard branches on a `closed` flag set by `close()`/`asyncDispose` instead of sniffing "disposed" out of rejection text, so a callback's own rejection is reported as itself.
- **Original failure preserved through claim release** (item 8): a new `releasingClaimOnFailure` combinator on the catch paths of `updateStackRecord`/`deleteStack` discards anything the release raises (interruption included) and re-raises the original cause; the interrupt re-raise stays scoped to the recording site.
- **Publication-poll runtime guard** (item 9): the poll's final answer is checked at runtime; a bounded schedule ending on `None` dies with a clear message instead of silently yielding `undefined`.
- **No fabricated report entries from interrupts** (item 10): the liveness-probe, runtime-inspection, and state-reclamation absorption points re-raise interrupt-only causes instead of recording them.
- **Claim-before-mask contract gap documented** (item 11): the provision mask now carries a comment pinning the constraint (sync adapters have no suspension point; the mask must extend over row creation before async repositories become real), cross-referencing the same shape in `deleteStack`.
- **`migrateSchema` reuses `runTransaction`** (item 12): migration no longer hand-rolls `BEGIN`/`COMMIT`; it runs on a freshly opened handle before any repository transaction exists, so the new reentrancy guard is not tripped.

## Items resolved without code

- **Item 1 (synchronous cold-start WAL retry)**: already fixed on the CLI-2106 branch — WAL conversion now retries via an Effect `Schedule.exponential` instead of a blocking `Atomics.wait`.
- **Item 6 (`credentials.*` fixture ownership)**: no existing issue owns credential-reference resolution; proposed CLI-2114 (first CLI slice wiring the managed layer) as the owner on Linear.


jgoux:
Addressed the external review finding (P3, literal-content misclassification): `stripComments` now masks string/template-literal text before the definition pattern runs, and `${...}` interpolations are re-scanned as code (comments stripped, nested literals masked, brace depth tracked). Both reported repros — `"class Fake extends Error"` in a string and a fake definition inside an interpolation comment — now extract nothing, with fail-before regression tests.

One nuance surfaced during the fix: the definition pattern captures tags from inside the double-quoted argument of `TaggedError("Tag")`-style calls, so that one string shape keeps its content (bounded `Error(` lookbehind); masking it too silently dropped the real coverage surface from >50 files to 6. Full guard suite: 136 tests green, real-codebase coverage assertions intact.

apps/cli/src/shared/telemetry/error-actionability-coverage.unit.test.ts
```diff
@@ -40,6 +40,188 @@ import {
...
+// length if the interpolation was left unterminated).
+function scanCode(
+  source: string,
+  start: number,
+  stopAtInterpolationClose: boolean,
+): { out: string; end: number } {
+  let out = "";
+  let i = start;
+  const n = source.length;
+  let braceDepth = 0;
```
avallete: **DX / MAINTAINABILITY** — the regex-literal gap the doc comment admits to fails in the *silent* direction for a guard test.

A regex literal whose body contains an escaped slash immediately followed by `*` starts a fake block comment here. `/\/*$/` — chars `/ \ / * $ /` — hits `c === "/" && next === "*"` at index 2, so everything from there to the file's next `*/` is blanked. Trailing-slash trimming (`path.replace(/\/*$/, "")`, `/\/*/g`) is an ordinary thing to write, and the blanked run typically spans to the next JSDoc terminator, i.e. potentially many declarations.

The consequence is that error classes stop being scanned and the actionability guard quietly passes for them, rather than failing. `expect(tagsByFile.size).toBeGreaterThan(50)` doesn't catch it, because the file still contributes its earlier tags. That's the opposite failure mode from the one this change fixes: a comment false positive breaks the test loudly and gets diagnosed in a minute.

I verified this is a latent gap, not a live one: extracting tags from every file under the five scanned roots with and without `stripComments` yields identical sets today (0 lost, 0 gained), so nothing regresses now.

Two ways to close it, smallest first:

1. Skip regex literals as opaque runs, using the standard "can a regex start here" heuristic on the previous significant character. The scanner already has the machinery for opaque runs; this is ~10 lines on top.
2. Drop the scanner and anchor the pattern to a declaration position instead, since the only false positive being fixed is a comment *mentioning* a definition:

```ts
const ERROR_DEFINITION_PATTERN =
  /^[ \t]*(?:export\s+)?(?:abstract\s+)?(?:declare\s+)?class\s+[A-Za-z0-9_]+\s+extends\s+[A-Za-z0-9_.]*(?:TaggedError|Error)\(\s*"([A-Za-z0-9_]+)"|^[ \t]*(?:export\s+)?(?:abstract\s+)?(?:declare\s+)?class\s+([A-Za-z0-9_]+)\s+extends\s+Error\b/gm;
```

I ran this against all five roots too: identical tag sets to the current pattern, and it rejects all six cases the new `stripComments` tests cover (comment lines start with `//`, `*`, or `/*`, so the line anchor never matches them) — for ~50 fewer lines and no tokenizer to maintain.

The honest trade-off: (2) has its own hole the scanner doesn't — a block-comment body line with no `*` prefix (`/*\nclass Fake extends Error {}\n*/`) would match. But that hole fails loudly, which is the right direction for a guard. Your call which shape you prefer; the thing I'd rather not leave as-is is the silent-loss direction.

---

packages/stack/src/managed/atomic-claim.ts
```diff
@@ -0,0 +1,78 @@
...
+ * rather than a claim anybody can observe.
+ */
+export const claimFileAtomically = async (
+  targetPath: string,
+  content: string,
+  options: FileClaimOptions = {},
+): Promise<FileClaimOutcome> => {
+  const linkFile = options.linkFile ?? link;
+  const temporaryPath = `${targetPath}.tmp.${options.temporaryId ?? randomUUID()}`;
+  await writeFile(temporaryPath, content, { flag: "wx", mode: options.mode });
```
avallete: **NON-BLOCKING OBSERVATION (theoretical)** — `flag: "wx"` on the *temporary* write is new for `identity.ts`, and it slightly undercuts the "stranded temp is harmless junk" claim in the doc block above.

The pre-PR identity path wrote the temp file with `{ mode: 0o600 }` and no exclusive flag; `StateManager` used `wx`. Unifying on `wx` is the stricter choice, which is fine, but it means a temp path that already exists now makes the claim throw raw `EEXIST` from the temp write rather than from `link`. That surfaces as a defect through `failsWithIdentity`, and as `reason: "io-error"` (not `"already-claimed"`) through `StateManager` — so the stranded sibling *is* observable after all, as a hard failure rather than junk.

Realistically this needs a repeating `temporaryId` — a SIGKILL between the temp write and its `unlink` plus a re-run with an injected factory that hands back the same value, or two concurrent claimants sharing one. Production uses `randomUUID`, and the two deterministic factories in the stack tests are a monotonic counter and a constant that fails UUID assertion before reaching this line, so nothing hits it today. Flagging it only because the option's doc explicitly invites deterministic ids and the primitive's doc asserts the stranded file is unobservable.

Smallest fix if you want the doc claim to hold literally: drop the exclusive flag on the temporary write (the path is unique by construction, and `link` is what settles the race — `wx` here buys nothing), or keep `wx` and reword the paragraph to say a stranded sibling blocks a re-run that reuses the same id.

---

avallete (commented):
## Overall

The implementation looks sound. Each of the twelve items maps to a change that actually does what the description says, the two extractions (`atomic-claim.ts`, `runTransaction` reuse in `migrateSchema`) remove real duplication rather than adding an abstraction layer, and the new tests exercise the behaviour rather than the plumbing. No blocking or significant issues found.

Codex didn't produce findings on this PR or on #6163 — both runs hit its usage limit — so this is a from-scratch read.

What I checked beyond reading the diff:

- `packages/stack`: `pnpm check:all` green (types, lint, fmt, knip); `--project integration` 212/212 green, including the four new cases.
- `apps/cli`: the coverage-guard test 132/132 green.
- The reentrancy guard's interaction with error classification: every `transaction(...)` call site passes `failsWith(...)` or `neverFails`, both of which rethrow an unmatched error, so the guard's plain `Error` lands as a defect rather than being laundered into a domain failure. That's the right channel for it.
- `publicationPollSchedule` is `Schedule.exponential` + `modifyDelay`, so the "this one is unbounded" premise behind the item-9 runtime check holds.
- Item 8's placement: in `deleteStack`, `releasingClaimOnFailure` sits outside `restore(...)` but inside `uninterruptibleMask`, so the release actually runs when the body is interrupted. `updateStackRecord`'s release is unmasked, but that is unchanged from the pre-PR `finishOperationBestEffort` call it replaces — not something this PR introduces.
- I ran the tag extraction over all five scanned roots with and without `stripComments`: identical sets (0 lost, 0 gained). So item 5 is purely preventive today, which is what you'd want.

## Findings

### 1. DX / MAINTAINABILITY — `stripComments` can silently shrink the guard's coverage

Inline on `error-actionability-coverage.unit.test.ts:88`. A regex literal containing `\/` followed by `*` (e.g. `/\/*$/`, a plausible trailing-slash trim) starts a fake block comment and blanks source to the next `*/`, so real error classes stop being scanned and the guard passes for them. Latent, not live — verified nothing regresses on the current tree — but it fails in the silent direction, whereas the comment false positive this change fixes fails loudly. Two fixes proposed inline, including an anchored-pattern variant I verified produces identical tags across all five roots and rejects all six new `stripComments` cases, for ~50 fewer lines.

### 2. NON-BLOCKING OBSERVATION — the description overstates what `stripComments` does

The PR body says it "strips comments and masks string/template-literal contents (scanning interpolation expressions as code)". The code does neither: string and template literals are copied through verbatim, and the function's own doc comment correctly says it "does not special-case regex literals or `${...}` interpolation inside template literals". So the pre-existing false positive where a *string* containing `class X extends Error` registers as a definition is still there — not a regression, the raw scan behaved the same, and I found no live instance. Worth correcting the body so the next reader doesn't assume masking is in place.

### 3. NON-BLOCKING OBSERVATION (theoretical) — `flag: "wx"` on the temporary write

Inline on `atomic-claim.ts:62`. New for the identity path, and it makes a SIGKILL-stranded `.tmp.<id>` observable as a hard failure when `temporaryId` repeats, which the primitive's doc says can't happen. Unreachable in production (`randomUUID`) and with both deterministic factories in the current tests.

### 4. NON-BLOCKING OBSERVATION — the reentrancy guard is sqlite-only

`repository-memory.ts` has no transaction concept, so a decision that re-enters the repository is refused on the sqlite adapter and silently tolerated on the in-memory one. Given the guard's premise that reentrancy is a caller bug, the asymmetry means such a bug can pass a memory-backed test and only surface on sqlite. The many `it.each(["in-memory", "persistent"])` cases here reduce the odds a lot, and adding a matching guard to the memory adapter is arguably its own change — noting it rather than asking for it.

### Unrelated, for the record

`packages/stack` `src/Stack.unit.test.ts > preserves current Functions and Edge Runtime settings across partial reloads` fails with `Unknown Error: undefined`. I confirmed it fails identically on the base commit (`5b1b94e`), so it predates this PR — mentioning it only so a red `test:unit` isn't attributed here.

## Verdict

Approve with non-blocking comments. Finding 1 is the only one I'd suggest acting on in this PR; 2 is a description edit; 3 and 4 are fine to leave or defer.

---

avallete (approved):

jgoux:
Additional scope (requested during review of the test surface): dropped implementation-detail unit tests in favor of stronger static guards — net −432 LoC.

- The error-actionability coverage guard now parses source with the TypeScript compiler API instead of a hand-rolled comment/string tokenizer; the tokenizer and its 10-case self-test suite are deleted. A parity harness confirmed byte-for-byte identical discovery before/after (119 files / 485 tags in `apps/cli/src`, all workspace package roots unchanged), so the guard's protection is intact while the entire class of literal-masking bugs is gone.
- `packages/stack/src/managed/model.ts` now enforces the error-code/tag contract at compile time (unions derived from the error classes + an exhaustive-array helper + `satisfies Record<...>`), replacing the reflection-based `managed-model.unit.test.ts`. Known residual gap, documented inline: two classes sharing an identical `code` literal would collapse the union undetected.
- Deleted `managed-stack.unit.test.ts` (legacy `StateManager` resolution pinned through a fake filesystem — pre-M2 code, implementation-detail coverage). The production module stays.

packages/stack/src/managed/atomic-claim.ts
jgoux: Good catch — fixed in 5722f5abb by dropping the exclusive flag on the temporary write (your smallest-fix option). The temp path is unique by construction and `link` is what settles the race, so `wx` there bought nothing; a stranded sibling is now genuinely unobservable — a retry reusing the same id just overwrites it. Added a regression test (`claims over a temporary file stranded by a killed run that reused the same id`) and extended the doc block to state why the temporary write is non-exclusive.

PR #6203, opened 2026-08-14
fix(cli): exec remaining start wrappers so docker stop is not 10s (CLI-2192)

avallete:
`supabase stop --no-backup` was gated by Docker's 10s SIGTERM grace: Kong, Vector, Edge Runtime, and Logflare kept `sh` as PID 1 (or hung in their own shutdown), and stops run in parallel so wall time = slowest container.

Postgres `exec` and postgres-meta v0.98.0 already landed on develop with #6184 / #6207. This PR is only the remaining wrappers, rebased onto that tip.

## What changed

**`exec` in every remaining CLI-authored wrapper** so the real process is PID 1 (deliberate divergence from Go; stop timing is outside the parity surface, ADR 0016). Command paths, output, exit codes, telemetry, and documented side effects unchanged; no `-t`, no `docker rm -f`, no grace-period env:

- **Kong**
- **Vector** — wget wait-loop kept; TERM during that wait now exits instead of burning the 10s grace
- **Edge Runtime** — long-lived `start` / `functions serve` container and the ephemeral script-runner

## `supabase stop --no-backup` wall time

| CLI | run 1 | run 2 | run 3 | min | avg | max |
|---|---:|---:|---:|---:|---:|---:|
| `@beta` 2.116.0-beta.6 | 12.665s | 12.743s | 12.502s | 12.502s | **12.64s** | 12.743s |
| PR 6203 @ 28c39f1 | 4.918s | 4.885s | 5.623s | 4.885s | **5.14s** | 5.623s |

**~2.5× faster** (avg **−7.5s / −59%**). Start time is unchanged (same ~30–45s after images are warm).

## Per-container `docker stop -t 10`

| Container | PID 1 `@beta` | PID 1 PR | Stop `@beta` | Stop PR |
|---|---|---|---:|---:|
| kong | `sh -c` … `docker-entrypoint.sh kong docker-start` | `nginx: master process` | 10.241s | **0.167s** |
| vector | `sh -c` … `vector --config` | `vector --config /etc/vector/vector.yaml` | 10.478s | **0.288s** |
| edge_runtime | `sh -c edge-runtime start …` | `edge-runtime start …` | 10.405s | **0.296s** |
| analytics (logflare) | `sh -c` … `sh run.sh` | supervised `sh run.sh` (TERM → 3s → KILL) | 10.274s | **3.227s** |
| db | `postgres` (already exec) | `postgres` (unchanged) | 0.571s | 3.4–4.5s (see below) |
| pg_meta | `node dist/server/server.js` | `node dist/server/server.js` | 0.257s | 0.297s |

Stops run in parallel, so wall time ≈ slowest container + CLI prune (`container prune` → optional volume prune → network prune). `@beta` is gated by the ~10.4s wrappers. This PR is gated by **Logflare’s ~3s supervisor** and **Postgres shutdown (~4s on a full local db)**, which matches the ~5s `stop --no-backup` times.

## Options to gain more speed (defered)

Just levers. After this PR, wall time is ~`max(logflare ≈ 3s, postgres ≈ 4s)` + ~1s prune. Cutting only Logflare leaves Postgres as the gate (~4s + prune ≈ 5s, i.e. little visible win). Beating ~3s wall needs **both** plus less prune.

### Logflare (~3s supervisor)

1. **Upstream:** image whose SIGTERM handler actually finishes (already tracked as the remaining nicety on CLI-2192). Then the 3s window can win without KILL.
2. **Shorter supervisor grace** (3s → 1s). Same TERM-then-KILL shape, ~2s less. More likely to KILL a BEAM that would have exited at 2.5s.
3. **KILL immediately** on TERM. Same end state as today’s post-timeout kill, ~3s sooner. Least graceful.

### Postgres (~4s smart shutdown)

4. **Fast shutdown** — trap SIGTERM and `pg_ctl stop -m fast` / SIGINT. Drops the wait-for-clients phase; still checkpoints. Helps most when Auth/REST/Realtime/Storage are still connected.
5. **Immediate shutdown** — `-m immediate` / SIGQUIT. Skips checkpoint; next `start` crash-recovers. Reasonable for `stop --no-backup` (volumes die anyway); bad if the volume is kept (`stop` without `--no-backup`).

### DIFF
--- a/apps/cli/src/legacy/commands/functions/serve/serve.integration.test.ts
+++ b/apps/cli/src/legacy/commands/functions/serve/serve.integration.test.ts
@@ -758,7 +758,7 @@ describe("legacy functions serve integration", () => {
           toDockerPath(tempRoot.current),
         ]);
         expect(dockerRun.args[dockerRun.args.length - 1]).toBe(
-          "edge-runtime start --main-service=/root --port=8081 --policy=per_worker\n",
+          "exec edge-runtime start --main-service=/root --port=8081 --policy=per_worker\n",
         );
 
         const envs = yield* Effect.promise(() => extractDockerEnvEntries(dockerRun));
@@ -1879,7 +1879,7 @@ describe("legacy functions serve integration", () => {
 
       const commandScript = dockerRun.args[dockerRun.args.length - 1] ?? "";
       expect(commandScript).toBe(
-        "edge-runtime start --main-service=/root --port=8081 --policy=per_worker\n",
+        "exec edge-runtime start --main-service=/root --port=8081 --policy=per_worker\n",
       );
 
       const cp = deployMockState.runCalls.find(
--- a/apps/cli/src/legacy/commands/start/services/kong.service.ts
+++ b/apps/cli/src/legacy/commands/start/services/kong.service.ts
@@ -37,8 +37,8 @@
  * disk when TLS is enabled AND both `cert_path`/`key_path` are configured —
  * see {@link LegacyKongContainerSpecInput.tlsCertContent}'s doc comment.
  * {@link legacyBuildKongEntrypointScript} reproduces the remaining
- * `custom_nginx.template` heredoc + exec line byte-for-byte; see its doc
- * comment for the exact shell mechanics.
+ * `custom_nginx.template` heredoc; the final command is `exec`'d so Kong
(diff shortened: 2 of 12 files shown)


avallete:
@codex review

jgoux (approved):

PR #6220, opened 2026-08-16
perf(cli): reuse shadow baseline cache in migration squash and pgadmin diff

avallete:
## Summary

Stacked on #6215 (← #6203 ← #6184 ← #6102).

`migration squash` and `db diff --use-pgadmin` were the last two shadow-database consumers still provisioning bare cold shadows — every run re-paid the full platform baseline (init schema + the PG15+ realtime/storage/auth one-shot migrate jobs, ~15s) that the shadow-baseline cache already eliminates for `db diff`/`db pull`. Both now provision through `legacyWithShadowDatabase` and warm-restore the cached pg_data tar when eligible.

The shared prologue (connect → prelude → `legacySetupDatabase`, with the warm/cold/cold-snapshotting branches) is extracted into `legacyOpenShadowBaselineSession` in `shadow-database.ts`; `legacySetupShadowDatabase` and the migrate path delegate to it, and squash calls it directly so its pre-migration auth/storage dumps stay exactly at the baseline seam. Squash keys config-following (its baseline always followed `config.toml`); pgadmin keys `webhooks: "enabled"` to share the legacy migra branch's forced-`pg_net` tars.

Two defects surfaced while wiring this in, both fixed here:

- **Cache-key/cluster-state mismatch (correctness):** `db pull` and `db diff`'s native branch hardcoded `{ webhooks: "enabled" }` as the cache-key opts even in pg-delta next mode, where the actual provisioning (`legacyMigrateNextShadowDatabase`) follows `config.toml`. On a project with webhooks disabled, a cold next-mode run published a `pg_net`-less tar under the "enabled" key — which the newly cached pgadmin branch (and the reverse direction) could then silently warm-restore. The key opts now follow the migration mode at all call sites; `legacyResolveMigrationsCatalogRef` in `legacy-pgdelta.cache.ts` had the same drift (keyed config-following, provisioned forced-on) and is fixed too. A regression test asserts a next-mode diff and a pgadmin diff on a webhooks-disabled config never share tars (verified failing pre-fix).
- **Readiness gate (perf):** both call sites gated on `legacyWaitForHealthyServices` (Docker HEALTHCHECK, first probe at t+10s, ~6.5s pure wait) instead of the `legacyWaitForShadowReady` connect probe every other cached consumer uses. Swapped, so the warm restore actually lands its win. The now-dead `neverHealthy` mock knob is deleted; the interrupt/timeout tests model a connect-refusing shadow instead.

Test-helper cleanups ride along: a shared `withLegacyShadowCacheEnabled` in `tests/helpers/legacy-mocks.ts` replaces per-file env stash/restore copies, and the squash test uses `legacyShadowBaselineCacheDir` instead of a hand-built path.

**Transcript note:** on a warm hit both commands skip the `Initialising schema...` / `Seeding globals from roles.sql...` lines, matching the existing warm behavior of `db diff`/`db pull`. All dumps, migration-replay lines, and result output are unchanged.

## Linked issue

Closes #

- [ ] The linked issue is **open** and carries the `open-for-contribution` label (or I'm a Supabase maintainer).

## Checklist


avallete:

PR #6221, opened 2026-08-16
refactor(cli): single-source the shadow webhooks policy on the setup input

avallete:
Stacked on #6220 (`avallete/shadow-cache-squash-pgadmin`).

## Summary

The shadow-baseline cache key and the provisioned baseline previously expressed the Webhooks/`pg_net` policy independently: callers passed `LegacyShadowCacheOpts.webhooks` for the key, while the `use` phase picked a provisioner that applied its own policy (`legacyMigrateShadowDatabase` forced `"enabled"`, `legacyMigrateNextShadowDatabase` followed config, `legacySetupShadowDatabase` took caller options). Nothing enforced agreement — three of five call sites had drifted before #6220 re-aligned them by hand, and nothing stopped a sixth call site from drifting again.

This makes that drift class unrepresentable. `LegacyShadowDbSetupInput` now carries a **required** `webhooks: LegacyShadowWebhooksPolicy` field that BOTH the cache-key resolution (`shadow-cache.ts`) and every baseline provisioner read, so a published snapshot's key can never describe a different baseline than the one actually provisioned.

The old parallel surfaces are deleted outright (no shims):

- `LegacyShadowCacheOpts.webhooks`
- `legacyMigrateNextShadowDatabase` — one migrate function remains, reading the input's policy
- the `LegacySetupDatabaseOptions` params on `legacyOpenShadowBaselineSession` / `legacySetupShadowDatabase`
- `legacyPrepareShadowSource`'s migrate dispatch and the pg-delta-next layer's webhooks-carrying `cacheOpts`

The per-engine semantics — legacy migra/pg-delta engines force `pg_net` on (Go parity), pg-delta next follows config — are named exactly once, in `legacyShadowSourceWebhooksPolicy(migrationMode)`, which `db diff`/`db pull` use when building their shadow inputs. pgAdmin, `migration squash`, the catalog exports, and pg-delta next's migrations/declarative shadows state their fixed policies at input construction, each with the rationale in place. The two raw-shadow sites (`db pull`'s legacy declarative export, `declarative.orchestrate.ts`) pass an explicitly inert `"config"` — they never run a baseline or touch the cache.

Effective policies, cache keys, and provisioned SQL are unchanged at every call site.


avallete:

PR #6223, opened 2026-08-16
perf(cli): warm-restore the local database baseline on db reset and fresh start

avallete:
## Summary

Stacked on #6220 (← #6215 ← #6203 ← #6184 ← #6102).

`supabase db reset` and a fresh-volume `supabase start`/`db start` ran the full platform baseline (init schema + the PG15+ realtime/storage/auth one-shot migrate jobs, ~15s) against the **real** `db` container on every invocation — the exact cost the shadow-baseline cache already eliminates for shadows, produced by the same shared `legacySetupDatabase`. The main database now shares that cache pool:

- **Warm (tar published for the config key):** the tar is injected as a `preStartArchives` entry into the `db` container spec (unpacked between `docker create` and `docker start`, the mechanics `pgdata-snapshot.ts`/`container-lifecycle.ts` already provide), the platform baseline is skipped entirely, and only the project's own migrations + seed replay. The transcript prints an explicit `Restoring cached baseline...` in place of `Initialising schema...`/`Seeding globals from roles.sql...`.
- **Cold:** the current flow runs unchanged, then publishes the baseline at the seam (after `legacySetupDatabase`, before migrate/seed) via a stop → `docker cp` → restart-with-connect-probe dance. Publication is best-effort: any failure warns and the command completes normally. Safe on `supabase start` because the DB bootstrap fully completes before any other service starts.
- **Fallback:** a warm restore that doesn't come up (corrupt/incompatible tar) warns, deletes the suspect tar on readiness timeout, removes the container **and named volume**, and recreates cold — with interrupt guards mirroring the shadow path so Ctrl-C never strands a half-restored db. The fallback's own export atomically replaces the bad tar.
- **Gates:** same eligibility as shadows (PG15+, `SUPABASE_SHADOW_CACHE` default-on env as escape hatch); `db start --from-backup` bypasses entirely; the PG≤14 reset path is untouched.

Structure: `main-db-baseline.ts` owns the whole bring-up (`legacyBringUpMainDbWithBaseline` — peek → spec decoration → create with warm-fallback), both `start-database.ts` and `recreate-local-database.ts` reduce to one call. `baseline-state.ts` hoists the cluster baseline state as a `warm | cold | uncached` discriminated union plus the session-opening branch both `db-setup.ts` and `shadow-database.ts` previously duplicated; `legacyRunFreshDbSetup` absorbs the deleted `legacyStartSetupLocalDatabase`. `shadow-cache.ts`'s key resolution, export step, and eviction are generalized structurally (`LegacyBaselineCacheInput`) so one implementation serves both cluster kinds; shadow warn/transcript strings are preserved verbatim.

Supporting fixes that fell out:

- The handlers' already-loaded `config.toml` values are threaded through instead of a redundant re-parse (fresh runs stay at the pre-existing two parses, and nothing runs when the cache is ineligible).
- `roles.sql` is read once: the peeked bytes that were hashed into the cache key are the exact bytes the seeding step executes, closing a narrow TOCTOU where the key could describe content that was never applied (also fixed for shadows).
- JWKS discovery is memoized at the source (`legacyMemoizeSuccess` on the input), so key resolution and the setup prelude share one resolution.
- `LegacyBaselineSnapshotRevivalFailure` is a `Data.TaggedError`.
- A documented invariant: shadows bootstrap under `-c max_worker_processes=0` while the main db does not, and this is deliberately unkeyed — safe today because the flag is command-line-only and no PG15+ baseline step persists worker-dependent state; any future baseline step that depends on live background workers must add the discriminator to the key.

The first commit is standalone test hygiene: four integration suites drove cold shadow provisions with the cache on and `SUPABASE_HOME` unpinned, writing stray tars into the developer's real `~/.supabase` on every test run.

New integration coverage: reset publish/reuse/broken-restore-recreate/export-failure-warns/PG14-never-consults, `db start` warm fresh start and `--from-backup` bypass, driven by real cold-run-then-warm-run cycles rather than hardcoded keys.

## Linked issue

Closes #

- [ ] The linked issue is **open** and carries the `open-for-contribution` label (or I'm a Supabase maintainer).

## Checklist


avallete:

PR #6224, opened 2026-08-16
fix(cli): batch migration statements (CLI-2182)

### LINKED ISSUE #6172: Supabase cli command slow on new version 2.113.0
# Bug report

- [x ] I confirm this is a bug with Supabase, not with my own application. 
- [ x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

npx --yes supabase@2.113.0 db reset --local takes over a half hour while npx --yes supabase@2.109.1 db reset --local take about 1 minute. 

## To Reproduce

1. Create the timestamped migration under `supabase/migrations/` in your repo.
2. Run npx --yes supabase@2.109.1 db reset --local
3. Note elapsed time
4. Repeat process with npx --yes supabase@2.113.0 db reset --local

## Expected behavior

I expect supabase@2.113.0 to be as fast as supabase@2.109.1

## System information

- OS: Windows
- Browser (if applies) [e.g. chrome, safari] - doesn't apply but I've used Chrome
- IDE: VS Code: code -v: 1.132.0 df53daabb18cd157bdb08c7f01c34df936cf12f4 x64
- node -v: v22.19.0
- npx supabase -v: 2.113.0

7ttp:
## TL;DR

fixes `supabase db reset` and `db push` taking minutes to run ( on windows )
bringing the port much closer to the Go CLI by batching migration and seed statements the same way it used to

## what was happening?

The native TypeScript port ran every SQL statement as its own database round trip
The Go CLI it replaced sent a whole file down the wire at once
so big dump style migrations that used to apply in seconds started taking minutes. Seed files had the same per statement loop...

## hows it fixed?

Go pipelined a whole file over the extended protocol and ended it with a single Sync,
 and we now follow that exactly, one Parse/Bind/Execute per statement and one Sync per batch. 
Statements that can't live inside a batch 
(CREATE INDEX CONCURRENTLY, VACUUM, ALTER SYSTEM, CLUSTER) still run standalone the way they always did
since Postgres refuses to run those inside a transaction. The history insert and the seed hash upsert ride in the final batch, so a file is only recorded once everything in it succeeded...

## ref:
- closes: https://github.com/supabase/cli/issues/6172


jgoux (approved):

avallete:
Few comments, okay as follow-ups:

1. **Connect failure misclassified as exec failure** (confirmed, new in this PR) — [legacy-db-connection.sql-pg.layer.ts:947](apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.layer.ts:947). If the backend is down when `acquireBatchClient` redials, the `ECONNREFUSED` gets wrapped as `LegacyDbExecError`: wrong telemetry category, the connect-suggestion text is dropped, and the user sees "At statement: 0" blaming their first SQL statement for a connectivity problem. The file's own comment on the sibling `acquireRawClient` path explicitly warns against exactly this.

2. **The optional `execBatch?` + dead fallback** (confirmed, flagged independently by five angles) — [legacy-migration-apply.ts:576](apps/cli/src/legacy/shared/legacy-migration-apply.ts:576). No production session ever lacks `execBatch`, so the BEGIN/COMMIT fallback is dead code — and worse, the `push`/`reset`/`squash`/`repair` integration suites all mock sessions *without* `execBatch`, so every command-level test exercises the dead path instead of what ships. Making `execBatch` required and updating the four mocks fixes both.

**Fine as follow-ups** (real but inherited from Go's identical arithmetic, so not regressions): a Sync-time deferred-constraint failure on a non-final batch names a statement that never ran; and an unmatched user `BEGIN` can leave the whole file uncommitted while the CLI reports success (pre-PR TS was accidentally safer than Go here; neither case has test coverage). Plus two minor cleanups: `finish()` rebuilds the error field-by-field, and the test file now has a third copy of the fake Postgres handshake loop.

PR #6234, opened 2026-08-17
fix(cli): classify batch connection failures as connect errors

avallete:
## TL;DR

Follow-ups to the two confirmed findings from the #6224 review ([comment](https://github.com/supabase/cli/pull/6224#issuecomment-5317097171)).

## 1. Batch connection failures were misclassified as exec failures

`execBatch` checks its own connection out of the pool, but mapped a failed checkout through `legacyToExecError`. A plain `ECONNREFUSED` during the redial surfaced as `LegacyDbExecError`: wrong telemetry actionability, the connect-suggestion text dropped, and `legacyFormatExecBatchError` rendering `At statement: 0` — blaming the migration's first statement for the database being unreachable.

Acquiring the batch connection now fails with `LegacyDbConnectError` (same pattern as the sibling `acquireRawClient`), and the migration-apply path surfaces it verbatim instead of routing it through the statement formatter. Mid-batch/Sync failures stay `LegacyDbExecError` — only connection *acquisition* moves to connect classification. Shadow-scoped callers fold it into their own connect-flavored errors (`LegacyShadowDbError{reason:"connect"}`, `LegacyDeclarativeShadowDbError`) so the suggestion text survives.

## 2. `execBatch` is now a required session capability

No production session ever lacked `execBatch`, so the `BEGIN`/`COMMIT` fallback in `execMigrationBatch` was dead code — and the push/reset/squash/repair integration suites all mocked sessions *without* it, so every command-level test exercised the dead path instead of what ships. `execBatch` is required on `LegacyDbSession`, the fallback is deleted, and the test mocks now go through a shared `legacySequentialExecBatch` helper (or batch recorders) so they exercise the batching path.

## ref
- follow-up to: https://github.com/supabase/cli/pull/6224


7ttp (approved):
Sand Surfing LGTM

jgoux (approved):

PR #6245, opened 2026-08-18
fix(stack): deflake service-state settling and control endpoint acquisition

avallete:
## Summary

This PR removes a family of intermittent stack test failures by fixing the underlying races rather than the tests, then verifies the result with five consecutive green CI runs and a local 2× CPU-oversubscribed stress gauntlet (8/8 integration runs green).

Runtime fixes:

- **`stopService` settling race** (the original `'Dormant' vs 'Stopped'` CI flake): `stopService` returned before the background projection fiber re-published the public state, so an immediate `getState` could observe the previous status. It now settles the projection before returning, matching every start path.
- **Control endpoint port collisions**: the control endpoint derived a single loopback port from two bytes of the stack id, so two live stacks could birthday-collide and the later acquirer hard-failed with `ControlAddressConflictError`. `acquireControl`/`probeControl` now walk a short deterministic candidate sequence — attach to a matching owner on any candidate (verified by `ownershipId`), bind the first free one, and conflict only when all candidates are foreign-occupied. `connectManagedStack` uses the probed endpoint, and exact service-port requests reserve every candidate. Protocol mismatch still fails closed.
- **Keep-alive livelock**: control status probes reused pooled connections, so a closed listener kept answering `/owner` on the poller's own hot connection and a scan-first acquirer could never bind the freed endpoint. Control reads and stop requests are now one-shot connections in both Node and Bun transports.
- **Retry budgets**: the bound-but-not-serving acquire retry is now duration-bounded (a count-based budget stretched a single acquire to 30–45s when reads hit the 500ms transport timeout), and `startStack`'s workspace-repair fence waits up to 30s instead of ~5s, which a realistic Git repair can exceed.

Test hardening:

- Managed test layers set a new `preferCatalogDefaults: false` plan option so parallel suites stop contending on the default ports (54321…), which sticky reuse re-reserves exactly; production behavior and the spec'd no-relocation semantics are unchanged.
- Write-failure injections are privilege-independent (FileSystem-seam gates / directory-as-file instead of chmod, which root bypasses).
- Supervisor test watchers re-arm on ENOENT from atomic-write temp files vanishing mid-scan; sub-second synchronization timeouts are widened into guards.

## Linked issue

Closes #

## Checklist

https://claude.ai/code/session_01McsNM9yxC5tiY6SusBALoq


jgoux (approved):

avallete:
Flake verification: re-running the Test workflow repeatedly on this head (`5d26196`) to confirm the `stopService` settling fix holds under CI load.

Results so far: attempts 1–3 fully green; the previously flaky `Stack.unit.test.ts` unit suite has passed in all 4 attempts.

Attempt 4 failed on an unrelated, pre-existing flake in `@supabase/stack:test:integration`:

- `managed-manager-ports.integration.test.ts > keeps automatic ports exclusive while stopped` died with `ControlAddressConflictError: Control endpoint http://127.0.0.1:23403 is occupied by a non-Supabase listener` (cause: endpoint owned by `d9dd6c8d…`, not `ce16dfe6…`).
- Mechanism: `controlEndpoint` (`managed/control.ts`) derives the port from 2 bytes of the stack id over a 22,768-port range on a fixed `127.0.0.1`, so two temp-workspace stack ids in concurrently running integration files can birthday-collide on the same port; the later acquirer correctly fails. This code path is untouched by this PR and reproduces on `develop` — flagging it here as a candidate for the ongoing deflaking work (#6239) rather than fixing it in this PR.

Re-run for attempt 5 is queued; will report the final tally.

---

avallete:
Flake verification complete — final tally over 5 full Test workflow attempts on `5d26196`:

| Attempt | Result |
|---|---|
| 1 | ✅ success |
| 2 | ✅ success |
| 3 | ✅ success |
| 4 | ❌ failure — unrelated `ControlAddressConflictError` port collision in `managed-manager-ports.integration.test.ts` (details in the previous comment) |
| 5 | ✅ success |

The `Stack.unit.test.ts` suite containing the previously flaky `preserves an explicitly stopped service across a stack restart` passed in **all 5 attempts**, so the `stopService` settling fix holds under CI load. The single attempt-4 failure is the pre-existing control-endpoint port collision that reproduces on `develop` and is independent of this change.

---

avallete:
Flake-fix verification complete: **5 consecutive green Test workflow runs** on head `5d80e4a` (run 32128584862, attempts 1–5), plus a green run on the intermediate head `65662f9` — six consecutive CI passes since the fixes landed. Locally, the integration suite also passed 8/8 under a 2× CPU-oversubscribed stress gauntlet that previously reproduced failures in 2–3 of every 6 runs.

Root causes fixed in this PR since the last report:

1. **Control endpoint port collisions** — `acquireControl`/`probeControl` now walk a deterministic candidate sequence instead of failing on a single hash-derived port (fixes the `managed-manager-ports` CI flake).
2. **Keep-alive livelock** — control status probes reused pooled connections, so a closed listener kept answering on the poller's own hot connection and a scan-first acquirer could never bind the freed endpoint; control transport requests are now one-shot in both Node and Bun transports.
3. **Count-bounded acquire retry** — could stretch one acquire call to 30–45s when reads hit their 500ms transport timeout; now duration-bounded (1.5s).
4. **Repair-fence budget** — `startStack` gave up on a held workspace-repair fence after ~5s; now waits up to 30s.
5. **Test hardening** — managed test layers opt out of catalog default ports (`preferCatalogDefaults` seam) so sticky reuse doesn't race parallel files' default-port probes; supervisor test watchers re-arm on ENOENT from vanishing atomic-write temp files; sub-second sync timeouts widened into guards.

---

PR #6246, opened 2026-08-18
fix(cli): restore postgres after role reverts (CLI-2205)

### LINKED ISSUE #6236: db push (passwordless login-role path): migration containing `reset role` fails the ledger INSERT with 42501 and silently no-ops `granted by current_user` cleanup
### Describe the bug

When `supabase db push --linked` runs **without a database password** (the temporary-login-role path, "Initialising login role..."), a migration that contains `set role` / `reset role` breaks in two ways:

1. **The push fails on the CLI's own ledger bookkeeping.** After the migration's `reset role`, the session is no longer `postgres` (see analysis below), so the CLI's appended  <br>`INSERT INTO supabase_migrations.schema_migrations(version, name, statements) VALUES($1, $2, $3)`  <br>fails with `42501 permission denied for schema supabase_migrations`, and the whole migration rolls back. Every statement in the migration file itself succeeds — the failure is always "At statement: N" where N == the number of top-level statements in the file, i.e. the ledger INSERT.
2. **Worse (silent correctness hazard): statements after** `reset role` **execute under the temporary login role, not** `postgres`**.** In our migration, cleanup of the form  <br>`revoke <role> from postgres granted by current_user;`  <br>became a **silent no-op** (grantor mismatch: the earlier `grant` was recorded with grantor `postgres`, but `current_user` at revoke time is the temp role). If the ledger INSERT had not happened to fail, the migration would have **committed with its role-hygiene cleanup quietly missing** — no error, no warning surfaced by the CLI.

Additionally, on **v2.110.0** (the first release where `db push` is served by the TypeScript port) the real error is swallowed: the CLI prints only

```
{"_tag":"Error","error":{"code":"LegacyDbPushApplyError","message":"effect/sql/SqlError: Failed to execute statement\nAt statement: 14\nINSERT INTO supabase_migrations.schema_migrations(version, name, statements) VALUES($1, $2, $3)"}}
```

with no SQLSTATE/message/detail, and `--debug` adds nothing. v2.114.0 correctly surfaces `ERROR: permission denied for schema supabase_migrations (SQLSTATE 42501)`, which is how we diagnosed this.

### To Reproduce

Against a hosted (linked) project, with **no** `SUPABASE_DB_PASSWORD` set and no `-p/--db-url` (so the CLI mints the temporary login role), push this migration:

```sql
-- repro migration: role choreography around a definer function
create role repro_writer nologin;

grant repro_writer to postgres;

set role repro_writer;

create or replace function public.repro_fn() returns int
  language sql security definer as 'select 1';

reset role;

revoke repro_fn_unrelated_priv_or_any_cleanup on schema public from repro_writer; -- any post-reset statement
revoke repro_writer from postgres granted by current_user;                        -- silent no-op under temp role
```

Observed:

* `supabase db push --linked` fails at "statement N" (the ledger INSERT) with `42501 permission denied for schema supabase_migrations` (visible on ≥ 2.114.0; swallowed into a generic `effect/sql/SqlError` on 2.110.0). The migration rolls back.
* Running the exact same push with `SUPABASE_DB_PASSWORD` set (session is genuinely `postgres`): the migration applies and registers cleanly, and the `granted by current_user` revoke actually takes effect.

Two sibling migrations in the same push containing plain DDL (no `set role`) applied and registered fine, both paths.

### Expected behavior

Either of:

* `RESET ROLE` inside a migration should restore the effective role the CLI intends to run as (`postgres`), not silently drop the session to the temporary login role. E.g. set the step-down role via connection-time options (`options=-c role=postgres`) so `RESET ROLE` restores `postgres`, or re-assert `SET ROLE postgres` before the ledger INSERT.
* At minimum, document that under the passwordless path, `set role` / `reset role` / `granted by current_user` inside migrations have different semantics than under a real `postgres` session — the silent-no-op cleanup case is the dangerous half of this, since it commits without any error.

(And thank you for the v2.114.0 error-reporting rebuild — v2.110.0's swallowed error made this significantly harder to diagnose.)

### System information

* OS: macOS (arm64)
* CLI: reproduced on **2.110.0** (error swallowed) and **2.114.0** (real SQLSTATE surfaced); both fail identically on the passwordless path, both succeed with `SUPABASE_DB_PASSWORD` set
* Hosted project, `db push --linked`, migration applied via the CLI only (no dashboard SQL)

### Analysis

`Initialising login role...` = the CLI minting a temporary login role via the Management API (`POST /v1/projects/{ref}/cli/login-role`) and connecting as it (`cli_login_postgres`), then stepping down to `postgres` with a session-level `SET ROLE` (the `db diff` template's "Step down from login role to postgres" comment is the same mechanism). Because the step-down is a plain `SET ROLE` statement rather than a connection-time default, a migration's own `RESET ROLE` restores the session to `cli_login_postgres`. From that point on, (a) the ledger INSERT runs as a role with no USAGE on `supabase_migrations`, and (b) role DDL that depends on `current_user` resolves against the temp role — grantor-mismatched `REVOKE ... GRANTED BY current_user` no-ops silently.

Possibly related cluster: supabase/cli#5912, supabase/cli#6116, supabase/cli#6094 (role-statement interactions on the hosted path in the same era).

7ttp:
## TL;DR

Passwordless `db push --linked` breaks on any migration containing `reset role`: 
the login role path relies on a session level `SET SESSION ROLE postgres` that the migration itself undoes. 
File runners now re-assert the step-down immediately after each role-reverting statement, at the end of each file, and before every CLI owned ledger write, so the whole file behaves the same on both auth paths.

## whats biting?

The passwordless path connects as a temp `cli_login_*` role and steps down to `postgres` once at connect. A migration's `reset role` reverts the session to the login role, so:

- the appended history insert fails with `permission denied for schema supabase_migrations (SQLSTATE 42501)` and rolls the migration back, even though every user statement succeeded
- authored transaction and pg-delta no-transaction files commit their statements but never record, so the next push re-applies them
- statements between the `reset role` and the end of the same file run as the login role, so `granted by current_user` cleanup silently no-ops while the push exits 0 (reproduced on staging: the stale `pg_auth_members` grant survives)
- later files and the `seed_files` upsert run as the login role too

## fixed now by:

- `LegacyDbSession.restoreRoleSql` (set only when the step-down ran) is injected by every file runner right after each top-level role revert (`RESET ROLE`, `SET [SESSION] ROLE [TO|=] NONE|DEFAULT` including a case-sensitively quoted `'none'`, `RESET SESSION AUTHORIZATION`, `SET SESSION AUTHORIZATION DEFAULT`, `DISCARD ALL`), and again at end of file and before the history insert and both `seed_files` upserts, so `current_user` matches a password session for the whole file
- injected restores never shift `At statement: N` and are never recorded in the history row; deliberate `set role <x>` choreography is untouched, and password, local and plain `--db-url` sessions see a byte identical statement stream
- the residual (dynamic SQL, `SET LOCAL ROLE NONE`, `session_user` itself) is documented in `docs/go-cli-divergences.md` with the end-of-file restore protecting every CLI owned write; both `SIDE_EFFECTS.md` tables record the new statements

## ref:
- closes: https://github.com/supabase/cli/issues/6236


avallete (commented):
Thank's for looking into this !

One issue raised by Claude on this one, dug a bit:

The end-of-file `restoreRoleSql` correctly fixes the filed bug: after `RESET ROLE`, the history insert no longer dies with `42501` and later files start as `postgres`. Password / local / non-stepped-down `--db-url` streams stay byte-identical. That part is good.

An issue is residual in `go-cli-divergences.md` as the product behavior.

Statements **between** `RESET ROLE` and the end of the **same** file still run as the login role. For the class of migrations in #6236 (`SET ROLE` / `RESET ROLE` / `REVOKE … GRANTED BY current_user`), this PR changes the failure mode the wrong way:

- **before:** fail-closed — `42501`, migration not recorded, retryable
- **after:** succeed, history row written, grant leftover, stderr `WARN:` that CI will ignore

The pooler rationale does not apply here. You do not need a connection-time `role=postgres` default. `restoreRoleSql` is already `SET SESSION ROLE postgres`. Inject it **after each top-level** `RESET ROLE` / `SET ROLE NONE` / `SET SESSION ROLE NONE` in the same batch/sequence, keep recording the original statements, and keep the trailing restore as a backstop. `At statement: N` should skip those internal ops the same way the trailing restore already does.

I reproduced this on a real passwordless `db push --linked` against staging (`Initialising login role…`, login role `cli_login_postgres`, `NOINHERIT`):

```sql
create table if not exists public.dogfood_probe (
  at text primary key,
  current_user_name text not null,
  session_user_name text not null
);
insert into public.dogfood_probe values ('start', current_user, session_user);
create role repro_writer nologin;
grant repro_writer to postgres;
grant usage on schema public to repro_writer;
grant insert on public.dogfood_probe to repro_writer;
set role repro_writer;
insert into public.dogfood_probe values ('as_writer', current_user, session_user);
reset role;
revoke repro_writer from postgres granted by current_user;
```

Result: `WARN: statements after RESET ROLE…`, exit 0, `schema_migrations` recorded `dogfood_reset_role`. Probe at start was `current_user=postgres` / `session_user=cli_login_postgres`. After the push, `pg_auth_members` still had `repro_writer` → `postgres` **grantor `postgres`**. The same `REVOKE … GRANTED BY current_user` on the password path (`current_user` actually `postgres`, no login-role init) dropped that row.

The warn can stay as defense-in-depth for dynamic SQL / siblings the detector misses. It should not be the answer for a static `RESET ROLE` already in the statement list.

**Request:** mid-file restore, then this is good to land.

avallete (approved):

PR #6277, opened 2026-08-19
fix(cli): detect dead db connections (CLI-2207)

### LINKED ISSUE #6244: supabase db reset/start: migration-apply SQL pipeline stops mid-stream with no server-side counterpart
# supabase db reset/start: migration-apply SQL pipeline stops mid-stream with no server-side counterpart

**Repository:** `LinkFault/linkfault-foundation`
**Workflow:** `LinkFault Verification` (`.github/workflows/verify.yml`)
**Runner:** `ubuntu-24.04` (GitHub-hosted)
**Failing step:** `supabase start --debug` or `supabase db reset --local --debug` (same underlying stall, surfaces in whichever step happens to be applying migrations at the time)

## Summary

Six `workflow_dispatch`/`push` runs, across four unrelated branches (including `main` itself, post-fix) with no shared content beyond a common ancestor, all stalled at the **identical point**: mid-way through a pipelined SQL statement stream the CLI sends while applying local migrations, specifically on

```
INSERT INTO supabase_migrations.schema_migrations(version, name, statements) VALUES($1, $2, $3)
```

immediately following the same migration file (`20260817160209_assessments.sql`). The step produces **no error text, no exception, no timeout message** — the log stream simply stops until the step's own external timeout (or GitHub's default) kills it. Reproduced on two pinned CLI versions (`2.109.1` and `2.114.0`), and both before and after an unrelated, fully-diagnosed, already-fixed content bug in this repo (a duplicate `CREATE TYPE` across two migrations) — ruling out both the CLI version and repo migration content as the cause.

A live database snapshot taken during one occurrence (run 6, below) shows the database itself idle and unblocked at the moment of the stall, with no connection matching the CLI's migration-apply session present at all. We can't see past that boundary — whether the stall is inside the CLI process itself, or in something between the CLI and the database (a Docker networking hiccup between client and container, for instance) isn't something we can distinguish from where we're looking. Filing here since the pipelined statement-batching behavior visible in `--debug` output is the last thing we can observe before the stall, not because we've localized the fault to this repo specifically.

## The six runs

| # | Run ID | Branch | Started (UTC) | Ended (UTC) | Duration | CLI version | Notes |
|---|---|---|---|---|---|---|---|
| 1 | [32091737259](https://github.com/LinkFault/linkfault-foundation/actions/runs/32091737259) | `feat/step-9-billing-enforcement` | 2026-08-18 02:24:47 | 02:27:45 | ~3m | 2.109.1 | |
| 2 | [32091982094](https://github.com/LinkFault/linkfault-foundation/actions/runs/32091982094) | `feat/step-9-billing-enforcement` | 2026-08-18 02:28:48 | 02:31:56 | ~3m | 2.109.1 | retry of #1 |
| 3 | [32092494820](https://github.com/LinkFault/linkfault-foundation/actions/runs/32092494820) | `feat/step-9-billing-enforcement` | 2026-08-18 02:37:15 | 02:39:49 | ~2.5m | 2.109.1 | retry of #1/#2 |
| 4 | [32093153813](https://github.com/LinkFault/linkfault-foundation/actions/runs/32093153813) | `chore/ci-resource-sampling` | 2026-08-18 02:48:11 | 02:50:56 | ~2.75m | 2.109.1 | host-resource telemetry captured (below) |
| 5 | [32095227294](https://github.com/LinkFault/linkfault-foundation/actions/runs/32095227294) | `feat/lean-admin` | 2026-08-18 03:22:17 | 03:25:04 | ~2.75m | 2.109.1 | unrelated migration content |
| 6 | [32099052341](https://github.com/LinkFault/linkfault-foundation/actions/runs/32099052341) | `main` | 2026-08-18 04:30:37 | 04:32:18 | ~1.75m | 2.109.1 | **post-fix, with DB-level watchdog — see below** |

Runs 1–3 are the same branch, retried per an internal "retry once on suspected flakiness" policy — three identical failures is what first raised this to a real investigation. Runs 4–5 are unrelated branches added specifically to test content-dependence; the stall is identical regardless of migration content. Run 6 is `main` itself, *after* fixing an unrelated duplicate-`CREATE TYPE` bug that had briefly been suspected as the cause, and *with* a container/DB-level watchdog added independently by another contributor — this is the run with the evidence below.

All six: the relevant step (`Start local Supabase` in runs 1–5, `Clean local database reset` in run 6 — see "Why the step differs" below) ends in `failure` with `RESULT_TESTS: skipped`; the job never reaches the pgTAP suite.

### Why the failing step differs between run 6 and runs 1–5

Runs 1–5 used a workflow without a `--debug` flag on `supabase db reset`. Run 6's workflow (updated independently, unrelated to this investigation) added `--debug` to the reset step and increased timeouts, which appears to have shifted *which* step the CLI's internal migration-apply sequence was mid-way through when the stall occurred — but it is the same stall, at the same statement, following the same migration.

## Evidence (run 6): the database shows no counterpart activity during the stall

Run 6's workflow includes a 30-second-interval background watchdog (added independently) that snapshots `docker ps`, `pg_stat_activity`, and blocking-lock state throughout the step, `if: always()` so it's captured even on failure. The snapshot taken during this stall:

```
--- pg_stat_activity (supabase_db_cjuamnnwhddycyicsqxu) ---
pid|state|wait_event_type|wait_event|query
384|active|Client|WalSenderWaitForWal|START_REPLICATION SLOT cainophile_5dciytpl LOGICAL ...
439|active|||SELECT pid, state, wait_event_type, wait_event, left(query, 200) AS query ... (the watchdog's own query)
205|idle|Extension|Extension|
230|idle|Client|ClientRead|LISTEN "pgrst"
... [23 more rows, all idle Supabase-internal services: Realtime, Logflare,
    PostgREST connection-pool housekeeping — none related to migration apply]
(27 rows)

--- blocking locks (supabase_db_cjuamnnwhddycyicsqxu) ---
pid|blocked_by|query
(0 rows)
```

**No row corresponds to the CLI's migration-applying connection at all**, and **zero blocking locks** exist. The database has no active or blocked query anywhere near the stalled `INSERT INTO supabase_migrations.schema_migrations` statement — it's sitting idle. Whatever's stopped making progress, it isn't visible from inside Postgres: no lock, no long-running query, no session matching the CLI's last `Parse`/`Bind`/`Describe`/`Execute` burst. That's as far as we can see from this side — we can't tell from here whether the CLI process itself stopped, or whether something between it and the database (e.g. a dropped or hung connection at the Docker networking layer) is why the database never saw the rest of the burst.

## Host-level resource telemetry (run 4)

Before the watchdog above existed, we independently added host-level sampling (`free -m` / `df -h` / `docker stats` / `docker ps`, every 15s) around the same step, to rule out resource exhaustion as a simpler explanation. From run 4:

- **Memory:** stable throughout — `available` stayed at 6.6–6.8GB of 7938MB total; `buff/cache` grew (normal page caching), no pressure.
- **Disk:** climbed from 82% used (14G avail) to 91% used (7.2G avail) over ~90s *before* the Postgres container started — consistent with Docker image pull/extraction. Never exhausted; bottomed at 7.2G free.
- **The DB container** appeared ~94s into the step, `health: starting`, then `healthy` 22s later, using a trivial 65–76MiB memory / 0.04–4.4% CPU throughout.
- **The stall occurs ~8–9s after the container reports healthy**, mid-pipeline in the same fast `Parse`/`Bind`/`Describe`/`Execute` burst as every other run. No OOM/kill/disk-full signature anywhere in any of the six full logs (grepped explicitly — zero matches).

Combined with run 6's finding, this rules out host-level resource exhaustion (memory/disk/CPU on the runner) and database-level lock contention as explanations. What's left is that the stall has no visible counterpart on either the host or the database side — consistent with something happening in the path between the CLI process and the database, but we can't localize it further than that from here.

## What's been ruled out

- **Memory exhaustion** — ruled out directly (run 4 telemetry).
- **Disk exhaustion** — real churn during image pull, never hit 0; not a clean explanation on its own.
- **Database-side lock contention** — ruled out directly (run 6: zero blocking locks, DB idle).
- **Migration content** — ruled out; identical stall across branches with materially different migration sets, always at the same pre-existing migration boundary, both before and after fixing an unrelated content bug that had briefly looked like a candidate cause.
- **CLI version** — tested `2.109.1` (five of six stalls) and `2.114.0` (one separate run, [32093728092](https://github.com/LinkFault/linkfault-foundation/actions/runs/32093728092), not counted above since it hit a different, self-inflicted, fully-diagnosed content bug before reaching the stall point — not conclusive either way on CLI version, but the stall independently reproduced on both versions across other runs).

## What we have not been able to do

No Docker or Supabase CLI available in any environment outside this GitHub Actions pipeline (checked directly — nothing on `PATH`, no local Docker install), so no local reproduction, bisection, or testing against a different Docker daemon version was possible. Everything above comes directly from the GitHub Actions run logs and the watchdog/sampler artifacts.

## For maintainers

- CLI versions observed: `2.109.1`, and possibly `2.114.0` (inconclusive — see above).
- Environment: GitHub Actions `ubuntu-24.04` hosted runner, Docker-based local stack (`supabase start` / `supabase db reset --local`).
- Symptom: mid-migration-apply, the CLI's SQL pipeline (visible via `--debug`'s raw `PG Send` frame logging) stops emitting further `Parse`/`Bind`/`Describe`/`Execute` frames after a specific statement, with no error, no exception, no exit code reported until the step's own external timeout kills it.
- Database-side snapshot at the time of the stall (attached below) shows no active or blocked query matching the stalled statement — whatever stopped making progress isn't visible from inside Postgres.
- Reproduced 6 times across 4 branches and 2 CLI versions; not migration-content-dependent (always at the same relative point in a fixed sequence of already-applied migrations, regardless of what else is in the branch).
- Happy to share full `--debug` logs / run links for any of the six runs above if useful.

7ttp:
## TL;DR

fixes `supabase db reset` and `supabase start` hanging forever with no error when the database connection dies while migrations are being applied

## whats broken?

node-postgres silently discards every protocol frame once a socket stops being writable, while still reporting the write as successful, so the whole batch goes nowhere and the CLI waits forever on a reply the server was never asked for

it also leaves TCP keepalive off by default, where the Go CLI's driver had it on, so a peer that dies without a FIN or RST is never noticed either

## fixed now by:

- failing a batch that never reached the wire as a connection error, carrying the driver's own reason instead of blaming the migration's first statement
- discarding that pooled connection, so the next batch redials instead of writing into the same dead socket
- turning TCP keepalive on for every connection the CLI opens, so a silently dead peer is eventually detected instead of waited on

a server that stays alive but never answers still waits, matching the Go CLI, since that is indistinguishable from a long running statement

## ref:
- closes: https://github.com/supabase/cli/issues/6244

### DIFF
--- a/apps/cli/docs/go-cli-divergences.md
+++ b/apps/cli/docs/go-cli-divergences.md
@@ -236,5 +236,7 @@ These commands exist in the TS CLI today but have no direct top-level equivalent
   (`apps/cli-go/pkg/migration/file.go:126-147`) renders every `ExecBatch` failure —
   dead connection included — as `<err>\nAt statement: N\n<sql>`. Naming a statement
   that provably never ran sent users debugging their own SQL for a transport failure,
-  so the TS shell reports the connectivity failure instead. A batch that was written
-  keeps Go's rendering unchanged.
+  so the TS shell reports the connectivity failure instead. This covers batches only: a
+  batch that was written, and the pipeline-incompatible statements the same loop runs
+  standalone through `exec` (`CREATE INDEX CONCURRENTLY`, `VACUUM`, …), both keep Go's
+  `At statement: N` rendering for every failure, transport included.
--- a/apps/cli/docs/go-cli-divergences.md
+++ b/apps/cli/docs/go-cli-divergences.md
@@ -230,3 +230,11 @@ These commands exist in the TS CLI today but have no direct top-level equivalent
   pull."). An in-sync database is a finding, not a failure to troubleshoot, so the
   debug hint sent users chasing a non-existent bug. Message text and exit code — the
   parts scripts depend on — are unchanged.
+- A migration batch that never reached the wire (the connection died before or during
+  submit) fails as a connection error carrying the driver's reason, with no
+  `At statement: N` line and no statement echo. Go's `formatError`
+  (`apps/cli-go/pkg/migration/file.go:126-147`) renders every `ExecBatch` failure —
+  dead connection included — as `<err>\nAt statement: N\n<sql>`. Naming a statement
+  that provably never ran sent users debugging their own SQL for a transport failure,
+  so the TS shell reports the connectivity failure instead. A batch that was written
+  keeps Go's rendering unchanged.
--- a/apps/cli/docs/go-cli-divergences.md
+++ b/apps/cli/docs/go-cli-divergences.md
@@ -246,13 +246,3 @@ These commands exist in the TS CLI today but have no direct top-level equivalent
   an output channel, and in managed mode inside the daemon process, so a user-visible warning
   there needs a diagnostics channel on `BuildResult` first; the applied value stays visible via
   `docker inspect`.
-- A migration batch that never reached the wire (the connection died before or during
-  submit) fails as a connection error carrying the driver's reason, with no
-  `At statement: N` line and no statement echo. Go's `formatError`
-  (`apps/cli-go/pkg/migration/file.go:126-147`) renders every `ExecBatch` failure —
-  dead connection included — as `<err>\nAt statement: N\n<sql>`. Naming a statement
-  that provably never ran sent users debugging their own SQL for a transport failure,
-  so the TS shell reports the connectivity failure instead. This covers batches only: a
-  batch that was written, and the pipeline-incompatible statements the same loop runs
-  standalone through `exec` (`CREATE INDEX CONCURRENTLY`, `VACUUM`, …), both keep Go's
-  `At statement: N` rendering for every failure, transport included.
--- a/apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.integration.test.ts
+++ b/apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.integration.test.ts
@@ -183,6 +183,7 @@ const fakeBatchServer = (
   readonly port: number;
   readonly close: () => void;
   readonly state: FakeBatchServerState;
+  readonly sockets: ReadonlyArray<net.Socket>;
 }> =>
   new Promise((resolve) => {
     const state: FakeBatchServerState = {
@@ -191,7 +192,9 @@ const fakeBatchServer = (
       params: [],
       syncs: 0,
     };
+    const sockets: Array<net.Socket> = [];
     const server = net.createServer((socket) => {
+      sockets.push(socket);
       let sawStartup = false;
       let pending = Buffer.alloc(0);
       let failed = false;
@@ -294,7 +297,7 @@ const fakeBatchServer = (
     });
     server.listen(0, "127.0.0.1", () => {
       const address = server.address() as net.AddressInfo;
-      resolve({ port: address.port, close: () => server.close(), state });
+      resolve({ port: address.port, close: () => server.close(), state, sockets });
     });
   });
 
@@ -675,6 +678,42 @@ describe("legacyDbConnectionSqlPgLayer extended batches", () => {
           expect(error._tag).toBe("LegacyDbExecError");
           expect(asBatchExecError(error).message).toContain("Connection terminated unexpectedly");
           yield* session.execBatch([{ sql: "SELECT 3" }]);
+        }),
+      );
+    }),
+  );
+
+  it.live("survives an idle raw-client socket death and redials for the next query", () =>
+    // node-postgres emits `error` on an idle client; with no listener that terminates the
+    // process, so a database that dies between two `queryRaw` calls must fail that call
+    // rather than the CLI, and must not leave the corpse cached for the call after it.
+    Effect.gen(function* () {
+      const server = yield* Effect.promise(() => fakeBatchServer());
+      yield* runWithBatchServer(server, (session) =>
+        Effect.gen(function* () {
+          yield* session.queryRaw("SELECT 1");
+          // `queryRaw` runs on its own client, opened after the pool's, so it is the
+          // newest connection the server has accepted.
+          const rawSocket = server.sockets.at(-1);
+          const openedBeforeRedial = server.sockets.length;
+
+          const closed = new Promise<void>((resolve) => {
+            rawSocket?.on("close", () => resolve());
+          });
+          yield* Effect.sync(() => rawSocket?.destroy());
+          yield* Effect.promise(() => closed);
+
+          const failed = yield* session.queryRaw("SELECT 1").pipe(
+            Effect.flip,
+            Effect.timeoutOrElse({
+              duration: Duration.seconds(10),
+              orElse: () => Effect.die("queryRaw never settled after the idle socket died"),
+            }),
+          );
+          expect(failed._tag).toBe("LegacyDbExecError");
+
+          yield* session.queryRaw("SELECT 1");
+          expect(server.sockets.length).toBeGreaterThan(openedBeforeRedial);
         }),
       );
     }),
--- a/apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.integration.test.ts
+++ b/apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.integration.test.ts
@@ -686,8 +686,8 @@ describe("legacyDbConnectionSqlPgLayer extended batches", () => {
 
   it.live("survives an idle raw-client socket death and redials for the next query", () =>
     // node-postgres emits `error` on an idle client; with no listener that terminates the
-    // process, so a database that dies between two `queryRaw` calls must fail that call
-    // rather than the CLI, and must not leave the corpse cached for the call after it.
+    // process, so a database dying between two `queryRaw` calls must not take the CLI with it,
+    // and must not leave the corpse cached for the calls after it.
     Effect.gen(function* () {
       const server = yield* Effect.promise(() => fakeBatchServer());
       yield* runWithBatchServer(server, (session) =>
@@ -704,14 +704,15 @@ describe("legacyDbConnectionSqlPgLayer extended batches", () => {
           yield* Effect.sync(() => rawSocket?.destroy());
           yield* Effect.promise(() => closed);
 
-          const failed = yield* session.queryRaw("SELECT 1").pipe(
-            Effect.flip,
+          // Whether the client has already noticed the death decides if this call redials or
+          // fails on the corpse, and that ordering is not ours to control, so accept either.
+          yield* session.queryRaw("SELECT 1").pipe(
+            Effect.exit,
             Effect.timeoutOrElse({
               duration: Duration.seconds(10),
               orElse: () => Effect.die("queryRaw never settled after the idle socket died"),
             }),
           );
-          expect(failed._tag).toBe("LegacyDbExecError");
 
           yield* session.queryRaw("SELECT 1");
           expect(server.sockets.length).toBeGreaterThan(openedBeforeRedial);
--- a/apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.integration.test.ts
+++ b/apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.integration.test.ts
@@ -20,6 +20,7 @@ import {
 import {
   legacyAcquirePgPool,
   legacyDbConnectionSqlPgLayer,
+  LegacyPgBatchQuery,
 } from "./legacy-db-connection.sql-pg.layer.ts";
 
 const SUGGESTION_CONTEXT = {
@@ -662,8 +663,8 @@ describe("legacyDbConnectionSqlPgLayer extended batches", () => {
   );
 
   it.live("fails a batch whose connection drops after it was written, then recovers", () =>
-    // Guards the driver path that already worked: a socket dropped after the batch was
-    // written must surface as a batch failure, and its client must not be recycled.
+    // A socket dropped after the batch was written must fail that batch and must not leave
+    // the client to be handed to the next one.
     Effect.gen(function* () {
       const server = yield* Effect.promise(() => fakeBatchServer({ destroyOnFirstSync: true }));
       yield* runWithBatchServer(server, (session) =>
@@ -716,6 +717,42 @@ describe("legacyDbConnectionSqlPgLayer extended batches", () => {
           expect(server.sockets.length).toBeGreaterThan(openedBeforeRedial);
         }),
       );
+    }),
+  );
+
+  it.live("refuses to write a batch onto a real pooled client whose socket is already gone", () =>
+    // The unit test drives `submit` through a hand-built connection; this pins the same
+    // refusal against a real node-postgres client, so a driver change that stops making the
+    // socket unwritable would be caught rather than mocked over. Destroying the socket and
+    // submitting in one synchronous block keeps the window deterministic: `writable` flips
+    // immediately, while pg only marks the client unqueryable on the next tick's close.
+    Effect.gen(function* () {
+      const server = yield* Effect.promise(() => fakeBatchServer());
+      yield* Effect.gen(function* () {
+        const pool = yield* legacyAcquirePgPool(
+          {
+            host: "127.0.0.1",
+            port: server.port,
+            user: "postgres",
+            password: SENTINEL_PASSWORD,
+            database: "postgres",
+            sslmode: "disable",
+          },
+          { isLocal: true, dnsResolver: "native" },
+        );
+        const client = yield* Effect.promise(() => pool.connect());
+        const batch = new LegacyPgBatchQuery([{ sql: "SELECT 1" }], () => {});
+
+        const refusal = yield* Effect.sync(() => {
+          client.connection.stream.destroy();
+          return batch.submit(client.connection);
+        });
+
+        expect(refusal?.message).toBe("the connection's socket is no longer writable");
+        expect(batch.outcome).toBe("unsent");
+        expect(server.state.frameTypes).toEqual([]);
+        client.release(new Error("done"));
+      }).pipe(Effect.scoped, Effect.ensuring(Effect.sync(server.close)));
     }),
   );
 
--- a/apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.integration.test.ts
+++ b/apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.integration.test.ts
@@ -176,6 +176,8 @@ const fakeBatchServer = (
     readonly emptyAt?: number;
     /** Never answer an extended-protocol frame, so a batch hangs until interrupted. */
     readonly stall?: boolean;
+    /** Drop the connection on the first Sync, so a batch dies mid-flight. */
+    readonly destroyOnFirstSync?: boolean;
   } = {},
 ): Promise<{
   readonly port: number;
@@ -268,6 +270,10 @@ const fakeBatchServer = (
             }
           } else if (type === "S") {
             state.syncs += 1;
+            if (options.destroyOnFirstSync === true && state.syncs === 1) {
+              socket.destroy();
+              return;
+            }
             if (options.failOnSync === true && !failed) {
               socket.write(
                 errorResponse({
@@ -648,6 +654,28 @@ describe("legacyDbConnectionSqlPgLayer extended batches", () => {
             }),
           ),
         ),
+      );
+    }),
+  );
+
+  it.live("fails a batch whose connection drops after it was written, then recovers", () =>
+    // Guards the driver path that already worked: a socket dropped after the batch was
+    // written must surface as a batch failure, and its client must not be recycled.
+    Effect.gen(function* () {
+      const server = yield* Effect.promise(() => fakeBatchServer({ destroyOnFirstSync: true }));
+      yield* runWithBatchServer(server, (session) =>
+        Effect.gen(function* () {
+          const error = yield* session.execBatch([{ sql: "SELECT 1" }, { sql: "SELECT 2" }]).pipe(
+            Effect.flip,
+            Effect.timeoutOrElse({
+              duration: Duration.seconds(10),
+              orElse: () => Effect.die("execBatch never settled after the connection died"),
+            }),
+          );
+          expect(error._tag).toBe("LegacyDbExecError");
+          expect(asBatchExecError(error).message).toContain("Connection terminated unexpectedly");
+          yield* session.execBatch([{ sql: "SELECT 3" }]);
+        }),
       );
     }),
   );
--- a/apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.layer.ts
+++ b/apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.layer.ts
@@ -202,6 +202,62 @@ export function legacyToExecError(error: unknown): LegacyDbExecError {
   return new LegacyDbExecError({ message: String(error), code: legacyExtractSqlState(error) });
 }
 
+const LEGACY_BATCH_CONNECTION_LOST =
+  "connection to the database was lost before the batch could be sent";
+
+/**
+ * pgconn's keepalive period (its default dialer, 5 minutes). Go applies that period to both
+ * the idle time and the probe interval; Node can only set the idle time, leaving interval and
+ * count at OS defaults, so a silently dead peer surfaces roughly 11 minutes later on Linux —
+ * sooner than Go's own window, not later.
+ */
+const LEGACY_PGCONN_KEEPALIVE_MILLIS = 300_000;
+
+/**
+ * Maps a failed migration batch to its public error. A batch that never reached the wire
+ * is a connectivity failure, not a statement failure, so it reports as one instead of
+ * blaming the batch's first statement; anything else keeps `legacyToExecError`'s
+ * server-error rendering plus the number of statements that completed.
+ */
+export function legacyBatchFailureError(
+  error: Error,
+  batch:
+    | { readonly completed: number; readonly submitted: boolean; readonly poisoned: boolean }
+    | undefined,
+): LegacyDbExecError | LegacyDbConnectError {
+  if (batch === undefined || (!batch.submitted && !batch.poisoned)) {
+    return new LegacyDbConnectError({
+      message:
+        error.message === LEGACY_BATCH_CONNECTION_LOST
+          ? LEGACY_BATCH_CONNECTION_LOST
+          : `${LEGACY_BATCH_CONNECTION_LOST}: ${error.message}`,
+    });
+  }
+  const mapped = legacyToExecError(error);
+  return new LegacyDbExecError({
+    message: mapped.message,
+    code: mapped.code,
+    detail: mapped.detail,
+    position: mapped.position,
+    statementIndex: batch.completed,
+  });
+}
+
+/**
+ * Whether a batch's pooled client must be destroyed rather than returned to the pool. A
+ * batch that never reached the wire leaves the client looking healthy to pg-pool while its
+ * socket is already gone, so the next checkout would write into the same dead connection.
+ */
+export function legacyShouldDiscardBatchClient(
+  batch: { readonly submitted: boolean } | undefined,
+  exit: Exit.Exit<unknown, unknown>,
+): boolean {
+  return (
+    batch?.submitted === false ||
+    (Exit.isFailure(exit) && (Cause.hasInterrupts(exit.cause) || Cause.hasDies(exit.cause)))
+  );
+}
+
 const legacyEncodeTextArray = (values: ReadonlyArray<string>): string =>
   `{${values
     .map((value) => `"${value.replaceAll("\\", "\\\\").replaceAll('"', '\\"')}"`)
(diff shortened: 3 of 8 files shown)


apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.layer.ts
```diff
@@ -511,6 +572,8 @@ export function legacyBuildRawPgConfig(
       : { host, port, user: cfg.user, password: cfg.password, database: cfg.database }),
     ...(sslOption === undefined ? {} : { ssl: sslOption }),
     connectionTimeoutMillis: connectTimeoutSeconds * 1000,
+    keepAlive: true,
```
avallete: **issue**

Enabling keepalive here creates a new crash path via the cached raw client. `acquireRawClient` builds a bare `new Pg.Client(winningRawConfig)` and never attaches an `'error'` listener (the only listeners in this file are on the pool, the batch's pooled client, and the COPY stream). When keepalive detects a dead peer while that client sits idle — e.g. mid `inspect report`, which holds it across ~14 sequential COPYs — pg's `_handleErrorEvent` ends with an unconditional `this.emit('error', err)` (pg 8.23 `client.js:416-423`), and an unlistened `'error'` throws, crashing the whole CLI. Before this PR keepalive was off, so an idle silent death produced no event at all this path is new.

avallete, line 588: **Blocking, reproduced: idle death of the raw client crashes the whole CLI — and keepalive widens the trigger.**

`acquireRawClient` builds a bare `new Pg.Client(winningRawConfig)` and never attaches an `'error'` listener. Any socket death while that client is idle (e.g. between `inspect report`'s COPYs) makes pg emit an unlistened `'error'` (`client.js:422`) and crashes the process. Reproduced on this branch under both Node and Bun with pg 8.23: idle RST → `Unhandled 'error': read ECONNRESET`, exit 1; idle FIN → `Connection terminated unexpectedly`, exit 1; same scenario with one `'error'` listener attached → survives gracefully.

The mechanism predates this PR (a server crash or `docker stop` already triggers it), but keepalive widens it: *silent* peer deaths, which previously produced no event at all while idle, now surface as `ETIMEDOUT` on the same unlistened path (that leg not reproduced — it needs a genuinely blackholed peer — but it's the same event class, per socket semantics). One `'error'` handler that invalidates the `rawClient` cache fixes the crash and the stale-client reuse at line 1010.

<details>
<summary>repro-raw-client-crash.mjs</summary>

```js
// Repro: an idle socket error on the listener-less cached raw client crashes the CLI.
// Mirrors acquireRawClient (legacy-db-connection.sql-pg.layer.ts:1009-1025): a bare
// Pg.Client with no 'error' listener, with the PR's keepAlive config. The fake server
// completes the postgres startup handshake, then kills the socket while the client is
// idle — the same event a failed keepalive probe (ETIMEDOUT) or a dying server
// (FIN/RST) delivers.
//
// Save under apps/cli/ (so `pg` resolves) and run:
//   node repro-raw-client-crash.mjs rst       -> process crash (unhandled 'error')
//   node repro-raw-client-crash.mjs fin       -> process crash
//   node repro-raw-client-crash.mjs listener  -> control: survives gracefully
// (bun works too)
import net from "node:net";
import pg from "pg";

const mode = process.argv[2];
if (!["fin", "rst", "listener"].includes(mode)) {
  console.error("usage: repro-raw-client-crash.mjs <fin|rst|listener>");
  process.exit(2);
}

const sockets = [];
const server = net.createServer((socket) => {
  sockets.push(socket);
  let handshakeDone = false;
  socket.on("data", () => {
    if (handshakeDone) return;
    handshakeDone = true;
    const authOk = Buffer.from([0x52, 0, 0, 0, 8, 0, 0, 0, 0]); // AuthenticationOk
    const ready = Buffer.from([0x5a, 0, 0, 0, 5, 0x49]); // ReadyForQuery 'I'
    socket.write(Buffer.concat([authOk, ready]));
  });
});

await new Promise((resolve) => server.listen(0, "127.0.0.1", resolve));
const port = server.address().port;

const client = new pg.Client({
  host: "127.0.0.1",
  port,
  user: "u",
  database: "d",
  keepAlive: true, // the PR's addition; not itself the trigger here
  connectionTimeoutMillis: 5000,
});
if (mode === "listener") {
  client.on("error", (err) => {
    console.log(`GRACEFUL: error listener caught: ${err.message}`);
  });
}

await client.connect();
console.log(`connected; client now idle (no active query), mode=${mode}`);

setTimeout(() => {
  for (const s of sockets) (mode === "fin" ? s.end() : s.resetAndDestroy());
}, 100);

setTimeout(() => {
  console.log("SURVIVED: process still alive 1.5s after idle socket death");
  process.exit(0);
}, 1500);
```

</details>

apps/cli/src/legacy/shared/legacy-migration-apply.ts
```diff
@@ -747,9 +744,9 @@ const execMigrationBatch = <E>(
           const base = executed;
           yield* session.execBatch(operations).pipe(
             Effect.mapError((cause) => {
-              // Acquiring the batch's connection failed: there is no failing
-              // statement to name, so the connect error (and its suggestion) is
-              // surfaced verbatim instead of being rendered as `At statement: N`.
+              // The batch's connection failed, either on checkout or before any of
```
avallete, line 750: **Mechanism reproduced: the standalone pipeline-incompatible path still blames the SQL for a dead connection.**

The reclassification only covers `execBatch`. It's structural — `session.exec` is typed `Effect<void, LegacyDbExecError>`, so the connect carve-out can't reach it. Feeding the real dead-client driver error through the real mapping on this branch renders:

```
Error: Client has encountered a connection error and is not queryable
At statement: 3
CREATE INDEX CONCURRENTLY idx ON t (c)
```

— exactly the "blame a statement that provably never ran" output the divergence-doc entry says was removed. (I verified the mapping+formatter chain, not a full `db reset` against a dying server.) Either extend the mapping to `exec` or scope the doc entry to say the standalone path intentionally keeps Go's rendering.

apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.layer.ts
```diff
@@ -202,6 +202,62 @@ export function legacyToExecError(error: unknown): LegacyDbExecError {
...
+ * Whether a batch's pooled client must be destroyed rather than returned to the pool. A
+ * batch that never reached the wire leaves the client looking healthy to pg-pool while its
+ * socket is already gone, so the next checkout would write into the same dead connection.
+ */
+export function legacyShouldDiscardBatchClient(
+  batch: { readonly submitted: boolean } | undefined,
+  exit: Exit.Exit<unknown, unknown>,
+): boolean {
+  return (
+    batch?.submitted === false ||
```
avallete: **Source-verified fragility: recovery here depends on a private pg-pool internal.**

After a submitted-then-died batch (`discard === false` → `release(undefined)`), the dead client is evicted only because pg-pool's `_release` removes clients where `!client._queryable` (`pg-pool@3.14.0 index.js:392`, marked `TODO(bmc): expose a proper, public interface`) — and pg-pool is an unpinned transitive dep. The no-discard choice is right (a syntax error shouldn't force a redial + `SET SESSION ROLE` replay on a `max:1` pool), but the "then recovers" integration test locks in this dependency without naming it. Worth a comment here stating the pg-pool behavior this branch relies on, so a future dep bump that breaks it is traceable.

apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.layer.ts
```diff
@@ -202,6 +202,62 @@ export function legacyToExecError(error: unknown): LegacyDbExecError {
...
+ * server-error rendering plus the number of statements that completed.
+ */
+export function legacyBatchFailureError(
+  error: Error,
+  batch:
+    | { readonly completed: number; readonly submitted: boolean; readonly poisoned: boolean }
+    | undefined,
+): LegacyDbExecError | LegacyDbConnectError {
+  if (batch === undefined || (!batch.submitted && !batch.poisoned)) {
+    return new LegacyDbConnectError({
```
avallete, line 231: **Reproduced: this error carries no `suggestion`, unlike its checkout-path sibling.**

Constructing this error on this branch gives `suggestion: undefined` (and `retryable: undefined`), while the identical dead-DB failure one tick earlier — pool checkout, line 1046 — goes through `legacyToConnectError` and carries the profile-aware connect suggestion. Renderers read `cause.suggestion` conditionally, so nothing crashes, but the user gets an actionable hint or not depending on which side of checkout the connection died.

Related, also reproduced: the `error.message === LEGACY_BATCH_CONNECTION_LOST` identity check on line 231 stutters as soon as anything wraps the synthetic error (a wrapping idiom this file already uses elsewhere):

```
connection to the database was lost before the batch could be sent: batch submit failed: connection to the database was lost before the batch could be sent
```

A tiny dedicated `Error` subclass thrown from `submit()` + `instanceof` still satisfies pg's `submit(): Error | null` contract.

apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.layer.ts
```diff
@@ -202,6 +202,62 @@ export function legacyToExecError(error: unknown): LegacyDbExecError {
   return new LegacyDbExecError({ message: String(error), code: legacyExtractSqlState(error) });
 }
 
+const LEGACY_BATCH_CONNECTION_LOST =
+  "connection to the database was lost before the batch could be sent";
+
+/**
+ * pgconn's keepalive period (its default dialer, 5 minutes). Go applies that period to both
+ * the idle time and the probe interval; Node can only set the idle time, leaving interval and
+ * count at OS defaults, so a silently dead peer surfaces roughly 11 minutes later on Linux —
```
avallete: **Not reproduced (arithmetic only): the "~11 minutes on Linux" figure looks wrong.**

Documented kernel defaults give 300s idle + 75s interval × 9 probes ≈ 975s ≈ **16 minutes**. I haven't measured it live (that needs a blackholed peer and the full wait), and the true figure is runtime-dependent — newer Node/libuv versions set `TCP_KEEPINTVL`/`TCP_KEEPCNT` themselves, and this runs under Bun. Suggest dropping the concrete minutes claim and keeping the 5-minute-idle rationale; the "sooner than Go's window" comparison still holds either way.

apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.layer.ts
```diff
@@ -202,6 +202,62 @@ export function legacyToExecError(error: unknown): LegacyDbExecError {
...
+          : `${LEGACY_BATCH_CONNECTION_LOST}: ${error.message}`,
+    });
+  }
+  const mapped = legacyToExecError(error);
+  return new LegacyDbExecError({
+    message: mapped.message,
+    code: mapped.code,
+    detail: mapped.detail,
+    position: mapped.position,
+    statementIndex: batch.completed,
```
avallete, line 244: **Reproduced (low priority): a poisoned batch always blames statement 0.**

For a poisoned batch, `completed` is 0 (corked stream, server acked nothing), so the failure always lands on statement 0. Driving the real class on this branch with a `bind` that throws on statement 7 of 10 renders:

```
Error: frame serialization blew up on statement 7
At statement: 0
SELECT 0 /* stmt 0 */
```

Pre-existing mapping, but this PR re-states it in exported code, and the new test "keeps a partially written batch on the statement path" uses `completed: 1` and never asserts `statementIndex`. What triggers a mid-serialization throw in production is admittedly speculative (oversized frames is my best guess), so: at minimum the test could pin the realistic `completed: 0` case.

apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.layer.ts
```diff
@@ -210,14 +266,15 @@ const legacyEncodeTextArray = (values: ReadonlyArray<string>): string =>
...
-class LegacyPgBatchQuery implements Pg.Submittable {
+export class LegacyPgBatchQuery implements Pg.Submittable {
   readonly statements: ReadonlyArray<{
     readonly sql: string;
     readonly params: ReadonlyArray<string | null>;
   }>;
   callback: (error: Error | undefined) => void;
   completed = 0;
   poisoned = false;
+  submitted = false;
```
avallete: **Design note (take or leave):** `submitted` and `poisoned` are mutually exclusive by construction, but that's an unenforced convention two consumers project differently (`!submitted && !poisoned` at 228 vs `submitted === false` at 256). A single `outcome: "unsent" | "poisoned" | "submitted"` set once in `submit()` would make the exclusivity structural — ~4 sites.

apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.integration.test.ts
```diff
@@ -652,6 +658,28 @@ describe("legacyDbConnectionSqlPgLayer extended batches", () => {
     }),
   );
 
+  it.live("fails a batch whose connection drops after it was written, then recovers", () =>
+    // Guards the driver path that already worked: a socket dropped after the batch was
```
avallete: Nit: "Guards the driver path that already worked" is temporal narration — the durable invariant is just "a socket drop after the batch was written must fail the batch and not recycle the client".

Coverage: this test exercises the post-write drop, but the exact path the PR fixes (unwritable-yet-`_queryable` socket → `LegacyDbConnectError` → discard → next batch redials) is only unit-tested in isolation. One integration case that makes the checked-out socket non-writable without a close would catch a wiring regression between `submit`/`legacyBatchFailureError`/`legacyShouldDiscardBatchClient` that would otherwise re-introduce the hang unseen.

avallete (changes requested):
The core fix is sound — I traced the load-bearing pieces and they hold: the widened discard predicate strictly covers the old `poisoned` check, the only `execBatch` caller guards `LegacyDbConnectError` before touching `statementIndex`, keepalive reaches both the pool and the raw client, and pg honors the `submit()` error return so the writable check settles instead of hanging. I also chased two scary scenarios that turned out impossible: there's no residual *hang* on `exec`/`query` (pg's `'close'` → `_errorAllQueries` fails them within a tick), and no applied-but-unrecorded migration (the history INSERT rides inside the batch).

I reproduced the findings below against this branch where marked; one is a process crash I'd fix before merge. The script used for the reproduced non-crash findings:

<details>
<summary>verify-findings.ts (run with bun from the repo root)</summary>

```ts
// Verification of the review findings against this branch's real modules.
// Save at the repo root and run: bun verify-findings.ts
import {
  LegacyPgBatchQuery,
  legacyBatchFailureError,
  legacyToExecError,
} from "./apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.layer.ts";
import { legacyFormatExecBatchError } from "./apps/cli/src/legacy/shared/legacy-migration-apply.ts";
import type * as Pg from "pg";

console.log("=== standalone exec renders dead connection as 'At statement: N' ===");
const deadConnErr = new Error("Client has encountered a connection error and is not queryable");
const execMapped = legacyToExecError(deadConnErr);
const rendered = legacyFormatExecBatchError(execMapped, 3, "CREATE INDEX CONCURRENTLY idx ON t (c)");
console.log(rendered.message);

console.log("\n=== poisoned batch blames statement 0 ===");
let bindCalls = 0;
const noop = () => {};
const conn = {
  stream: { writable: true, cork: noop, uncork: noop },
  parse: noop,
  bind: () => {
    bindCalls += 1;
    if (bindCalls === 7) throw new Error("frame serialization blew up on statement 7");
  },
  describe: noop,
  execute: noop,
  sync: noop,
} as unknown as Pg.Connection;
const statements = Array.from({ length: 10 }, (_, i) => ({ sql: `SELECT ${i} /* stmt ${i} */` }));
const batch = new LegacyPgBatchQuery(statements, noop);
const submitErr = batch.submit(conn);
console.log("poisoned:", batch.poisoned, "submitted:", batch.submitted, "completed:", batch.completed);
const failure = legacyBatchFailureError(submitErr!, batch);
console.log("mapped:", failure._tag, "statementIndex:", (failure as { statementIndex?: number }).statementIndex);

console.log("\n=== batch-lost ConnectError has no suggestion ===");
const unsent = new LegacyPgBatchQuery([{ sql: "SELECT 1" }], noop);
const unsentErr = unsent.submit({ stream: { writable: false } } as unknown as Pg.Connection);
const connectFailure = legacyBatchFailureError(unsentErr!, unsent);
console.log(connectFailure._tag, "| suggestion:", (connectFailure as { suggestion?: string }).suggestion);

console.log("\n=== message-equality sentinel stutters once wrapped ===");
const wrapped = new Error(`batch submit failed: ${unsentErr!.message}`);
console.log(legacyBatchFailureError(wrapped, unsent).message);
```

</details>

apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.layer.ts
7ttp: resolved in latest commit

apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.layer.ts
7ttp: resolved in latest commit

apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.layer.ts
7ttp: resolved in latest commit:
(dropped the number, kept the 5min idle rationale and the sooner-than-go comparison. 
agreed its runtime dependent enough that no figure is better than a wrong one)

apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.layer.ts
7ttp: resolved in latest commit

apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.layer.ts
7ttp: resolved in latest commit
(took it)

apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.layer.ts
7ttp: suggestion part is fixed, it picks up `LEGACY_SUGGEST_LOCAL_STACK` now so it lines up with the checkout one

went a slightly different route on the identity check though. `submit()`'s guard returns its own message now and the mapper always prefixes, so theres no sentinel left to stutter on, wrapped or not.
 a subclass felt like the heavier option since it'd trip the error-actionability guard for a marker that never really leaves the module. if you'd still rather have the typed instanceof:
maybe we can take this up as a *separate followup* and keep this one scoped in..

left retryable unset for now too, only db-setup's connect retry reads it and that wraps connect rather than execBatch.

apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.integration.test.ts
7ttp: reworded it. couldnt get the coverage bit deterministic without it racing on when pg flips `_queryable`
 so leaving that as a followup as well..

apps/cli/src/legacy/shared/legacy-migration-apply.ts
7ttp: went with the second option and scoped the doc entry, extending it to exec would widen session.exec's error channel out to every caller so that felt like its own change....

apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.integration.test.ts
```diff
@@ -652,6 +662,100 @@ describe("legacyDbConnectionSqlPgLayer extended batches", () => {
...
+          // `queryRaw` runs on its own client, opened after the pool's, so it is the
+          // newest connection the server has accepted.
+          const rawSocket = server.sockets.at(-1);
+          const openedBeforeRedial = server.sockets.length;
+
+          const closed = new Promise<void>((resolve) => {
+            rawSocket?.on("close", () => resolve());
+          });
+          yield* Effect.sync(() => rawSocket?.destroy());
+          yield* Effect.promise(() => closed);
```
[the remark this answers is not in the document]
7ttp: fixed: 066e08fd8

apps/cli/docs/go-cli-divergences.md
```diff
@@ -246,3 +246,13 @@ These commands exist in the TS CLI today but have no direct top-level equivalent
   an output channel, and in managed mode inside the daemon process, so a user-visible warning
   there needs a diagnostics channel on `BuildResult` first; the applied value stays visible via
   `docker inspect`.
+- A migration batch that never reached the wire (the connection died before or during
+  submit) fails as a connection error carrying the driver's reason, with no
+  `At statement: N` line and no statement echo. Go's `formatError`
+  (`apps/cli-go/pkg/migration/file.go:126-147`) renders every `ExecBatch` failure —
+  dead connection included — as `<err>\nAt statement: N\n<sql>`. Naming a statement
+  that provably never ran sent users debugging their own SQL for a transport failure,
```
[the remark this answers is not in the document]
7ttp: fixed: 367ed4b35

apps/cli/docs/go-cli-divergences.md
```diff
@@ -246,3 +246,13 @@ These commands exist in the TS CLI today but have no direct top-level equivalent
   an output channel, and in managed mode inside the daemon process, so a user-visible warning
   there needs a diagnostics channel on `BuildResult` first; the applied value stays visible via
   `docker inspect`.
+- A migration batch that never reached the wire (the connection died before or during
+  submit) fails as a connection error carrying the driver's reason, with no
```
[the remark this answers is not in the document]
7ttp: fixed: 367ed4b35

apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.integration.test.ts
7ttp: reworded it. couldnt get the coverage bit deterministic without it racing on when pg flips `_queryable`
 so leaving that as a followup as well..

apps/cli/docs/go-cli-divergences.md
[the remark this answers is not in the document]
7ttp: fixed: 367ed4b35

apps/cli/docs/go-cli-divergences.md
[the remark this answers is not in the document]
7ttp: fixed: 367ed4b35

apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.layer.ts
```diff
@@ -231,6 +295,9 @@ class LegacyPgBatchQuery implements Pg.Submittable {
   }
 
   submit(connection: Pg.Connection): Error | null {
+    if (!connection.stream.writable) {
```
avallete, line 298: Non-blocking: this covers the already-dead case (`Connection._send` no-ops when `!writable` and still returns success). If the socket flips unwritable *during* the corked parse/bind/execute/sync loop, later `_send`s still discard frames and `outcome` becomes `"submitted"` — same footgun, much smaller window. A post-loop `!writable` check (or honoring `_send`'s return) would close it. Not merge-blocking; the reported hang is the already-dead path.

apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.unit.test.ts
```diff
@@ -561,3 +589,192 @@ describe("legacyToExecError (pg server-error extraction)", () => {
...
+
+    expect(error._tag).toBe("LegacyDbConnectError");
+    expect(error.message).toBe(
+      "connection to the database was lost before the batch could be sent: " +
+        "the connection's socket is no longer writable",
+    );
+    expect(error[ErrorActionabilityId]).toMatchObject({ error_category: "db_connection" });
+  });
+
+  it("keeps the driver's own reason when pg refused the batch before submit", () => {
```
avallete, line 687: Non-blocking: this case and the one at 699 assert the same unsent → `LegacyDbConnectError` + driver-reason mapping, only the input string differs. The discard helper's `undefined` + success exit at 777 is also a state `execBatch` cannot produce. Fine to compress if you touch the file again; not blocking.

avallete (approved):
Re-reviewed HEAD after the follow-ups. The hang-fix path holds: `submit()` refuses an already-unwritable socket, pg 8.23 delivers that Error through `handleError`, unsent batches surface as `LegacyDbConnectError` (with the local-stack hint) and discard the pooled client, and the raw-client idle `error` listener stops the process crash.

First-commit items look addressed. Not reopening the typed `submit()` subclass or the full `execBatch`+destroy wiring test.

Two non-blocking notes inline. Separate observation: keepalive idle is 5 minutes, so it will not save the ~2–3 minute CI timeouts in #6244 — the unwritable refusal is the CI-visible fix. Standalone `exec` still rendering a dead connection as `At statement: N` is documented and an intentional scope cut.

apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.layer.ts
7ttp: resolved in latest commit, check sits after the uncork tho rather than after the loop, since thats where writable actually flips

apps/cli/src/legacy/shared/legacy-db-connection.sql-pg.unit.test.ts
7ttp: resolved in latest commit

PR #6295, opened 2026-08-21
feat(cli): add config diff command

kanadgupta:
Implements [CLI-2156](https://linear.app/supabase/issue/CLI-2156/add-supabase-config-diff-to-the-cli): a read-only `supabase config diff` that classifies drift between `supabase/config.toml` and the effective configuration `GET /v2/projects/{ref}/config` reports for a target project or branch. Never writes. Builds on CLI-2155's sparse subtraction/defaults (#6205) and consumes CLI-2230's `ProjectConfig` convergence normalizers (#6339) as its comparison operands.

## What changed

**`packages/config` — the comparison core (ADR 0022)**
- `config-diff.ts`: a pure classifier producing a typed `ConfigChangeSet` (`update` / `remote_only` / `local_only`, plus `masked` and per-class `counts`), reusable by `config pull` without the command layer.
- Both operands are CLI-2230 convergence projections (ADR 0021): `fromConfigDocument({config, document})` locally (raw-presence-masked, canonicalized, secret-omitting) and `fromApiProjectConfig(response)` remotely. All wire knowledge — renames, boolean inversions, duration/byte-size conversions, the GoTrue key table — lives in the shared `projectConfigMappingRows` registry, so the managed surface is `isComparableProjectConfigPath` **by construction**: a path with no registry row (`[studio]`, ports, image pins, `[realtime]` locals, `workers`) can never be reported.
- Classification is driven by the raw document's *declared*-key set — the one signal a decoded config cannot recover — so "the file wrote the default" and "the file is silent" classify differently (`update` vs suppressed/`remote_only`).
- `remote_only` suppression baseline: the default config's own convergence projection, falling back to the raw schema default for push-gated containers (network restrictions' allow-all default is exactly the platform's unconfigured state), then to the type's zero value. An untouched project diffs clean.
- Secrets (the registry's `isSecret` rows) are "present, unknown": both normalizers omit them, they never classify or count, and locally-declared ones surface via `masked` so a clean diff is visibly a partial claim.
- Residual equality is meaning-based: multiset array comparison (`additional_redirect_urls` order is not drift) and string/number, string/boolean scalar tolerance.
- `io.ts`/`lib/env.ts`: value origins now record the resolving env-var name, so a change on an `env()`-fed property names the variable.

**`apps/cli` — legacy-shell command**
- `legacy/commands/config/diff/`: command + handler + errors + `SIDE_EFFECTS.md`. Target resolution: `--target <branch-name|uuid|ref>` (same acceptance as `link`; 404 → "run `supabase branches list`"), `--project-ref`, else the linked ref; `--target` + `--project-ref` together is a hard error. When the resolved ref matches a `[remotes.*]` block's `project_id`, the local operand is the branch's merged effective config (ADR 0018), otherwise the base config — the echoed line always says which.
- Output: text (unset renders `(unset)` / `(not returned)`; `(from env VAR)` annotations; masked note) and `--output-format json|stream-json` (structured payload with `schema_version`, `target`, `scope`, `changes[]`, `masked[]`, `counts`). The comparison-scope line lists which response blocks were carried. The Go-compat `-o/--output` flag is **rejected outright** (every value, `pretty` included) with an error pointing at `--output-format` — per the ticket-thread decision that net-new commands carry no Go parity contract. `--exit-code` sets exit 1 on drift via `ProcessControl.setExitCode` after the payload is emitted.
- `legacy/shared/legacy-branch-ref.resolver.ts`: the branch name/UUID/ref resolver hoisted out of the branches family (cross-family use) with injected error mappers; the branches family keeps a thin binding so its call sites are unchanged.

**Docs**: ADR 0022 (classification + managed surface, incl. the registry consolidation and its relationship to ADR 0019/0020/0021), `go-cli-divergences.md` TS-only command entry (replacing the ticket's stale `go-cli-porting-status.md` criterion), per-command `SIDE_EFFECTS.md`.

## History note for reviewers

The branch was first implemented with a self-contained translation table (a ~900-line port of the Go CLI's `FromRemoteAuthConfig`). After #6339 landed the registry-driven normalizers on `develop` — with this command as their named consumer — the merge commit (`766182f`) brought `develop` in and the follow-up (`cae9c14`) deleted the tables and rebuilt the classifier on the registry, per the repo's no-parallel-code-paths policy. ADR 0022's "Considered Alternatives" records both designs.

## Decisions & assumptions worth reviewing

1. **Managed surface = the shared registry** (vs schema annotations or response-key walking): single source of truth shared with Studio and the future push mapper; a missing row fails safe (silently unmanaged, never misreported). Alternatives in ADR 0022.
2. **Suppression baseline chain** (default projection → raw schema default → zero value) is a judgment call: without it, every unconfigured OAuth provider and the platform's allow-all network restrictions read as drift on untouched projects. The cost is that a remote override that happens to equal a schema default on an undeclared path is not reported.
3. **`schema_version`** in the payload is the file's `$schema` ref (falling back to the current schema URL) — CLI-2155 shipped no separate version token.
4. **Rendered "local" values are convergence projections** (ADR 0021), i.e. what pushing the file would produce hosted — canonicalized durations/byte sizes, push-gated omissions — not necessarily the file's literal spelling. Documented in `SIDE_EFFECTS.md`.
5. **Legacy `-o` support was implemented per the original acceptance criteria, then removed** after Colum confirmed on the ticket that parity isn't a goal for net-new commands. The flag now fails fast with a bespoke invalid-input error; the JSON payload always carries explicit `null` for unset sides.
6. **Masked credentials are transparent**: listed in `masked[]` / a text note rather than silently skipped, and never affect `--exit-code`.
7. **Partial responses degrade, never error**: comparable paths the response doesn't carry are `local_only` when declared locally, silent otherwise; the scope line calls out missing blocks. (Today's v2 schema requires all six blocks, so this is belt-and-braces for API evolution and permission-trimmed keys.)
8. The classifier inherits ADR 0021's limits: unconditionally-mapped fields with no local-silence signal can surface as honest-but-push-unactionable `remote_only` entries (tracked on CLI-2266).


kanadgupta:
Merged `develop` in (merge commit, no rebase) and consolidated the comparison core onto CLI-2230's registry:

- **Merge commit** (`766182f`): conflict resolution + mechanical adaptation — our ADR renumbered 0019→0022 (develop took 0019–0021), the env-var-name threading re-applied to the relocated `CliConfigValueOrigin`, `CliConfig`/entrypoint-split renames, the live test rewritten for the fixture-based live harness, and the v2 test fixture gaining the new required `database.major_version`.
- **Consolidation** (`cae9c14`): #6339 built `fromConfigDocument`/`fromApiProjectConfig` explicitly as this diff's operands (ADR 0021), so the branch's own ~900-line translation tables (`config-diff.{managed,auth,read}.ts`) are deleted and `diffProjectConfig` now compares the two convergence projections over `isComparableProjectConfigPath`, with masked secrets from the registry's `isSecret` rows and `remote_only` suppression baselined on the default config's projection (falling back to raw schema defaults for push-gated containers, e.g. network restrictions' allow-all). Classification semantics (declared-set-driven `update`/`remote_only`/`local_only`, masked note, env-var naming, `--exit-code`) are unchanged. ADR 0022 records the consolidation and its relationship to ADR 0019/0020/0021.

apps/cli/src/legacy/commands/config/diff/diff.command.ts
```diff
@@ -0,0 +1,49 @@
...
+    Flag.withDescription("Project ref of the Supabase project."),
+    Flag.optional,
+  ),
+  target: Flag.string("target").pipe(
+    Flag.withDescription(
+      "Branch name, branch ID, or project ref to compare against. Mutually exclusive with --project-ref.",
+    ),
+    Flag.optional,
+  ),
+  exitCode: Flag.boolean("exit-code").pipe(
```
Coly010, line 21: **Blocker:** `Flag.boolean("exit-code")` without `Flag.withDefault(false)` makes this a *required* flag, so the command cannot be invoked as documented:

```
$ supabase config diff
Error: required flag(s) "exit-code" not set
```

That's the exact invocation in the EXAMPLES block below and in the generated docs spec (which even emits `default_value: "false"`). The integration suite can't catch this because every test hands a pre-built flags object to the handler and never goes through the parser — please ship a parser-level test with the one-line fix (e.g. an e2e assertion that `config diff` with no args doesn't emit `required flag(s)`), or the next boolean flag will regress the same way.

packages/config/src/config-diff.ts
```diff
@@ -0,0 +1,301 @@
...
+  );
+}
+
+/**
+ * Order-insensitive, type-aware value equality: arrays compare as multisets
+ * (`additional_redirect_urls` in a different order is not a difference), and
+ * scalars tolerate string/number and string/boolean representation skew.
+ */
+export function isEqualConfigValue(a: unknown, b: unknown): boolean {
+  if (Array.isArray(a) && Array.isArray(b)) {
```
Coly010, line 238: **Blocker:** comparing *every* array as a multiset false-negatives on order-significant config. `api.schemas` (PostgREST's default schema is the first entry) and `api.extra_search_path` (a literal `search_path`, where order is resolution order) are sequences, not sets. Reproduced: local `["public","extensions"]` vs remote `db_extra_search_path: "extensions,public"` → `changes: []`, so `--exit-code` exits 0 on a difference that changes runtime behaviour — in a drift detector.

"Is this array a set or a sequence?" is per-field wire semantics, i.e. registry knowledge: suggest a row property (`arrayEquality: "set" | "sequence"`) on `projectConfigMappingRows`, defaulting to sequence (over-report rather than under-report), with `additional_redirect_urls` opting into set semantics.

packages/config/src/config-diff.ts
```diff
@@ -0,0 +1,301 @@
...
+  }
+  if (typeof value === "number" || typeof value === "boolean") {
+    // Scalars fold to their string form so "1" and 1 compare equal, matching
+    // the scalar type-awareness above.
+    return `s:${String(value)}`;
+  }
+  return `j:${JSON.stringify(value)}`;
+}
+
+function isZeroValue(value: unknown): boolean {
```
Coly010: **Blocker:** this fallback tests JS zeros, but the operand is already canonicalized by the registry — GoTrue's `sessions_timebox: 0` arrives here as the *string* `"0s"` and escapes the check. Confirmed live against staging: a near-default project reports `auth.sessions.timebox` / `auth.sessions.inactivity_timeout` as `[remote only] remote: "0s"`, plus 13 more noise lines for `auth.email.template.*.subject` / `notification.*.subject` (platform-reported defaults with no baseline in the default CLI config; the three `storage.*` entries look like the same class). On the staging bench project, 15 of the 18 `remote_only` entries were this noise — an untouched project fails `--exit-code`, which is exactly the flooding ADR 0022 says the baseline prevents.

Suggest not inferring "unconfigured" from JS zeros at all: put the platform's unconfigured value on the registry row (e.g. `unconfiguredValue`), and add a registry-driven test enumerating every comparable path whose baseline is `undefined`, asserting its zero form suppresses.

packages/config/src/config-diff.ts
```diff
@@ -0,0 +1,301 @@
...
+    if (remoteValue !== undefined) {
+      // The local projection is silent: the file doesn't declare it, or push
+      // cannot communicate the declared state (ADR 0021's unmanaged-by-push
+      // families). Suppress the remote value when it matches the default
+      // config's own projection; for paths that projection is also silent on
+      // (push-gated containers), fall back to the raw default config's value
+      // (e.g. `db.network_restrictions.allowed_cidrs`'s allow-all default is
+      // exactly the platform's unconfigured state), then to the type's zero
+      // value (the platform's report of an unconfigured feature).
+      const baseline = valueAtPath(defaults, path) ?? valueAtPath(getDefaultCliConfig(), path);
```
Coly010: **Blocker (false clean):** `applyPushUnmanagedOmissions` unconditionally drops `auth.oauth_server` from the local projection, but `auth.oauth_server.enabled` is a comparable registry path the API reports. Reproduced end-to-end: file declares `[auth.oauth_server] enabled = true`, remote reports `oauth_server_enabled: false` → local is silent, remote equals the raw default, suppressed here — and the command prints `No config differences found.` A declared local value that genuinely disagrees with the remote vanishes with no change entry, no masked note, no signal.

Suggestion (additive): a third bucket alongside `masked` — `unmanaged: ReadonlyArray<string>` for declared paths the local projection dropped — surfaced like the masked note ("N declared property(ies) cannot be pushed and were not compared: …"). The classifier already has both inputs it needs (`declared` + local silence).

Related, same line: the `?? valueAtPath(getDefaultCliConfig(), path)` fallback is hard-coded, so the `defaults` option (which no call site currently passes) only controls one of the three baseline tiers. Either make the option the complete baseline or delete it.

packages/config/src/config-diff.ts
```diff
@@ -0,0 +1,301 @@
...
+      changes.push(
+        declared
+          ? {
+              path,
+              class: "update",
+              local: localValue,
+              remote: remoteValue,
+              ...(envVariable === undefined ? {} : { envVariable }),
+            }
+          : { path, class: "remote_only", local: undefined, remote: remoteValue },
```
Coly010: **Major:** on an undeclared path the classifier *had* the local effective value (the materialized default it just compared) and throws it away. The user sees:

```
api.max_rows [remote only]
  local:  (unset)
  remote: 250
```

…when the local effective value is `1000` and a `config push` would overwrite the remote `250` with it. That's the "someone changed it in the dashboard" case — the primary reason this command exists — and `[remote only]` reads as "key exists only remotely", which is false for anything with a schema default. It also collapses two states the future `config pull` consumer needs distinguished (file-silent vs materialized-default-disagrees).

Suggestion: keep `local` populated and add `readonly declared: boolean` to `ConfigChange` (already computed above); render `local: 1000 (schema default — not declared in config.toml)`.

packages/config/src/config-diff.ts
```diff
@@ -0,0 +1,301 @@
...
+}
+
+function isPlainRecord(value: unknown): value is Readonly<Record<string, unknown>> {
+  return typeof value === "object" && value !== null && !Array.isArray(value);
+}
+
+/** Walks a dotted path through records with own-key checks only. */
+function valueAtPath(root: unknown, path: string): unknown {
+  let current: unknown = root;
+  for (const segment of path.split(".")) {
```
Coly010: Dotted-string paths are lossy for record keys: this `path.split(".")` (and `isDeclaredAtPath`) round-trips paths that `collectLeafPaths` joined with `"."`, while the package itself deliberately keys paths as segment arrays (`pathKey = JSON.stringify(path)`) for exactly this reason. An `auth.sms.test_otp` key containing a `.` resolves `undefined` on both sides and the drift is *silently dropped* — reproduced: local `test_otp["415.2127777"] = "111111"` vs remote `"999999"` → `changes: []`, exit 0.

Suggest `readonly path: ReadonlyArray<string>` on `ConfigChange` (also what `config pull` will need to write back into a document), joining for display only in `diff.format.ts`.

apps/cli/src/legacy/commands/config/diff/diff.handler.ts
```diff
@@ -0,0 +1,216 @@
...
+    ref = flags.target.value;
+  } else {
+    ref = yield* resolver.resolve(flags.projectRef);
+  }
+
+  yield* Effect.gen(function* () {
+    // 1. Load the local config, merging a matching `[remotes.*]` block over
+    // the base document when the target ref names a declared branch (ADR
+    // 0018). Never writes — this command is read-only by contract.
+    const loaded = yield* loadCliConfig(runtimeInfo.cwd, {
```
Coly010: **Major — `--workdir` is silently ignored:** `runtimeInfo.cwd` is `process.cwd()`, while the project ref (`LegacyProjectRefResolver`) and the linked-project cache both resolve against `cliSettings.workdir` — so `config diff --workdir ../other` compares the *wrong* `config.toml` against `../other`'s linked project (reproduced with two disjoint roots). `SIDE_EFFECTS.md` documents the workdir-relative behaviour, so it's a doc/impl mismatch too. `config push` shares the bug (`push.handler.ts:134`) — worth fixing both in the same change.

apps/cli/src/legacy/commands/config/diff/diff.handler.ts
```diff
@@ -0,0 +1,216 @@
...
+    );
+    yield* fetching?.clear() ?? Effect.void;
+
+    // 3. Project both sides through CLI-2230's convergence normalizers (ADR
+    // 0021): `fromConfigDocument` gets the loaded config WITH its raw
+    // document so raw-presence masking applies, and `fromApiProjectConfig`
+    // canonicalizes the response into the same post-push shape. A response
+    // the registry cannot narrow (out-of-domain mapped values) is an API
+    // problem, not a transport one.
+    const remote = yield* Effect.try({
```
Coly010, line 328: Two issues at this boundary:

1. A schema-mapping failure gets labelled a *network* error, the message stutters (`failed to read project config: ProjectConfigParseError: Could not read the project config from the Management API response: …`), the upstream `suggestion` ("Try upgrading the Supabase CLI…" — the one actionable sentence) is dropped by `String(cause)`, and `externalActionabilityByTag.ProjectConfigParseError` is bypassed — the adapter that deliberately splits caller-misuse from genuine response problems, whose own comment warns that collapsing both into `api_status` corrupts the external-failure KPI.
2. The sibling `fromConfigDocument(loaded)` call below isn't wrapped at all, so its throw becomes a defect.

Suggestion: keep `ProjectConfigParseError` in the typed channel (or a dedicated `LegacyConfigDiffResponseDecodeError` carrying `reason` and forwarding `suggestion`), and treat both normalizer calls the same way.

apps/cli/src/legacy/commands/config/diff/diff.handler.ts
```diff
@@ -0,0 +1,216 @@
...
+      const message =
+        total === 0 ? "No config differences found." : `${total} config difference(s) found.`;
+      yield* output.success(message, legacyConfigDiffPayload(changeSet, scope, context));
+    } else {
+      yield* output.raw(legacyRenderConfigDiffText(changeSet));
+    }
+
+    // 6. `--exit-code`: differences flip the exit status after the payload is
+    // out, without an error envelope corrupting machine output.
+    if (flags.exitCode && changeSet.changes.length > 0) {
```
Coly010: `--exit-code` drift → 1, but every failure also exits 1, so in text mode a script can't distinguish "config drifted" from "token expired" / 403 / malformed TOML — `config diff --exit-code || alert` fires on auth failures. The convention gives drift its own code (`git diff --exit-code`: errors → 128; `diff(1)`: 1 vs 2; `terraform plan -detailed-exitcode`: 0/1/2). Suggest exit **2** on drift with 1 reserved for errors; if 1 stays, the `SIDE_EFFECTS.md` exit-code table needs an explicit "drift and failure are indistinguishable in text mode; use `--output-format json`" note, because that table currently reads as a contract.

apps/cli/src/legacy/commands/config/diff/diff.handler.ts
```diff
@@ -0,0 +1,216 @@
...
+      yield* processControl.setExitCode(1);
+    }
+  }).pipe(Effect.ensuring(linkedProjectCache.cache(ref)), Effect.ensuring(telemetryState.flush));
```
Coly010: **Legacy Shell Invariant #1:** these `Effect.ensuring`s wrap only the inner gen, so every pre-resolution failure — the `-o` rejection, the `--target`/`--project-ref` conflict, branch 404 and branch-resolve network errors (which *do* make an API call) — returns without flushing telemetry or writing the linked-project cache. Reproduced: `config diff -o json` → flushed = false. `SIDE_EFFECTS.md` says both files are written on "success **and** failure", and the branches family deliberately keeps resolution *inside* the wrapper (`branches/delete`). Suggest an outermost `Effect.ensuring(telemetryState.flush)` around the whole handler with `linkedProjectCache.cache(ref)` on the inner scope that has a resolved ref — plus mirror `telemetry.flushed === true` assertions for the three failure paths.

Related: with `--target <ref>` in an *unlinked* directory, the cache writes `linked-project.json` for a project the user never linked (the mismatch guard only protects linked dirs).

apps/cli/src/legacy/commands/config/diff/diff.handler.ts
```diff
@@ -0,0 +1,216 @@
...
+  const processControl = yield* ProcessControl;
+  const goOutputFlag = yield* LegacyOutputFlag;
+
+  // Net-new TS command with no Go parity contract: the Go-compat `-o/--output`
+  // flag is rejected outright (every value, `pretty` included) rather than
+  // honored — machine output goes through `--output-format` only (CLI-2156,
+  // per Colum). Checked first so no target resolution or network call runs.
+  if (Option.isSome(goOutputFlag)) {
+    return yield* new LegacyConfigDiffOutputFlagUnsupportedError({
+      message:
```
Coly010: Not relitigating the rejection itself — but three papercuts on this path: (1) `--help` still advertises `-o` under GLOBAL FLAGS, so help promises what the handler kills; worth a sentence in the command description ("machine output via `--output-format json|stream-json`; the global `-o/--output` is not supported"). (2) The message is lowercase while the flag-validation error one keystroke away is capitalized, and the appended "Try rerunning the command with --debug" can't help a usage error. (3) The error class carries no `suggestion` field, so `--output-format json` consumers get no remedy even though it's a known literal string. Also: `db diff -o <file>` muscle memory lands on a choice-validation error that never mentions this command has no file output.

apps/cli/src/legacy/commands/config/diff/diff.format.ts
```diff
@@ -0,0 +1,156 @@
...
+
+function maskedNote(masked: ReadonlyArray<string>): string {
+  return `Note: ${masked.length} credential value(s) not compared (masked by the API): ${masked.join(", ")}\n`;
+}
+
+/** Human-readable diff body for text mode (stdout). */
+export function legacyRenderConfigDiffText(changeSet: ConfigChangeSet): string {
+  const lines: Array<string> = [];
+  for (const change of changeSet.changes) {
+    lines.push(`${change.path} [${CLASS_LABELS[change.class]}]`);
```
Coly010: **Security (medium):** values are safe (`renderValue` routes strings through `JSON.stringify`, which escapes ESC), but `change.path` here, `envVariable` two lines down, and `appliedRemote`/`branch`/`projectRef` in the stderr lines above are interpolated raw. Path segments are attacker-influenced: `auth.sms.test_otp` and `[remotes.*]` keys are unconstrained TOML keys (quoted keys accept `\u001b`), and the API's `sms_test_otp` record keys come straight from `envToMap` with no validation — so a hostile repo or endpoint can emit raw ANSI or forge output lines (e.g. a remotes name ending `\nNo config differences found.` makes a drifting config look clean in scrollback/CI logs).

This exact threat is documented on `legacySanitizeInlineName` (PR #6168 review), and the handler already applies it to the branch-404 message — apply it to every non-constant string in the text renderer. JSON mode needs no change (`JSON.stringify` escapes control chars).

apps/cli/src/legacy/commands/config/diff/diff.format.ts
```diff
@@ -0,0 +1,156 @@
+import type { ConfigChange, ConfigChangeSet, CliConfigValueOrigin } from "@supabase/config";
+
+/**
+ * Pure formatters, payload builders, and input adapters for `config diff` —
+ * no Effect, no services, unit-testable in isolation.
+ */
+
+/** The per-service blocks of the v2 project-config resource. */
+const REMOTE_CONFIG_BLOCKS = ["api", "auth", "database", "pooler", "realtime", "storage"] as const;
```
Coly010: Two on the scope machinery:

1. This hand-copies the field set of `ProjectConfigApiAttributesSchema` into a CLI formatter (and the unit test pins it). When the platform adds a block and the package schema learns it, the scope line reports it "not returned" forever, test-green. The package owns the response shape — export the section keys from `@supabase/config`, or have `fromApiProjectConfig` report the carried-block set.
2. `isRemoteBlockRecord` accepts `{}`, so the most likely partial response (permission-truncated `auth: {}`) passes: the scope line swears `auth` was compared while 38 auth keys print `(not returned)` as `local_only` — and with `--exit-code`, that's a permanently red CI no file edit can fix. Treat empty records as not-returned (or render key counts).

(See also the SIDE_EFFECTS.md thread: with the current strict response contract, a genuinely *missing* block never reaches this code at all.)

apps/cli/src/legacy/commands/config/diff/diff.format.ts
```diff
@@ -0,0 +1,156 @@
...
+  scope: LegacyConfigDiffScope,
+  context: LegacyConfigDiffContext,
+): Record<string, unknown> {
+  const valueEntry = (key: string, value: unknown): Record<string, unknown> => ({
+    [key]: value === undefined ? null : value,
+  });
+
+  const { update, remote_only, local_only } = changeSet.counts;
+  return {
+    schema_version: context.schemaVersion,
```
Coly010: `schema_version` here is the user's `$schema` URL — user-controlled, per-repo, and splatted at the envelope top level right next to `message`, exactly where machine consumers expect the *payload contract* version. Any jq/agent consumer gating on `.schema_version` for forward-compat is writing a bug. Suggest `config_schema` for the document reference and a real integer `schema_version: 1` for the payload shape — cheap now, unfixable once scripts key off it.

Nit on the same payload: `env_variable` is a comma-joined string when an array literal drew on several vars (`env.ts:280` does `envNames.join(", ")`), which consumers can't split safely — an array field would be cleaner on a brand-new payload with no compat constraint. And `message` in JSON mode drops the masked-secrets caveat that text mode prints, so an agent echoing `.message` reports "in sync" on a project whose SMTP password drifted.

apps/cli/src/legacy/commands/config/diff/diff.command.ts
```diff
@@ -0,0 +1,49 @@
...
+import { legacyManagementApiRuntimeLayer } from "../../../shared/legacy-management-api-runtime.layer.ts";
+import { withLegacyCommandInstrumentation } from "../../../telemetry/legacy-command-instrumentation.ts";
+import { legacyConfigDiff } from "./diff.handler.ts";
+
+const config = {
+  projectRef: Flag.string("project-ref").pipe(
+    Flag.withDescription("Project ref of the Supabase project."),
+    Flag.optional,
+  ),
+  target: Flag.string("target").pipe(
```
Coly010: `--target` is the only flag of this name in the legacy tree, and it re-invents vocabulary `link` already settled (positional `[ref-or-branch]`, plus `--project-ref` itself accepting branch names, with the documented "values that are exactly 20 lowercase letters are always treated as project refs" rule). As shipped, `--target` is a strict superset of `--project-ref` (now redundant on this command), the two are mutually exclusive — which forbids the coherent "branch `staging` of project X" and forces a `link` in unlinked CI checkouts — and the 20-lowercase-letters rule isn't in this description, so a branch literally named like a ref silently resolves as a project and 404s. Suggest either `link`'s shape (positional + branch-accepting `--project-ref`) or keeping `--target` and dropping `--project-ref`; either way, copy `link`'s sentence about ref-shaped values.

apps/cli/src/legacy/commands/config/diff/diff.command.ts
```diff
@@ -0,0 +1,49 @@
...
+      description: "Diff against the linked project",
+    },
+    {
+      command: "supabase config diff --target staging --exit-code",
+      description: "Diff against the 'staging' branch, exiting 1 on drift",
+    },
+  ]),
+  Command.withHandler((flags) =>
+    legacyConfigDiff(flags).pipe(
+      withLegacyCommandInstrumentation({ flags, safeFlags: ["project-ref"] }),
```
Coly010: `safeFlags: ["project-ref"]` logs the ref verbatim while `config push` (same family, same flag) redacts it, and the established safe list in `apps/cli/CLAUDE.md` doesn't include the config family. Telemetry drift is silent and breaks dashboards — either drop this or add it to `config push` and extend the documented list in the same change.

apps/cli/src/legacy/commands/config/diff/diff.errors.ts
```diff
@@ -0,0 +1,94 @@
...
+      ? { ...actionability.apiStatus, fingerprint_suffix: "api_response" }
+      : actionability.externalNetwork;
+  }
+}
+
+export class LegacyConfigDiffReadStatusError extends Data.TaggedError(
+  "LegacyConfigDiffReadStatusError",
+)<StatusErrorArgs> {
+  get [ErrorActionabilityId](): CliErrorActionabilityDeclaration {
+    return statusCodeActionability(this.status);
```
Coly010: `/v2/projects/{ref}/config` names a user-selected resource, so a 404 means "wrong project ref" — user-actionable. Without `{ notFoundIsInvalidInput: true }` (which `LegacyConfigDiffBranchResolveStatusError` above and all 11 `push.errors.ts` status errors on ref-addressed routes pass), this classifies a 404 as an external-service problem and skews the KPI split.

apps/cli/src/legacy/commands/config/diff/SIDE_EFFECTS.md
```diff
@@ -0,0 +1,112 @@
...
+  never plaintext; the registry's `isSecret` rows) are treated as "present, unknown" — never
+  reported as differences and never counted for `--exit-code`; they are surfaced via the
+  masked note / `masked[]`.
+- **Values are convergence projections (ADR 0021):** both sides are normalized through
+  `@supabase/config`'s `fromConfigDocument`/`fromApiProjectConfig`, so a reported "local"
+  value is what pushing the file would produce hosted (canonicalized durations/byte sizes,
+  push-gated omissions), not necessarily the file's literal spelling.
+- **Partial responses:** a managed property the response does not carry is `local_only`
+  when the file declares it and silent otherwise; a missing block is called out on the
+  scope line rather than treated as an error.
```
Coly010: This claim (and ADR 0022's "partially-populated responses degrade … instead of an error") doesn't hold: `V2GetProjectConfigOutput` makes all six blocks — and their keys — *required*, so a missing block or key fails the typed decode inside the API client before any of this leniency runs. Confirmed live: staging doesn't return `storage.database_pool_mode` yet, and the command hard-fails on every invocation with

```
failed to read project config: SchemaError(Missing key
  at ["data"]["attributes"]["storage"]["database_pool_mode"])
```

— i.e. the command is currently broken against staging, and a permission-truncated response (the case the ADR names) surfaces as an opaque SchemaError. Consequences: the scope line's "(not returned: …)" branch is unreachable in production (it prints the constant six-block list on every run), and two `diff.format.unit.test.ts` cases exercise unreachable states.

Pick one: loosen the contract (make blocks/keys optional, matching `auth`'s leniency — the stated intent) and keep the scope machinery, or delete the scope machinery and correct this doc + ADR 0022. Don't leave the doc asserting behaviour the contract forbids.

apps/cli/docs/go-cli-divergences.md
```diff
@@ -12,13 +12,14 @@ not a compatibility promise.
...
+| `config diff`     | [`../src/legacy/commands/config/diff/diff.command.ts`](../src/legacy/commands/config/diff/diff.command.ts)         | Read-only drift report between `supabase/config.toml` and `GET /v2/projects/{ref}/config` (CLI-2156). TS-only: the old Go CLI had no config diff. `--target` accepts a branch name/UUID/ref; `--exit-code` exits 1 on drift. Rejects the Go-compat `-o/--output` flag outright — machine output is `--output-format json\|stream-json` only (no Go parity contract for net-new commands, per the CLI-2156 discussion). Comparison core lives in `@supabase/config` (ADR 0022). |
```
Coly010: `apps/cli/CLAUDE.md` declares this file "a **frozen historical record** … Do not add new entries — new flags and features are simply new CLI behavior." This row was flagged independently by every review pass. Suggest dropping it (the `-o` rationale already lives in `SIDE_EFFECTS.md` and the error text; ADR 0022 is the right home for the decision), which also undoes the table reflow that turned a 1-line change into a 15-line diff. `SIDE_EFFECTS.md:9` points here too and should stop.

apps/cli/src/legacy/commands/config/diff/diff.integration.test.ts
```diff
@@ -0,0 +1,642 @@
...
+      pool_mode: "transaction",
+      ignore_startup_parameters: "",
+      server_idle_timeout: 0,
+      server_lifetime: 0,
+      query_wait_timeout: 0,
+      reserve_pool_size: 0,
+      default_pool_size: 20,
+      max_client_conn: 100,
+    },
+    auth: {},
```
Coly010: This `auth: {}` is the structural blind spot that let both classifier blockers through: v2 `attributes.auth` is an open record, so an empty object is schema-valid — meaning the entire GoTrue mapping surface (~200 keys; all the duration/inversion/sentinel logic) is never exercised end-to-end, and the fixture docstring's "an empty config.toml diffs clean" claim is only true *because* auth is empty.

Highest-value test additions:
1. A realistic all-defaults `auth` block in this fixture, asserting a fresh config diffs clean (this is where the `"0s"` suppression bug lives).
2. The same cleanliness assertion in `diff.live.test.ts` — its own comment names "the GoTrue-keyed auth record … classifying cleanly" as the one thing mocks can't prove, then asserts only `exitCode === 0`.
3. A `not.toContain(<secret>)` assertion on the masking scenario (it already seeds `GITHUB_SECRET=shh` and an HMAC-shaped remote value without asserting absence), so the "secrets never leak" claim is pinned against formatter changes.
4. Telemetry-flush assertions on the `-o`, flag-conflict, and branch-404 paths (currently they'd fail).

Coly010 (changes requested):
Ran an adversarial review of this PR: four independent passes (implementation, architecture, security, DX) plus live runs from source against staging (`api.supabase.green`) with a real linked project. Every finding was reproduced by running code, not read off the diff.

**Verdict: request changes.** The design underneath — the registry-driven managed surface, the secret triple-gating, convergence-projection operands — held up under genuinely hostile probing (16 secret-leak probes across all output modes failed; unknown future secret API fields fail *safe*; read-only is proven by mtime+content pinning; machine stdout stays payload-pure; the branch-resolver hoist is behaviour-identical with all 88 branches tests green). The blockers are concentrated in the classifier's equality/suppression edges, the flag wiring, and documented claims the code doesn't keep.

### Blockers (all inline)
1. **`--exit-code` is accidentally required** → plain `supabase config diff` (the help's own first example) errors out. One-line fix; needs a parser-level test.
2. **Order-insensitive array equality** false-negatives on `api.schemas` / `api.extra_search_path` → `--exit-code` exits 0 on real drift.
3. **The `remote_only` suppression baseline misses canonicalized zeros and platform-default subjects** → untouched projects report drift. Live on staging: 15 of 18 remote-only entries on a near-default project were this noise (`"0s"` session values + 13 template/notification subjects + 3 storage defaults).
4. **False clean:** a declared `auth.oauth_server.enabled` disagreeing with the remote prints "No config differences found."

### Confirmed live against staging
Beyond blocker 1, the command hard-fails on staging today: `SchemaError(Missing key at ["data"]["attributes"]["storage"]["database_pool_mode"])` — the generated contract requires every block key, so the documented "partial responses degrade, never error" behaviour is unreachable (inline on `SIDE_EFFECTS.md`). I only obtained a successful run by locally patching the contract to make that key optional. With that patch, the happy path works well end-to-end: 29 classified changes, correct `counts`, `(from env VAR)` annotations, masked-credentials note, machine payload and exit codes all as designed.

### Majors (all inline)
- `remote_only` erases the local value it just compared — the output can't answer "what would `config push` change?".
- `--workdir` is silently ignored (`config push` shares the bug).
- Telemetry flush + linked-project cache skipped on every pre-resolution failure path (Legacy Shell Invariant #1).
- ANSI/control-character injection via unsanitized path segments and names in text output (`legacySanitizeInlineName` exists for exactly this and is used 14 lines away).
- `--exit-code` conflates drift with failure (both exit 1).
- JSON `schema_version` is the user's `$schema` URL, not a payload contract version.
- Response-decode failures mislabeled as network errors, dropping the upstream suggestion and bypassing the purpose-built actionability adapter.
- Dotted-path round-tripping silently drops record keys containing `.`.

### Smaller items not carried by an inline thread
- `DiffProjectConfigOptions` asks for `local` and `declared` separately — two params that must come from the same load, with nothing enforcing it. Consider accepting the loaded pair and deriving both.
- `counts` is derived state computed in three places (`config-diff.ts`, both formatters + `changes.length` in the handler); one will drift. Either drop it from the package type or make `total` part of it and use it everywhere.
- No docs-site overlay (`config push` has `docs/supabase/config/push.md`), so the published reference page for a semantically subtle command falls back to one sentence — nothing on `--exit-code`, `(unset)` vs `(not returned)`, masking, or the fact that rendered local values are convergence projections (a user who writes `"1m"` and sees `"1m0s"` will grep their file and file an issue; worth a one-line note in the output or docs).
- One concept, three spellings: `[remote only]` / `remote-only` / `remote_only` across label, summary, and JSON. And `N difference(s)` where the count is known at render time.
- `--target <uuid>` is echoed as a quoted display name (`Comparing against '1111…-…'`); the branch's actual name is never shown.
- `SIDE_EFFECTS.md` "config.toml is read before any network call" is false for `--target <branch>` — branch resolution runs first, so a broken TOML burns an API round-trip, and in a fresh directory the "run `supabase link`" error wins over the friendlier "run `supabase init`" one.
- JSON `scope` lists only present blocks; consumers must re-derive the missing set from a hardcoded list — consider `scope: {present, missing}`.
- ADR 0022 ships as `proposed` (README row too) while its body says it "was first accepted", and it's silent on three shipped decisions: `--target`, the `-o` rejection, and the `remote_only` local-value nulling.

### Test-suite structure (why the suite missed the blockers)
The parser is never exercised (blocker 1); the `auth: {}` fixture means the largest, most transform-heavy mapping surface never runs end-to-end (blocker 3); the live test asserts only exit 0 where its own comment says cleanliness is the point; and no test asserts a secret *string* is absent from output. Details inline on the fixture.

### What's genuinely good here
The `(from env VAR)` annotation, the `(unset)`/`(not returned)`/`null` distinctions, byte sizes rendered in the user's units, masked secrets surfaced-but-never-counted, the read-only proof in tests, and the registry consolidation over a parallel translation table are all exactly right. This is close to a really good command — it just can't currently be invoked, and each of its two core promises (no false drift, no false clean) has a reproduced counterexample.

apps/cli/src/legacy/commands/config/diff/diff.command.ts
kanadgupta: Fixed in 04a174af1: `Flag.withDefault(false)`, and — since you're right that the integration suite structurally can't catch this — a new `diff.e2e.test.ts` pins the parser at the subprocess boundary (`config diff` with no args must not emit `required flag(s)`), so the next boolean flag can't regress the same way.

packages/config/src/config-diff.ts
kanadgupta: Adopted your design verbatim in 3f23ca8ce: `arrayEquality: "set" | "sequence"` on `projectConfigMappingRows`, defaulting to sequence (over-report), with `additional_redirect_urls` opting into set semantics. Your exact repro — local `["public","extensions"]` vs `db_extra_search_path: "extensions,public"` — is now a pinned unit test (both `api.schemas` and `api.extra_search_path`), and `isEqualConfigValue`'s exported default flipped to sequence with the mode as an explicit parameter.

packages/config/src/config-diff.ts
kanadgupta: Adopted in 3f23ca8ce: zero inference is gone entirely — the suppression baseline is now default-projection → raw-default → the row's declared `unconfiguredValue`, and a path with no baseline at any tier reports. Rows declare: `"0s"` for both session bounds, the 13 provisioning-default subject strings, and `false` for the 7 notification toggles. For provenance I didn't guess the subjects: they're pinned byte-for-byte from the recorded real responses already in this repo (`apps/cli-e2e/fixtures/recorded/GET_v1_projects___PROJECT_REF___config_auth/`), and the notification default comes from supabase/auth's `NotificationsConfiguration` (`default:"false"`). The registry-driven test you asked for walks every comparable path whose config-side baseline is undefined: rows with `unconfiguredValue` must classify their unconfigured report clean, and absence-class rows must REPORT a zero-form value rather than swallow it.

packages/config/src/config-diff.ts
kanadgupta: Both adopted in 3f23ca8ce. `ConfigChangeSet.unmanaged` is the third bucket: declared comparable paths the local projection dropped (your `auth.oauth_server.enabled` repro, sentinel-pruned siblings, unselected SMS providers), surfaced like the masked note — text prints `Note: 1 declared property cannot be pushed and was not compared: auth.oauth_server.enabled`, and the machine `message` carries the caveat too. Your exact end-to-end repro is a pinned test. The `defaults` option is deleted — with no call site and only one of three tiers under its control it was dead optionality, and the tiering is now fully internal.

packages/config/src/config-diff.ts
kanadgupta: Adopted in 3f23ca8ce: `local` stays populated with the materialized default and `ConfigChange` carries `declared: boolean`; text renders `local:  1000 (schema default — not declared in config.toml)` (pinned integration test with your exact `max_rows: 250` scenario), and the JSON payload carries `declared` per change so the future `config pull` consumer gets the file-silent vs default-disagrees distinction.

packages/config/src/config-diff.ts
kanadgupta: Adopted in 3f23ca8ce: `ConfigChange.path` is `ReadonlyArray<string>` end to end (as are `masked`/`unmanaged`), joined for display only in `diff.format.ts`, and the JSON payload emits segment arrays for the same reason. Your `test_otp["415.2127777"]` repro is a pinned unit test — the drift now classifies as an update instead of vanishing.

apps/cli/src/legacy/commands/config/diff/diff.handler.ts
kanadgupta: Fixed in 27b62ef0b, both commands in the same change as you suggested: diff and push now load config against `cliSettings.workdir` (push's project-env/root derivation too), matching the resolver and the linked-project cache. Pinned by an integration test whose ambient cwd points at a directory with no `supabase/` at all.

apps/cli/src/legacy/commands/config/diff/diff.handler.ts
kanadgupta: Fixed in 3f23ca8ce, taking your first option: `ProjectConfigParseError` stays in the typed channel for **both** normalizer calls (the sibling `fromConfigDocument` now runs inside `diffProjectConfig` and gets the same boundary), so the message stutter is gone, the upstream `suggestion` survives to the renderer, and the purpose-built `externalActionabilityByTag` adapter classifies it instead of `api_status`. Anything non-`ProjectConfigParseError` escaping the normalizers stays a defect. The integration test now asserts the tag, the message, and the retained suggestion.

apps/cli/src/legacy/commands/config/diff/diff.handler.ts
kanadgupta: Adopted in 27b62ef0b: drift exits **2**, `1` stays the CLI-wide failure code (the terraform `-detailed-exitcode` mapping you suggested; the `diff(1)` inversion would have flipped exit 1's meaning relative to every other command). Flag description, examples, SIDE_EFFECTS exit-code table, and the docs overlay all state the 0/1/2 contract.

apps/cli/src/legacy/commands/config/diff/diff.handler.ts
kanadgupta: Fixed in fba78b708: `Effect.ensuring(telemetryState.flush)` now wraps the whole handler body (there is no pre-wrapper failure path left — even flag normalization is inside), and the cache finalizer fires exactly when a target ref resolved, per your suggested split. Mirror assertions added for the failure paths that exist after the `-o` rejection and flag conflict were retired: branch 404, missing config, and malformed TOML with a branch target (`telemetry.flushed === true`, plus `cachedRef` undefined when unresolved).

On the related note (cache write for an explicit ref in an unlinked directory): that's family-wide behavior today — `backups/list` (the canonical Invariant #1 pattern) and `config push` both `cache(ref)` unconditionally with whatever ref resolved, and the cache itself is fire-and-forget, fills only an absent file, and never fails the command. I've kept diff consistent with the family rather than diverging in one command; if you think the family semantics should change, I'd rather do that in its own issue across all callers. Happy to file it.

apps/cli/src/legacy/commands/config/diff/diff.handler.ts
kanadgupta: Went one step further than the papercuts in 04a174af1: the rejection itself is gone. Codex flagged that Legacy Shell Invariant #6 in AGENTS.md flatly requires both output mechanisms honored (`--output` first), and re-reading it I couldn't justify carving this command out — so `-o json|yaml|toml|env` now encode the structured payload through the shared encoders (backups/list's exact pattern) and `pretty` falls through to `--output-format`. That kills all three papercuts at the root (help now promises exactly what the handler does, and there's no error to capitalize or carry a suggestion), plus the `db diff -o <file>` muscle-memory case degrades to a real encoding instead of a dead end. SIDE_EFFECTS documents the encodings' null/array behavior.

apps/cli/src/legacy/commands/config/diff/diff.format.ts
kanadgupta: Fixed in 27b62ef0b: `legacySanitizeInlineName` is applied to every non-constant string in the text renderer — path segments (joined display form), env-var names, `appliedRemote`, `branch`, and `projectRef` in the stderr lines — with the rule documented at the top of the formatter. Your forged-clean-line scenario is pinned by an integration test using a `[remotes."evil[31mred"]` block: the escape byte must not reach either stream. JSON mode left as-is per your note.

apps/cli/src/legacy/commands/config/diff/diff.format.ts
kanadgupta: Both fixed in 27b62ef0b: (1) the block set is now exported from `@supabase/config` (`projectConfigApiBlockKeys`, derived from the response mirror's own key set) so it moves with the package instead of being hand-copied; (2) empty block records count as not-returned — the shared fixture's `auth: {}` case became its own test asserting `(not returned: auth)`, and the fixture itself now carries a realistic populated auth record (977e3001d). With a32ada78e relaxing the response contract, the not-returned branch is also reachable in production now rather than dead code.

apps/cli/src/legacy/commands/config/diff/diff.format.ts
kanadgupta: All three adopted: `schema_version` is now the integer payload-contract version (`1`, a named constant) with the `$schema` URL moved to `config_schema` (27b62ef0b); `env_variables` is an array end to end — I took it to the source rather than splitting at the edge, so `CliConfigValueOrigin.envVariables` is a list and the comma-join in `env.ts` is gone entirely (3f23ca8ce); and the machine `message` carries the masked (and new unmanaged) caveats, pinned by a test asserting a `.message` echoer never claims full sync while a secret is masked (27b62ef0b, 977e3001d).

apps/cli/src/legacy/commands/config/diff/diff.command.ts
kanadgupta: Adopted the second of your two shapes in fba78b708: `--target` is gone and `--project-ref` is branch-accepting with `link`'s exact description sentence (including the 20-lowercase-letters rule), which keeps diff flag-compatible with `config push` and retires the conflict error and the redundancy. The UUID path resolves through `GET /v1/branches/{id}` with the parent passed lazily, so it now works in an unlinked directory — which also softens the unlinked-CI limitation you noted (a CI checkout can target any branch by UUID without linking; the positional-parent shape remains available later if the named-branch-of-explicit-project case materializes).

apps/cli/src/legacy/commands/config/diff/diff.command.ts
kanadgupta: Fixed in fba78b708, taking your "add it to config push and extend the documented list" option — with one wrinkle the flag reshape introduced: diff's `--project-ref` now accepts branch names, so it uses `link`'s ref-shape-gated guard (verbatim only when `PROJECT_REF_PATTERN` matches; a user-created branch name never reaches PostHog). Push's ref-only flag logs verbatim unconditionally. The documented safe list in `apps/cli/CLAUDE.md` now names the config family and the branch-accepting guard rule.

apps/cli/src/legacy/commands/config/diff/diff.errors.ts
kanadgupta: Fixed in 27b62ef0b: `LegacyConfigDiffReadStatusError` passes `{ notFoundIsInvalidInput: true }`, matching the branch-resolve error and the ref-addressed `push.errors.ts` convention, with a comment stating the rule.

apps/cli/src/legacy/commands/config/diff/SIDE_EFFECTS.md
kanadgupta: Took the first option in a32ada78e: the contract is loosened rather than the scope machinery deleted. All 13 object-level `required` arrays under `data.attributes` (the six blocks and their nested containers) are relaxed through the established `openapi-overrides.json` test+replace mechanism — the same pattern as the SAML `attribute_mapping` and custom-hostname entries — so your staging failure shape (missing `storage.database_pool_mode`) and permission-truncated blocks decode and degrade as ADR 0022 intended. `@supabase/config`'s lenient mirror was already optional-everywhere, so only the generated side moved. A new `packages/api` client test pins the exact staging shape decoding, and the two formerly-unreachable scope-line states are now real (and empty blocks count as not-returned — see the scope thread). The envelope (`data`/`type`/`id`/`attributes`) and array-item shapes stay strict: a partial response omits fields, not halves of array elements.

apps/cli/docs/go-cli-divergences.md
kanadgupta: Dropped in 04a174af1 by restoring the file to develop's version outright, which also undoes the table reflow. The `-o` rationale is moot now that the flag is honored (see that thread); the surviving decisions live in SIDE_EFFECTS.md and ADR 0022 (now `accepted`, a4dc035ba).

apps/cli/src/legacy/commands/config/diff/diff.integration.test.ts
kanadgupta: All four adopted in 977e3001d (with the fixture landing on the classifier fixes from 3f23ca8ce):

1. The shared fixture now carries a ~65-key realistic fresh-project auth record at platform defaults — sessions zeros, all 13 subjects, notification toggles, durations, inversions — and the existing clean-config test proves the whole surface classifies cleanly against an empty config.toml.
2. `diff.live.test.ts` asserts no `auth.` change lines on a fresh project, not just exit 0.
3. The masking scenario now asserts `not.toContain` for both the resolved plaintext and an HMAC-shaped remote digest, on both streams, in text and JSON modes (including the payload serialization).
4. Telemetry-flush assertions cover the failure paths that still exist after the `-o` rejection and flag conflict were retired: branch 404, missing config, and malformed TOML with a branch target (the last also pinning zero API requests, per the read-before-network fix).

apps/cli/src/legacy/commands/config/diff/diff.handler.ts
```diff
@@ -0,0 +1,216 @@
...
+        "the -o/--output flag is not supported by config diff; use --output-format json|stream-json instead.",
+    });
```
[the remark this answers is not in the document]
kanadgupta: Done in 04a174af1 — `-o json|yaml|toml|env` now encode the structured payload via the shared encoders with `--output` taking priority, and `pretty` falls through to `--output-format`, per Invariant #6 and the backups/list pattern.

apps/cli/src/legacy/commands/config/diff/diff.handler.ts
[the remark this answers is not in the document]
kanadgupta: Done in fba78b708 — `Effect.ensuring(telemetryState.flush)` wraps the full handler body (config load and branch resolution included), with the ref-dependent cache finalizer gated on a resolved ref. Failure-path assertions added in 977e3001d.

apps/cli/src/legacy/commands/config/diff/diff.handler.ts
```diff
@@ -0,0 +1,216 @@
...
+  // Resolve the comparison target to a project ref. `--target` accepts a
+  // branch name, a branch UUID, or a raw project ref (same acceptance as
+  // `link`); a ref-shaped value skips the parent-project resolution entirely
+  // so it works in an unlinked directory.
+  let ref: string;
+  let branch: string | undefined;
+  if (Option.isSome(flags.target) && !LEGACY_BRANCH_PROJECT_REF_PATTERN.test(flags.target.value)) {
+    const target = flags.target.value;
+    branch = target;
+    const parentRef = yield* resolver.resolve(Option.none());
```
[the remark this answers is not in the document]
kanadgupta: Done in fba78b708 — the parent ref is passed to the branch resolver lazily and evaluated only for branch-NAME lookups, so a UUID `--project-ref` resolves through `GET /v1/branches/{id}` in an unlinked directory. Pinned by an integration test with `projectId: Option.none()`.

apps/cli/docs/go-cli-divergences.md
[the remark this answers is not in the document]
kanadgupta: Done in 04a174af1 — the file is restored to develop's version (row and reflow both gone).

apps/cli/src/legacy/commands/config/diff/diff.handler.ts
```diff
@@ -0,0 +1,216 @@
...
+  const linkedProjectCache = yield* LegacyLinkedProjectCache;
+  const telemetryState = yield* LegacyTelemetryState;
+  const runtimeInfo = yield* RuntimeInfo;
+  const processControl = yield* ProcessControl;
+  const goOutputFlag = yield* LegacyOutputFlag;
+
+  // Net-new TS command with no Go parity contract: the Go-compat `-o/--output`
+  // flag is rejected outright (every value, `pretty` included) rather than
+  // honored — machine output goes through `--output-format` only (CLI-2156,
+  // per Colum). Checked first so no target resolution or network call runs.
```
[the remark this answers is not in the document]
kanadgupta: Done — the framing disappeared with the code that carried it: the `-o` rejection comment/error (04a174af1) and the SIDE_EFFECTS "no Go CLI equivalent" line. The command's behavior is now described in its own terms throughout.

apps/cli/src/legacy/commands/config/diff/diff.handler.ts
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  yield* Effect.gen(function* () {
+    // 1. Load the local config, merging a matching `[remotes.*]` block over
+    // the base document when the target ref names a declared branch (ADR
+    // 0018). Never writes — this command is read-only by contract.
+    const loaded = yield* loadCliConfig(runtimeInfo.cwd, {
+      projectRef: ref,
+      goViperCompat: true,
+    }).pipe(
```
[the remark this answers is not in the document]
kanadgupta: Done in fba78b708 — the config is loaded and validated before any network call, with the `[remotes.*]` overlay applied by a re-load once the resolved ref is known (only configs that declare remotes reload). A malformed TOML with a branch target now pins zero API requests in the integration suite.

apps/cli/src/legacy/commands/config/diff/diff.handler.ts
```diff
@@ -0,0 +1,216 @@
...
+  let ref: string;
+  let branch: string | undefined;
+  if (Option.isSome(flags.target) && !LEGACY_BRANCH_PROJECT_REF_PATTERN.test(flags.target.value)) {
+    const target = flags.target.value;
+    branch = target;
+    const parentRef = yield* resolver.resolve(Option.none());
+    ref = yield* legacyResolveBranchProjectRef(target, parentRef, {
+      mapGetError: mapBranchResolveError,
+      mapFindError: mapBranchResolveError,
+    }).pipe(
```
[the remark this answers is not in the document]
kanadgupta: Done in fba78b708 — branch resolution runs under `output.task("Resolving branch...")` with fail/clear on every exit, matching the config fetch.

kanadgupta:
Thank you for the adversarial review — every blocker reproduced against running code made this round very direct to act on. All feedback is addressed in seven follow-up commits (no force pushes):

| Commit | What it carries |
| --- | --- |
| a32ada78e `fix(api)` | Relaxes all 13 object-level `required` arrays under `V2ProjectConfigResponse.data.attributes` via the established `openapi-overrides.json` mechanism, so the staging shape (missing `storage.database_pool_mode`) and permission-truncated responses decode instead of hard-failing. Pinned by a client test with the exact staging shape. |
| 3f23ca8ce `fix(config)` | The classifier cluster, expressed as registry row knowledge: `arrayEquality: "set" \| "sequence"` (sequence default; `additional_redirect_urls` opts into set), `unconfiguredValue` replacing JS-zero inference (sessions `"0s"`, the 13 provisioning-default subjects — values pinned from the recorded `config_auth` fixtures in this repo —, notification toggles `false`), the `unmanaged` bucket for declared-but-unpushable paths, segment-array paths end to end, `remote_only` keeping the materialized local + `declared`, counts computed once with `total`, the `defaults` option deleted, `DiffProjectConfigOptions` taking the loaded pair, `envVariables` as a list, and `ProjectConfigParseError` kept in the typed channel for both normalizer calls. |
| 04a174af1 `fix(cli)` | `--exit-code` gets `Flag.withDefault(false)` (plus a parser-level e2e pin), and `-o/--output` is now **honored** per Legacy Shell Invariant #6 instead of rejected — which also retires the rejection error, its three papercuts, and the frozen-record row (file restored to develop's version, undoing the reflow). |
| fba78b708 `refactor(cli)` | `--target` folded into a branch-accepting `--project-ref` (link's vocabulary, with the 20-lowercase-letters sentence); config loaded and validated before any network call with the `[remotes.*]` overlay re-applied post-resolution; UUID targets resolve without a linked parent; branch resolution under an `output.task`; UUID echoed as an identifier, not a quoted name; telemetry flush outermost with the cache write gated on a resolved ref; safeFlags aligned across the config family (ref-shape-gated on diff, plain on push, documented list extended). |
| 27b62ef0b `fix(cli)` | `--workdir` honored (diff **and** push), drift exits 2 with 1 reserved for errors, `legacySanitizeInlineName` on every non-constant string in text output, `schema_version: 1` (integer payload contract) + `config_schema` (the `$schema` URL), `scope: {present, missing}` with the block set exported from `@supabase/config` and empty records counting as not-returned, the masked/unmanaged caveats carried in the machine `message`, 404 on the config route classified as invalid input, and one spelling per concept with real pluralization. |
| 977e3001d `test(cli)` | The structural blind spots: a realistic fresh-project auth record in the shared fixture (clean end to end), the live test asserting auth-record cleanliness rather than just exit 0, `not.toContain(<secret>)` on both streams and the machine payload, and telemetry-flush/no-cache assertions on the failure paths. |
| a4dc035ba `docs(cli)` | The docs-site overlay (`docs/supabase/config/diff.md`) and ADR 0022 → accepted, with the review-driven decisions recorded (including the three it was silent on — all three changed in this round). |

Verification: `pnpm check:all` green (12/12 tasks), full CLI unit (5433) + integration (3423) suites green, `packages/api`/`packages/config` suites green, and the targeted `diff.e2e.test.ts` green against the built binary. I couldn't run the live suite from this environment (no staging credentials here) — the staging repro is covered by the new contract test with your exact `database_pool_mode` shape, and the live test now asserts the auth-record cleanliness your review flagged.

One deliberate non-change, called out inline as well: the linked-project cache write for an explicit `--project-ref` in an unlinked directory is pre-existing family-wide behavior (`backups/list`, `config push` both `Effect.ensuring(cache(ref))` with whatever ref resolved; the cache itself is fire-and-forget and only fills an absent file). Diverging in one command felt worse than the status quo — happy to open a separate issue on the family semantics if you think it should change.

Also worth noting: the three `storage.*` remote-only entries from your staging run are, as far as I can tell, genuine plan-provisioned state (`image_transformation.enabled: true` etc. on the platform vs the schema's `false` default) — a `config push` of a default file would genuinely try to change them, so they report by design (ADR 0022's "platform defaults that diverge from schema defaults surface as drift"). The row-level `unconfiguredValue` mechanism is there if a plan-independent unconfigured value emerges for them.

kanadgupta:
Develop is merged in (merge commit `7c63ce697`, no force pushes) and all 13 AI-review findings are addressed across three follow-up commits.

**The merge, and one deliberate reversal.** Develop's CLI-2234 surface trim moved the diff engine's exports (`diffProjectConfig`, the `ConfigChange*` types, `projectConfigApiBlockKeys`) onto the apps/cli-only `./internal` subpath — they stay out of the semver-covered contract until `config pull` settles the shape (`isEqualConfigValue` left the barrels entirely; only the package's own tests consume it). The upstream spec sync also renamed `V2ProjectConfigResponse` → `V2ProjectConfigResponse_Output`, stranding the 26 contract-relaxation override entries from the last review round — and rather than re-pointing them, the merge **drops the relaxation entirely**, because the AI review's forward-compatibility finding is right on the underlying architecture: ADR 0019 rule 2 says this fetch must go through `executeRaw` ("required, not incidental"), and `17254efdd` now does exactly that, following the `projects/list` pattern. That supersedes the contract-loosening we settled on in the previous round — same goal, the ADR-designed mechanism: partial responses *and* unknown enum values (a new `pool_mode`) both degrade through `@supabase/config`'s one lenient mirror, pinned by a new integration test, and the generated contract stays strict for typed-client consumers. The PR is ~800 lines smaller for it.

**Finding → commit map:**

| Finding | Commit |
| --- | --- |
| 🔴 CRITICAL: push logs any `--project-ref` verbatim | `44572340c` — the whitelist is gated on `PROJECT_REF_PATTERN` (link/diff's guard); the wiring is exported (`legacyConfigPushHandler`, link's precedent) and both directions are pinned (verbatim ref / `"<redacted>"` arbitrary string). The context-merging analytics mock those tests need existed as three identical local copies — hoisted to `tests/helpers/mocks.ts` per the hoist rule. |
| 🟠 MAJOR: branch-name resolution uses the linked branch ref as parent | `44572340c` — the lazy parent is now `legacyResolveParentScopedProjectRef` (the branches family's resolver), with a regression test where `.temp/project-ref` holds a branch ref and the lookup must hit the parent from `linked-project.json`. |
| 🟡 strict decode before lenient projection | `17254efdd` → `executeRaw` per ADR 0019 (see above). |
| 🟡 remotes configs loaded twice (duplicate warnings) | `b02b19487` — reload only when a remote's `project_id` actually matches the resolved ref, checked on the already-loaded document. |
| 🟡 set mode is multiset | `b02b19487` — true membership equality via `Set`, duplicate-membership coverage added. |
| 🟡 `target.branch: null` vs documented-optional | `b02b19487` — omitted when absent. |
| 🟡 literal NUL byte in test source | `b02b19487` — escaped `"\u0000"` spelling. |
| 🟡 push subdirectory test no longer exercises discovery | `b02b19487` — it varies `cliSettings.workdir` (what `--workdir` resolves to) instead of the now-unused mocked cwd. |
| ⚪ stale `target` flag registration | `b02b19487` — removed. |
| ⚪ e2e temp-dir leak | `b02b19487` — removed in a `finally`. |
| ⚪ formatter imports a service-bearing module for a regex | `b02b19487` — patterns hoisted to a service-free `legacy-ref-patterns.ts`. |
| ⚪ unused `LEGACY_CONFIG_DIFF_PAYLOAD_VERSION` export | `b02b19487` — consumed by the integration test asserting the payload's `schema_version`. |
| ⚪ em-dash JSON escapes in the overrides | resolved by the merge itself — the overrides file reverted to develop's version with the relaxation drop. |

Verification: `pnpm check:all` green (13/13), full CLI unit (5532) + integration (3439) suites green, `packages/config` (1119) and `packages/api` suites green, targeted `diff.e2e.test.ts` green against the built binary.

Coly010 (approved):

PR #6309, opened 2026-08-24
fix(cli): mount external scope targets

### LINKED ISSUE #35021: fail to deploy functions that use deno workspace packages
# Bug report

- [X] I confirm this is a bug with Supabase, not with my own application. 
- [X] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

functions deploy fails to deploy functions with workspace dependencies

## To Reproduce

See the following project: https://codesandbox.io/p/devbox/elastic-chatterjee-forked-6496py?workspaceId=ws_VGZMLHNu2hWPYZLrXEYbsU

you can see the supabase function in supabase/functions/example. Attempting to deploy this will fail. However `deno run index.ts` works as expected

## Expected behavior

`pnpx supabase functions deploy` deploys the function and resolves the workspace dependencies in the same way that deno does.

## System information

- Version of supabase-js: 2.20.12
- Version of Node.js: v22.14.0

## Additional context

note that it fails with the following error:
```
Relative import path "@wovenlibs/core" not prefixed with / or ./ or ../ and not in import map
```

because deno can resolve the workspace packages without a scheme. But it seems supabase cannot resolve workspace files in the same way.

7ttp:
## TL;DR

Local Function bundling now mounts the workspace paths declared under an import map's `scopes`, even when they resolve outside the Supabase project.

## what's biting the users?

`supabase functions serve`, `supabase start`, and `supabase functions deploy`
 silently dropped `scopes` targets that resolve above the project, 
so an enclosing Deno workspace never reached the Edge Runtime container and each Function died with `Relative import path "@acme/thing" ... not in import map`. 
This bites monorepos where the Supabase project is a nested repository or a submodule, and nothing was logged at any level, so the failure surfaced as a Deno resolution error rather than a missing mount....

## now fixed by: 
targets declared under `scopes` that resolve outside the project are mounted read only, and each surviving mount prints a warning naming the host path. 
Those mounts are excluded from the file watch set, so a mounted workspace directory cannot become a watch root.
 Server side bundling, reached with `--use-api` or when the Docker daemon is stopped, is unchanged and still does not carry scope targets...

## ref:
- closes: https://github.com/supabase/supabase/issues/35021
- extends: https://github.com/supabase/cli/pull/3773


apps/cli/src/shared/functions/deploy.ts
```diff
@@ -1265,10 +1267,18 @@ export async function buildDockerBinds(
...
-      await appendBindWithinRoots(importMapAllowedRoots, target);
-      if ((await stat(target)).isDirectory()) {
+      const hostPath = await realpath(target);
+      const contained = isContainedInAnyPath(importMapAllowedRoots, hostPath);
+      if (contained) {
+        extraBinds.push(`${hostPath}:${toDockerPath(hostPath)}:ro`);
+      }
+      const isDirectory = (await stat(target)).isDirectory();
+      if (!contained && kind === "scope") {
+        explicitScopeBinds.add(`${hostPath}:${toDockerPath(target)}:ro`);
```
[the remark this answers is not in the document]
7ttp: intentional
 restores the behaviour go kinda had ([#3773](https://github.com/7ttp/cli/issues/3773))
and this is local-only: serve/start already run the checkout's own code with env secrets by this point, so the repo is trusted. mounts are :ro and deduped...

7ttp:
@codex review

apps/cli/src/shared/functions/serve.ts
avallete, line 1683: **issue**

The new out-of-root scope-mount warning never reaches the user in `functions serve` or `start`: both consumers collect onWarning messages into bindWarnings and only ever inspect them for the substring "failed to read file:"

> A user runs `supabase functions serve` with a deno.json scope value like "../../secrets/"; the entire out-of-repo directory is bind-mounted read-only into the Edge Runtime container with zero visible output. The per-mount warning the PR description promises only prints for `functions deploy --use-docker` (whose onWarning forwards to stderr). start's SIDE_EFFECTS.md documents the suppression but serve's SIDE_EFFECTS.md — touched by this PR — does not, despite sharing the identical code path.

apps/cli/src/shared/functions/deploy.ts
```diff
@@ -1316,7 +1326,25 @@ export async function buildDockerBinds(
...
+    const containerPath = dockerBindContainerPath(bind);
+    if (occupiedContainerPaths.has(containerPath)) {
+      continue;
+    }
+    occupiedContainerPaths.add(containerPath);
+    uniqueScopeBinds.push(bind);
+    options.onScopeBindOutsideRoots?.(bind);
+    await (options.onWarning ?? (async () => {}))(
+      `WARN: Mounting import map scope target outside the project root: ${dockerBindHostPath(bind)}\n`,
+    );
```
avallete: **nitpick**

The PR adds a fourth inline copy of `options.onWarning ?? (async () => {})` inside buildDockerBinds (existing copies at lines 1268, 1299, 1305) instead of hoisting a single `const warn = ...` local.

apps/cli/src/shared/functions/deploy.ts
```diff
@@ -1207,6 +1207,7 @@ export async function buildDockerBinds(
   options: {
     readonly additionalModuleRoots?: ReadonlyArray<string>;
     readonly onWarning?: (message: string) => Promise<void>;
+    readonly onScopeBindOutsideRoots?: (bind: string) => void;
```
avallete, line 1210: **nipick(non-blocking) - follow-up**

Watch exclusion rides on a side callback plus byte-identical string set-subtraction across modules, with formatted bind strings re-parsed via dockerBindContainerPath/dockerBindHostPath in three places, instead of buildDockerBinds returning structured bind metadata.

> Correctness holds only while the callback's string and the returned string stay byte-identical across deploy.ts and serve.ts. Any future bind-format change (a SELinux :z flag, mode normalization, canonicalization in sanitizeDockerBinds) silently breaks `scopeBinds.has(bind)` and re-adds the external monorepo root to the watch set — the watcher blow-up this PR exists to prevent with no type error or failing test. Returning {hostPath, containerPath, mode, source} entries and formatting -v strings at the docker edge removes the class, and also subsumes the diverging inline copy of appendBindWithinRoots (deploy.ts:1270-1280) and the second dedup loop beside sanitizeDockerBinds.

apps/cli/src/shared/functions/serve.ts
```diff
@@ -1237,10 +1238,15 @@ const waitForRestartSignal = Effect.fnUntraced(function* (watchSpecs: ReadonlyAr
...
   const stream = Stream.mergeAll(
     watchSpecs.map((spec) =>
-      fileWatcher.watch(spec.root, { ignore: watchIgnoreGlobs }).pipe(
-        Stream.map((events) => events.filter((event) => eventMatchesSpec(spec, event))),
-        Stream.filter((events) => events.length > 0),
-      ),
+      fileWatcher
+        .watch(spec.root, {
+          ignore: watchIgnoreGlobs,
+          recursive: spec.matchPaths === undefined,
```
avallete: **nitpick**

recursive is re-derived from matchPaths presence at the watch call site the third place inferring meaning from that optional field instead of WatchSpec carrying an explicit recursive flag set once in buildWatchSpecs.

> A future WatchSpec variant (a glob match, or a directory spec restricted by matchPaths) updated in eventMatchesSpec but not at line 1244 silently flips recursion, surfacing as platform-dependent missed restarts (fs.watch recursive semantics differ per OS). A `readonly recursive: boolean` assigned in buildWatchSpecs gives the invariant one owner.

apps/cli/src/shared/functions/deploy.unit.test.ts
```diff
@@ -347,6 +348,163 @@ describe("buildDockerBinds — import-map key matching (spec-strict) and the fil
...
+  it("mounts an out-of-root file-valued scope target itself, warns, and does not follow its imports", async () => {
+    const workspaceRoot = await realpath(await mkdtemp(join(tmpdir(), "deploy-external-scope-")));
+    const libsDir = join(workspaceRoot, "libs", "thing");
+    const projectRoot = join(workspaceRoot, "infra", "my-project");
+    const functionsDir = join(projectRoot, "supabase", "functions");
+    const functionDir = join(functionsDir, "hello");
+    const outputDir = join(projectRoot, "out");
+    const scopeEntrypoint = join(libsDir, "mod.ts");
+    const scopeDependency = join(libsDir, "util.ts");
+    const warnings: Array<string> = [];
```
avallete, line 415: **nitpick**

New tests hand-roll fixture setup that existing helpers already provide: the out-of-root unit test rebuilds createFunctionProjectWithDenoJson's mkdtemp/mkdir/writeFile/config sequence, and the serve integration test re-implements writeProjectFile with eight manual mkdir+writeFile pairs.

> The helpers and the inline copies drift independently as the base project fixture shape evolves. The only genuinely new requirement — a .git-marked workspace root above the project dir — could be an optional parameter on createFunctionProjectWithDenoJson, and the serve test's file writes are writeProjectFile-shaped calls sitting next to that helper (serve.integration.test.ts:423).

apps/cli/src/shared/functions/serve.ts
7ttp: fixed: 91fdadb00

apps/cli/src/shared/functions/deploy.ts
7ttp: fixed: b9335b4fa

apps/cli/src/shared/functions/serve.ts
7ttp: fixed: d1b7bd4d1

avallete (approved):

PR #6315, opened 2026-08-24
refactor(config): split entrypoints into pure default, /io and /effect (CLI-2231)

Coly010:
## Summary

Splits `@supabase/config`'s public surface into three strictly layered entrypoints, per [CLI-2231](https://linear.app/supabase/issue/CLI-2231/refactor-config-package-to-default-export-node-with-effect-as-subpath) (including its 2026-08-24 amendment for browser consumers):

- **`.` — pure, browser-safe.** Schema, types, document encode functions, defaults/sparse mapping, error classes, and constants only. No Effect-returning functions are exported, and no `@effect/platform-*`, `node:`, or `bun:` module can enter its transitive graph — Studio (see supabase/supabase#48906) can bundle this platform-free. The pure document layer was physically partitioned out of `io.ts`/`functions-manifest.ts` (new `config-document.ts`, `functions-manifest-model.ts`) so the invariant is structural, and a static import-graph test (`entrypoint-purity.unit.test.ts`) enforces it: file allowlist, bare-specifier allowlist (`effect`/`smol-toml`/`dedent`), dynamic-`import()`/`require` detection, and inline-snapshot pinning of both export surfaces. Browser bundle drops 387 KB → ~355 KB with zero file-IO code reachable.
- **`./effect` — Effect-native superset.** Re-exports everything in `.` plus the Effect-native surface (`loadProjectConfig`, `saveProjectConfig`, discovery, env loading, `inferFunctionsManifest`, `ProjectConfigStore`, `projectConfigStoreLayer`, …), so Effect-native consumers need one specifier (Option B from the issue). The 22 `apps/cli` files that import Effect-native symbols were repointed with byte-identical symbol lists; the other 37 consumer files keep the bare specifier unchanged. On the issue's open decision: `resolveProjectValue`/`resolveProjectSubtree` stay Effect-typed in `./effect` only (the amendment requires `.` to carry plain-sync signatures; unwrapping can be revisited in CLI-2234).
- **`./io` — Promise facade for non-Effect consumers**, replacing the unused `./bun`/`./node` subpaths (verified zero consumers) with `bun`/`node`/`browser`/`default` export conditions. The `browser` condition resolves to a side-effect-free stub exporting the same seven names, each rejecting at invocation with a pointer to `.`, instead of dragging `@effect/platform-node` into a browser bundle (kept side-effect-free so isomorphic/edge bundles that never call config IO still build and load). All three condition targets re-export the pure surface, so `./io` consumers never need a second specifier for parameter/return types. Both facades now share one lazily built `ManagedRuntime` via a common factory — previously each call did `ManagedRuntime.make(...)` fresh and never disposed it — and the runtime layer is narrowed to `FileSystem + Path` (the former full `*Services.layer` attached a permanent `process.stdin` listener via Terminal).
- **`./schema.json`** — unchanged.

Dependency reshape: `effect` moves to `peerDependencies` — Effect v4 resolves services by string key, so this isn't about tag identity; the real grounds are a single `effect` version across the dependency graph while rc-to-rc internals churn, dedupe/bundle size, and letting the consumer own the version — and `@effect/platform-bun`/`-node` become optional peers (a consumer only needs the one matching their runtime), each mirrored in `devDependencies` for local build/test. `dedent` and `smol-toml` stay regular dependencies.

Docs: ADR 0009's open "config package architecture" bullet is resolved with the entrypoint contract, and `packages/config/AGENTS.md` records the monorepo import rule (Effect symbols → `/effect`; pure-only → `.`; `./io` is for external non-Effect consumers only).

Out of scope (tracked separately): npm publishing/compiled dist output (CLI-2232/CLI-2233) and the public-surface audit (CLI-2234).

## Linked issue

Linear: [CLI-2231](https://linear.app/supabase/issue/CLI-2231/refactor-config-package-to-default-export-node-with-effect-as-subpath)

## Checklist


packages/config/src/io-browser.ts
```diff
@@ -0,0 +1,10 @@
+// Resolved by bundlers targeting `browser` for the `@supabase/config/io`
+// subpath. There is no browser-safe file-IO implementation — browser
+// consumers must use the pure `@supabase/config` entrypoint instead.
+function unavailableInBrowser(): never {
+  throw new Error(
+    '@supabase/config/io is not available in browser bundles; import the pure surface from "@supabase/config" instead.',
+  );
+}
+
+unavailableInBrowser();
```
Coly010: **should-fix:** This stub never delivers the error message it was written for, and it breaks bundles it shouldn't. Two independently reproduced problems:

1. **Named imports are a build-time failure.** The stub exports nothing while `bun.ts`/`node.ts` export seven functions plus `ProjectConfigIo`. `bun build --target=browser` on `import { loadProjectConfig } from "@supabase/config/io"` fails with `No matching export in "io-browser.ts" for import "loadProjectConfig"` — Rollup/Vite/webpack produce equivalent generic diagnostics. The curated message pointing at the pure entrypoint is unreachable on that path.
2. **This top-level call throws at module-evaluation time.** A namespace or side-effect import bundles fine but crashes the whole entry on load, even along code paths that never touch config IO — verified with `import * as io` behind an `if (isServer)` guard. That's the worst case for exactly the isomorphic/edge consumers a `browser` condition exists for (Next edge conditions include `browser`).

The purity test only asserts the `./io` condition targets *exist on disk*, and the bun↔node export-parity check in `promise-facade.unit.test.ts` exempts this file, so the shape mismatch passes CI.

Suggested shape: export the same seven names as async functions that each call `unavailableInBrowser()` at invocation, add `export type { ProjectConfigIo } from "./promise-facade.ts";`, delete the top-level call so the module is side-effect-free, and extend the parity test to a three-way check including this file (which would have caught this).

packages/config/src/promise-facade.unit.test.ts
```diff
@@ -0,0 +1,312 @@
...
+});
+
+describe("promise-facade stdin-leak regression (CLI-2231)", () => {
+  // `BunServices.layer`/`NodeServices.layer` pull in Terminal, which attaches
+  // a permanent `process.stdin` "end" listener on first use — this facade
+  // only needs `FileSystem | Path` (see `bun.ts`/`node.ts`), so a facade call
+  // must never grow that listener count.
+  test("a facade call does not attach a process.stdin 'end' listener", async () => {
+    const cwd = makeTempProject();
+    const before = process.stdin.listenerCount("end");
```
Coly010: **should-fix:** This regression guard does not fail when the regression it names is reintroduced. Reverting `bun.ts` to `makeProjectConfigIo(BunServices.layer)` keeps this file 15/15 green: the facade is a module-level singleton whose runtime is built on the *first* facade call, and 14 earlier tests in this file already made that call — so the Terminal stdin listener is attached before `before` is snapshotted here, and the delta is 0 either way. Running the same file with `-t "stdin"` (no earlier tests) fails with `expected 1 to be +0`, which pins the ordering dependence. CI runs whole files, so the protection is illusory.

The invariant itself is real and worth guarding (`NodeTerminal.make` does `stdin.once("end", …)` at layer-build time with only a scope finalizer to remove it). Suggested fix: assert the structural property instead — build `Layer.mergeAll(BunFileSystem.layer, BunPath.layer)` in the test and assert the resulting context contains no `Terminal` — or move this assertion into its own test *file* so the measured call is genuinely the first, with a comment that ordering is load-bearing.

packages/config/src/entrypoint-purity.unit.test.ts
```diff
@@ -0,0 +1,409 @@
...
+
+// Relative specifiers in this package always carry an explicit `.ts`
+// extension, so a resolved path never needs extension probing. The
+// lookbehind excludes an occurrence of the word "from" that is itself the
+// entire contents of a quoted string argument (e.g. `missing("vonage",
+// "from")` in auth/sms.ts) — a real `from` clause is never immediately
+// preceded by a quote character.
+const fromClauseSpecifier = /(?<!["'])\bfrom\b\s*["']([^"']+)["']/g;
+const sideEffectImportSpecifier = /^\s*import\s*["']([^"']+)["']\s*;?\s*$/gm;
+const dynamicImportOrRequireSpecifier = /\b(?:import|require)\s*\(\s*["']([^"']+)["']\s*\)/g;
```
Coly010: **should-fix:** The specifier scanner has two reproducible false negatives, and false negatives are the dangerous direction here — this test is the sole enforcement of the PR's headline invariant, and both holes keep it silently green.

1. **Backtick specifiers are invisible.** `` const m = await import(`./io.ts`); `` extracts nothing — this regex's character class only accepts `["']`. A formatter-stable, perfectly valid template-literal dynamic import would pull the entire IO layer into the pure graph with every assertion still passing. Non-literal `import(expr)` is likewise ignored rather than failed.
2. **`blankOutTypeOnlyStatements` over-erases.** `export type Foo = number` (no semicolon) followed by `import { b } from "./bad.ts";` extracts nothing: `[^;]*?` runs past the semicolon-less type alias and swallows the following *real* import along with it.

Since vitest already runs under Bun here (per this package's AGENTS.md), the cleanest fix deletes most of the hand-rolled scanning: `new Bun.Transpiler({ loader: "ts" }).scanImports(src)` reproduces the exact intended semantics (verified: type-only statements excluded, mixed inline-`type` kept, `dynamic-import` and `require-call` reported) — a `ts.createSourceFile` walk works too, since `typescript` is a devDependency. If keeping the regexes, at minimum add backticks to the specifier class, tighten the type-only blanking to require a `from` clause before any statement boundary, and hard-fail on any dynamic `import(`/`require(` whose argument is not a string literal.

docs/adr/0009-configuration-schema-and-validation.md
```diff
@@ -18,8 +18,24 @@
...
+- `@supabase/config/schema.json` — the generated JSON Schema for `ProjectConfig`, a `dist/` build output.
+
+`effect` is a peer dependency, not a regular one: `ProjectConfigStore`'s `Context.Tag` identity must be the same module instance a consumer's own Effect program resolves against, which a bundled or duplicated `effect` copy would silently break. `@effect/platform-bun`/`@effect/platform-node` are optional peers — only `./io` needs a concrete platform implementation, a given consumer only ever runs on one of Bun or Node, and requiring both unconditionally would be dead weight for whichever runtime isn't in use.
```
Coly010: **should-fix:** The recorded mechanism for the `effect` peer-dependency move is factually wrong for Effect v4, and ADRs are the durable artifact future dependency decisions will cite. Service identity in v4 is the **string key**, not module-instance tag identity: `Context.ts` in the pinned source states "At runtime, services are stored by each key's string `key`" and "The string key is the runtime identity of the service" — v4 deliberately uses string TypeIds so duplicated copies interoperate. `ProjectConfigStore` (key `"@supabase/config/ProjectConfigStore"`) would resolve correctly across a duplicated `effect` install.

The *decision* is still right — restate it on grounds that are true: a single `effect` version across the dependency graph while rc-to-rc internals churn, dedupe/bundle size for a large library, and letting the consumer own the version rather than pinning it transitively. The optional-platform-peer half of this paragraph is accurate as written.

packages/config/src/bun.ts
```diff
@@ -1,67 +1,16 @@
...
+);
+
+export const loadProjectConfig = projectConfigIo.loadProjectConfig;
+export const findProjectRootFor = projectConfigIo.findProjectRootFor;
+export const findProjectPathsFor = projectConfigIo.findProjectPathsFor;
+export const loadProjectConfigFile = projectConfigIo.loadProjectConfigFile;
+export const loadProjectEnvironmentFor = projectConfigIo.loadProjectEnvironmentFor;
+export const saveProjectConfig = projectConfigIo.saveProjectConfig;
+export const loadFunctionsManifest = projectConfigIo.loadFunctionsManifest;
+export type { ProjectConfigIo } from "./promise-facade.ts";
```
Coly010, line 16: **consider:** `./io` is the one entrypoint that isn't self-sufficient. `./effect` is a strict, test-enforced superset of `.`, but `./io` exports only the seven functions plus `ProjectConfigIo` — while every parameter and return type in that interface (`LoadedProjectConfig`, `LoadProjectConfigOptions`, `SaveProjectConfigOptions`, `ProjectPaths`, `ProjectEnvironment`, `FunctionsManifest`, `ProjectConfig`, …) is only nameable from `.`. So the external, least-Effect-sophisticated audience this entrypoint exists for is forced into a dual-import pattern the package otherwise avoids.

It's free to fix: those pure modules are already in `./io`'s transitive graph, and `.`'s pinned export list collides with none of the seven facade names, so `export * from "./index.ts";` here, in `node.ts`, and in the reworked `io-browser.ts` is unambiguous. A superset assertion mirroring the existing `./effect` one would make the contract symmetric and enforced rather than prose.

packages/config/src/errors.ts
```diff
@@ -1,5 +1,5 @@
 import { Data } from "effect";
-import type { ConfigFormat } from "./io.ts";
+import type { ConfigFormat } from "./config-document.ts";
```
Coly010: **consider:** This inverts the dependency direction and creates a type-level import cycle through the new file: `errors.ts` (the package's most primitive module — `base.ts`, `project.ts`, and `io.ts` all sit above it) now points up at `config-document.ts`, which imports `type { ProjectEnvironment }` from `project.ts`, which imports `ProjectEnvParseError` from here — `errors → config-document → project → errors`. It's benign today because both back-edges are whole-statement `import type` and erase under `verbatimModuleSyntax`, but it's structurally invisible (the purity walker blanks type-only edges by design, so nothing will ever report it) and becomes a live constraint for declaration emit in CLI-2232.

`ConfigFormat` is a two-member string union with zero dependencies; moving it (and arguably `ProjectConfigValueSource`) to a leaf both `errors.ts` and `config-document.ts` import from breaks the cycle for free.

packages/config/AGENTS.md
```diff
@@ -0,0 +1,51 @@
...
+## Monorepo import rule
+
+- A file anywhere in this monorepo that needs **any** Effect-native symbol (the `ProjectConfigStore`
+  service, `loadProjectConfig`/`saveProjectConfig`, project-environment resolution, functions-manifest
+  inference, etc.) imports everything it needs from `@supabase/config/effect` — never mix that with
+  importing the same package's default entrypoint in the same file.
+- A file that needs only pure symbols (the schema, types, encoding, sparse defaults, errors) imports
+  from `@supabase/config`.
+- `@supabase/config/io` is exclusively for external consumers outside this monorepo that aren't
+  Effect-native. Do not add an internal consumer of it.
```
Coly010, line 33: **consider:** This PR introduces two monorepo-wide invariants — nothing in-repo imports `@supabase/config/io`, and any file needing an Effect-native symbol takes everything from `./effect` — and enforces neither mechanically, in a PR that otherwise builds a bespoke import-graph walker for the sibling invariant. Today's state is clean (verified: all 23 repointed files take Effect symbols from `./effect`; the remaining bare-specifier importers use only pure symbols; zero `./io` consumers), but `./io` looks like the friendliest entrypoint and this is exactly the rule a future contributor breaks first.

A small unit test next to the purity test that scans `apps/` and `packages/` for `@supabase/config/io` and deep `@supabase/config/src/…` specifiers, failing with a pointer to this rule, would make the drift loud.

Coly010 (commented):
Review of the three-entrypoint split. What was verified before commenting: `packages/config` unit tests (183/183), lint/fmt/knip, and `tsc --noEmit` clean; `apps/cli` `tsc --noEmit` clean plus targeted unit/integration runs over the repointed areas; `pnpm install --frozen-lockfile` clean with a byte-identical lockfile; a scripted scan of all 92 `@supabase/config` importers confirming zero value-imports of the 14 symbols removed from `.`, no file mixing `.` and `./effect`, and zero consumers of the removed `./bun`/`./node` subpaths; empirical `bun build --target=browser` runs against each `./io` condition; and the `ManagedRuntime` lifecycle against the pinned effect source (build fiber memoized, scope finalizers removed via `addObserver`, no accumulation — the never-disposed singleton is sound).

The layering itself is right and the partition is a verbatim move with no behavior drift. Seven inline comments below: four **should-fix** (the `browser` condition of `./io` breaks builds instead of producing the documented runtime error; the stdin-leak regression test does not fail when the regression is reintroduced; the purity walker has two reproducible false negatives; ADR 0009 records a factually wrong mechanism for the `effect` peer-dep move) and three **consider** (`./io` is not type-self-sufficient; a new type-level import cycle through `errors.ts`; the two monorepo-wide import rules are prose-only). All findings were reproduced or source-verified, not pattern-matched.

packages/config/src/io-browser.ts
Coly010: Accepted — fixed in ef5139085. `io-browser.ts` is now side-effect-free and exports the same seven names as async functions typed against `ProjectConfigIo`, each rejecting at invocation with the curated message; `export type { ProjectConfigIo }` added. The parity check is now three-way (bun/node/io-browser), and two new tests pin the module contract: a dynamic import of the module resolves without throwing, and `loadProjectConfig` rejects with the documented message. Verified the original repro: a standalone browser build with a named `{ loadProjectConfig }` import now builds successfully (141 modules) and evaluates without throwing.

packages/config/src/promise-facade.unit.test.ts
Coly010: Accepted — fixed in ef5139085. The guard moved to its own file (`promise-facade.stdin.unit.test.ts`) whose only facade interaction is the measured call, with a comment stating the first-call ordering is load-bearing and relies on vitest's per-file isolation. Mutation-verified: reverting `bun.ts` to `makeProjectConfigIo(BunServices.layer)` now fails the file with `expected 1 to be +0`. Also added your suggested structural companion: `Layer.mergeAll(BunFileSystem.layer, BunPath.layer)` builds a context with no `Terminal` service (`Context.getOption` + `Option.isNone`).

packages/config/src/entrypoint-purity.unit.test.ts
Coly010: Accepted — fixed in ef5139085. The hand-rolled stripping/blanking/regex extraction is replaced with `new Bun.Transpiler({ loader: "ts" }).scanImports(source)` (with a clear error if `Bun` is undefined). Both reproduced holes are now regression tests and pass: backtick dynamic imports are surfaced as `dynamic-import`, and the semicolon-less type alias no longer swallows the following real import (real parser, not regex). One behavior we verified empirically and hardened around: `scanImports` silently drops non-literal dynamic imports (`import(someVar)`) and interpolated template literals, so the walker now hard-fails on any dynamic `import()`/`require()` whose argument isn't a static string — checked against the transpiler's comment-free, type-erased output to avoid the old false-positive class (SQL-string and comment cases re-verified).

docs/adr/0009-configuration-schema-and-validation.md
Coly010: Accepted — fixed in ef5139085. The ADR now cites the real v4 mechanism (service identity is the string key, per Context.ts in the pinned source) and restates the peer-dep decision on true grounds: one `effect` version across the graph while rc-to-rc internals churn, dedupe/bundle size, and consumer ownership of the version. Grepped AGENTS.md and README for the same claim — neither repeats it. The PR description carried the same wrong justification; updating it as well.

packages/config/src/bun.ts
Coly010: Accepted — fixed in ef5139085. All three `./io` condition targets (`bun.ts`, `node.ts`, `io-browser.ts`) now `export * from "./index.ts"` (collision-free, re-verified), and a superset assertion mirroring the `./effect` one enforces it: every runtime export of `.` is present with an identical, non-shadowed binding in all three modules, and the three modules' full export-name sets are asserted identical to each other.

packages/config/src/errors.ts
Coly010: Accepted — fixed in ef5139085. `ConfigFormat` moved to a new zero-import leaf (`config-format.ts`); `errors.ts`, `config-document.ts`, `io.ts`, and `index.ts` all point at the leaf directly (no re-export shim through `config-document.ts`), and `errors.ts` now imports only from `effect` and the leaf — cycle gone. Scoped the move to `ConfigFormat` only: `ProjectConfigValueSource`/`Origin` aren't used by `errors.ts`, so moving them wasn't needed for the cycle and would have mixed file-format and value-provenance concepts in one leaf. One interesting consequence: the leaf never appears in the purity walker's file allowlist because every importer consumes it via `import type`/`export type`, which fully erases — zero runtime footprint, matching real bundler behavior.

packages/config/AGENTS.md
Coly010: Accepted — fixed in ef5139085. New `monorepo-import-contract.unit.test.ts` next to the purity test walks `apps/` and `packages/` sources (excluding node_modules/dist/.repos and this package's own src; forbidden strings built by concatenation) and fails naming exact file paths on any `@supabase/config/io` or `@supabase/config/src/…` import, with a pointer to AGENTS.md's rule. Scans 1,831 files in ~235ms; sanity-checked it catches a planted violation. Known limitation documented in the file: Nx caching means it only re-runs when packages/config changes.

packages/config/src/monorepo-import-contract.unit.test.ts
```diff
@@ -0,0 +1,66 @@
...
+function collectTsFiles(dir: string, into: string[]): void {
+  for (const entry of readdirSync(dir, { withFileTypes: true })) {
+    const fullPath = join(dir, entry.name);
+
+    if (entry.isDirectory()) {
+      if (EXCLUDED_DIR_NAMES.has(entry.name) || fullPath === thisPackageSrcDir) {
+        continue;
+      }
+      collectTsFiles(fullPath, into);
+    } else if (entry.isFile() && entry.name.endsWith(".ts")) {
```
avallete: Non-blocking: this is the right guard, but it will not run when the violation is actually introduced.

`test:unit` inputs are `{projectRoot}/**/*` (`tools/nx-plugins/src/test.plugin.ts`), so `nx affected` / a warm cache skips this when only `apps/cli` (or another consumer) adds `@supabase/config/io`. CI today still executes it via a cold `nx run-many`, which is why this is not a merge blocker — but that is the exact “contributor grabs the friendliest entrypoint” case the test exists for.

Smallest fix: the same `extraInputs` hatch `packages/stack` already uses in that plugin — `{workspaceRoot}/apps/**/*.ts` and `{workspaceRoot}/packages/**/*.ts`. While there, `endsWith(".ts")` also misses a future `.tsx` importer (none today).

packages/config/src/promise-facade.unit.test.ts
```diff
@@ -0,0 +1,361 @@
...
+  "loadProjectConfigFile",
+  "loadProjectEnvironmentFor",
+  "saveProjectConfig",
+];
+
+describe("promise-facade parity between bun.ts, node.ts, and io-browser.ts", () => {
+  // A prior two-way check here (bun.ts vs node.ts only) exempted io-browser.ts
+  // from parity, which is exactly why it was allowed to ship as a bare
+  // top-level `throw` exporting nothing — this three-way check closes that
+  // gap.
```
avallete: Non-blocking: this narrates a prior review (“A prior two-way check…”) rather than the current invariant. Same for the footer at 357–361 (“previously lived here”).

The three-way parity check and the dedicated stdin file already lock the behavior. A one-liner of current intent is enough — e.g. `io-browser` must export the same facade names as bun/node.

avallete (approved):
The three-entrypoint split holds on HEAD: `.` stays a physical, browser-safe partition; `./effect` is a real unshadowed superset; `./io` is a lazy Promise facade over `FileSystem | Path` only, with a side-effect-free browser stub. The seven earlier self-review threads are still fixed. CLI Effect call sites moved to `/effect`; remaining `@supabase/config` imports are schema/types/constants.

Two non-blocking nits inline — neither needs to land before merge:

1. The monorepo import-contract test will not re-run when another workspace adds a forbidden `/io` or deep import (`test:unit` inputs are project-local). Add workspace `extraInputs` the way `packages/stack` already does; scanning `.tsx` too would close the other documented hole.
2. A couple of leftover review-history comments in the facade tests can shrink to the current invariant.

Approve as-is.

packages/config/src/monorepo-import-contract.unit.test.ts
Coly010: Accepted — fixed in 7b23f744f. `packages/config:test:unit` now gets workspace-wide `extraInputs` (`{workspaceRoot}/apps/**/*.{ts,tsx}`, `{workspaceRoot}/packages/**/*.{ts,tsx}`) via the same hatch `packages/stack` uses in `test.plugin.ts`, and the walker scans `.tsx` too. Verified behaviorally, not just by inspection: warmed the cache, then added a new file under `apps/cli/src/` — the target re-ran (198 tests executed) instead of hitting the cache; a plain `touch` doesn't invalidate since nx hashes contents, so the probe used a real content change.

packages/config/src/promise-facade.unit.test.ts
Coly010: Accepted — fixed in 7b23f744f. Both comments now state the current invariant only: the parity comment says the browser stub must export every facade name so named imports exist at build time, and the footer is a one-liner pointing at `promise-facade.stdin.unit.test.ts` and why it needs its own vitest-isolated file.

PR #6327, opened 2026-08-25
ci: shard package-local test suites

jgoux:
## Summary

- Move TypeScript unit, integration, and end-to-end execution to package-local Vitest scripts orchestrated by Turbo while keeping live tests separate.
- Run unit and integration suites in independent CI jobs so integration tests can be sharded separately later.
- Apply each end-to-end shard to every package that owns an e2e suite, executing packages sequentially within a runner.
- Encode Go and CLI build prerequisites for local e2e entrypoints, then isolate Vitest passthrough arguments from those build tasks.
- Keep Go unit tests package-local and in their dedicated race-and-coverage workflow while the root Vitest fan-out remains uniformly shardable.
- Right-size the unit job to a 4-vCPU Blacksmith runner without preemptively capping workers.
- Remove the `test:core` alias and document explicit unit and integration commands.

## Context

The first e2e shard previously ran every non-CLI-harness e2e suite unsharded, making it consistently much slower than the other shards. Propagating the same Vitest shard to each package distributes those suites uniformly.

Turbo currently provides only the test task graph and prerequisite ordering. Test caching remains disabled until its inputs and outputs are designed separately.


apps/cli/package.json
```diff
@@ -32,8 +32,10 @@
     "docs:spec": "bun scripts/generate-docs-spec.ts",
     "dev:next": "pnpm exec bun src/next/main.ts",
     "dev:legacy": "pnpm exec bun src/legacy/main.ts",
-    "test": "nx run-many -t test:core test:e2e --projects=$npm_package_name",
-    "test:core": "nx run-many -t test:unit test:integration --projects=$npm_package_name --coverage.enabled",
+    "test": "pnpm run test:unit --coverage.enabled && pnpm run test:integration --coverage.enabled && pnpm run test:e2e",
+    "test:unit": "bun --bun vitest run --project unit --coverage.reportsDirectory=coverage/unit",
+    "test:integration": "bun --bun vitest run --project integration --coverage.reportsDirectory=coverage/integration",
+    "test:e2e": "bun --bun vitest run --project e2e",
```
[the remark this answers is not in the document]
jgoux: Fixed by routing package e2e entrypoints through Turbo. The graph now runs `@supabase/cli-go#build` → `supabase#build` before both CLI e2e suites; CI uses `--only` after its existing build steps. The normal root e2e help/shard path completes all five graph tasks.

jgoux:
@codex review

apps/cli-go/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@supabase/cli-go",
+  "private": true,
+  "scripts": {
+    "build": "go build -o supabase-go ."
```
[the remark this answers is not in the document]
jgoux: Handled in the updated design. `apps/cli-go` now has its own package-local `test:unit` command, while the root unit fan-out explicitly selects the TypeScript/Vitest workspaces. This keeps `--coverage.enabled` and future Vitest shard flags from being forwarded to `go test`. Go changes remain covered by the dedicated Go workflow, which runs the stronger race-and-coverage suite. The split TypeScript unit/integration jobs no longer unlock the Go keyring, and the contributor docs describe the boundary.

Coly010 (approved):

CONTRIBUTING.md
```diff
@@ -141,20 +141,19 @@ pnpm run fix:all     # run all fixers across every project
...
+| `test`             | Run the package's declared test suites                   |
+| `test:unit`        | Run unit tests                                           |
+| `test:integration` | Run integration tests where applicable                   |
+| `test:e2e`         | Run end-to-end tests where applicable                    |
 | `check:all`        | Run all check targets for this project                   |
 | `fix:all`          | Run all fix targets for this project                     |
 | `types:check`      | Type-check with `tsc --noEmit` _(inferred by Nx plugin)_ |
 
-The inferred scripts (`test:unit`, `test:integration`, `test:e2e`, `types:check`) are not declared in `package.json` — they are injected by local Nx plugins in `tools/nx-plugins/`. They are fully cached and can be discovered via `nx show project <name>`.
+The test scripts are declared in each package's `package.json`, so package-local test commands are directly discoverable and can be sharded independently. Type-checking remains an Nx-inferred target.
```
avallete, line 156: Non-blocking, but this line is ahead of the code: `tools/nx-plugins/src/test.plugin.ts` is still registered in `nx.json` and still infers `test:unit`/`test:integration`/`test:e2e` targets from every vitest config with `cache: true` (and `go.plugin.ts` still infers `test:unit` for cli-go; `nx.json` also keeps the `test`/`test:e2e` targetDefaults). So `nx run-many -t test:unit` remains a second, *cached* test path that invokes vitest directly and bypasses the Turbo orchestration — a stale-green footgun. Since nothing in CI uses those targets anymore, it'd be worth stripping the test-target inference (keeping `types:check` and project registration) here or in a fast follow-up.

.github/workflows/test.yml
```diff
@@ -158,40 +163,23 @@ jobs:
...
       # The ts-legacy/ts-next harnesses invoke `node apps/cli/dist/supabase.js`
       # with `SUPABASE_CLI_BINARY_OVERRIDE` pointing at the compiled per-shell
-      # binary in `apps/cli/dist/`. The `nx run @supabase/cli-e2e:test:e2e`
-      # target normally builds the supabase umbrella via `dependsOn`, but we
-      # bypass nx for sharding so we build it explicitly.
+      # binary in `apps/cli/dist/`. Build the CLI explicitly before invoking
+      # every package-local e2e suite.
       - name: Build CLI
-        if: steps.detect.outputs.cli_e2e == 'true'
         run: pnpm exec nx run supabase:build
```
avallete, line 175: Nit: CI builds the CLI via `nx run supabase:build` while the local `test:e2e` scripts build via `turbo run supabase#build` → `@supabase/cli-go#build`. Both work, and I get that Turbo only owns tests for now, but worth converging on one build orchestrator eventually. Related: the Nx project is named `cli-go` (go.plugin default) while the new `apps/cli-go/package.json` names it `@supabase/cli-go` — two names for the same workspace invites confusion.

avallete (approved):
LGTM — verified the risky spots and they all check out:

- The 8 `@parcel/watcher-*` platform packages land in `devDependencies` only; the published `supabase` package still ships just `eciesjs` + `jose` with `files: ["dist/supabase.js"]`, so no platform-mismatch risk for npm consumers (and this quietly fixes the `pnpm build` break on missing watcher binaries, including win32 which `supportedArchitectures` doesn't cover).
- Forwarding `--shard=N/3` to every e2e package is safe: all three e2e-owning vitest configs set `passWithNoTests: true` (packages/stack only has 4 e2e files).
- Dropping the keyring-unlock step from test.yml is correct: `cli-go-ci.yml` keeps the unlock and its `merge_group` trigger has no path filter, so Go tests still gate every merge even when a PR doesn't touch `apps/cli-go/`.
- No test files orphaned by the removed vitest project stanzas (api/config have zero integration/e2e files, process-compose zero e2e), and every package with unit tests declares `test:unit:run`.
- `test-summary` preserves the branch-protection check name and mirrors the existing `test-e2e-summary` skipped-handling pattern with identical `if` conditions.
- No stale `test:core` / `nx affected` / `NX_BASE` references remain.

Two non-blocking comments inline about leftover Nx test-target inference and the dual build orchestrators.

CONTRIBUTING.md
jgoux: Addressed in 58fd79378. Removed the obsolete Vitest inference plugin, the Nx test and test:e2e defaults, and Go test:unit inference. Ordinary test targets now come only from package scripts that invoke Turbo; test:live is the sole test target that retains Nx orchestration. CONTRIBUTING.md no longer advertises Nx for ordinary tests.

.github/workflows/test.yml
jgoux: Keeping the CI build on Nx is intentional while this PR scopes Turbo to tests. Package-local e2e scripts use the Turbo build dependency so clean local runs prepare artifacts, whereas CI already handles the Go binary cache/build before its existing Nx CLI build. I also verified the resolved Nx graph: the canonical project name is @supabase/cli-go, and both cli-go and @supabase/cli-go selectors resolve to that same project. Build-orchestrator convergence remains part of the eventual Nx removal.

PR #6328, opened 2026-08-25
refactor(config): rename config document type ProjectConfig to CliConfig (CLI-2235)

Coly010:
## Summary

Repo-wide vocabulary rename settling `@supabase/config`'s core naming before first npm publish ([CLI-2235](https://linear.app/supabase/issue/CLI-2235/rename-config-document-type-projectconfig-cliconfig-free-projectconfig)).

One prefix rule across the package: **`Cli*` = the local checkout side; bare `Project*` = the hosted Supabase project.** After this PR no exported name uses a bare "project" to mean the local checkout, and `ProjectConfig` is freed for the hosted-project subset type the API-mapping work (CLI-2230 / CLI-2156) will introduce. Renames are only free while the package is `private: true`, which is why this blocks CLI-2169 (first publish).

No functional behavior changes: the config-file surface (`project_id`, section names), the generated `schema.json`, and the schema URL value are unchanged. **Four renamed error tags are user-visible and change deliberately with this release**: the stderr and `--output-format json` error code changes (`ProjectConfigParseError`→`CliConfigParseError`, `ProjectEnvParseError`→`CliProjectEnvParseError`, `MissingProjectConfigValueError`→`MissingCliConfigValueError` on malformed-config/env failures, and `ProjectHomeNotDirectoryError`→`CliProjectHomeNotDirectoryError` when `.supabase/` is blocked by a file or symlink loop), and the telemetry `error_fingerprint` series fork accordingly (`tag:<old>` goes flat, `tag:<new>` starts) — annotate dashboards/alerts keyed on the old values at release. A `normalize-error` unit test now pins the user-facing codes so the next such rename is a visible review event. Service tag strings were renamed with their services (app-internal identity keys) and moved out of the package's namespace: `supabase/config/*` → `supabase/cli/*` for all apps/cli-owned services.

ADRs 0009/0018/0019 keep their pre-rename symbol names as historical records — the tables below are the mapping.

### Step 0 — free the `CliConfig` name (apps/cli)

| Old | New |
| --- | --- |
| `CliConfig` (service, tag `supabase/config/CliConfig`) | `CliSettings` (tag `supabase/cli/CliSettings`) |
| `LegacyCliConfig` (tag `supabase/legacy/CliConfig`) | `LegacyCliSettings` (tag `supabase/legacy/CliSettings`) |
| `cliConfigLayer`, `mockCliConfig`, `cliConfig` vars, … | `cliSettingsLayer`, `mockCliSettings`, `cliSettings`, … |
| `cli-config.service.ts` / `cli-config.layer.ts` / `legacy-cli-config.*` | `cli-settings.*` / `legacy-cli-settings.*` |

### Config-document family (`ProjectConfig*` → `CliConfig*`)

| Old | New |
| --- | --- |
| `ProjectConfig`, `ProjectConfigSchema`, `ProjectConfigJson`, `toProjectConfigJsonSchema` | `CliConfig`, `CliConfigSchema`, `CliConfigJson`, `toCliConfigJsonSchema` |
| `loadProjectConfig`, `loadProjectConfigFile`, `saveProjectConfig`, `LoadedProjectConfig`, `Load/SaveProjectConfigOptions` | `loadCliConfig`, `loadCliConfigFile`, `saveCliConfig`, `LoadedCliConfig`, `Load/SaveCliConfigOptions` |
| `encodeProjectConfigToJson/Toml`, `projectConfigValueSourceAt`, `ProjectConfigValueOrigin/Source` | `encodeCliConfigToJson/Toml`, `cliConfigValueSourceAt`, `CliConfigValueOrigin/Source` |
| `ProjectConfigParseError`, `MissingProjectConfigValueError` | `CliConfigParseError`, `MissingCliConfigValueError` |
| `ProjectConfigStore`, `projectConfigStoreLayer` (+ file renames) | `CliConfigStore`, `cliConfigStoreLayer` (`cli-config.service.ts` / `cli-config.layer.ts`) |
| `PROJECT_CONFIG_SCHEMA_URL` (value unchanged) | `CLI_CONFIG_SCHEMA_URL` |
| `getDefaultProjectConfig`, `subtractProjectConfig` | `getDefaultCliConfig`, `subtractCliConfig` |
| compounds re-grepped post-CLI-2231: `writeProjectConfig`, `decodeProjectConfig`, `ProjectConfigIo`, `loadFunctionsProjectConfig`, `renderProjectConfigTemplate`, `BaseProjectConfig`, `SparseProjectConfig`, … | same names with `CliConfig` substituted |

### Local-checkout family (`Project*` → `CliProject*`)

| Old | New |
| --- | --- |
| `ProjectEnvironment`, `loadProjectEnvironment(For)`, `LoadProjectEnvironmentOptions`, `LiveProjectEnvironment` | `CliProjectEnvironment`, `loadCliProjectEnvironment(For)`, … |
| `ProjectPaths`, `findProjectPaths(For)`, `FindProjectPathsOptions`, `findProjectRoot(For/FromRepoState)` | `CliProjectPaths`, `findCliProjectPaths(For)`, …, `findCliProjectRoot(…)` |
| `ProjectEnvParseError` | `CliProjectEnvParseError` |
| `resolveProjectValue(AtPath)`, `resolveProjectSubtree`, `ResolvedProjectValue`, `ResolveProjectOptions` — config family: they resolve `env()` refs in config values | `resolveCliConfigValue(AtPath)`, `resolveCliConfigSubtree`, `ResolvedCliConfigValue`, `ResolveCliConfigOptions` |
| `ProjectContext` (apps/cli, tag `supabase/config/ProjectContext`) + `projectContextLayer(For)`, `mockProjectContext`, … + file renames | `CliProjectContext` (tag `supabase/cli/CliProjectContext`) + `cliProjectContextLayer(For)`, `mockCliProjectContext`, … (`cli-project-context.*`) |
| `ProjectHome` (apps/cli) + `projectHomeLayer(For)`, `ProjectHomeNotDirectoryError`, … + file renames | `CliProjectHome` (tag `supabase/cli/CliProjectHome`) + `cliProjectHomeLayer(For)`, … (`cli-project-home.*`) |
| `ProjectLocalServiceVersions` (apps/cli) + layer/mocks + file renames | `CliProjectLocalServiceVersions` (tag `supabase/cli/CliProjectLocalServiceVersions`) (`cli-project-local-service-versions.*`) |
| `projectCommandBaseLayer`, `provideProjectCommandRuntime` (apps/cli) | `cliProjectCommandBaseLayer`, `provideCliProjectCommandRuntime` |
| `LoadCliConfigOptions.projectEnv` (published options field) + the `resolveCliConfig*` resolvers' `projectEnv` param | `cliProjectEnv` (apps/cli-internal `CliProjectContext.projectEnv` deliberately keeps its name — filed on CLI-2234) |

### Deliberately unchanged

- `DuplicateRemoteProjectIdError` / `InvalidRemoteProjectIdError` — "project id" there genuinely means the hosted project ref (the *new* meaning of "project"). (`remoteProjectConfig`, initially kept under the same carve-out, was renamed to `remoteCliConfigBlock` per review — it binds the reserved `ProjectConfig` spelling for what is a `CliConfig` sub-document.)
- `packages/api` generated names (`v2GetProjectConfig`, …) — hosted-project endpoint operations.
- Legacy-internal `Legacy*` / `legacy*ProjectEnv*` / `LegacyLocalProjectContext` symbols — already namespaced by their `Legacy` prefix and pending deletion with the legacy shell.
- Go-source provenance comments (`getProjectRoot`, `internal/utils/misc.go`) and `apps/cli-go/**`.

### Docs

`packages/config/docs/project-config-loading.md` — the doc that defined the old vocabulary — is rewritten as `cli-config-loading.md` around the final naming, with stale behavioral claims corrected against current source along the way:

- missing `env(NAME)` resolution does not fail with `MissingProjectEnvVarError` (that class doesn't exist); the resolvers leave unresolved literals as plain strings
- `loadCliConfig` substitutes most `env(NAME)` refs eagerly at load; only `functions.<name>.env.*` stays literal for on-demand resolution
- `supabase init` writes a commented `config.toml` template, not a `$schema`-only `config.json`
- managed-stack state lives under global `SUPABASE_HOME`, not repo-local `.supabase/`

## Review round 1 (2026-08-25)

Addressed from the multi-pass review:

- **Truthful user-visible statement** (above) + `normalize-error` unit test pinning the error codes.
- **Finished the vocabulary in `apps/cli/src/next/config/`**: `CliProjectHome`, `CliProjectLocalServiceVersions`, `cliProjectCommandBaseLayer`/`provideCliProjectCommandRuntime`. Remaining bare `Project*` there (`ProjectLinkRemote`, `ProjectLinkState`, `ProjectRefResolver`) describe the hosted project or the link to it — now stated in the vocabulary doc.
- **Retagged all apps/cli services out of the package's namespace**: `supabase/config/*` → `supabase/cli/*` (each tag literal verified single-occurrence).
- **`remoteProjectConfig` → `remoteCliConfigBlock`**; `LoadCliConfigOptions.projectEnv` → `cliProjectEnv` (+ resolver params).
- **ADR 0009/0018**: dated rename addenda (old→new tables + pointer to the vocabulary doc); ADR bodies untouched.
- **Docs**: README vocabulary blockquote; vocabulary doc gains the `/io` facade section, `LegacyCliSettings` bullet, `project_id` dual-meaning note, corrected `SUPABASE_HOME` keying + `identity.json` inventory; CLI-2110 plan doc symbols updated; `packages/stack/README.md` `cliSettings.supabaseHome`; stale comment fixes (`paths.ts`, `project.unit.test.ts`, `push.handler.ts`, `run.ts` defect strings).

Filed as explicit pre-publish decisions rather than rushed here: `/io` facade naming convention → [CLI-2234](https://linear.app/supabase/issue/CLI-2234); `BaseCliConfig` as the hosted-diff operand type → [CLI-2230](https://linear.app/supabase/issue/CLI-2230).

## Linked issue

[CLI-2235](https://linear.app/supabase/issue/CLI-2235/rename-config-document-type-projectconfig-cliconfig-free-projectconfig)

- [x] I'm a Supabase maintainer working from a Linear ticket.
- [x] The PR title follows [Conventional Commits](https://www.conventionalcommits.org/).


packages/config/src/errors.ts
```diff
@@ -1,7 +1,7 @@
 import { Data } from "effect";
 import type { ConfigFormat } from "./config-format.ts";
 
-export class ProjectConfigParseError extends Data.TaggedError("ProjectConfigParseError")<{
+export class CliConfigParseError extends Data.TaggedError("CliConfigParseError")<{
```
Coly010, line 4: **[should-fix] These three renamed tag literals are user-visible — the PR body's "all user-visible output are unchanged" is false.**

The tag strings changed, not just the class names (`CliConfigParseError` here, `CliProjectEnvParseError` L51, `MissingCliConfigValueError` L56), and they surface verbatim in three places:

- **stderr** — reproduced on this branch: a malformed `config.toml` + `supabase status` prints `failed to read config: CliConfigParseError` (was `ProjectConfigParseError`), via `String(cause)` at `apps/cli/src/legacy/shared/legacy-local-project-context.ts:86,100,179` and `gen.signing-keys-config.ts:498,502`. Blast radius: `start`/`stop`/`status`/`gen signing-key`/`functions`/db-bootstrap. Same for `CliProjectEnvParseError` on a malformed `supabase/.env`.
- **`--output-format json`** — `normalize-error.ts:173-174` uses `_tag` as `code` and falls back to it for `message` (none of the three carries a `message`/`detail` field), so the JSON envelope changes too.
- **telemetry** — `error_fingerprint` is `` `tag:${_tag}` `` (`error-actionability.ts:746-752`), so three PostHog series fork at this release: `tag:ProjectConfigParseError`→`tag:CliConfigParseError`, `tag:ProjectEnvParseError`→`tag:CliProjectEnvParseError`, `tag:MissingProjectConfigValueError`→`tag:MissingCliConfigValueError`. Any dashboard/alert keyed on the old values goes flat with a new series beside it.

The rename is the right call at the right time (pre-publish, internal identifiers) — but it should be a stated decision, not a silent side effect. Ask: correct the PR body ("one stderr/JSON error-code substring changes on malformed-config failures; three fingerprint transitions: old→new") so the telemetry owner can annotate before release day.

Also worth noting: no test pins the user-facing string — `failed to read config:` has zero matches in any test file, which is why this slipped through green CI. A `normalize-error.unit.test.ts` case pinning `code: "CliConfigParseError"` would make the next such rename a visible review event. (Separate follow-up: `failed to read config: CliConfigParseError` is unhelpful either way — the error already carries `path`/`format`/`cause`; rendering those would stop the tag being load-bearing forever.)

packages/config/docs/cli-config-loading.md
```diff
@@ -0,0 +1,349 @@
...
+  access token, telemetry flags, `supabaseHome`, `noKeyring`, `debug`). Lives in `apps/cli`, not
+  this package.
+- `CliProjectEnvironment`: the merged env map for the active project (`supabase/.env` +
+  `.env.local` + ambient `process.env`).
+- `CliProjectPaths`: the discovered filesystem paths for the active project.
+- `CliProjectContext`: apps/cli's runtime bundle of discovered paths + merged env for the active
+  project.
+
+The `Cli*` prefix names the local checkout side — what the CLI reads, writes, or resolves about
+itself on disk. A bare `Project*` name is reserved for the hosted Supabase project side.
```
Coly010: **[should-fix] This rule is stated absolutely, but sibling services in `apps/cli/src/next/config/` break it.**

Bare `Project*` names still bound to pure local-checkout concepts: `ProjectHome` (tag `supabase/config/ProjectHome`; shape is `projectRoot`/`supabaseDir`/`projectHomeDir`), `ProjectLocalServiceVersions` (reads `.supabase/local-versions.json`), plus `projectCommandBaseLayer`/`provideProjectCommandRuntime`. `project-runtime.layer.ts` now names the same local-checkout layer graph in both vocabularies in one 40-line file (`discoveredCliProjectContextLayer` next to `discoveredProjectHomeLayer`), and `project-home.layer.ts:10` defines `findCliProjectRootFromRepoState` inside the `ProjectHome` service.

(`ProjectLinkRemote`/`ProjectRefResolver` are genuinely hosted — fine. `ProjectLinkState` is a defensible gray area.)

Either finish the rename (`CliProjectHome`, `CliProjectLocalServiceVersions`, `cliProjectCommandBaseLayer`) or scope this sentence and list the known exceptions with a ticket (CLI-2234?). An absolute rule that the next contributor immediately sees broken in the neighbouring directory reads as aspirational.

packages/config/AGENTS.md
```diff
@@ -1,30 +1,30 @@
 # @supabase/config
 
 Supabase project configuration package built on Effect V4 Schema — owns the canonical
-`ProjectConfig` schema, config file loading/saving, and JSON Schema generation.
+`CliConfig` schema, config file loading/saving, and JSON Schema generation.
 
 ## Entrypoints
 
 Three entrypoints plus a generated artifact (see ADR 0009's 2026-08-24 decision for the full
```
Coly010, line 8: **[should-fix] This routes readers to ADR 0009 as the authoritative rationale, but this PR doesn't touch `docs/` and that ADR's 2026-08-24 decision section now specs a surface that no longer exists.**

ADR 0009:25-30 is written in present tense as a live interface spec: `ProjectConfigSchema`, `encodeProjectConfigToJson`, `loadProjectConfig`/`saveProjectConfig`, `ProjectConfigStore`/`projectConfigStoreLayer` — and line 30 documents the runtime service key as `"@supabase/config/ProjectConfigStore"` when the actual key is now `"@supabase/config/CliConfigStore"`. That string is the crux of the effect-as-peer-dep argument, so its exactness matters.

ADR 0018 is worse: 8 uses of `ProjectConfig` in the *old* sense ("a locally decoded `ProjectConfig`", `SparseProjectConfig`/`BaseProjectConfig` at :20/:24) while ADR 0019 next door already uses `ProjectConfig` in the *new* hosted-subset sense. One word, two opposite meanings across adjacent ADRs — and 0018 is the design reference for exactly the tickets (CLI-2156/CLI-2064) that will introduce the hosted `ProjectConfig`; a reader there will conclude the subtraction core is typed on the hosted subset when it's `BaseCliConfig`.

Suggest appending a dated rename note (old→new table + pointer to `docs/cli-config-loading.md`) to 0009 and 0018 in this PR. Not a rewrite — preserves the historical record, kills the trap. The "ADRs are immutable" defence covers rationale, not a cross-linked live interface spec.

packages/config/src/config-document.ts
```diff
@@ -86,8 +86,8 @@ export interface LoadProjectConfigOptions {
    * also reads `.env.<SUPABASE_ENV>` files) resolve it themselves and pass it in
    * so loading does not re-read those files or depend on `process.env` mutation.
    */
-  readonly projectEnv?: ProjectEnvironment;
-  /** See {@link FindProjectPathsOptions.search}. */
+  readonly projectEnv?: CliProjectEnvironment;
```
Coly010: **[pre-publish judgment] `projectRef` (hosted, line 80) and `projectEnv` (local checkout) now sit nine lines apart on a published options interface — the exact ambiguity this PR exists to eliminate.**

The type annotation already spells `CliProjectEnvironment`, so the field name is the only thing left in the old vocabulary. The same name recurs on `CliProjectContext.projectEnv` (`apps/cli/src/next/config/cli-project-context.service.ts:7`) and as the positional parameter of `resolveCliConfigValue`/`resolveCliConfigSubtree` (`project.ts:399,417`).

Interface property names are effectively frozen after first publish. Consider `cliProjectEnv` (or `environment`) now; the positional resolver params are lower stakes but nearly free while you're in here.

packages/config/src/node.ts
```diff
@@ -1,21 +1,21 @@
...
 
-export const loadProjectConfig = projectConfigIo.loadProjectConfig;
-export const findProjectRootFor = projectConfigIo.findProjectRootFor;
-export const findProjectPathsFor = projectConfigIo.findProjectPathsFor;
-export const loadProjectConfigFile = projectConfigIo.loadProjectConfigFile;
-export const loadProjectEnvironmentFor = projectConfigIo.loadProjectEnvironmentFor;
-export const saveProjectConfig = projectConfigIo.saveProjectConfig;
-export const loadFunctionsManifest = projectConfigIo.loadFunctionsManifest;
-export type { ProjectConfigIo } from "./promise-facade.ts";
+export const loadCliConfig = cliConfigIo.loadCliConfig;
```
Coly010, line 9: **[pre-publish judgment] The `/io` facade names follow no rule a consumer can infer.**

Three are identical to `/effect` (`loadCliConfig`, `saveCliConfig`, `loadCliConfigFile` — same import name returns `Promise` or `Effect` depending only on subpath), three take a `For` suffix (`findCliProjectRootFor`, `findCliProjectPathsFor`, `loadCliProjectEnvironmentFor`), and one renames the verb (`loadFunctionsManifest` vs `/effect`'s `inferFunctionsManifest`).

Pre-existing from CLI-2231 — but this PR is explicitly the pre-publish naming settlement, and these seven names are the entire surface a non-Effect consumer types. Pick one convention (drop `For` everywhere, or apply it everywhere; align the manifest verb either way), here or as a filed decision that lands before first publish.

packages/config/src/sparse.ts
```diff
@@ -22,26 +22,26 @@ type DeepPartial<T> =
...
- * The root-scope fields of a {@link ProjectConfig}, without the nested
+ * The root-scope fields of a {@link CliConfig}, without the nested
  * `remotes` record. Subtraction accepts this shape to keep `remotes` out of
  * its contract: the operands are root-scope *effective* configs — e.g. the
  * merged base config, or a branch's effective config translated from the
  * Management API, which has no `remotes` of its own — so neither operand has
  * to fabricate a `remotes` field to type-check.
  */
-export type BaseProjectConfig = Omit<ProjectConfig, "remotes">;
+export type BaseCliConfig = Omit<CliConfig, "remotes">;
```
Coly010, line 35: **[pre-publish judgment — the one expensive-to-reverse finding in this review] `BaseCliConfig` is the operand type of the CLI↔hosted comparison core, and this docstring already names the future hosted config as an operand** ("a branch's effective config translated from the Management API").

Two consequences worth deciding before publish rather than after:

1. Under the vocabulary rule this PR installs, a `Cli`-prefixed type designed to accept a hosted-project value is a contradiction hiding in a docstring.
2. Concretely: if `ProjectConfig` lands as a genuine *subset* (no `studio`/ports/`edge_runtime`, per the vocabulary doc), it will **not** be structurally assignable to `BaseCliConfig`, and CLI-2156 needs either a widening cast (banned by the no-`as` rule) or a signature change to the published `subtractCliConfig`/`omitDefaultValues` (`:193`, `:221`).

Options: a family-neutral operand name (`ConfigBase`/`EffectiveConfig`), or keep `BaseCliConfig` and commit to `ProjectConfig` being a *widening* of the shared root scope. Either is fine — but it should be a decision, not a discovery during CLI-2156.

apps/cli/src/next/config/cli-settings.service.ts
```diff
@@ -16,6 +16,6 @@ interface CliConfigShape {
   readonly doNotTrack: Option.Option<string>;
 }
 
-export class CliConfig extends Context.Service<CliConfig, CliConfigShape>()(
-  "supabase/config/CliConfig",
+export class CliSettings extends Context.Service<CliSettings, CliSettingsShape>()(
+  "supabase/config/CliSettings",
```
Coly010: **[should-fix, free today] `CliSettings` lives in apps/cli but registers under `"supabase/config/CliSettings"` — one `@` away from the package's real `"@supabase/config/CliConfigStore"`, and under the new vocabulary the namespace claim is precisely backwards: this service holds no `CliConfig`.**

Effect v4 service identity *is* the tag string (ADR 0009:30), so this is a live runtime identifier in a namespace owned by a package about to publish. Both this and `"supabase/config/CliProjectContext"` are apps/cli-internal (verified: each tag literal appears exactly once, at its definition site — no comparisons, no persistence), so retagging to `supabase/cli/CliSettings` / `supabase/cli/CliProjectContext` costs nothing now and removes the near-collision permanently.

apps/cli/src/legacy/commands/config/push/push.handler.ts
```diff
@@ -123,20 +123,20 @@ export const legacyConfigPush = Effect.fn("legacy.config.push")(function* (
     // 1. Load config.toml (TOML parse error aborts before any network call).
     //
     // NOTE (CLI-1489): `config push` needs the fully decoded config (every
-    // service subset), so it uses `loadProjectConfig` rather than the tolerant
-    // `LegacyProjectConfig` subtree reader. `loadProjectConfig` raises
-    // `ProjectConfigParseError` on `env(...)` refs over numeric/bool fields,
+    // service subset), so it uses `loadCliConfig` rather than the tolerant
+    // `LegacyCliConfig` subtree reader. `loadCliConfig` raises
```
Coly010: **[nit] `LegacyCliConfig` names a symbol that doesn't exist** — pre-existing (`LegacyProjectConfig` was equally dangling), but the renamed spelling is now actively confusing: it's one cluster away from `LegacyCliSettings`, which is a real and unrelated service. The thing being referenced is the tolerant `legacy-db-config.toml-read.ts` subtree reader — name that instead.

packages/config/README.md
```diff
@@ -1,46 +1,47 @@
 # @supabase/config
 
-Supabase project configuration package built on Effect V4 Schema.
+Supabase project configuration package built on Effect V4 Schema — owns the canonical `CliConfig`
```
Coly010, line 3: **[should-fix, given "before first npm publish" is this PR's whole premise] The npm landing page never explains the one thing a newcomer will ask.**

"owns the canonical `CliConfig` document schema" is exactly where an external consumer thinks *"CliConfig? I'm not writing a CLI, I'm reading my project's config.toml."* The vocabulary rule only exists in `docs/cli-config-loading.md`, linked at the bottom under Architecture Docs. Three lines near the top would carry the whole scheme:

```md
> `CliConfig` is the config *document* (`supabase/config.toml`/`.json`) — the full local
> superset the CLI reads and writes. `ProjectConfig` is reserved for the hosted-project
> subset (not implemented yet). `CliSettings` is the CLI's own runtime settings and lives
> in the CLI, not this package.
```

packages/config/src/index.ts
```diff
@@ -7,28 +7,28 @@
...
+  type CliConfig,
+  type CliConfigJson,
 } from "./base.ts";
 export {
   DuplicateRemoteProjectIdError,
   InvalidRemoteProjectIdError,
-  MissingProjectConfigValueError,
-  ProjectConfigParseError,
-  ProjectEnvParseError,
+  MissingCliConfigValueError,
```
Coly010, line 20: **[nit]** This export block was alphabetical before the rename; the renamed entries kept their old positions (`CliConfigParseError`/`CliProjectEnvParseError` now sort after `Duplicate…`/`Invalid…`/`Missing…`). Purely cosmetic.

Coly010 (commented):
Multi-pass review of the CLI-2235 rename (engineer / architect / DX lenses), run against head `b5c88da72`.

## Verdict

**The mechanical rename is verifiably clean — mergeable as-is on the code.** The asks are: make the PR body truthful about the three user-visible error tags (see `errors.ts` comment), reconcile the docs/ADRs that now contradict the tree, and settle four pre-publish naming judgment calls while they're still free.

## What was verified (all passing)

- **Reverse-rename proof**: replayed all 99 symbol mappings + 12 file renames in reverse over HEAD and diffed against the merge base — only 5 files had residual content (all intentional or flagged inline), and exactly **one** non-symbol string literal changed in the entire 403-file diff (a test temp-dir prefix, `cli-settings.layer.unit.test.ts:15` — flagging only so it's a conscious choice). The reflow commit is genuinely formatting-only.
- **Identifier census balances exactly** (e.g. `loadProjectConfig` 167→0 / `loadCliConfig` 0→167; `mockLegacyCliConfig` 313→0 / `mockLegacyCliSettings` 0→313). Zero stragglers of old names outside the declared carve-outs; carve-outs (`apps/cli-go`, generated `packages/api`, `RemotesSchema`, `Legacy*`) respected. The stringly-typed `Effect.catchTag("CliProjectEnvParseError", …)` in `secrets/set.handler.ts:262` was correctly updated.
- **`ProjectConfig` is genuinely free**: no exported binding on any entrypoint (`.`, `/io` node+bun+browser, `/effect`); export-surface snapshots are **set-equal** to the base after reverse-rename — nothing added, removed, or mixed; reordering is the correct consequence of `.sort()`.
- **Generated `schema.json` is byte-identical** between merge base and HEAD (built both, diffed: 5,975 lines identical); `$schema` URL value unchanged; zero changes to any `.toml`/`.json`/`.snap` fixture; `supabase init` template untouched.
- **Checks**: types:check across config/cli/e2e/test-helpers/stack, lint, fmt, knip all green; 198/198 config unit, 5,059 CLI unit, 3,251 integration tests pass.
- **Doc fact-check**: all four corrected behavioral claims in `cli-config-loading.md` verified true against source — unresolved `env()` stays literal (`project.ts:341-343`), eager-vs-deferred interpolation (`io.ts:531` / `lib/env.ts:29`, only `functions.<name>.env.*` deferred), `init` writes a commented `config.toml` (`project-init.ts:308`), managed state under global `SUPABASE_HOME`. Also confirmed `MissingCliConfigValueError` is constructed nowhere, as the doc now says.

## Findings not anchorable inline (files outside the diff)

- `packages/stack/README.md:48` — `cliConfig.supabaseHome` missed by the sweep (`supabaseHome` is a `CliSettings` field → `cliSettings.supabaseHome`).
- `docs/adr/0009-configuration-schema-and-validation.md:25-30` + `docs/adr/0018-sparse-config-subtraction.md` — need dated rename notes; details on the `AGENTS.md` comment.
- `docs/superpowers/plans/2026-08-16-cli-2110-managed-port-allocation.md:78,113-134` — live plan (0/44 steps checked) still instructs an implementer to create `ProjectConfigValueSource` / `projectConfigValueSourceAt` and extend `LoadedProjectConfig` — all renamed by this PR. Rename the symbols or mark it superseded.
- `apps/cli/src/shared/cli/run.ts:712-714` — the `CliConfigStore` fallback stub's defect strings still say "unexpected root project config access / file access / write"; these can surface in a crash report naming the freed concept.
- `packages/config/src/base.ts:43` — internal `const remoteProjectConfig` binds the reserved identifier for what is a `CliConfig` sub-document (leaks into a comment at `io.ts:319`). Not exported, so low cost — but it's the one name the PR reserves. Suggest `remoteCliConfigBlock`.

## Severity summary

- **Should-fix (in this PR)**: errors.ts claim/telemetry statement · vocabulary-rule exceptions (`ProjectHome`, `ProjectLocalServiceVersions`) · ADR 0009/0018 rename notes · `cliConfig.workdir` comments (×2) · README vocabulary note · `CliSettings`/`CliProjectContext` tag namespace.
- **Pre-publish judgment calls (decide before first publish, not necessarily here)**: `LoadCliConfigOptions.projectEnv` field name · `/io` facade naming convention · `BaseCliConfig` as the hosted-diff operand type (the one that's expensive to reverse).
- **Nits**: everything else inline + the unanchorable list above.

packages/config/src/paths.ts
```diff
@@ -67,9 +67,9 @@ export interface FindProjectPathsOptions {
   readonly search?: boolean;
```
Coly010, line 67: **[should-fix, two words — actual site is line 59, just above this hunk]** The comment "the legacy `stop`/`status` ports' `cliConfig.workdir`" was correct pre-rename, when `cliConfig` was the settings service. Post-rename, `cliConfig` means the config document — which has no `workdir`; the field lives on `CliSettings` (`legacy-cli-settings.service.ts:41`). Should read `cliSettings.workdir`.

Twin occurrence in the same wording: `packages/config/src/project.unit.test.ts:57`.

packages/config/docs/cli-config-loading.md
```diff
@@ -0,0 +1,349 @@
...
+
+## CLI-owned Repo State
+
+The CLI also keeps machine-local project state outside `@supabase/config`'s scope, in two places
+(see `apps/cli/docs/supabase-home.md` for the full layout):
+
+- a repo-local `.supabase/` directory, sibling to `supabase/`, holding checkout-specific caches:
+  linked remote project metadata (`project.json`) and checkout-local service-version overrides
+  (`local-versions.json`)
+- the global `SUPABASE_HOME` directory, holding managed-stack metadata and runtime state, keyed by
```
Coly010, line 323: **[nit] "keyed by the discovered project root" overstates it.** Per `apps/cli/docs/supabase-home.md:67-69`, stack identity is the canonical local-project key *relative to the enclosing Git checkout* plus workspace identity — not the discovered project root (nearest `supabase/config.*` ancestor). Tighten or drop the "keyed by" clause. Also, the `.supabase/` inventory at 281-283 omits `identity.json` (written for non-Git folders, `supabase-home.md:45`).

packages/config/docs/cli-config-loading.md
```diff
@@ -0,0 +1,349 @@
+# CLI Config Loading
+
+This document explains how the CLI's on-disk config document loading works, across
+`@supabase/config` and the CLI.
+
+## Vocabulary
+
+- `CliConfig`: the persisted config-file document (`supabase/config.toml` / `supabase/config.json`)
```
Coly010, line 8: **[nit — completeness of the canonical vocabulary reference]** Three gaps for the doc that is now the only place the naming rule is written down:

1. **`/io` is never mentioned**, yet it's the entrypoint whose names *diverge* (`findCliProjectRootFor`, `findCliProjectPathsFor`, `loadCliProjectEnvironmentFor`, and `loadFunctionsManifest` vs `/effect`'s `inferFunctionsManifest`). A vocabulary reference that skips one of three published entrypoints — the one with the exceptions — undercuts "canonical". A short `/io` section listing the seven facade names would close it.
2. **Glossary omits `LegacyCliSettings`** — the seventh name a grepping reader immediately hits (`apps/cli/src/legacy/config/legacy-cli-settings.service.ts`). One bullet: same role as `CliSettings`, legacy shell, pending deletion.
3. Worth one sentence on **`project_id`'s dual meaning** — the most user-visible instance of the ambiguity this doc exists to settle: root `project_id` is *local* (defaults to the directory name, `base.ts:16-19`) while `remotes.*.project_id` is the *hosted* project ref (`base.ts:24`).

packages/config/src/errors.ts
Coly010: Accepted. The PR body now states the three user-visible transitions explicitly (stderr + `--output-format json` code, and the `tag:<old>`→`tag:<new>` fingerprint forks) instead of claiming output is unchanged. Added pinning tests in `normalize-error.unit.test.ts` for all three tags with literal-string assertions — regression-verified by temporarily reverting a tag and watching the test fail with the exact diff. The render-`path`/`format`/`cause` improvement is a good follow-up but out of scope here. Fixed in 9bebebcbc.

packages/config/docs/cli-config-loading.md
Coly010: Accepted — finished the rename rather than scoping the sentence: `CliProjectHome`, `CliProjectLocalServiceVersions`, `cliProjectCommandBaseLayer`/`provideCliProjectCommandRuntime`, with file renames (`cli-project-home.*`, `cli-project-local-service-versions.*`). `projectHomeDir` stays — it's a path field naming the `.supabase/` dir, not a service. The vocabulary doc now lists the deliberate exceptions (`ProjectLinkRemote`, `ProjectLinkState`, and legacy's actual `LegacyProjectRefResolver`) as hosted/link-side names. 9bebebcbc + 7f5f92c1f.

packages/config/AGENTS.md
Coly010: Accepted — dated addenda appended to ADR 0009 and ADR 0018 (old→new tables + pointer to `docs/cli-config-loading.md`), following the existing addendum pattern from ADR 0016; bodies untouched. The 0009 service-key line is corrected to the verified current tag `"@supabase/config/CliConfigStore"`. 7f5f92c1f.

packages/config/src/config-document.ts
Coly010: Accepted for the published surface: `LoadCliConfigOptions.projectEnv` → `cliProjectEnv`, plus the `resolveCliConfig*` positional params. App-internal `CliProjectContext.projectEnv` deliberately keeps its name for now (it sits next to an identically-named legacy-context field across ~38 sites) — filed on CLI-2234 to settle. Worth noting the rename immediately earned its keep: `serve.ts` passed the option through a conditional spread that bypasses excess-property checks and silently dropped the pre-resolved env — the serve integration suite caught it and it's fixed. 9bebebcbc.

packages/config/src/node.ts
Coly010: Agreed, filed as an explicit pre-publish decision on CLI-2234 with your enumeration (drop `For` everywhere vs apply everywhere; align `loadFunctionsManifest`/`inferFunctionsManifest`). The vocabulary doc now has a `/io` facade section listing the seven names and pointing at that decision. Not settled in this PR — the audit is where final names get recorded.

packages/config/src/sparse.ts
Coly010: Agreed this is the expensive-to-reverse one — filed on CLI-2230 (the mapping ticket) so the operand-type story (`ConfigBase`/`EffectiveConfig` vs committing to `ProjectConfig` widening the shared root scope) is decided before CLI-2156 implements against it. Not renamed here: the right name falls out of that decision, and renaming twice pre-publish is worse than once.

apps/cli/src/next/config/cli-settings.service.ts
Coly010: Accepted and extended: all six apps/cli service tags moved out of the package namespace, `supabase/config/*` → `supabase/cli/*` (`CliSettings`, `CliProjectContext`, `CliProjectHome`, `CliProjectLocalServiceVersions`, `ProjectLinkState`, `ProjectLinkRemote`) — each literal verified single-occurrence at its definition site before retagging. 9bebebcbc.

apps/cli/src/legacy/commands/config/push/push.handler.ts
Coly010: Accepted — the comment now names the real module: `legacy-db-config.toml-read.ts`. 9bebebcbc.

packages/config/README.md
Coly010: Accepted — your three-line blockquote added verbatim after the opening paragraph. 7f5f92c1f.

packages/config/src/index.ts
Coly010: Accepted — errors export block re-alphabetized. 9bebebcbc.

packages/config/src/paths.ts
Coly010: Accepted — both sites (`paths.ts:59` and the twin in `project.unit.test.ts`) now read `cliSettings.workdir`. 9bebebcbc.

packages/config/docs/cli-config-loading.md
Coly010: Accepted — reworded per `supabase-home.md`: stack identity is the canonical local-project key relative to the enclosing Git checkout plus workspace identity, and the `.supabase/` inventory now includes `identity.json` (non-Git folders). 7f5f92c1f.

packages/config/docs/cli-config-loading.md
Coly010: Accepted, all three: `/io` facade section with the seven verified names (+ CLI-2234 pointer for the convention settlement), a `LegacyCliSettings` vocabulary bullet, and the `project_id` dual-meaning note (root = local, defaults to directory name; `remotes.*.project_id` = hosted ref). 7f5f92c1f.

Coly010:
Review round 1 addressed in 9bebebcbc + 7f5f92c1f (replies on every thread). The findings without inline anchors:

- **`packages/stack/README.md:48`** — `cliConfig.supabaseHome` → `cliSettings.supabaseHome`. Fixed.
- **ADR 0009 / 0018** — dated rename addenda appended (see the AGENTS.md thread).
- **CLI-2110 plan doc** (`docs/superpowers/plans/2026-08-16-…`) — agreed it's live, not historical: its symbol references updated to the post-rename names (`CliConfigValueSource`, `cliConfigValueSourceAt`, `LoadedCliConfig`, …).
- **`run.ts` fallback stub defect strings** — reworded to `unexpected root cli-config access/file access/write` so a crash report can't surface the freed concept.
- **`remoteProjectConfig`** — accepted your reasoning over the issue's blanket "RemotesSchema internals keep-as-is" carve-out: renamed to `remoteCliConfigBlock` (+ the `io.ts` comment). It was the one binding of the reserved spelling left in the tree.
- **Temp-dir prefix** (`cli-settings.layer.unit.test.ts`) — conscious choice, kept: it's a filesystem prefix for test isolation, not vocabulary.

Two pre-publish judgment calls filed as explicit decisions rather than rushed here: `/io` facade naming convention → CLI-2234; `BaseCliConfig` as the hosted-diff operand type → CLI-2230.

One bonus find from your `LoadCliConfigOptions.projectEnv` ask: `serve.ts` passed the option through a conditional spread that bypasses TS excess-property checking, so the field rename silently dropped serve's pre-resolved env — caught by the serve integration suite (the `.env.development` numeric-`env()` case), fixed in 9bebebcbc.

Validation after the round: types:check ×5 projects, config 198 tests, CLI test:core (unit + integration incl. the new normalize-error pins), oxlint, oxfmt, knip — all green with verified exit codes.

apps/cli/src/next/config/cli-project-home.service.ts
```diff
@@ -6,23 +6,25 @@ import {
   ErrorActionabilityId,
 } from "../../shared/telemetry/error-actionability.ts";
 
-export class ProjectHomeNotDirectoryError extends Data.TaggedError("ProjectHomeNotDirectoryError")<{
+export class CliProjectHomeNotDirectoryError extends Data.TaggedError(
+  "CliProjectHomeNotDirectoryError",
```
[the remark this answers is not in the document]
Coly010: Accepted — good catch: this was a fourth user-visible tag transition, missed because the sweep for these was scoped to `packages/config` errors. Verified it's the last one: a diff-wide sweep of changed `Data.TaggedError` literals finds exactly four. Went with migrate-and-pin rather than keeping the old literal (a class/tag mismatch is the kind of confusion this PR exists to remove, and this error is next-shell-only on a rare path, so the fingerprint fork cost is minimal): the PR body now lists all four transitions, and `normalize-error.unit.test.ts` pins `code: "CliProjectHomeNotDirectoryError"` (with its real message, since this class carries one). 03ec58fe6.

avallete (approved):
Before it get cold.

PR #6337, opened 2026-08-26
fix(config): default auto_expose_new_tables to true to match platform

avallete:
## Summary

The 2026-05-30 flip of `[api].auto_expose_new_tables` (#5524) made *unset* mean "revoke the default Data API GRANTs on new `public` entities", anticipating a cloud default change that never landed — platform projects still auto-expose new entities. A local stack bootstrapped with the flag unset therefore diverged from remote and polluted the user's first `db diff` with spurious GRANT/REVOKE noise.

This flips the effective default back so local matches the platform:

- **unset ≡ explicit `true`** → keep the bundled initial-schema GRANTs (auto-expose), at every consumer: next `start`, legacy `legacyApplyApiPrivileges`, and the shadow-db / pg-delta cache keys (unset and `true` now hash identically, so a no-op config edit can't force a re-snapshot).
- **explicit `false`** → run the revoke SQL, matching a cloud project with the "Default privileges for new entities" toggle turned off.
- The deprecation warning and the 2026-10-30 removal narrative are removed from both shells, the init template, the schema description, and `db start`'s `SIDE_EFFECTS.md`, since their premise collapsed.
- The field stays a tri-state optional key, so `config push` keeps sending it omitempty and unset stays distinguishable. The init template documents the `true` default as a commented example; fresh projects stay on the unset fallback.

Reviewer notes:

- Users with the flag unset take a one-time shadow re-snapshot (~90MB) because the baked cluster genuinely changes (the revoke SQL no longer runs). Cache keys diverging from Go-era `.temp/pgdelta` entries is intentional for the same reason.
- Volumes initialized under the unset-means-revoke era keep their revoked GRANTs until a `db reset`/fresh volume; this change only affects new setups.
- `apps/cli-go` is intentionally untouched (residual delegation surface). `apps/docs/public/cli/config.schema.json` gets a surgical description edit only — the committed file is already stale at HEAD for unrelated toolchain reasons (oxfmt version drift + effect rc `$ref` inlining), so a full regen is left for its own commit.

## Linked issue

Closes # — none; Supabase maintainer working from an internal report.

## Checklist


jgoux (approved):

avallete (commented):
Dogfooded this against a freshly created project plus a dispatched Live E2E run on this branch. The fix is correct and demonstrably improves the first-diff/first-pull story, but it breaks `pull.live.test.ts` deterministically — that test needs updating in this PR before the daily live run goes red.

## Premise verified against a real fresh project

Created a brand-new project (`jrugjsnitqygggmdjutd`, PG 17.6.1.165) and inspected it:

- `pg_default_acl` for role `postgres` in schema `public` still carries `anon`/`authenticated`/`service_role` entries for tables (`arwdDxtm`), sequences (`rwU`), and functions (`X`).
- A newly created `public` table gets `relacl = {postgres=arwdDxtm/postgres,anon=arwdDxtm/postgres,authenticated=arwdDxtm/postgres,service_role=arwdDxtm/postgres}`.

So the platform never adopted revoke-by-default, and the local unset-means-revoke default was genuinely divergent. Flipping unset ≡ `true` is right.

## Live E2E on this branch: the fix visibly works — and breaks one live test

Dispatched [Live E2E run 32958709109](https://github.com/supabase/cli/actions/runs/32958709109) on this branch: **23/24 passed**. The one failure is the strongest possible confirmation of the fix:

```
Applying migration 17877404358914036_e2e_pull.sql...
Diffing schemas...
No schema changes found
The remote database is already in sync with your local migrations — nothing to pull.
: expected 1 to be +0
```

`pull.live.test.ts` pushes one migration to a fresh staging project and expects the subsequent `db pull` to exit 0. Pre-flip, that pull **always** found something — the spurious default-privilege/GRANT noise between the revoked shadow and the fresh remote (the test's own cleanup comment mentions the "generated grant statements"). With this PR the shadow finally matches the remote, the pull is genuinely empty, and `db pull`'s established in-sync contract is a non-zero exit (`LegacyDbPullInSyncError`, see `db/pull/SIDE_EFFECTS.md`). Today's scheduled develop run passed the suite, so this failure is caused by this branch and will fail every live run until fixed.

Suggested fix that keeps the test exercising a real pull: create remote-only drift before pulling, e.g.

```ts
const drift = await cli([
  "db", "query",
  `create table if not exists e2e_pull_drift_${version} (id int);`,
  "--db-url", project.dbUrl,
]);
requireLiveSuccess(drift, "db query drift setup");

const result = await cli(["db", "pull", "--db-url", project.dbUrl, "--yes"]);
expect(result.exitCode, result.stderr).toBe(0);
```

(or, less useful, assert the in-sync outcome explicitly). Worth a comment in the test either way: a clean post-push pull is now the *expected* state for a fresh project — that's the point of this PR.

Story-level follow-up worth considering separately: now that in-sync first pulls become the common case instead of the noisy exception, `db pull`'s non-zero in-sync exit starts biting scripted workflows that never hit it before. Out of scope here, but the flip changes how often users see it.

## Code review

Traced every consumer of the tri-state; the flip is consistent everywhere it matters:

- `legacyApplyApiPrivileges` covers `db start` fresh volume, full `start`, and both `db reset` paths (PG14 `initDatabase` and the PG15+ recreate pipeline all route through the one helper).
- Cache keys: checked all four pre/post-PR × unset/explicit transitions. Unset configs get a new key (correct — the baked cluster genuinely changes, no stale revoked-snapshot reuse); explicit `true`/`false` keys collide with their pre-PR counterparts only where the baked cluster is byte-identical (correct reuse); unset ≡ `true` share a key so a no-op config edit can't force the ~90MB re-snapshot. Same reasoning holds for the pg-delta `setupInputsToken`.
- `config push` still sends the field omitempty; unset stays distinguishable (`api.sync.ts`).
- `next/` now agrees with `packages/stack`'s resolver default (`?? true`), removing the previous contradiction where the stack default said `true` but the next start command forced `false`.
- Ran the touched unit/integration suites locally: all pass (one unrelated failure in my sandbox — a chmod-based permission-denied test that can't fail when running as root).

**Residual Go surface now contradicts the new default** (non-blocking, but deserves a tracked follow-up since both binaries ship in one package):

1. Go `config.Validate` (`apps/cli-go/pkg/config/config.go:1010`) still prints the 2026-10-30 deprecation WARN for `auto_expose_new_tables = true` on every proxied command that loads config (`db diff --use-pg-schema`, `db remote changes|commit`, `db branch *`). Post-PR that warning tells users to "set it to false to adopt the new default" — the exact opposite of the new story.
2. Those proxied diff paths bake their shadow via Go `SetupDatabase` → `ApplyApiPrivileges` (`diff.go:171` → `start.go:426`), which still treats unset as revoke. With the flag unset, `db diff --use-pg-schema` and `db remote changes` will now emit exactly the spurious default-privilege noise this PR eliminates on the native paths — pre-PR they were at least consistent with the local cluster. It's a two-line fix in cli-go (flip the nil-check, delete the WARN) plus a sidecar rebuild, if you'd rather not accept the inconsistency on the deprecated paths.

Nit: the init template's commented example is `# auto_expose_new_tables = true`, which is a no-op to uncomment now that `true` is the default; since the surrounding text pitches `false` as the opt-out, `false` may be the more useful example line.

---

PR #6345, opened 2026-08-26
feat(cli): inject function slug into served functions

raulb:
## Summary

Adds `SUPABASE_FUNCTION_SLUG` to every locally served Edge Function.

The value always matches the requested function name. This works in both the legacy and next stacks.

**Context**

- `@supabase/server` uses this value in `withOAuthProtectedResource`.
- See [supabase/server#117](https://github.com/supabase/server/pull/117).

**Shared function folders**

- Edge Runtime reuses workers by their service path.
- Two functions can share one source folder. They must not share one worker environment.
- The CLI now gives each shared function a stable and unique worker path.
- The CLI still passes the real source file as the entrypoint. Imports and file access remain unchanged.
- Functions in separate folders keep their current service path.
- The change does not use queues, owner maps, or `forceCreate`.

## Linked issue

Fixes AI-1129

- [x] The linked issue is **open** and carries the `open-for-contribution` label, or I am a Supabase maintainer.

## Checklist

- [x] The PR title follows [Conventional Commits](https://www.conventionalcommits.org/).
- [x] Tests were added or updated for the change.
- [ ] From the repository root, `pnpm check:all` passes, and all required workspace checks pass.


apps/cli/src/shared/functions/serve.main.ts
```diff
@@ -331,6 +331,8 @@ Deno.serve({
           ([name, _]) => !name.startsWith("SUPABASE_"),
         ),
       ),
+      // Listed after the spreads so neither the container env nor function config can shadow it
+      SUPABASE_FUNCTION_SLUG: functionName,
```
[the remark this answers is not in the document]
raulb: Thanks, this was valid. `SUPABASE_FUNCTION_SLUG` is part of the worker env, while Edge Runtime reuses workers by `servicePath`, so two configured function names that share one entrypoint directory can observe the first-created slug.

I resolved this without adding a slug map/cache layer: the serve paths now detect entrypoint directories used by more than one configured function and set `forceCreate` only for those shared `servicePath`s. That keeps the normal worker reuse behavior for the common one-function-per-directory case, while preventing a cached shared-directory worker from exposing another function’s slug. I kept this scoped to the CLI serve paths; a true `(servicePath, slug)` worker cache identity would be a separate Edge Runtime/API change.

apps/cli/src/shared/functions/serve.main.ts
```diff
@@ -347,7 +362,8 @@ Deno.serve({
       ([name, _]) => !EXCLUDED_ENVS.includes(name) && !name.startsWith("SUPABASE_INTERNAL_"),
     );
 
-    const forceCreate = false;
+    // Shared entrypoint directories need a fresh worker for the per-function slug.
+    const forceCreate = sharedServicePaths.has(servicePath);
```
jgoux: **[P1] Avoid forcing a new worker on every request for shared paths**

`sharedServicePaths` is static, so once a directory is shared this passes `forceCreate: true` for every request, including A → A requests to the same function. In Edge Runtime v1.74.3, `force_create` makes [`maybe_active_worker` return no reusable worker](https://github.com/supabase/edge-runtime/blob/47d04fdd22e33ea3fd904576cf3248d963d903a9/crates/base/src/worker/pool.rs#L741-L748), and when permits are exhausted it can [create without acquiring the concurrency semaphore](https://github.com/supabase/edge-runtime/blob/47d04fdd22e33ea3fd904576cf3248d963d903a9/crates/base/src/worker/pool.rs#L340-L347). Under the default `per_worker` policy, repeated or concurrent traffic to either slug therefore keeps cold-creating isolates, loses reusable module state, and can exhaust runtime resources. Please preserve a stable worker identity/cache per `(servicePath, functionName)` (or equivalent scoped lifecycle coordination) instead of enabling `forceCreate` unconditionally for the whole shared path. The current env-only tests do not exercise worker creation/reuse, so this also needs a shared-directory integration case.

packages/stack/src/services/edge-runtime-main.ts
```diff
@@ -184,18 +181,40 @@ async function serveFunction(req: Request, config: any, functionName: string, fu
...
+      servicePath,
       memoryLimitMb: 256,
       workerTimeoutMs: 400000,
       noModuleCache: false,
       noNpm: false,
       importMapPath: functionConfig.importMapPath ?? undefined,
       envVars,
-      forceCreate: false,
+      // Shared entrypoint directories need a fresh worker for the per-function slug.
+      forceCreate: hasSharedServicePath(config, servicePath),
```
jgoux: **[P1] Apply the same worker-reuse remediation in the stack runtime**

This duplicates the unconditional shared-path `forceCreate` behavior described in [the CLI serve-path finding](https://github.com/supabase/cli/pull/6345#discussion_r3879696907), so the stack implementation also creates a new worker for every request to either function and can bypass the pool semaphore under load. Please keep the two runtime paths aligned when introducing the stable per-`(servicePath, functionName)` worker identity/cache, and cover this path with the same shared-directory integration scenario.

apps/cli/src/shared/functions/serve.main.ts
raulb: @jgoux yeah, this is correct. I was able to confirm it against the Edge Runtime source you linked. I fixed it by removing `sharedServicePaths` and giving each function its own directory instead, so `forceCreate` stays false everywhere and normal worker reuse is fully restored.

apps/cli/src/shared/functions/serve.main.ts
```diff
@@ -364,26 +388,53 @@ Deno.serve({
...
+      const currentSlot = new Promise<void>((resolve) => {
+        releaseSlot = resolve;
       });
+      sharedWorkerQueues.set(
+        servicePath,
+        previousSlot.then(() => currentSlot),
+      );
+      await previousSlot;
+      releaseSharedWorkerSlot = releaseSlot!;
+      forceCreate = sharedWorkerOwners.get(servicePath) !== functionName;
```
jgoux: **[P1] The owner map does not identify the worker Edge Runtime will reuse**

`forceCreate: true` only skips the active-worker lookup; it does not retire workers already registered for this `servicePath`. After A creates worker A and B force-creates worker B, both remain in the same registry. A later B request reaches this line with `forceCreate: false`, and Edge Runtime can round-robin back to worker A, exposing A’s code/state/environment and `SUPABASE_FUNCTION_SLUG` under B. See [`maybe_active_worker`](https://github.com/supabase/edge-runtime/blob/47d04fdd22e33ea3fd904576cf3248d963d903a9/crates/base/src/worker/pool.rs#L741-L748), [`add_user_worker`](https://github.com/supabase/edge-runtime/blob/47d04fdd22e33ea3fd904576cf3248d963d903a9/crates/base/src/worker/pool.rs#L546-L559), and the registry’s [round-robin selection](https://github.com/supabase/edge-runtime/blob/47d04fdd22e33ea3fd904576cf3248d963d903a9/crates/base/src/worker/pool.rs#L173-L203). Alternating A/B requests also keep forcing new workers, so accumulation remains reachable. A last-owner map cannot provide a stable `(servicePath, functionName)` cache identity; that identity must exist in the worker pool/caller-owned worker cache, or shared paths need to be rejected until it does.

apps/cli/src/shared/functions/serve.main.ts
```diff
@@ -364,26 +388,53 @@ Deno.serve({
...
+      sharedWorkerOwners.set(servicePath, functionName);
```
jgoux: **[P2] Do not record ownership before worker creation succeeds**

If B’s forced `userWorkers.create()` fails, this map still records B while the only reusable worker may belong to A. The next B request therefore computes `forceCreate: false` and can reuse A’s worker. Ownership must only change after successful creation while coordination is still held. This fixes the failure-state corruption, although it does not address the separate P1 that multiple same-path workers remain registered and can be selected later.

packages/stack/src/services/edge-runtime-main.ts
```diff
@@ -184,26 +181,86 @@ async function serveFunction(req: Request, config: any, functionName: string, fu
...
+    const currentSlot = new Promise<void>((resolve) => {
+      releaseSlot = resolve;
     });
+    sharedWorkerQueues.set(
+      servicePath,
+      previousSlot.then(() => currentSlot),
+    );
+    await previousSlot;
+    releaseSharedWorkerSlot = releaseSlot!;
+    forceCreate = sharedWorkerOwners.get(servicePath) !== functionName;
```
jgoux: **[P1] Apply the stable worker-identity fix in the stack path too**

This owner comparison duplicates the CLI serve-path bug: after multiple forced creations, Edge Runtime retains multiple active workers for this `servicePath`, so a later `forceCreate: false` call may round-robin to a worker with another function’s slug. See the full lifecycle trace in [the corresponding CLI finding](https://github.com/supabase/cli/pull/6345#discussion_r3880967000). Please keep both runtime paths aligned, with integration coverage for A → B → repeated B and alternating A/B requests.

packages/stack/src/services/edge-runtime-main.ts
```diff
@@ -184,26 +181,86 @@ async function serveFunction(req: Request, config: any, functionName: string, fu
...
+  // Only shared directories pay for coordination: serialize worker creation for this
+  // servicePath so a concurrent request for a different function can't race the pool's
+  // "which slug currently owns this worker" decision, then force a fresh worker only
+  // when the owner actually changed.
+  let forceCreate = false;
+  let releaseSharedWorkerSlot: (() => void) | undefined;
+  if (sharedServicePaths.has(servicePath)) {
+    const previousSlot = sharedWorkerQueues.get(servicePath) ?? Promise.resolve();
+    let releaseSlot: () => void;
+    const currentSlot = new Promise<void>((resolve) => {
```
jgoux: **Blocking: this manual Promise gate currently fails the required quality check**

`@supabase/root#lint:effect:check` rejects this `new Promise(...)` construction, which is why the PR’s `Check code quality` job is red. The stack runtime is Effect-scoped and its coordination must use the applicable Effect primitive rather than a Promise gate. Given the worker-identity P1 above, the coordination design should be corrected first instead of mechanically wrapping this gate just to satisfy lint.

apps/cli/src/shared/functions/serve-main-offline.e2e.test.ts
```diff
@@ -125,6 +146,28 @@ function containerLogs(container: string): string {
   return `${result.stdout ?? ""}\n${result.stderr ?? ""}`;
 }
 
+async function fetchColdFunction(
+  url: string,
+  diagnosticContainers: readonly string[],
+  init?: RequestInit,
+): Promise<Response> {
+  try {
+    const response = await fetch(url, {
```
jgoux: **[P2] Retry transient failures while cold-starting these workers**

This helper performs only one `fetch`, so a transport error or transient 502/503 immediately fails the test even though the 60-second signal suggests it waits for readiness. The three call sites cold-start previously untouched workers concurrently, while the preceding auth readiness probe never creates any of them. The repository’s `fetchFunctionWhenReady` helper documents that Edge Runtime can briefly disconnect or return 502/503 during this phase. Please retry only those transient outcomes until the deadline, and retain these container diagnostics when the deadline expires.

jgoux (approved):

apps/cli/src/shared/functions/serve-main-offline.e2e.test.ts
raulb: https://github.com/supabase/cli/pull/6345/commits/b43cee86c9782bc8d86e61152e573a6c4a59bc49

PR #6371, opened 2026-08-28
feat(workers push): make the build wait opt-out with `--no-wait` flag

johnstonmatt:
## Summary

`supabase experimental workers push` blocks on the server-side container build. That build routinely runs for minutes, so the common case — a deploy that builds fine — is the slowest thing in the loop.

`--no-wait` returns once the platform accepts the deploy, which is the last thing the command can learn without waiting: the deploy response arrives only after the spec and the uploaded context are accepted, and it carries the accepted spec back. Waiting stays the default, so a plain push still reports the build's verdict and existing invocations are unchanged.

- Under `--no-wait` the details block leads with a `State` row — the one row that says the worker is not serving yet — and drops `Image`, since no image exists until the build produces one.
- A success trailer then points at `experimental workers status` for the build's outcome. Text output only; machine callers read `build_state` from the payload.
- A deploy answered with a spec already in `failed` is reported as a failure whether or not the build was waited on, rather than exiting zero on a worker that will never come up.

The second commit is a separate fix that fell out of the review: the wait now runs only when the deploy response left `build_state` at `building`. `V2DeployAWorkerOutput` permits a terminal `active` or `failed` on the deploy response itself, and polling on top of that could only contradict it — `awaitWorkerBuild` reads a post-deploy 404 as "still building", so an already-failed deploy could burn the full poll budget and surface as a timeout rather than the failure the platform had already reported.

## Stack

Top of the workers stack, on top of `workers logs` (#6410). Everything below it — the name prompt (#6349), the output polish (#6389) and `workers logs` — is independent of this flag, so this PR can be rejected on its own without holding any of them up.

## Linked issue

FUNC-848 (Linear). Supabase maintainer, exempt from the `open-for-contribution` flow.

## Checklist

- [x] The PR title follows [Conventional Commits](https://www.conventionalcommits.org/)


apps/cli/src/legacy/cli/legacy-boolean-flag-defaults.unit.test.ts
```diff
@@ -0,0 +1,44 @@
...
+function booleanFlagsRequiringAValue(command: Command.Command.Any): ReadonlyArray<string> {
+  // Throws rather than skipping if effect's internal shape moves, so this
+  // cannot quietly degrade into a test that inspects nothing.
+  const own = legacyCommandInternals(command).config.flags.flatMap((flag) => {
+    const unwrapped = legacyUnwrapParam(flag);
+    if (unwrapped === undefined) {
+      throw new Error(`Unrecognizable flag param on "${command.name}".`);
+    }
+    const { single, isOptional } = unwrapped;
+    return single.primitiveType._tag === "Boolean" && !isOptional
```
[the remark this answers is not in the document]
johnstonmatt: Fixed in dc11ce14f.

Swapped the `_tag` read for `Primitive.getTypeName`, the published getter — its own doc example pins `Primitive.getTypeName(Primitive.boolean)` to `"boolean"`. `Primitive.isBoolean` does exist in the effect source, but it is absent from `effect@4.0.0-rc.111`'s published `.d.ts` (same gap this tree already documents for `Param.extractSingleParams` and `choiceKeys`), so calling it would need a cast.

The expected name is derived from `Primitive.boolean` rather than written as the literal `"boolean"`. Were that name to change upstream, a hardcoded literal would match nothing and leave the guard silently passing every command — the one failure mode a regression test must not have.

Verified by mutation, not just by the suite staying green: removing `Flag.withDefault(false)` from `workers push --wait` still turns the test red with `push --wait`.

apps/cli/src/legacy/cli/legacy-boolean-flag-defaults.unit.test.ts
```diff
@@ -0,0 +1,44 @@
...
+ * `workers push --wait` shipped without either and made a plain
+ * `supabase workers push` unusable. Nothing caught it: handler integration
+ * tests build their flags record directly, so they never touch the parser, and
+ * the required-ness is invisible to the type checker because a required boolean
+ * flag still infers as `boolean`.
+ */
+
+function booleanFlagsRequiringAValue(command: Command.Command.Any): ReadonlyArray<string> {
+  // Throws rather than skipping if effect's internal shape moves, so this
+  // cannot quietly degrade into a test that inspects nothing.
```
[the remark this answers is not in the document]
johnstonmatt: Fixed in dc11ce14f.

The guard now walks all three parameter sets a command can be parsed with — its own `config.flags`, the inherited `contextConfig.flags`, and the scoped global flags via the existing `legacyUserGlobalFlagParams` helper.

Verified by mutation: dropping `Flag.withDefault(false)` from the global `--debug` fails the test with `supabase --debug`, which it did not before. Worth noting the globals land on the root only, so there is no per-command duplication in the failure list.

Every boolean in the tree already carries a default (`--debug`, `--yes`, `--experimental`, `--create-ticket` globally, plus the command-local ones), so the expansion is green today rather than papering over an existing break.

apps/cli/src/legacy/commands/workers/push/push.handler.ts
```diff
@@ -356,13 +373,38 @@ const deployOneWorker = Effect.fnUntraced(function* (input: {
...
+      // more than for those: pushing several workers would otherwise bury each
+      // worker's hint under the next worker's packaging and deploy output.
+      //
+      // One short sentence per line, with the command and the flag aqua'd the
+      // way every other follow-up hint in this shell writes them. The single
+      // wrapped paragraph this replaced re-flowed differently at every terminal
+      // width and buried both commands mid-sentence.
+      yield* emitSuccessTrailer(
+        `\nYour build was submitted successfully.\n` +
+          `Run ${legacyAqua(`supabase workers status ${name}`)} to check on it.\n` +
```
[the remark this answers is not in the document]
johnstonmatt: Fixed in 5072b3ec0.

Confirmed, and it was three hints rather than one — all in the same copy-paste-verbatim family:

- the still-building success trailer (`push.handler.ts:411`)
- the retry suggestion on `WorkerBuildFailedError` (`push.handler.ts:360`)
- the give-up suggestion on `WorkerBuildTimeoutError`, raised from `awaitWorkerBuild` in `shared/workers/workers-api.ts:487`

`deployOneWorker` now takes a `refSuffix` its caller computes once from `legacyWorkersProjectRefSuffix(flags.projectRef)` — the helper you pointed at, the same one `status` and `delete` use — and `awaitWorkerBuild` takes it as an option for the timeout it raises. Keyed off the flag, so a ref that came from the link is not repeated back as noise.

Four integration tests under `carries an explicit --project-ref into its hints`: one per hint from an unlinked checkout, plus the mirror case asserting a linked run still emits no `--project-ref`. The three positive cases verified red against the un-suffixed strings and green with them.

apps/cli/src/legacy/commands/workers/push/SIDE_EFFECTS.md
```diff
@@ -72,6 +76,13 @@ payload always carries a `workers` array, which a flat `KEY=value` list cannot
 express, and discovering that at the end would fail the command with the remote
 project already changed.
 
+Without `--wait` the deploy returns with the build still running, so the follow-up hint (`workers status`, and `--wait`) is emitted as a success trailer: stderr, once, at the end of the run rather than between workers.
```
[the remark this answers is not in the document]
johnstonmatt: Fixed in 5072b3ec0.

Confirmed — the trailer is inside `output.format === "text" && !input.machineOutput` in `deployOneWorker`, so no machine format emits it. The section now says text-only explicitly and points machine callers at `build_state` in the payload.

On the second half of your note: `reportUnattempted` really does have no format guard, so the "every format, machine ones included" paragraph one below was accurate as written. I left the claim and stated the contrast with the trailer instead, since that difference is deliberate — a CI run needs to know what still needs deploying. The section also now records that the hint carries an explicit `--project-ref`, per the other thread.

apps/cli/src/legacy/commands/experimental/workers/push/push.handler.ts
```diff
@@ -325,18 +333,27 @@ const deployOneWorker = Effect.fnUntraced(function* (input: {
...
 
-  const settled = yield* awaitWorkerBuild(api, projectRef, name, {
-    schedule: input.pollSchedule,
-    retrySchedule: input.pollRetrySchedule,
-    refSuffix: input.refSuffix,
-    onPoll: (polled) =>
-      polled.buildState === "building" ? deploying.message("Building worker...") : Effect.void,
-  }).pipe(Effect.tapError(() => deploying.fail()));
-
+  const settled = input.wait
```
kanadgupta: Re: the earlier Codex comment about honoring terminal deploy responses before polling — I verified it and agree, and it is still unaddressed. `V2DeployAWorkerOutput` permits `build_state: "failed"` and `"active"` on the deploy response itself, and `awaitWorkerBuild` treats a 404 poll as "still building" and keeps going. So with `--wait`, a deploy that already answered `failed` can, in the worst case (post-deploy 404 race, transient read errors), burn the whole 10-minute poll budget and surface as `WorkerBuildTimeoutError` instead of the failure the platform already reported. A fresh poll can also return a stale terminal state from the previous deploy and contradict the verdict the deploy response carried. Only polling when the accepted response says `building` is a one-line fix and also skips a redundant GET on the common terminal cases:

```suggestion
  const settled = input.wait && accepted.buildState === "building"
```

Non-blocking, but I would take this before merge since the fix is tiny.

*🤖 AI-drafted comment (Claude), pending human review.*

apps/cli/src/legacy/commands/experimental/workers/push/push.handler.ts
```diff
@@ -367,13 +384,38 @@ const deployOneWorker = Effect.fnUntraced(function* (input: {
...
+      // `gen signing-key` all route through `emitSuccessTrailer` so it prints
+      // once at the end of the run instead of scrolling away. It matters here
+      // more than for those: pushing several workers would otherwise bury each
+      // worker's hint under the next worker's packaging and deploy output.
+      //
+      // One short sentence per line, with the command and the flag aqua'd the
+      // way every other follow-up hint in this shell writes them. The single
+      // wrapped paragraph this replaced re-flowed differently at every terminal
+      // width and buried both commands mid-sentence.
+      yield* emitSuccessTrailer(
```
kanadgupta: Non-blocking: Re: the earlier Codex comment about accepted-build hints being lost when a later worker fails — I verified this is real. `run.ts` only drains `SuccessTrailer` when the exit code is 0 (`afterSuccess` is gated on `code === 0` before `successTrailer.takeAll`), so in a multi-worker text-mode push without `--wait`, a worker that was accepted here loses its follow-up hint entirely if any later worker fails — even though its remote build is still running. The integration tests do not catch this because they run the handler without the `SuccessTrailer` service, which makes `emitSuccessTrailer` fall back to writing stderr immediately.

Whether failed runs should print success trailers at all is a design call (the service is named for success, and the failure path already emits the error plus `Not attempted:`), so I would at least record the tradeoff here — the comment above explains why a trailer beats an inline line, but not that the trailer silently evaporates on partial failure, which is exactly the multi-worker case it cites as its motivation.

*🤖 AI-drafted comment (Claude), pending human review.*

apps/cli/src/legacy/commands/experimental/workers/push/push.handler.ts
```diff
@@ -367,13 +384,38 @@ const deployOneWorker = Effect.fnUntraced(function* (input: {
...
       legacyRenderWorkerDetails([
+        // Labelled `State`, and placed first, the way `workers status` renders
+        // the same field: without `--wait` it is the one row that says the
+        // worker is not serving yet, so it should not be hunted for at the
+        // bottom of the block.
+        ["State", settled.buildState],
         ["Runtime", runtime],
         ["Size", formatApiSize(settled.spec.size)],
+        // Empty without `--wait`: no image exists until the build produces one,
+        // and `legacyRenderWorkerDetails` drops an empty-valued row.
```
kanadgupta: Non-blocking question: "no image exists until the build produces one" is only true for a first deploy. `image_version` is optional-but-permitted on the deploy response (`V2DeployAWorkerOutput`), so a re-push of an already-active worker may well echo the *currently serving* image while `build_state` is `building` — in which case this row renders the old image next to `State building`, and the machine payload below carries the same stale `image_version` alongside `build_state: "building"`, which a script could easily misread as the new build. If the API is known to omit `image_version` while a build is in flight, ignore this; otherwise consider blanking the row (and dropping the payload field) whenever `settled.buildState === "building"`, which also makes the comment accurate for redeploys. The test fixtures only cover the first-deploy shape (deploy response without `image_version`), so this case is currently unexercised either way.

*🤖 AI-drafted comment (Claude), pending human review.*

apps/cli/src/legacy/commands/experimental/workers/push/push.handler.ts
kanadgupta: Follow-up candidate (small enough to take now): Re: the earlier Codex comment about honoring terminal deploy responses before polling — I verified it and agree, and it is still unaddressed. `V2DeployAWorkerOutput` permits `build_state: "failed"` and `"active"` on the deploy response itself, and `awaitWorkerBuild` treats a 404 poll as "still building" and keeps going. So with `--wait`, a deploy that already answered `failed` can, in the worst case (post-deploy 404 race, transient read errors), burn the whole 10-minute poll budget and surface as `WorkerBuildTimeoutError` instead of the failure the platform already reported. A fresh poll can also return a stale terminal state from the previous deploy and contradict the verdict the deploy response carried. Only polling when the accepted response says `building` is a one-line fix and also skips a redundant GET on the common terminal cases:

```suggestion
  const settled = input.wait && accepted.buildState === "building"
```

apps/cli/src/legacy/commands/experimental/workers/push/push.handler.ts
kanadgupta, line 441: Follow-up candidate: Re: the earlier Codex comment about accepted-build hints being lost when a later worker fails — I verified this is real. `run.ts` only drains `SuccessTrailer` when the exit code is 0 (`afterSuccess` is gated on `code === 0` before `successTrailer.takeAll`), so in a multi-worker text-mode push without `--wait`, a worker that was accepted here loses its follow-up hint entirely if any later worker fails — even though its remote build is still running. The integration tests do not catch this because they run the handler without the `SuccessTrailer` service, which makes `emitSuccessTrailer` fall back to writing stderr immediately.

Whether failed runs should print success trailers at all is a design call (the service is named for success, and the failure path already emits the error plus `Not attempted:`), so I would at least record the tradeoff here — the comment above explains why a trailer beats an inline line, but not that the trailer silently evaporates on partial failure, which is exactly the multi-worker case it cites as its motivation.

apps/cli/src/legacy/commands/experimental/workers/push/push.handler.ts
kanadgupta: Follow-up candidate (question): "no image exists until the build produces one" is only true for a first deploy. `image_version` is optional-but-permitted on the deploy response (`V2DeployAWorkerOutput`), so a re-push of an already-active worker may well echo the *currently serving* image while `build_state` is `building` — in which case this row renders the old image next to `State building`, and the machine payload below carries the same stale `image_version` alongside `build_state: "building"`, which a script could easily misread as the new build. If the API is known to omit `image_version` while a build is in flight, ignore this; otherwise consider blanking the row (and dropping the payload field) whenever `settled.buildState === "building"`, which also makes the comment accurate for redeploys. The test fixtures only cover the first-deploy shape (deploy response without `image_version`), so this case is currently unexercised either way.

apps/cli/src/legacy/commands/experimental/workers/push/push.handler.ts
```diff
@@ -367,13 +384,38 @@ const deployOneWorker = Effect.fnUntraced(function* (input: {
...
+      // worker's hint under the next worker's packaging and deploy output.
+      //
+      // One short sentence per line, with the command and the flag aqua'd the
+      // way every other follow-up hint in this shell writes them. The single
+      // wrapped paragraph this replaced re-flowed differently at every terminal
+      // width and buried both commands mid-sentence.
+      yield* emitSuccessTrailer(
+        `\nYour build was submitted successfully.\n` +
+          `Run ${legacyAqua(`supabase experimental workers status ${name}${input.refSuffix}`)} to check on it.\n` +
+          `Add ${legacyAqua("--wait")} to block on the build next time.\n`,
```
kanadgupta: (unrelated) nice DX work with these lines! i'm a big fan of CTAs like this

kanadgupta (approved):
LGTM, confirmed this works as expected! claude left three pieces of feedback that seem worth looking into, but happy with whatever call you make

> > [!NOTE]
> > This review was drafted by an AI (Claude) and is pending human vetting. Refreshed after new commits landed (head `6a0bab9`; the delta was mostly a merge of the stacked base, plus a few in-PR touches reviewed below).
> 
> ## What this PR does
> 
> Makes `workers push` return as soon as the platform accepts the deploy, instead of blocking minutes on the server-side container build; `--wait` opts back into the old blocking behavior. A deploy answered with `build_state: failed` still exits 1 in both modes. The no-wait text output leads the details block with a `State` row, drops the (nonexistent) `Image` row, and emits a success trailer pointing at `workers status` and `--wait`; machine callers read `build_state` from the payload. SIDE_EFFECTS.md and the exit-code table are updated to match. Since the last review round the PR also gained a `output.format === "text"` gate on the per-worker `Deploying Worker n/N` progress line (keeping it out of `--output-format json`, with a test), a telemetry span rename to `legacy.experimental.workers.push`, and a comment reword in the boolean-flag-defaults guard — all clean, no new findings.
> 
> ## Does it deliver FUNC-848?
> 
> Yes. The ticket asks for exit-on-optimism by default and blocking under `--wait`, and that is exactly what ships. The default flip (previously push blocked and failed on a bad build; now it exits 0 while the build runs) is handled carefully: the `State` row, the trailer hint, `build_state` in the machine payload, the flag description naming the default, and updated exit-code docs. Anyone gating CI on the old blocking behavior needs to add `--wait`, but on an experimental-namespace alpha command that is an acceptable, well-signposted change. Tests sit at the right places on the pyramid (integration-level, schedule seams injected, no sleeps), and the polling itself still runs through the pre-existing `Schedule`-based `awaitWorkerBuild`.
> 
> ## Findings
> 
> Graded for an experimental command: "fix before merge" is reserved for correctness/data-loss/security issues of real note; everything else is a follow-up candidate.
> 
> **Fix before merge:** none.
> 
> **Follow-up candidates (inline):**
> 1. With `--wait`, a deploy response that already carries a terminal `build_state` (`failed`/`active`) still enters the poll loop; in the 404-race / flaky-read cases this can convert an already-reported failure into a 10-minute `WorkerBuildTimeoutError`. One-line fix (`input.wait && accepted.buildState === "building"`). This amplifies an open Codex thread that I verified is correct and still unaddressed.
> 2. In a multi-worker no-wait push, an accepted worker's trailer hint is silently dropped if a later worker fails — `run.ts` drains `SuccessTrailer` only on exit code 0. Also an open, correct Codex thread; at minimum worth documenting as a tradeoff.
> 3. On a re-push of an existing worker, the deploy response may echo the previous `image_version` while `build_state` is `building`, so the no-wait output (and machine payload) can show a stale image. Question-grade; depends on API behavior the fixtures don't cover.
> 
> ## CI
> 
> Green. The three failures noted in the earlier round (knip unused export and a color-tinting unit test, both in files owned by the logs PR below this one) were inherited from the stacked base and have since been fixed there.
> 
> ## Existing review threads
> 
> The author addressed the earlier Codex rounds thoroughly (the `_tag` guard, the `--project-ref` suffix in hints, the text-only trailer docs). Two open Codex P2s on `push.handler.ts` (lines 345 and 416) remain unaddressed and are both correct — my inline comments 1 and 2 amplify them. Three other unresolved Codex threads (delete/status SIDE_EFFECTS.md, `new.handler.ts`) point at files under the old pre-`experimental/` paths that are not part of this PR's incremental diff — stack drift; they belong to the PRs below this one. (#6432 now stacks on top of this branch; its changes are out of scope here.)
> 
> **Overall:** 0 fix-before-merge issue(s), 3 follow-up candidate(s).

apps/cli/src/legacy/commands/experimental/workers/push/push.handler.ts
```diff
@@ -322,18 +330,27 @@ const deployOneWorker = Effect.fnUntraced(function* (input: {
...
-  const settled = yield* awaitWorkerBuild(api, projectRef, name, {
-    schedule: input.pollSchedule,
-    retrySchedule: input.pollRetrySchedule,
-    refSuffix: input.refSuffix,
-    onPoll: (polled) =>
-      polled.buildState === "building" ? deploying.message("Building worker...") : Effect.void,
-  }).pipe(Effect.tapError(() => deploying.fail()));
-
+  const settled = input.wait
+    ? yield* awaitWorkerBuild(api, projectRef, name, {
```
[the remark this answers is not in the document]
johnstonmatt: Fixed in 486a47bf9.

Confirmed, and taken now rather than as a follow-up. The wait is gated on the deploy response having left the build unresolved:

```ts
const settled =
  input.noWait || accepted.buildState !== "building"
    ? accepted
    : yield* awaitWorkerBuild(...)
```

`V2DeployAWorkerOutput` permits a terminal `active` or `failed` on the deploy response, and that verdict is this deploy's — a fresh `GET` could only contradict it. `awaitWorkerBuild` reads a post-deploy 404 as "still building", so an already-`failed` deploy could burn the full poll budget and surface as `WorkerBuildTimeoutError` instead of the failure the platform had already reported; a concurrent deployment could answer with a state belonging to someone else's build. It also drops a redundant `GET` on the common terminal cases.

Two integration tests under `honours a terminal deploy response instead of polling`: a deploy answered `active` reports the image without a `GET`, and one answered `failed` raises `WorkerBuildFailedError` without a `GET`. Both verified red against the old condition and green with the new one.

Note the flag also flipped in 486a47bf9~1: `--wait` is now `--no-wait`, so waiting is the default and this gate is on the default path rather than the opt-in one.

apps/cli/src/legacy/commands/experimental/workers/push/push.handler.ts
johnstonmatt: Fixed in 486a47bf9.

Confirmed, and taken now rather than as a follow-up. The wait is gated on the deploy response having left the build unresolved:

```ts
const settled =
  input.noWait || accepted.buildState !== "building"
    ? accepted
    : yield* awaitWorkerBuild(...)
```

`V2DeployAWorkerOutput` permits a terminal `active` or `failed` on the deploy response, and that verdict is this deploy's — a fresh `GET` could only contradict it. `awaitWorkerBuild` reads a post-deploy 404 as "still building", so an already-`failed` deploy could burn the full poll budget and surface as `WorkerBuildTimeoutError` instead of the failure the platform had already reported; a concurrent deployment could answer with a state belonging to someone else's build. It also drops a redundant `GET` on the common terminal cases.

Two integration tests under `honours a terminal deploy response instead of polling`: a deploy answered `active` reports the image without a `GET`, and one answered `failed` raises `WorkerBuildFailedError` without a `GET`. Both verified red against the old condition and green with the new one.

Note the flag also flipped in 486a47bf9~1: `--wait` is now `--no-wait`, so waiting is the default and this gate is on the default path rather than the opt-in one.

apps/cli/src/legacy/commands/experimental/workers/push/push.handler.ts
johnstonmatt: Fixed in f9647a083 — it's real (V2DeployAWorkerOutput has `image_version` as optionalKey beside a `"building"` build_state), so both the Image row and the payload field are now omitted while the build is still running

apps/cli/src/legacy/commands/experimental/workers/push/push.handler.ts
johnstonmatt: Fixed in f9647a083 — rather than document the tradeoff, the failure path now names them itself: "Still building: api", alongside "Not attempted: web"

PR #6374, opened 2026-08-28
test(stack): qualify complete slim Docker service graph

jgoux:
## Summary

- consolidate fragmented Docker coverage into one public 13-service stack journey
- verify canonical slim images, representative product behavior, and restart persistence
- prove two automatically allocated sibling stacks keep ports, data, ownership, and cleanup isolated

## Context

This is a test-only qualification change for CLI-2113. The existing stack implementation already satisfied the complete Docker graph and sibling-isolation contracts; the new journey makes those release-critical guarantees observable through the public package surface.


avallete (approved):

packages/stack/tests/createStack-docker.e2e.test.ts
```diff
@@ -1,22 +1,77 @@
...
+  `supabase-${service}-${identity}`;
+
+const ownedDockerContainers = (
+  identity: string,
+): ReadonlyArray<{ readonly name: string; readonly image: string }> =>
+  SERVICE_NAMES.flatMap((service) => {
+    const name = ownedDockerContainerName(service, identity);
+    const output = execFileSync(
+      "docker",
+      ["ps", "-a", "--filter", `name=^${name}$`, "--format", "{{.Names}}\t{{.Image}}"],
```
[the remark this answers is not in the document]
jgoux: Reproduced this against the current Docker daemon: for an existing named container, both `name=^supabase-...$` and `name=^/supabase-...$` returned exactly one match. The targeted Docker journey also passes the assertions that require all 13 exact owned names, so the existing filter is not returning an empty set. Leaving this unchanged because the proposed slash does not fix a reproducible defect on the supported Docker boundary.

packages/stack/tests/createStack-docker.e2e.test.ts
```diff
@@ -1,22 +1,77 @@
...
-import { activationTimeoutSecondsForService } from "../src/ServiceActivation.ts";
 import { createStack, type StackHandle } from "../src/node.ts";
 import { dependencyTimeoutSecondsForServices } from "../src/services/health-budgets.ts";
-import { DEFAULT_VERSIONS } from "../src/versions.ts";
+import { SERVICE_NAMES, type ServiceName } from "../src/versions.ts";
 import { setupTestTable } from "./helpers/e2e.ts";
 
-const STACK_DOCKER_E2E_TEST_TIMEOUT_MS = 180_000;
-const STACK_DOCKER_E2E_SETUP_OVERHEAD_MS = 90_000;
+const STACK_DOCKER_E2E_TEST_TIMEOUT_MS = 240_000;
```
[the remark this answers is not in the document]
jgoux: Fixed in 6e332dd54. The test timeout is now derived as two full-start budgets plus a named 240-second journey overhead; the beforeAll primary start keeps its independent full-start hook budget. This covers the sibling start and primary restart without adding another test.

packages/stack/tests/createStack-docker.e2e.test.ts
```diff
@@ -1,22 +1,79 @@
...
+  realtime: "ghcr.io/supabase/cli/realtime:v2.129.9",
+  storage: "ghcr.io/supabase/cli/storage:v1.71.0",
+  imgproxy: "ghcr.io/supabase/cli/imgproxy:v3.8.0",
+  mailpit: "ghcr.io/supabase/cli/mailpit:v1.30.2",
+  pgmeta: "ghcr.io/supabase/cli/pgmeta:v0.98.0",
+  studio: "ghcr.io/supabase/cli/studio:2026.08.24-sha-8ec45b2",
+  analytics: "ghcr.io/supabase/cli/analytics:v1.50.6",
+  vector: "ghcr.io/supabase/vector:0.53.0-alpine",
+  pooler: "ghcr.io/supabase/supavisor:2.9.7",
+};
```
avallete: These 13 URLs are identical to `dockerImageForService(service, DEFAULT_VERSIONS[service])`. A catalog version bump will fail this journey even when the stack is correct. Derive the expected map from those helpers instead of pinning the tags here.

Same reuse note: `ownedDockerContainerName` below is `dockerContainerName` from `StackIdentity.ts`.

packages/stack/tests/createStack-docker.e2e.test.ts
```diff
@@ -65,192 +148,339 @@ dockerDescribe("createStack e2e (docker mode)", () => {
...
+    // Verify all exact owned Docker containers are cleaned up, including stopped containers.
+    if (apiPort !== undefined) {
+      const remaining = ownedDockerContainers(apiPort).map((container) => container.name);
+      expect(remaining).toEqual([]);
+    }
 
     try {
       rmSync(dataDir, { recursive: true, force: true });
     } catch {}
   }, 30_000);
```
avallete: This hook is still 30s, but process-compose's orchestrator shutdown budget is 60s (`shutdownTimeoutSeconds`). A slow 13-service dispose can get cut off mid-cleanup and leak exact-owned containers onto the shared daemon.

Raise the `afterAll` budget to cover a full stop plus the leftover check. After listing leftovers, `docker rm -f` those exact names (best-effort), then fail — same gap as the sibling `finally` below.

packages/stack/tests/createStack-docker.e2e.test.ts
```diff
@@ -65,192 +148,339 @@ dockerDescribe("createStack e2e (docker mode)", () => {
...
+          try {
+            if (siblingApiPort !== undefined) {
+              const remaining = ownedDockerContainers(siblingApiPort).map(
+                (container) => container.name,
+              );
+              expect(remaining).toEqual([]);
+            }
+          } catch (error) {
+            siblingContainerCleanupError = error;
+          }
```
avallete, line 433: When `sibling.dispose()` rejects or leaves a survivor, this only records the leftover names. The containers stay on the shared daemon and can name-collide with a later sequential e2e file that is handed the same API port.

Preserve the failure, but `docker rm -f` the exact owned names after capturing them.

packages/stack/tests/createStack-docker.e2e.test.ts
```diff
@@ -65,192 +148,339 @@ dockerDescribe("createStack e2e (docker mode)", () => {
...
+          expect.arrayContaining([
+            expect.objectContaining({ name: "pgmeta", status: "Healthy" }),
+            expect.objectContaining({ name: "studio", status: "Healthy" }),
+          ]),
+        );
+        expect(afterRestart).toEqual(
+          expect.arrayContaining(
+            EAGER_SERVICES.map((name) => expect.objectContaining({ name, status: "Healthy" })),
+          ),
+        );
```
avallete: Leftover from the old Studio-graph test. `pgmeta` and `studio` are already in `EAGER_SERVICES`, which is asserted immediately below.

avallete (approved):
The single journey locks the stated contracts: 13 slim images, mixed eager/lazy, sibling isolation, restart persistence, and exact-name cleanup. Policy overrides that force analytics/vector/pgmeta eager are intentional — do not replace them with `DEFAULT_SERVICE_POLICIES`.

**Land gate:** latest e2e red is this suite's `beforeAll`, from the slim-postgres `/run/postgresql` socket-dir crash. That is already fixed on develop in #6401; this branch does not contain it. Merge develop and land only on green e2e.

Non-blocking comments below: derive expected images from the catalog, raise the `afterAll` budget and force-remove leftover exact-owned names on the failure path, drop the leftover pgmeta/studio restart assert.

packages/stack/tests/createStack-docker.e2e.test.ts
jgoux: Fixed in 61098438f. The sibling failure path now captures exact-owned survivors, force-removes only those names best-effort, preserves all journey/dispose/cleanup failures, and then reports them.

packages/stack/tests/createStack-docker.e2e.test.ts
```diff
@@ -65,192 +148,339 @@ dockerDescribe("createStack e2e (docker mode)", () => {
...
+          } catch (error) {
+            siblingDisposeError = error;
+          }
+
+          try {
+            if (siblingApiPort !== undefined) {
+              const remaining = ownedDockerContainers(siblingApiPort).map(
+                (container) => container.name,
+              );
+              expect(remaining).toEqual([]);
```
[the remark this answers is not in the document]
jgoux: Validated and fixed in 61098438f. Sibling teardown now force-removes only the exact captured owned container names before reporting the preserved cleanup failure.

packages/stack/tests/createStack-docker.e2e.test.ts
jgoux: Fixed in 61098438f. The redundant pgmeta/studio restart assertion was removed; the immediately following EAGER_SERVICES assertion remains the single coverage point.

packages/stack/tests/createStack-docker.e2e.test.ts
jgoux: Fixed in 61098438f. The journey now derives each expected image with dockerImageForService(service, DEFAULT_VERSIONS[service]) and reuses dockerContainerName for exact ownership.

packages/stack/tests/createStack-docker.e2e.test.ts
jgoux: Fixed in 61098438f. The primary afterAll now has a named 90-second budget covering the 60-second shutdown allowance plus inspection overhead. It captures dispose and cleanup failures, force-removes only captured exact-owned survivors best-effort, and always attempts the exact data-directory cleanup.

packages/stack/tests/createStack-docker.e2e.test.ts
```diff
@@ -65,192 +148,339 @@ dockerDescribe("createStack e2e (docker mode)", () => {
...
     await stack?.dispose();
 
-    // Verify all Docker containers are cleaned up after dispose
-    const remaining = execSync(`docker ps -q --filter name=supabase-.*-${apiPort}`)
-      .toString()
-      .trim();
-    expect(remaining).toBe("");
+    // Verify all exact owned Docker containers are cleaned up, including stopped containers.
+    if (apiPort !== undefined) {
+      const remaining = ownedDockerContainers(apiPort).map((container) => container.name);
```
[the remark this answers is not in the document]
jgoux: Validated against the 60-second orchestrator shutdown budget and fixed in 61098438f. The primary teardown now has a 90-second hook budget and exact-owned fallback removal while preserving cleanup failures.

jgoux:
Addressed the review round in 61098438f. The branch includes develop commit cbc47375f from #6401, and the replacement CI run is green, including the complete Docker journey in e2e shard 1. All inline threads are resolved.

PR #6379, opened 2026-08-28
test(stack): qualify native Postgres, Auth, and PostgREST core

jgoux:
## Summary

Qualifies the first Dockerless native vertical slice through the public `@supabase/stack` surface: isolated native prefetch, lazy and eager activation, real Auth and PostgREST flows, PostgreSQL extensions, restart preservation, retryable typed preparation failure, and exact resource cleanup.

Adds a consumer-selected cache root to Promise-based prefetch, exposes the public stack error type, and preserves lazy proxy activation across a whole-stack restart without weakening explicit per-service stops.


packages/stack/tests/createStack-native.e2e.test.ts
```diff
@@ -1,59 +1,409 @@
...
+    }
   }, 30_000);
 
+  test("serves Auth signup and password sessions through native JIT activation", async () => {
+    const client = createClient(stack.url, stack.publishableKey);
+    await client.auth.signOut();
+    const signIn = await client.auth.signInWithPassword({
+      email: authEmail,
+      password: authPassword,
+    });
```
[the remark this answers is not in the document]
jgoux: Fixed in d17f9bf9c. Signup, sign-out, and password sign-in now share one self-contained JIT retry scenario with local credentials; the suite-scoped fixture and order-dependent test were removed. The scenario passes when selected alone.

packages/stack/src/LocalStack.ts
```diff
@@ -957,9 +994,11 @@ export const localStackLayer = (
             return;
           }
           if (runtimeState === undefined) {
+            yield* Ref.set(wholeStackStoppedServicesRef, new Set());
             yield* Ref.set(phaseRef, "stopped");
             return;
           }
+          yield* rememberWholeStackStoppedServices(runtimeState);
```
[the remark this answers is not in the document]
jgoux: Fixed in 41be77567. LocalStack.stop now returns when the lifecycle phase is already stopped, under the existing lifecycle lock, so a repeated stop preserves the original whole-stack service snapshot. The native restart scenario now calls stop twice and still reactivates lazy services through proxy requests.

packages/stack/tests/createStack-native.e2e.test.ts
```diff
@@ -1,59 +1,403 @@
...
+      );
+      expect(rows).toHaveLength(1);
+      expect(rows[0]?.title).toBe("native restart persistence");
+    } finally {
+      await check.close();
+    }
+    const statuses = await stack.getStatus();
+    expect(statuses.find((state) => state.name === "postgres")?.status).toBe("Healthy");
+    expect(statuses.find((state) => state.name === "auth")?.status).toBe("Stopped");
+    expect(statuses.find((state) => state.name === "postgrest")?.status).toBe("Stopped");
```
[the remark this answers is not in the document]
jgoux: Fixed in 41be77567. The restart scenario now activates Auth and PostgREST through real proxy requests and asserts both are Healthy before stopping, so its stopped-policy and post-restart JIT assertions are self-contained. The scenario passes when selected alone.

packages/stack/src/LocalStack.ts
```diff
@@ -1011,6 +1064,9 @@ export const localStackLayer = (
               return { runtime, targets: [service] };
             }).pipe(withLifecycleLock);
             yield* waitForTargets(started).pipe((effect) => withReadinessPolicy(effect, name));
+            yield* clearWholeStackStopAllowanceAfterSuccess(
+              lifecycleTargetsForService(enabledServices, service),
+            );
```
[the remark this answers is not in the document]
jgoux: Fixed in ccbf5c57e. restartService now clears whole-stack recovery allowance only for the requested service it actually restarted, leaving stopped owned companions recoverable by the next activation. The regression uses lazy Analytics/Vector through the public Stack and activator surfaces, reproduces the prior explicit-stop failure, and does not invoke Docker.

packages/stack/src/LocalStack.ts
```diff
@@ -626,6 +630,22 @@ export const localStackLayer = (
...
+        });
+      const wholeStackStopAllowance = Ref.get(wholeStackStoppedServicesRef);
+      const rememberWholeStackStoppedServices = (runtime: RuntimeState) =>
+        Effect.gen(function* () {
+          const running = (yield* runtime.orchestrator.getAllStates).flatMap((state) => {
+            if (state.desired !== "running") return [];
+            const service = SERVICE_NAMES.find((candidate) => candidate === state.name);
+            return service !== undefined && enabledServices.includes(service) ? [service] : [];
+          });
+          yield* Ref.set(wholeStackStoppedServicesRef, new Set(running));
```
avallete: **BLOCKER** — overwrite, not merge.

`Ref.set(..., new Set(running))` keeps only services whose orchestrator `desired` is currently `"running"`. After `stop` → `start`, a lazy service that was in the allowance stays `desired: "stopped"` (mixed-policy start only brings up postgres + eager). The **second** whole-stack `stop` then drops it from the allowance.

```
start → activate auth → stop → start → stop → start → first auth request
```

That last activate hits `Cannot activate auth because dependency auth was explicitly stopped` / 503, even though nobody called `stopService`. Same `Ref.set` also wipes a good snapshot if `stop` is interrupted after desired states flip and then retried (phase is `"stopping"`, so the new early return does not help).

Smallest fix: union into the existing set instead of replacing it:

```ts
Ref.update(wholeStackStoppedServicesRef, (current) => new Set([...current, ...running]))
```

Explicit `stopService` still clears the owned closure, so those names never re-enter via union. Add a unit test for two stop/start cycles without an activate in between, and one that interrupts `stop` mid-teardown then retries.

packages/stack/src/LocalStack.ts
```diff
@@ -956,10 +993,16 @@ export const localStackLayer = (
...
+          }
           if (runtimeState === undefined) {
+            yield* Ref.set(wholeStackStoppedServicesRef, new Set());
             yield* Ref.set(phaseRef, "stopped");
             return;
           }
+          yield* rememberWholeStackStoppedServices(runtimeState);
           yield* Ref.set(phaseRef, "stopping");
           yield* runtimeState.orchestrator.stop;
           yield* Ref.set(phaseRef, "stopped");
```
avallete, line 1011: **SIGNIFICANT** — same overwrite, interrupted-stop path.

The `phase === "stopped"` guard makes a completed `stop()` idempotent (good; the e2e `stop(); stop(); start()` path relies on it). It does not cover `"stopping"`.

If `orchestrator.stop` is interrupted after it has flipped `desired` to `"stopped"` (in-tree: `stack.stop.pipe(Effect.timeout("1 second"))`), a retry re-enters here and `rememberWholeStackStoppedServices` replaces the first snapshot with an empty/partial set. `dispose` is already `uninterruptible`; `stop` is not.

The union fix on the snapshot function is enough. Optionally also skip re-snapshot when phase is already `"stopping"`, or single-flight the teardown.

packages/stack/tests/createStack-native.e2e.test.ts
```diff
@@ -1,59 +1,417 @@
...
+      await sql.unsafe(
+        `INSERT INTO public.todos (title, completed) VALUES ('native restart persistence', true)`,
+      );
+    } finally {
+      await sql.close();
+    }
+
+    await stack.stop();
+    await stack.stop();
+    await stack.start();
```
avallete, line 373: This only exercises a completed `stop` twice (phase already `"stopped"`, snapshot preserved), then one `start`. It does not catch the second real stop/start cycle, where `rememberWholeStackStoppedServices` overwrites the allowance and drops still-stopped lazy services.

A regression for `stop → start → stop → start → activate auth/postgrest` (no activate between the two cycles) would lock the merge fix. The analytics unit test has the same one-cycle gap.

avallete (changes requested):
The native qualification and prefetch `cacheRoot` / `StackError` surface look right. The allowance model (whole-stack stop vs explicit `stopService`, companion `restartService`) is also the right shape.

One bookkeeping bug still breaks the feature this layer is meant to qualify: `rememberWholeStackStoppedServices` overwrites the snapshot with currently `desired === "running"` services. That drops lazy services that are still recoverable after the first restart, and it is the same hole as an interrupted/retried `stop` while phase is `stopping`.

**Changes requested** on that root cause. Details on the lines.

packages/stack/src/LocalStack.ts
jgoux: Fixed in 1d8d64401 with the suggested Ref.update union. Explicit stopService still clears the owned lifecycle allowance. Coverage now exercises two real stop/start cycles with no activation between them, plus an interrupted-stop retry; both regressions reproduced the StackBuildError before the fix.

packages/stack/tests/createStack-native.e2e.test.ts
jgoux: Fixed in 1d8d64401. The native journey explicitly activates Auth and PostgREST before the first stop, then runs stop, idempotent stop, start, stop, start with no activation between the two real cycles, and finally activates both without another download. The analytics/vector unit journey retains restartService coverage and adds the same repeated-cycle regression.

packages/stack/src/LocalStack.ts
jgoux: Fixed in 1d8d64401. The snapshot is now preserved by union when a stop retry enters from phase stopping. The regression waits for the service to enter Stopping and its cleanup gate, delivers interruption immediately, retries stop, restarts, and verifies lazy companion activation succeeds.

packages/stack/src/LocalStack.ts
```diff
@@ -956,10 +993,16 @@ export const localStackLayer = (
...
+          const phase = yield* Ref.get(phaseRef);
+          if (phase === "stopped") {
+            return;
+          }
           if (runtimeState === undefined) {
+            yield* Ref.set(wholeStackStoppedServicesRef, new Set());
             yield* Ref.set(phaseRef, "stopped");
             return;
           }
+          yield* rememberWholeStackStoppedServices(runtimeState);
```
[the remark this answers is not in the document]
jgoux: Confirmed and fixed in 1d8d64401. rememberWholeStackStoppedServices now unions newly observed running services into the existing whole-stack snapshot, so retrying an interrupted stop cannot erase the original lazy-service allowance. A deterministic gated-cleanup regression interrupts stop, retries it, restarts the stack, and successfully reactivates the lazy analytics/vector pair.

avallete (approved):

PR #6389, opened 2026-08-29
feat(workers): bring the command family's output onto one shape

johnstonmatt:
## Summary

The workers commands each grew their own way of saying "here is what happened" and "here is what to run next". This settles them on the shapes the rest of the legacy shell already uses. **No command changes what it does** — this is output, plus the coverage that pins it.

- **Success trailers.** "What to run next" lines in `new`, `push`, `delete` and `status` move to `emitSuccessTrailer`, the way `stop`, `bootstrap`, `migration repair` and `gen signing-key` already emit theirs: printed once at the end of the run rather than inline, so a multi-worker push does not bury each worker's hint under the next worker's output. The commands within them are aqua'd.
- **`list` advisories.** Both take the yellow `WARNING:` prefix and the two-line consequence shape `start`'s Docker notice uses. Each was one long sentence that re-flowed at a different width, directly under a table that lines its columns up.
- **`list` drops the URL column.** Every worker's URL is the same host and prefix with the name on the end, and carrying it pushed the table past 130 columns for one derivable field — `renderGlamourTable` sizes each column to its widest cell and never wraps. `status` still renders it vertically, and every machine format still carries `url` per worker.
- **`push` progress.** Per-worker announcements are counted (`Deploying Worker 1/2:`) and a multi-worker run closes with a summary. Each worker takes minutes; the name alone said nothing about how much of the run was left.
- **`push` names what it never attempted.** The loop stops at the first failure and the error only names the worker that broke, leaving the rest to be reconstructed from argument order. On stderr in every format, machine ones included: that run is a CI run.
- **`--project-ref` survives into `push`'s retry suggestions**, via the `legacyWorkersProjectRefSuffix` helper `status` and `delete` already use. A suggestion is copy-pasted verbatim, so one that dropped it re-resolved against whatever this checkout was linked to.

Also adds unit coverage for `legacyRenderWorkerDetails`, pins the shared `-o env` refusal, and adds a guard (own commit) asserting no legacy boolean flag ships required — `Flag.boolean` alone builds a *required* param, and nothing in the existing suites notices.

## Stack

On top of the `workers new` name prompt (#6349). Above it: `workers logs` (#6410), then `push --wait` (#6371) last, so the output work can ship independently of both.

## Linked issue

[FUNC-851](https://linear.app/supabase/issue/FUNC-851/general-output-polish). Supabase maintainer, exempt from the `open-for-contribution` flow.

## Checklist

- [x] The PR title follows [Conventional Commits](https://www.conventionalcommits.org/)


apps/cli/src/legacy/commands/experimental/workers/list/list.handler.ts
```diff
@@ -68,18 +78,25 @@ function runtimeLabel(row: WorkerRow): string {
...
     runtimeLabel(row),
     row.deployed === undefined ? "-" : formatApiSize(row.deployed.spec.size),
     stateLabel(row),
     row.deployed === undefined ? "-" : String(row.deployed.spec.instances),
-    row.url ?? "-",
   ];
 }
 
-export const legacyWorkersList = Effect.fn("legacy.experimental.workers.list")(function* (
+export const legacyWorkersList = Effect.fn("legacy.workers.list")(function* (
```
kanadgupta: Blocking. Re: the earlier Codex comment about span names — I verified it, and it's right. The base branch has these five handlers at `legacy.experimental.workers.*` (renamed deliberately when the commands moved under the `experimental` parent), and this PR silently renames them all back to `legacy.workers.*`. Neither the PR body nor the commit message mentions it, and it contradicts the "no command changes what it does" claim — the span path is how local traces and the NDJSON exporter identify the command, and it no longer matches the invoked route `supabase experimental workers list`. This looks like a rebase artifact rather than a decision; please restore the `legacy.experimental.workers.*` names in all five handlers (list, status, push, delete, new), or state the rationale if it's intentional.

```suggestion
export const legacyWorkersList = Effect.fn("legacy.experimental.workers.list")(function* (
```

*🤖 AI-drafted comment (Claude), pending human review.*

apps/cli/src/legacy/commands/experimental/workers/push/push.handler.ts
```diff
@@ -439,26 +469,46 @@ export const legacyWorkersPush = Effect.fn("legacy.experimental.workers.push")(f
...
             ? {}
             : { pollRetrySchedule: options.pollRetrySchedule }),
-        }),
+        }).pipe(Effect.tapError(() => reportUnattempted(names.slice(index + 1)))),
+      );
+    }
+
+    // Only for a run that deployed several: one worker already said so itself,
+    // and repeating it as a summary reads like a second deploy.
+    if (names.length > 1 && !machineOutput && output.format === "text") {
```
kanadgupta: This new summary is gated on `!machineOutput && output.format === "text"`, and the per-worker "Deployed Worker" block inside `deployOneWorker` (line 361) is gated on `output.format === "text" && !input.machineOutput` — but the `Deploying Worker n/N:` announcement at line 485 checks only `!machineOutput`. `legacyWorkersMachineOutputRequested` covers just the legacy `-o` flag, so a multi-worker run under `--output-format json`/`stream-json` still prints `Deploying Worker 1/2:` to stderr while printing nothing else human. That's also what makes the new SIDE_EFFECTS.md paragraph ("text-only by contrast") inaccurate — the same mismatch the Codex thread on push/SIDE_EFFECTS.md flagged. I'd gate the announcement the same way this summary and the per-worker block are gated, so all three prose emissions agree and the doc becomes true as written.

*🤖 AI-drafted comment (Claude), pending human review.*

apps/cli/src/legacy/commands/experimental/workers/push/push.integration.test.ts
```diff
@@ -143,6 +156,7 @@ describe("legacy workers push", () => {
       expect(out.stdoutText).toContain("Deployed Worker api");
       expect(out.stdoutText).toContain("Runtime");
       expect(out.stdoutText).toContain(`https://${WORKERS_PROJECT_REF}.supabase.co/workers/v1/api`);
+      expect(out.stdoutText).toContain("v1");
```
kanadgupta: Non-blocking: this assertion is vacuous — the line above already asserts the full URL, whose path contains `/workers/v1/`, so `toContain("v1")` can never fail independently. The default `workerResource` stub carries no `image_version`, so if the intent was to pin the Image row of the details block, it doesn't. Either drop it or assert the row it means to pin.

```suggestion
```

*🤖 AI-drafted comment (Claude), pending human review.*

apps/cli/src/legacy/cli/legacy-boolean-flag-defaults.unit.test.ts
```diff
@@ -0,0 +1,71 @@
...
+ */
+const BOOLEAN_TYPE_NAME = Primitive.getTypeName(Primitive.boolean);
+
+function booleanFlagsRequiringAValue(command: Command.Command.Any): ReadonlyArray<string> {
+  const internals = legacyCommandInternals(command);
+  // All three parameter sets a command can be parsed with, not just its own:
+  // `Command.withSharedFlags` puts inherited flags on `contextConfig`, and the
+  // root's persistent flags arrive as `globalFlags`. A bare boolean introduced
+  // through either would break every command that inherits it while a guard
+  // reading only `config.flags` stayed green.
```
kanadgupta: Non-blocking: the guard itself looks solid (it walks own, shared, and global flags, uses the published introspection helpers rather than `_tag`, and fails loud on an unrecognizable param). But it's a repo-wide CLI-parser guard with no connection to workers output — it's already its own commit, and as its own `test(cli)` PR it would land independently of this stack and keep this diff on-subject. Worth splitting out if this PR needs another revision anyway.

*🤖 AI-drafted comment (Claude), pending human review.*

apps/cli/src/legacy/commands/experimental/workers/push/push.handler.ts
```diff
@@ -436,26 +466,46 @@ export const legacyWorkersPush = Effect.fn("legacy.workers.push")(function* (
     yield* legacyRejectWorkersEnvOutput();
 
     const machineOutput = yield* legacyWorkersMachineOutputRequested();
+    // Computed once for the whole run, the way `status` and `delete` do: an
+    // explicit `--project-ref` has to survive into every hint this push emits.
+    const refSuffix = legacyWorkersProjectRefSuffix(flags.projectRef);
```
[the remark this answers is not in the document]
johnstonmatt: should we reject commands with empty --project-ref flags up front?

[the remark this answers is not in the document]
johnstonmatt: Fixed in de593920d.

`LegacyProjectRefResolver` reads the flag as `Option.isSome(flagValue) && flagValue.value.length > 0`, and the suffix helper only checked `isSome`, so `--project-ref ""` echoed a valueless flag back into every suggestion that command emits. `legacyWorkersProjectRefSuffix` now applies the same `.length > 0` test, with a unit test in `workers.output.unit.test.ts` pinning all three cases.

On rejecting an empty `--project-ref` up front instead: not here. "Empty means absent" is the resolver's deliberate reading and it is shared by every command that takes the flag, so turning it into an error is a behaviour change across the whole shell rather than an output fix — and it would belong in `LegacyProjectRefResolver`, not in `workers`.

apps/cli/src/legacy/commands/experimental/workers/status/SIDE_EFFECTS.md
```diff
@@ -61,11 +61,11 @@ wrapper emits for every command.
...
-| `-o pretty` / `table` / `csv` | the text rendering — these fall through rather than encoding                                           | as above                     |
-| `-o env`                      | refused before any request; the payload nests an instance tally a flat `KEY=value` list cannot express | the error                    |
+| Mode                          | stdout                                                                                                 | stderr                                                              |
+| ----------------------------- | ------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------- |
+| text (default)                | the details block                                                                                      | an unreadable instance tally, and the build-retry hint on a failure |
+| `--output-format json`        | one structured result carrying every reported field                                                    | as above                                                            |
```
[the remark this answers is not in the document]
johnstonmatt: Fixed in 71af01ded.

Both handlers return at `legacyEmitWorkersMachineOutput` or at `output.success`, so the trailer and the stderr notice beside it are only reachable in the text branch. The structured rows in both matrices now say so rather than "as above", and a note under each table explains why — including that `-o pretty`, `table` and `csv` *do* get them, since they encode nothing and fall through to the same text branch.

apps/cli/src/legacy/commands/experimental/workers/push/SIDE_EFFECTS.md
```diff
@@ -72,6 +72,17 @@ payload always carries a `workers` array, which a flat `KEY=value` list cannot
 express, and discovering that at the end would fail the command with the remote
 project already changed.
 
+A multi-worker run stops at the first failure, and names the workers it never
+attempted on stderr in **every** format, machine ones included: that run is a
+CI run, where nobody watched the loop and "what still needs deploying" is the
+question the failure raises. The per-worker `Deploying Worker n/N:` announcement
+is text-only by contrast, since it is progress rather than an outcome.
```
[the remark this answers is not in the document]
johnstonmatt: Fixed in d99e6b518 — gated the announcement rather than documenting the leak.

You are right that `legacyWorkersMachineOutputRequested()` reads `-o` alone, so `--output-format json` left `machineOutput` false. The gate now also checks `output.format === "text"`, which is what the run's closing summary six lines below was already doing. `push.integration.test.ts` covers it, and the test does fail without the gate.

The "not attempted" report stays in every format on purpose: it names what still needs deploying, which is the question an unwatched CI run has to answer.

apps/cli/src/legacy/commands/experimental/workers/list/list.handler.ts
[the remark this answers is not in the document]
johnstonmatt: Fixed in 1f103f99a.

Confirmed against `develop`, which carries `legacy.experimental.workers.*` on all five handlers — this branch had reverted them. Restored across `list`, `push`, `status`, `new` and `delete`.

apps/cli/src/legacy/commands/experimental/workers/push/push.integration.test.ts
kanadgupta: (nit) redundant assertion given the line above it

```suggestion
```

apps/cli/src/legacy/cli/legacy-boolean-flag-defaults.unit.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    return Primitive.getTypeName(single.primitiveType) === BOOLEAN_TYPE_NAME && !isOptional
+      ? [`${command.name} --${single.name}`]
+      : [];
+  });
+
+  return [...own, ...legacyFlattenSubcommands(command).flatMap(booleanFlagsRequiringAValue)];
+}
+
+describe("legacy boolean flag wiring", () => {
+  it("gives every boolean flag a default, so omitting it is not a parse error", () => {
```
kanadgupta: Follow-up candidate: the guard itself looks solid (it walks own, shared, and global flags, uses the published introspection helpers rather than `_tag`, and fails loud on an unrecognizable param). But it's a repo-wide CLI-parser guard with no connection to workers output — it's already its own commit, and as its own `test(cli)` PR it would land independently of this stack and keep this diff on-subject. Fine to ship here too; just noting the cleaner split.

*🤖 AI-drafted comment (Claude), pending human review.*

kanadgupta (commented):
LGTM!

> > [!NOTE]
> > This review was drafted by an AI (Claude). Refreshed after the new commits (head `b3ea7df57`; previously reviewed at `1d7e06b6a`).
> 
> ## What this PR does
> 
> Settles the `experimental workers` family's human output onto the shell's shared shapes: "what to run next" hints in `new`, `push`, `delete`, and `status` move to `emitSuccessTrailer` (verified against how `stop`, `bootstrap`, `migration repair`, and `gen signing-key` emit theirs — usage, text-mode gating, and stderr flush semantics all match); `list`'s two advisories take the `WARNING:` two-line shape and the table drops the URL column; `push` counts its per-worker announcements, closes multi-worker runs with a summary, names never-attempted workers on failure, and carries an explicit `--project-ref` into its retry suggestions. The rest of the diff is tests pinning existing behavior, plus a repo-wide boolean-flag-default guard (own commit).
> 
> ## What the new commits addressed
> 
> All four previously open issues are fixed, and I verified each in the code:
> 
> - **Span names restored** to `legacy.experimental.workers.*` across all five handlers (`1f103f99a`) — this was my one real concern with the previous revision.
> - **`Deploying Worker n/N:` now gated on `output.format === "text"` as well as `-o`** (`d99e6b518`), with a clear comment on why both axes are needed and a new integration test pinning json mode; the push SIDE_EFFECTS "text-only" paragraph is now true as written.
> - **Empty `--project-ref ""` no longer echoed into retry suggestions** (`de593920d`): `legacyWorkersProjectRefSuffix` now mirrors the resolver's empty-means-absent reading, with a unit test. This also resolves what I had flagged as a cross-PR routing — the helper fix landed in this PR.
> - **Status and delete SIDE_EFFECTS matrices corrected** (`71af01ded`): structured rows now say the trailer and stderr notices are text-only, with an accurate note about the `-o pretty`/`table`/`csv` fall-through. I checked the new rows against the handlers' early returns; they match.
> 
> All four Codex threads are answered by the author with the fixing commits and can be resolved.
> 
> ## Remaining findings
> 
> These commands are experimental, so the bar applied here is: only correctness/data-loss/security issues of real note hold the merge.
> 
> **Fix before merge:** none.
> 
> **Follow-up candidates (2):**
> 
> 1. Vacuous `toContain("v1")` assertion in `push.integration.test.ts:159` — already implied by the URL asserted on the previous line, and the default stub carries no image version to pin.
> 2. ~~The repo-wide boolean-flag guard test is solid but unrelated to workers output; as its own `test(cli)` PR it would land independently of this stack.~~ **kanad edit**: removed this feedback, don't think it's worth separating out into its own PR
> 
> ## Does "no behavior change" hold?
> 
> Yes, now without caveats for command semantics: the non-test changes are output text, output streams, hint content, and the two output-gating fixes above. The span rename that previously contradicted the claim is reverted.
> 
> **Overall:** 0 fix-before-merge issue(s), ~2~ 1 follow-up candidate(s).
>

kanadgupta (approved):
whoops lol

PR #6403, opened 2026-08-31
feat(cli): make shadow baseline cache opt-out (default ON)

avallete:
## Summary

Changes the shadow baseline cache (`SUPABASE_SHADOW_CACHE`) from opt-in (default OFF) to opt-out (default ON). This improves performance for `db diff`, `db pull`, and schema declarative operations by caching the baseline PGDATA snapshot by default, while allowing users to disable it by explicitly setting `SUPABASE_SHADOW_CACHE=0` or `false`.

### Key Changes

1. **Environment variable semantics**: `SUPABASE_SHADOW_CACHE` now defaults to enabled when unset. Setting it to any non-viper-true value (`0`, `false`, empty, or garbage) disables the cache.

2. **New `whenUnset` option**: Extended `legacyViperEnvBoolWithProjectFallback` with an optional `whenUnset` parameter that resolves unset keys to a specified boolean, enabling opt-out gate semantics while preserving exact `ParseBool` behavior for present values.

3. **Updated documentation**: All SIDE_EFFECTS.md files and code comments reflect the new default-ON behavior and opt-out mechanism.

4. **Test coverage**: Added tests for the new `whenUnset` option, flipped the default-gate integration test to assert the default-ON cold export + warm restore, and added a project-dotenv opt-out test.

### Behavior

- **Unset or true**: Cache is enabled (cold run snapshots baseline, warm runs restore from snapshot)
- **Explicitly set to `0`/`false`/empty/garbage**: Cache is disabled (uncached create/remove pair, identical to old default)
- **Project dotenv support**: The setting is honored from both shell environment and project `.env` files
- **`sync --no-cache`** still bypasses restore and publish per invocation

Users who explicitly disabled the cache keep their behavior; all other runs now leave up to 3 ~90MB baseline tars (2-day TTL) under `~/.supabase/cache/shadow-baseline/`.

## Linked issue

Closes #

- [x] The linked issue is **open** and carries the `open-for-contribution` label (or I&#39;m a Supabase maintainer).

## Checklist

https://claude.ai/code/session_01Uj7BjiXEnbcWy41ToxCnPs


apps/cli/src/legacy/shared/db-bootstrap/shadow-cache.ts
```diff
@@ -915,6 +917,7 @@ export const legacyAcquireShadowDatabase = <E>(
       !legacyViperEnvBoolWithProjectFallback(
         LEGACY_SHADOW_CACHE_ENV,
         input.setup.projectEnvValues ?? {},
+        { whenUnset: true },
```
jgoux, line 921: **[P3] Fall back early for unusable cache roots.** With the cache now enabled when the variable is unset, an unwritable or root-squashed `SUPABASE_HOME` selects the cold cached lifecycle before discovering that the cache directory cannot be created. Every affected invocation consequently provisions a non-auto-remove container, stops it, fails the cache write, restarts it, waits for readiness, and warns. Could we probe/create the cache root before choosing `legacyColdCachedShadow` and degrade to the plain uncached shadow when it is unusable?

apps/cli/tests/helpers/cli.ts
```diff
@@ -327,7 +327,7 @@ export function spawnSupabase(
     SUPABASE_HOME: homeDir,
     SUPABASE_NO_KEYRING: "1",
     SUPABASE_TELEMETRY_DISABLED: "1",
-    // Isolate e2e from a developer/CI soak (`SUPABASE_SHADOW_CACHE=1`). Cache-subject tests opt in via `options.env`.
+    // Isolate e2e from the default-ON shadow cache. Cache-subject tests opt in via `options.env`.
```
jgoux: **[P3] Make the default-unset state testable at the subprocess boundary.** `spawnSupabase` always injects `SUPABASE_SHADOW_CACHE=0`, while `options.env` only permits string values. The cache e2e can therefore test explicit `1` but cannot run with the key genuinely absent, which is the new shipped default. Could the helper accept `undefined` to remove a base-env key, then use that in the existing cold/warm e2e instead of explicit `1`? This would cover the exact default without adding another e2e run.

jgoux (commented):
Two non-blocking suggestions from an extensive review; no P0-P2 defects found.

jgoux (approved):

apps/cli/src/legacy/shared/db-bootstrap/shadow-cache.ts
avallete: Done in d7a1315. `legacyAcquireShadowDatabase` now probes the cache root (the same `mkdir -p` with mode 0700 the export performs) right after the key-input eligibility checks, before choosing the cached lifecycle. On failure it warns once (`Warning: shadow baseline cache unavailable (cannot create <dir>: …); continuing uncached.`) and returns the plain uncached shadow — `--rm` intact, no stop/export/restart cycle. The export keeps its own mkdir as a safety net for a root that vanishes mid-run. Covered by a new integration test that occupies the cache root's path with a regular file.

---

apps/cli/tests/helpers/cli.ts
avallete: Done in d7a1315. `spawnSupabase`/`runSupabase` now accept `Record<string, string | undefined>` for `options.env`; an `undefined` value removes the key from the merged child env entirely (pins and ambient values alike, since `options.env` spreads last and undefined entries are filtered before spawn). The shadow-cache e2e's cold/warm run now passes `SUPABASE_SHADOW_CACHE: undefined`, so it exercises the genuinely-unset shipped default instead of an explicit `1`.

---

apps/cli/src/legacy/shared/db-bootstrap/shadow-cache.ts
```diff
@@ -923,11 +927,28 @@ export const legacyAcquireShadowDatabase = <E>(
...
+    // The cache root must be usable BEFORE committing to the cached lifecycle: the cold path
+    // drops `--rm` and pays a stop → export → restart cycle whose write is already doomed when
+    // this directory cannot be created (an unwritable or root-squashed `SUPABASE_HOME`) — and
+    // with the cache on by default, every affected invocation would pay that cycle just to warn.
+    // The probe is the exact mkdir the export performs; the export keeps its own as a safety net
+    // for a root that vanishes mid-run.
+    const cacheDir = legacyShadowBaselineCacheDir(input.path);
+    const cacheRoot = yield* Effect.result(
+      input.fs.makeDirectory(cacheDir, { recursive: true, mode: 0o700 }),
     );
```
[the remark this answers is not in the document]
avallete: Confirmed and fixed in b19aee2. The probe now chains `access(W_OK)` after the mkdir, so a pre-existing unwritable root (EACCES for the current user, EROFS on a read-only mount, an NFS root-squash denial) also degrades to the plain uncached shadow before any container is created. Went with `access(2)` rather than a create-and-delete temp file: it demonstrates the same denial cases without leaving any artifact for a crashed run to strand in the shared cache directory. Added an integration test for the pre-existing read-only root (guarded like the workers-push suite's, since chmod cannot revoke write access from a privileged user).

---

PR #6408, opened 2026-08-31
feat(workers logs): FUNC-853 - add supabase workers logs command

johnstonmatt:
Adds a way to read a worker's runtime output, closing the gap between a successful deploy and diagnosing a misbehaving worker. Along the way it makes `workers new`'s name prompt-driven, moves the command family's output onto shared conventions, and makes `workers push` return without waiting on the build by default.

- Add `supabase workers logs <name>`, reading the project's unified logs endpoint (there is no worker-scoped log route) filtered on `log_attributes` rather than the empty top-level `source` column, with `--source`, `--tail`, and a `--follow` mode that polls at the rate limit's floor, dedupes overlapping windows, and streams via `stream-json`
- Print log timestamps in local time in text output, matching the `--debug` HTTP logger, while machine payloads keep UTC and epoch forms
- Make `supabase workers new`'s name argument optional, prompting for it (validated against the same checks the command already enforces) when omitted and there's a terminal to ask on, gated on both stdout and stdin being a TTY
- Add `--wait` to `supabase workers push`, so the command returns once the deploy is accepted rather than blocking on the server-side build; report unattempted workers on a run that fails partway through
- Unify the workers command family's "what to run next" hints onto `emitSuccessTrailer`, standardize `list`'s advisories on the `WARNING:` shape, and drop the URL column from `list`'s table since it made every row 130+ columns wide for a derivable field
- Add a repo-wide test guarding that every legacy CLI boolean flag carries a default, after `--wait` initially shipped without one and broke plain `workers push` parsing

### DIFF
--- a/apps/cli/src/legacy/commands/workers/logs/logs.handler.ts
+++ b/apps/cli/src/legacy/commands/workers/logs/logs.handler.ts
@@ -138,6 +138,10 @@ export const legacyWorkersLogs = Effect.fn("legacy.workers.logs")(function* (
     // `--tail 0` means "no history", not "no new lines".
     const pollTail = Math.max(flags.tail, 1);
 
+    // The stream tag only earns its width when streams are actually mixed; with
+    // `--source` every line would carry the same one.
+    const showStream = Option.isNone(flags.source);
+
     /**
      * Write a batch of lines out, in whichever form the format calls for.
      *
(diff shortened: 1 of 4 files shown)


johnstonmatt:
Closed by a botched stack reorder on my side, not merged — none of these commits ever reached `develop`.

GitHub marked this merged because its head branch's commits became reachable from its base branch (`FUNC-848`) during a force-push, before the PR bases were repointed. A merged PR can't be reopened, so the work continues in #6410 with the same commits, rebased onto the `experimental` move (#6409).

PR #6409, opened 2026-08-31
feat(cli): move workers commands under experimental parent

johnstonmatt:
Add a `supabase experimental` command family, registered with
`Command.unlisted`, and move `workers` underneath it so the family stays
invocable but is explicitly marked unstable and excluded from help, shell
completions, the wizard, and generated docs.

- Add `legacy/commands/experimental/experimental.command.ts` as the new unlisted parent and mount it in `root.ts`, replacing the direct `workers` registration
- Relocate the `workers` command tree (`delete`, `list`, `new`, `push`, `status`, shared helpers, and tests) under `experimental/workers/`, updating relative imports accordingly
- Update every user-facing invocation string (examples, suggestions, error messages, comments) in handlers, tests, and shared worker utilities from `supabase workers ...` to `supabase experimental workers ...`
- Drop the `supabase-workers` entries from the legacy docs tables so the family no longer appears in generated docs


apps/cli/src/legacy/cli/root.ts
```diff
@@ -70,8 +70,8 @@ export const legacyRoot = Command.make("supabase").pipe(
     legacyDbCommand,
     legacyDomainsCommand,
     legacyEncryptionCommand,
+    legacyExperimentalCommand.pipe(Command.unlisted),
```
kanadgupta: Non-blocking: the ticket's core guarantee — that this family stays out of the generated docs reference (and completions) — currently rests on the `.unlisted` filters with no test pinning it, and the same is true of the existing `db test|branch|remote` precedent. Since `legacy-docs-spec.unit.test.ts` already exercises the real command tree, one cheap assertion there that no `supabase-experimental*` slug appears in the generated spec would guard against a future regression quietly publishing the experimental family to the public docs site. Fine as a follow-up.

kanadgupta (approved):
## Review summary

This is a clean, mechanical execution of FUNC-854, and it fully delivers the ticket: the `workers` family is relocated under a new `supabase experimental` parent, registered with `Command.unlisted`, and every discovery surface the repo controls (help via the unlisted flag, shell completions in `legacy-complete.ts`, flag suggestions, and the generated docs spec in `legacy-docs-spec.ts:282`) filters on `.unlisted`, following the established `db test|branch|remote` precedent. I verified there are zero remaining `supabase workers` invocation strings anywhere in `apps/cli` — the sweep across examples, error suggestions, comments, SIDE_EFFECTS.md files, and test assertions is complete.

**Codex's telemetry finding (P2) is fully addressed** by 417755b: all five runtime layers now pass the `["experimental", "workers", <cmd>]` path (`delete`/`list`/`push`/`status` via `legacyManagementApiRuntimeLayer`, `new` via `commandRuntimeLayer`), and all five `Effect.fn` span names carry the `legacy.experimental.workers.*` prefix. Nothing was missed.

Other things I checked and am happy with:

- No stale imports of the old `commands/workers/` path; `root.ts` swaps the registration cleanly.
- The docs-table cleanup (`supabase-workers` tag and `supabase-workers-push instances` override) is consistent with the family no longer being generated.
- The telemetry command series changes from `workers <cmd>` to `experimental workers <cmd>` — a deliberate series break that matches the invocation-path change, per the Codex thread.
- The doc comment on `legacyExperimentalCommand` explaining that graduation is a breaking rename (and that user-facing strings must keep the full path) is exactly the right context to leave behind.
- I considered whether the ~20 repeated `supabase experimental workers` literals should be a shared constant (DRY), and concluded no: they are human-readable message text that tests assert on verbatim, and a prefix constant would hurt grep-ability for no real win.

One non-blocking inline suggestion about pinning the hidden-from-docs guarantee with a cheap unit assertion.

**Verdict: recommend approve.** The inline comment is a nice-to-have follow-up, not a blocker. CI is green across the board.

apps/cli/src/legacy/cli/root.ts
kanadgupta: Additional non-blocking feedback: I'd probably keep the `Command.unlisted` combinator and move it to [this file](https://github.com/supabase/cli/blob/856e786b85c09a95db7e0af9b1d6b11799efbbdb/apps/cli/src/legacy/commands/experimental/experimental.command.ts) to conform to how we do it elsewhere[^1]

[^1]: https://github.com/supabase/cli/blob/856e786b85c09a95db7e0af9b1d6b11799efbbdb/apps/cli/src/legacy/commands/db/db.command.ts#L29-L31

PR #6431, opened 2026-09-02
refactor(workers): consolidate the command family onto shared helpers

johnstonmatt:
Pure refactor of the `workers` command family, one consolidation per commit, stacked on #6371. The duplication it removes was hiding a handful of bugs, fixed here as a side effect:

- `legacyWorkersRun` owns the shared command scaffold — service acquisition, ref resolution and suffix, and the two finalizers — with the ordering written once. Four handlers resolved the ref above both finalizers, so an unlinked non-interactive run failed before the telemetry flush was installed and wrote no post-run event.
- `legacyEmitWorkersPayload` makes the "emit a structured payload" decision once, and `legacyWorkersRendersText` names the "renders human text" condition. Both fix the same precedence bug across ten hand-written call sites: `-o pretty --output-format json` asks for text but was treated as a machine run.
- `legacyWorkersCommand`, with `legacyWorkersPushCommand` and `legacyWorkersStatusCommand` for the two recurring suggestions, replaces the `supabase experimental workers` literal in roughly thirty call sites. A unit test pins the spelling, since these strings get copy-pasted out of a terminal.
- `legacyWorkerNotDeployed` builds the sentence that is identical in three commands and takes the way out that is not: `status` and `logs` point at `push`, `delete` deliberately points at `list`.
- `workers logs`: the follow loop's two always-paired `Ref`s become one `FollowCursor`, and the paging walk moves to `drainSince`, so the poll body reads as its three steps — read the cursor, drain the window, emit what is new.
- `workers push`: the sixty-line source check lifts out of `deployOneWorker` into `assertDeployableSource`, leaving the deploy readable as its steps.

Also trims comments that explained a change rather than the code. Rationale about what used to be wrong, and how many callers had it wrong, lives in these commit messages rather than in a docblock every later reader pays for.


Coly010:
Heads up: this branch now conflicts with `develop` after a couple of recent merges — the legacy-shell rename (#6525, strips the `legacy`/`LEGACY` prefix across ~1400 files) and the `src/legacy/*` flattening (#6486). I looked at auto-resolving it (as I've done for a few other affected open PRs), but the bulk of the conflict here isn't mechanical: this branch built `supabase workers logs` itself (starting `1a2236d61`), and `develop` separately absorbed a materially different implementation of the same command via #6410 after this branch forked — different error types, different output shape, different structure.

Reconciling those is a product/design call (which implementation's behavior wins, or a hybrid), not something I should guess at, so I've left this branch untouched rather than force a merge. Whenever you're ready to pick this back up, worth deciding up front whether to drop this branch's `workers logs` work in favor of develop's (#6410) version or keep yours and reconcile deliberately — that'll make the rest of the merge conflict (which is mechanical) much more mechanical too. Happy to help once that's decided — just flag it.

PR #6432, opened 2026-09-02
feat(workers): add exposure control and new --instances flag

johnstonmatt:
Adds an `exposure` dial (`public`/`private`) for worker deployments alongside the existing runtime/size/instances dials, and lets `workers new` set instance count up front instead of only through `push`.

- Add `exposure` as a closed set (`public`/`private`) on `workers new` and `workers push`, recorded in `config.toml`/`config.json` and resolved with the same precedence as size/runtime: flag override, then recorded value, then `public` default; an unrecognized recorded value refuses the deploy rather than silently coercing it.
- Add `--instances` to `workers new`, written to `config.toml` only when it differs from the default of 1, and rendered as a bare TOML number rather than a quoted string.
- Extend `toml-section.ts` and `worker-config.ts` to support writing numeric values, and update the config schema/docs (`packages/config/src/workers.ts`, `config.schema.json`, `project-config.schema.json`) to describe `exposure`.
- Add `UnknownWorkerExposureError` and corresponding unit/integration test coverage across both commands.


laktek (approved):

apps/cli/src/legacy/commands/experimental/workers/push/push.command.ts
```diff
@@ -24,6 +25,17 @@ const config = {
     ),
     Flag.optional,
   ),
+  exposure: Flag.choice("exposure", WORKER_EXPOSURES).pipe(
```
kanadgupta, line 28: Fix before merge: `exposure` needs to be added to `VALUE_CONSUMING_LONG_FLAGS` in `apps/cli/src/legacy/shared/legacy-db-target-flags.ts`. The completeness test in `legacy-db-target-flags.unit.test.ts` statically scans every `*.command.ts` under `legacy/commands/` for `Flag.choice`/`Flag.string`/... literals and asserts each name is registered — `runtime`, `size` and `instances` are all in the set, `exposure` is not, so that test fails. It doesn't show here because the `Test` job was skipped on this stacked run; it will go red once the unit suite actually runs. Beyond the test, the set is what the telemetry argv scanner uses to know that bare-form `--exposure private` consumes the next token, so the flag's value is mis-parsed in telemetry until it's registered. The github-actions review thread on this line already flags this correctly and it's unaddressed — chiming in to second it.

apps/cli/src/shared/workers/worker-config.ts
```diff
@@ -107,6 +108,10 @@ export function readWorkersSection(workers: unknown): WorkersSection {
     entries[key] = {
       runtime: stringOrUndefined(value["runtime"]),
       size: stringOrUndefined(value["size"]),
+      // Left as whatever string was written, like `runtime` and `size`: `push`
+      // is what names the accepted values, and dropping an unrecognized one here
+      // would silently deploy a worker at the default exposure instead.
+      exposure: stringOrUndefined(value["exposure"]),
```
kanadgupta: Fix before merge: `stringOrUndefined` maps `""` to `undefined`, so a recorded `exposure = ""` reads as absent and `push` silently deploys the worker at the default — public — instead of refusing. That contradicts the comment directly above (an unrecognized recorded value is supposed to reach `push`, which refuses it by name) and it fails in exactly the direction this dial exists to prevent. For `runtime`/`size` the shared helper's empty-string fallback is benign; for `exposure` it can re-expose a worker somebody blanked out deliberately. Passing any string through for this key preserves the refusal contract (`parseWorkerExposure("")` already rejects it):

```suggestion
      exposure: typeof value["exposure"] === "string" ? value["exposure"] : undefined,
```

The depthfirst and adjudicated-review threads on this line already raise this and are unaddressed — agreeing with them.

apps/cli/src/legacy/commands/experimental/workers/push/push.handler.ts
```diff
@@ -142,6 +147,40 @@ function resolveInstances(options: {
...
+ * {@link DEFAULT_WORKER_EXPOSURE}. Never left unset, because every deploy sends a
+ * complete spec and an omitted exposure would re-expose a worker somebody had
+ * deliberately made private.
+ *
+ * `--exposure` is a `Flag.choice`, so only a recorded value can be unrecognized
+ * — and that is refused rather than coerced, the same way `resolveSize` treats a
+ * size it does not know: silently deploying a `private`-typo'd worker as public
+ * is the one outcome nobody asked for.
+ */
+const resolveExposure = Effect.fnUntraced(function* (options: {
```
kanadgupta, line 169: Follow-up candidate: `--exposure` being a one-deploy override matches `--instances` and the help text says so, but for this dial the cost of forgetting is a worker silently going public: `push api --exposure private` on a worker whose config records `public` (which is every worker the new `new` scaffolds) deploys private once, and the next bare `push` re-exposes it. `resolveRuntime` above already has a precedent for a cheap stderr nudge ("Pin it down by adding [workers.<name>] runtime = ... to supabase/config.toml"); emitting the same nudge when the override differs from the recorded value would make the persistence rule hard to miss, without changing the override semantics.

apps/cli/src/legacy/commands/experimental/workers/new/new.handler.ts
```diff
@@ -174,6 +182,55 @@ const resolveSize = Effect.fnUntraced(function* (options: {
...
 });
 
+/**
+ * Recorded on every scaffold, not just when it is asked for: `push` sends a
+ * complete spec each time, so a worker whose `exposure` is absent from
+ * `config.toml` is deployed public by the next bare `push`. Writing the value
+ * down — default included, the way `runtime` and `size` are — is what makes
+ * `--exposure private` stick past the deploy that chose it.
+ */
+const resolveExposure = Effect.fnUntraced(function* (options: {
```
kanadgupta, line 192: Follow-up candidate: this is the third copy of the same explicit → prompt → default resolver in this file (`resolveRuntime`, `resolveSize`, now `resolveExposure`) — identical shape modulo the value set, default, question and label/hint mapping. A single parameterized closed-set dial resolver would collapse roughly forty lines and make the next dial a one-liner. It's consistent with the existing per-dial pattern, so fine to ship as-is; noting it because the file now has three of them.

kanadgupta (changes requested):
works as expected in my lightweight testing 👍 but there are two valid issues that claude flagged that i believe should be addressed prior to merge, see below

> > [!NOTE]
> > This review was drafted by an AI (Claude).
> 
> This PR adds an `exposure` dial (`public`/`private`) to `workers new` and `workers push`, recorded in `config.toml` and resolved with the same precedence as the size/runtime dials (flag override → recorded value → `public` default, with an unrecognized recorded value refusing the deploy), plus an `--instances` flag on `workers new` that is written only when it differs from the default of 1 and rendered as a bare TOML number. The `toml-section.ts`/`worker-config.ts` numeric extension is minimal (one `renderPair` branch and a widened patch type), the closed-set/parse/refuse machinery mirrors the sibling `runtime`/`size` dials closely, and integration/unit coverage is thorough — including the case-insensitive read, the typo refusal before anything is uploaded, and the bare-number round-trip. FUNC-859 is delivered as specced.
> 
> **Fix before merge**
> 
> 1. `--exposure` is missing from `VALUE_CONSUMING_LONG_FLAGS`, so the static completeness test in `legacy-db-target-flags.unit.test.ts` fails — masked on this PR because the `Test` job was skipped in this stacked run — and bare-form `--exposure private` is mis-parsed by the telemetry argv scanner. (Amplifies the unaddressed github-actions thread on `push.command.ts`.)
> 2. A recorded `exposure = ""` is read as absent by `readWorkersSection` and silently deploys at the `public` default, contradicting this PR's own refuse-don't-coerce contract for this dial. (Amplifies the unaddressed depthfirst and adjudicated-review threads on `worker-config.ts:114`.)
> 
> **Follow-up candidates**
> 
> 3. `push --exposure private` is a one-deploy override by design, but the next bare `push` silently re-exposes the worker; a stderr nudge to record the value (precedent: the runtime-guess nudge in `resolveRuntime`) would be cheap.
> 4. `resolveExposure` in `new.handler.ts` is the third copy of the explicit → prompt → default resolver shape; a parameterized closed-set dial resolver could collapse the three.
> 5. The `renderPair` doc comment overclaims: a finite non-integer (e.g. `2.5`) renders as *valid* TOML that only fails on the next config load, so the re-parse does not catch it. Current callers are `Flag.integer`-bounded, so this is a comment/latent-contract fix — the existing bot thread on `toml-section.ts` has it right.
> 6. `new` records `exposure` in `config.toml` only, while the project loader prefers `config.json` when one exists — in such a project a scaffolded `--exposure private` is invisible to `push`. That TOML-only writer is a documented pre-existing constraint from lower in the stack, but `exposure` raises its stakes; worth revisiting when the JSON path is addressed.
> 
> **Existing threads:** the adjudicated review's CRITICAL (empty exposure) and MAJOR (flag registration) findings are correct and unaddressed — items 1–2 above chime in on them. Its refutation of the "explicit `--instances 1` should be written" finding is right: the sparse-config treatment is deliberate, tested, and documented. The two NITs (missing output assertions, `SIDE_EFFECTS.md` table wording on `source`) are fair but minor.
> 
> **Overall:** 2 fix-before-merge issue(s), 4 follow-up candidate(s).

apps/cli/src/legacy/commands/experimental/workers/push/push.command.ts
johnstonmatt: Fixed in 4d0cc0897

apps/cli/src/shared/workers/worker-config.ts
johnstonmatt: Fixed in 35ebbaf78

apps/cli/src/legacy/commands/experimental/workers/push/push.handler.ts
johnstonmatt: Done in d67d67b5f

apps/cli/src/legacy/commands/experimental/workers/new/new.handler.ts
johnstonmatt: i gave this a shot, it wasn't worth the helper function

apps/cli/src/legacy/commands/experimental/workers/push/push.handler.ts
```diff
@@ -144,6 +149,75 @@ function resolveInstances(options: {
...
+      // stderr, so it never lands inside a payload stdout is carrying — and
+      // unguarded by format, like the runtime nudge: a CI run is exactly where
+      // a one-deploy exposure quietly reverting matters most.
+      yield* output.raw(
+        `--exposure ${chosen} applies to this deploy only: supabase/config.toml ${
+          options.recorded === undefined
+            ? `records no exposure for ${options.name}`
+            : `records exposure = "${options.recorded}"`
+        }, so the next bare push will not use ${chosen}. ` +
+          `Pin it down by adding [workers.${options.name}] exposure = "${chosen}" to supabase/config.toml.\n`,
```
kanadgupta: small copy edit suggestion below 👇 i imagine it adds some complexity to this logic for what is a minor change so i'm cool with whatever you run with here

> Follow-up candidate: this nudge fires in three cases, and in two of them `[workers.<name>]` — and usually the `exposure` key itself — already exists in `config.toml` (the `records exposure = "public"` and unreadable-recorded branches). Taking `Pin it down by adding [workers.api] exposure = "private"` literally there appends a duplicate table or key, which makes the file unparseable on the next load. The refusal path below already has the executable verb for that situation (`Set [workers.api] exposure to …`); switching to "set"/"change" when `options.recorded` is defined — keeping "adding" for the no-entry case, matching the runtime nudge — would keep the instruction correct in all three branches.

kanadgupta (approved):
one tiny follow-up below worth noting but happy with this as is!

> > [!NOTE]
> > This review was drafted by an AI (Claude).
> 
> Refresh pass following the submitted CHANGES_REQUESTED review (head was 2f4f76610). Head is now d67d67b5f, retargeted onto `develop` after the stack below merged, and the previously-skipped `Test` job now runs and is green.
> 
> **Prior findings, accounted for:**
> 
> 1. `--exposure` missing from `VALUE_CONSUMING_LONG_FLAGS` — **fixed in 4d0cc0897** (one-line registration; the completeness guard now runs in CI and passes, and the telemetry argv scan parses bare `--exposure private` correctly).
> 2. Blank recorded `exposure = ""` failing open to public — **fixed in 35ebbaf78**. A dedicated `recordedStringOrUndefined` reader keeps the empty string for `exposure` only, so it reaches `push`'s refusal, now with its own "blank exposure" wording; unit + integration coverage including the nothing-was-uploaded assertion. Keeping the collapsing reader for `runtime`/`size`/`source` (whose absence-fallbacks widen nothing) is the right scoping, and the rationale is documented at the helper.
> 3. (follow-up a) One-deploy `--exposure` silently reverting — **done in d67d67b5f**, stronger than suggested: the stderr nudge fires on any drift from what a bare push would resolve to (including an unreadable recorded value), and stays quiet on case-insensitive agreement or a flag restating the default. Five integration tests cover the matrix. One wording issue in the nudge text — see the inline comment.
> 4. (follow-up b) Parameterizing the three dial resolvers — **rebutted by the author, and I agree**: he tried it and dropped it, and the nudge logic added in d67d67b5f makes `resolveExposure` diverge further from `resolveRuntime`/`resolveSize`, so a shared resolver would now need more knobs than it saves.
> 5. (follow-up c) `renderPair` doc overclaim — **fixed in ce7c54dd4**, beyond the comment fix: `isRenderableTomlNumber` guards `planWorkerEntry` before rendering, refusing `1.5`/`-1`/`1e21`/`NaN` with `WorkerConfigWriteUnsafeError` before anything reaches disk, with table-driven tests. This also settles both bot threads on `toml-section.ts`.
> 6. (follow-up d) TOML-only entry writer vs the JSON-preferring project loader — **stands as a follow-up**. No code change; the constraint remains documented in `new/SIDE_EFFECTS.md` as pre-existing from lower in the stack. Acceptable under the experimental bar, but worth revisiting when the JSON write path is addressed — the depthfirst thread on `new.handler.ts` was resolved without a reply, so flagging it here so it isn't lost.
> 
> The two NITs from the adjudicated review (missing output assertions, `SIDE_EFFECTS.md` table wording) were both addressed in d9eca6295.
> 
> **Delta since the last review:** five tightly scoped commits (the four fixes above plus the test/doc commit). I re-checked the empty-exposure reader's blast radius — `entry.exposure` is consumed only by `push`'s `resolveExposure`, so the pass-through cannot leak a blank string anywhere else — and the number guard's placement ahead of both the render and the syntax-only re-parse. Effect usage, error typing, and the testing pyramid all hold; no `as` casts in production code.
> 
> **Retarget sanity:** the diff vs `develop` contains exactly this PR's 20 files (exposure + instances work, their tests, docs, and generated schemas) — no leftovers from the merged stack.
> 
> **New findings:**
> 
> *Follow-up candidates*
> - The drift nudge says "Pin it down by **adding** `[workers.<name>] exposure = …`" even in the two branches where the table and often the key already exist; following it literally there produces a duplicate table/key (inline comment on `push.handler.ts`).
> - Prior finding 6 above (config.json projects can't see a scaffolded `exposure`) remains open.
> 
> **Overall:** 0 fix-before-merge issue(s), 2 follow-up candidate(s).

PR #6464, opened 2026-09-04
ci(cli): install release dependencies fresh on Windows

jgoux:
## Summary

The Windows release smoke job fails during dependency setup after restoring the pnpm global virtual-store cache because `effect-tsgo` cannot resolve its Windows TypeScript binary. The initial cache-miss run succeeded, while both cache-hit attempts reproduced the failure.

Skip only the pnpm store cache on Windows in the shared setup action so fresh installs reconstruct the package links. Keep Go caching on Windows and preserve all dependency caching on other platforms.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -23,14 +23,15 @@ runs:
       with:
         version: 2026.9.0
 
+    # actions/cache-restored pnpm global virtual-store links are unusable on Windows; fresh installs reconstruct them.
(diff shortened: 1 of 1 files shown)


jgoux:
Addressed both automated review findings in `8901734ef`: the Windows guard now lives in the shared setup action, and it skips only the pnpm cache steps. Windows Go caches remain enabled, while every shared-setup caller gets the same safe pnpm behavior.

Coly010 (approved):

PR #6465, opened 2026-09-04
chore: removed the next folder from the cli directory

Prashansa-K:
## Summary

next version of the CLI is a scraped project at the moment. Thus, removing these files so as to reduce confusion.
If required in future, we can always access these via git history.

#### Brief of changes
- Restored genuinely-shared infrastructure from `next/` into `src/shared/` (auth, config, cli/command-docs) with corrected import paths, since legacy actually depends on it.
- Fixed all downstream breakage: import paths across shared/cli/run.ts, telemetry layers, test helpers (mocks.ts), and scripts/generate-docs.ts.
- Modified release tooling to drop the next/alpha shell entirely.
- Removed unused dependencies.
- Rewrote a few things so that references of `next` could be removed from tests, test-helpers or docs.

#### Deliberately left untouched
`apps/cli/AGENTS.md` and `apps/cli/CLAUDE.md` — describes the stale two-shell architecture
`apps/cli/docs/platform-command-generation.md` — documents supabase api routes/api request, a feature that only ever existed in `next/`.
`apps/cli/docs/go-cli-divergences.md`
The above documents can be changed later once we shift the legacy CLI directly under `src`.

## Checklist

- [x] Tests added or updated for the change - NA.


apps/cli/src/shared/cli/run.ts
```diff
@@ -716,9 +715,8 @@ export interface RunCliOptions {
   readonly analyticsLayer: AnyAnalyticsLayer;
   /**
    * Extra command paths (on top of the shared `selfManagedSignalCommands` list) that must NOT
-   * be wrapped in the global signal-interrupt handler for this shell specifically — see
-   * `next/cli/main.ts`'s own `start` exemption for why a shell needs this instead of just
-   * adding to the shared list.
+   * be wrapped in the global signal-interrupt handler for this shell's invocation specifically,
+   * rather than for every shell sharing this runner.
    */
   readonly additionalSelfManagedSignalCommands?: ReadonlyArray<ReadonlyArray<string>>;
```
Coly010: `additionalSelfManagedSignalCommands` existed only to give `next/`'s divergent `start` command tree its own signal-handling exemption (per the comment just above). With `next/` gone, no caller passes a value for it anymore — worth deleting this option as part of the same cleanup, per the repo's Refactoring Policy on obsolete scaffolding.

apps/cli/package.json
```diff
@@ -82,14 +69,11 @@
     "dotenv": "^17.4.2",
     "effect": "catalog:",
     "esbuild": "^0.28.2",
-    "ink": "^7.1.1",
-    "ink-spinner": "^5.0.0",
     "pg": "^8.23.0",
     "pg-copy-streams": "^7.0.0",
     "posthog-node": "^5.51.2",
     "prettier": "3.9.6",
     "react": "^19.2.8",
```
Coly010: Once `shared/runtime/ink.layer.ts` is deleted, `shared/runtime/ink.service.ts` (the `Ink` tag) has zero consumers left anywhere in `apps/cli/src`, and it's the only remaining importer of `react`. So `react` here is now dead weight too — worth removing `ink.service.ts` and this dependency together. Neither is caught by `knip-bun`, so it's an easy one to miss.

Coly010 (commented):
Reviewed the full diff against `origin/develop` (06af58c23) in a local worktree: `pnpm types:check`, `pnpm exec knip-bun`, and the full unit (324 files / 5593 tests) + integration (141 files / 3382 tests) suites all pass clean. The `next/` → `shared/` moves and the release/CI workflow changes (dropping the `alpha` channel) are internally consistent — traced every non-mechanical rename and found no dangling references.

A few things outside this diff worth a follow-up, since they aren't on the PR description's own "deliberately left untouched" list:

- **Root `README.md:132`** — the onboarding quickstart still runs `pnpm dev:next -- --help`, which now fails with `ERR_PNPM_NO_SCRIPT` since `dev:next` is removed from `apps/cli/package.json` (reproduced locally).
- **`apps/cli/CLAUDE.md` / `AGENTS.md`** — understood these are intentionally deferred per the PR description, but flagging that they currently describe a three-tree `legacy/next/shared` architecture and a next/-dual-write obligation that no longer applies, and claim next/ is "moving to a dedicated branch" — there's no such branch in the remote; it's just being deleted with git history as the fallback. Worth prioritizing given how directly these files steer both contributors and coding agents working in this package.
- **`apps/cli/docs/ui.md`** — still documents `@effect/atom-react` / `ink` / `ink-spinner` as "the recommended approach" for terminal UI, with examples lifted from the deleted `StartDashboard.tsx`. None of those packages remain in `package.json` after this PR.

Also worth calling out explicitly in the PR description rather than folding into the general chore: `scripts/generate-docs.ts` now builds the docs-site command reference from `legacyRoot` instead of `nextRoot` — looks like this silently fixes a latent bug where the public CLI reference docs were being generated from the experimental v3 tree instead of the actually-shipped legacy CLI.

pnpm-workspace.yaml
```diff
@@ -12,7 +12,6 @@ allowBuilds:
   sharp: true
 
 catalog:
-  "@effect/atom-react": "4.0.0-rc.112"
```
Coly010, line 15: The corresponding `minimumReleaseAgeExclude` entry (`"@effect/atom-react@4.0.0-rc.111"`, now around line 101) wasn't cleaned up when this catalog entry was removed. Harmless, but stale.

Prashansa-K:
Added a disclaimer line in `cli/AGENTS.md` about scraping of `next/` for now. 
Not revising it yet, as I we may have to revisit it again once we get rid of legacy and move everything directly under src.

Coly010 (approved):

PR #6467, opened 2026-09-04
fix(cli): honor `SUPABASE_API_*` overrides (CLI-2318)

### LINKED ISSUE #6452: seed buckets ignores SUPABASE_API_PORT and calls the storage API on the config.toml port
### Affected area

Storage

### Supabase CLI version

v2.116

### Operating system

Arch Linux

### Installation method

pnpm

### Command

```shell
SUPABASE_API_PORT=54331 supabase seed buckets
```

### Actual output

```shell
{"_tag":"Error","error":{"code":"LegacyStorageGatewayNetworkError","message":"failed to execute http request: HttpClientError: Transport error (GET http://127.0.0.1:54321/storage/v1/bucket)"}}
```

### Expected behavior

The storage gateway should call the resolved `api.port`, the same value `SUPABASE_API_PORT` sets for every other consumer of that setting.

### Steps to reproduce

1. `mkdir /tmp/repro && cd /tmp/repro`
2. `supabase init --yes`
3. Append to `supabase/config.toml`:
   ```toml
   [storage.buckets.app]
   public = false
   ```
4. `SUPABASE_API_PORT=54331 supabase seed buckets`
5. The error names `http://127.0.0.1:54321/storage/v1/bucket`

**NOTE:** The same error happens when running `supabase db reset`.

### Crash report ID

_No response_

```shell

```

### `supabase services`
------------------------|------------------------|--------
supabase/postgres      | 17.6.1.084             | -
supabase/gotrue        | v2.187.0               | -
postgrest/postgrest    | v16.1                  | -
supabase/realtime      | v2.129.3               | -
supabase/storage-api   | v1.37.7                | -
supabase/edge-runtime  | v1.74.3                | -
supabase/studio        | 2026.08.17-sha-0c1da8f | -
supabase/postgres-meta | v0.98.0                | -
supabase/logflare      | 1.50.4                 | -
supabase/supavisor     | 2.9.7                  | -

### `docker version`
```shell
Client:
 Version:           29.7.2
 API version:       1.55
 Go version:        go1.26.5-X:nodwarf5
 Git commit:        a7dcaa6fdb
 Built:             Thu Aug  6 16:23:26 2026
 OS/Arch:           linux/amd64
 Context:           default

Server:
 Engine:
  Version:          29.7.2
  API version:      1.55 (minimum version 1.40)
  Go version:       go1.26.5-X:nodwarf5
  Git commit:       6a43e3d5af
  Built:            Thu Aug  6 16:23:26 2026
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v2.3.4
  GitCommit:        db8809540e1a7a9da5d518876894933ff55692ab.m
 runc:
  Version:          1.5.1
  GitCommit:
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

7ttp:
## TL;DR

fixes `seed buckets`, `db reset`, and `storage --local` ignoring the `SUPABASE_API_*` overrides and calling the local storage gateway on the raw `config.toml` port...

## whats broken?

`SUPABASE_API_PORT=54331 supabase seed buckets` fails with a transport error on `http://127.0.0.1:54321/storage/v1/bucket`
the shared credentials resolver reads `api.port`, `api.external_url`, and `api.tls.*` raw, while `start`/`status`/`stop` resolve the same fields post-override. the CLI can start a stack it then cannot reach.

## now fixed by:

folding the `SUPABASE_API_*` overrides once in `resolveLocalApiConfig` 
thru the same `legacyEnvOverride*` helpers and dotenv walk the sibling commands use
so the gateway URL and TLS gate read the resolved view. a malformed override now fails with the same invalid config error `start`/`status`/`stop` already raise. `--linked` untouched...

## ref:

- closes: https://github.com/supabase/cli/issues/6452


apps/cli/src/legacy/shared/legacy-storage-credentials.ts
```diff
@@ -123,12 +132,75 @@ export const legacyResolveStorageCredentials = Effect.fnUntraced(function* (opts
...
+          api.enabled,
+          "api.enabled",
+          projectEnvValues,
+        ),
+        external_url: legacyEnvOverride(
+          "SUPABASE_API_EXTERNAL_URL",
+          api.external_url,
+          projectEnvValues,
+        ),
+        port: legacyEnvOverridePort("SUPABASE_API_PORT", api.port, "api.port", projectEnvValues),
```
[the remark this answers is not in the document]
7ttp: fixed: af1061b31

apps/cli/src/legacy/commands/seed/buckets/buckets.integration.test.ts
```diff
@@ -1061,6 +1061,177 @@ describe("legacy seed buckets", () => {
     );
   });
 
+  it.live("calls the gateway on the SUPABASE_API_PORT override, not the config.toml port", () => {
+    // #6452: the resolved api.port must reach the storage gateway like every
+    // other consumer of that setting.
+    const previous = process.env["SUPABASE_API_PORT"];
+    process.env["SUPABASE_API_PORT"] = "55511";
```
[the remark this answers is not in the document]
7ttp: fixed: af1061b31

apps/cli/src/legacy/commands/seed/buckets/buckets.integration.test.ts
```diff
@@ -1061,6 +1061,177 @@ describe("legacy seed buckets", () => {
...
+      routes: [
+        { method: "GET", match: "/storage/v1/bucket", body: [] },
+        { method: "POST", match: "/storage/v1/bucket", body: { name: "images" } },
+      ],
+    });
+    return Effect.gen(function* () {
+      const exit = yield* legacySeedBuckets(DEFAULT_FLAGS).pipe(Effect.provide(layer), Effect.exit);
+      expect(Exit.isSuccess(exit)).toBe(true);
+      expect(requests.length).toBeGreaterThan(0);
+      expect(requests.map((r) => new URL(r.url).port)).toEqual(requests.map(() => "55511"));
```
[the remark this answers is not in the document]
7ttp: fixed: af1061b31

Coly010 (commented):
Verified the core fix: `resolveLocalApiConfig`'s zero-port gate matches `legacyValidateResolvedConfig`'s message/ordering exactly, and the override-precedence semantics line up with `start`/`status`/`stop`. Test coverage in `buckets.integration.test.ts` is thorough (port, external_url, TLS enable/cert/key, malformed values, zero-port enabled/disabled). Left inline notes on the items below; two more that fall outside this diff's changed files:

- `legacy-seed-buckets.ts`'s empty-config short-circuit (around the `if (projectRef === "" && bucketNames.length === 0 && !hasVectorBuckets)` check) returns before `legacyResolveStorageCredentials` ever runs, so a malformed `SUPABASE_API_PORT`/zero-port-while-enabled override silently exits 0 when there's nothing configured to seed. Low real-world impact (nothing observable happens either way) — noting for awareness, not blocking.
- Consider filing a follow-up issue for the `resolveLocalServiceRoleKey` dotenv blind spot noted inline below — same root cause class as this fix (#6452), but the vault-decryption path involved makes it more than a one-line change, so it's reasonable to scope it out of this PR.

Nice fix overall — thanks for tracking down #6452.

apps/cli/src/legacy/shared/legacy-storage-credentials.ts
```diff
@@ -123,12 +132,85 @@ export const legacyResolveStorageCredentials = Effect.fnUntraced(function* (opts
...
+      new LegacyStorageConfigError({
+        message: cause instanceof Error ? cause.message : String(cause),
+      }),
+  });
+  // An enabled API with a resolved port of 0 is rejected with the canonical
+  // missing-field message (`legacyValidateResolvedConfig`) — the override could
+  // otherwise smuggle a zero port past config validation into the gateway URL.
+  if (resolved.enabled && resolved.port === 0) {
+    return yield* new LegacyStorageConfigError({
+      message: "Missing required field in config: api.port",
```
Coly010: This duplicates a validation branch that already lives in `legacyValidateResolvedConfig` (`legacy-config-validate.ts`, the `input.api?.enabled && input.api.port === 0` check), including a hand-copied message string. `apps/cli/CLAUDE.md` calls this out explicitly: "Config validation is implemented exactly once... do not add per-command reimplementations of these checks" — and there's a dedicated `legacy-config-validate.parity.unit.test.ts` built specifically to catch message drift across callers, which this third call site isn't part of. If the canonical message ever changes there, this copy can silently diverge with nothing to catch it.

Would it be feasible to factor a small exported helper (e.g. `legacyValidateApiPort`) out of `legacyValidateResolvedConfig` that both call, and extend the parity test to drive this path too? `legacyValidateResolvedConfig`'s full `LegacyConfigValidationInput` is heavier than this call site needs (requires `db`, `storageBucketNames`, etc.), so a narrow shared helper seems better than calling the whole validator with fabricated inputs.

apps/cli/src/legacy/shared/legacy-storage-credentials.ts
```diff
@@ -123,12 +132,85 @@ export const legacyResolveStorageCredentials = Effect.fnUntraced(function* (opts
...
-function resolveLocalBaseUrl(config: LegacyStorageConfigView): string {
-  return legacyResolveApiExternalUrl(config.api, legacyGetHostname());
-}
+const resolveLocalApiConfig = Effect.fnUntraced(function* (
+  fs: FileSystem.FileSystem,
+  path: Path.Path,
+  workdir: string,
+  api: LegacyStorageConfigView["api"],
+) {
+  const projectEnvValues = yield* legacyLoadProjectEnv(fs, path, workdir).pipe(
```
Coly010: This re-reads the project dotenv files from disk independently. Two of this function's callers already have a resolved env map in scope by this point: `seed buckets` already calls `legacyLoadProjectEnv` once (for `SUPABASE_YES`) and discards it, and `start` resolves its own via the richer, ancestor-aware `legacyResolveProjectEnvironmentValues` before folding it into `resolvedConfig`.

Beyond the redundant I/O, `legacyLoadProjectEnv` here always walks `<workdir>/supabase` + `workdir` verbatim with no ancestor project-root search, while `legacyResolveProjectEnvironmentValues` (what `start` uses) can resolve a different directory pair when the project root is discovered by searching ancestors rather than equal to `cliSettings.workdir`. That means the doc comment's claim just above ("a caller that already folded these overrides (start) re-resolves to the same values, so the fold is idempotent") holds in the common case but isn't strictly guaranteed.

Worth threading an optional `projectEnvValues` parameter through `legacyResolveStorageCredentials`, following the existing `resolvedConfig` passthrough pattern on `legacySeedBucketsRun`? Would remove the duplicate read and make the idempotency claim provable rather than argued.

apps/cli/src/legacy/shared/legacy-storage-credentials.ts
```diff
@@ -97,7 +105,8 @@ export const legacyResolveStorageCredentials = Effect.fnUntraced(function* (opts
 
   const fs = yield* FileSystem.FileSystem;
   const path = yield* Path.Path;
-  const baseUrl = resolveLocalBaseUrl(opts.config);
+  const api = yield* resolveLocalApiConfig(fs, path, cliSettings.workdir, opts.config.api);
+  const baseUrl = legacyResolveApiExternalUrl(api, legacyGetHostname());
   const apiKey = yield* resolveLocalServiceRoleKey(opts.config.auth);
```
Coly010, line 128: Pre-existing, not introduced by this PR, but worth flagging while you're in this exact code path: `resolveLocalServiceRoleKey` below reads `SUPABASE_AUTH_JWT_SECRET`/`SUPABASE_AUTH_SERVICE_ROLE_KEY` straight from `process.env` (see its body a few lines down), never from the `projectEnvValues` map that `resolveLocalApiConfig` just loaded two lines above. So a value set only in `supabase/.env` for either auth var stays invisible here, even though the map is already sitting in scope. Same bug class as #6452, just on `auth.*` instead of `api.*`. Might be worth a follow-up issue rather than folding into this PR, since the vault-decryption (`encrypted:`) path makes a faithful fix more than a few lines.

apps/cli/src/legacy/shared/legacy-storage-credentials.errors.ts
```diff
@@ -13,7 +13,10 @@ import {
...
  * `LegacyStorageConfigError` covers the config-load-time validations run
  * before the Storage API client is built (`auth.jwt_secret` length, Kong TLS cert/key pairing
- * and readability). The remaining three mirror `tenant.GetApiKeys` failure
+ * and readability, a malformed `SUPABASE_API_*` port/bool override, an enabled
+ * API whose resolved `api.port` is `0`, and an unreadable/malformed project
+ * dotenv file — see `resolveLocalApiConfig`).
+ * The remaining three mirror `tenant.GetApiKeys` failure
  * modes on the `--linked` path.
  */
 export class LegacyStorageConfigError extends Data.TaggedError("LegacyStorageConfigError")<{
```
Coly010, line 22: `LegacyStorageConfigError` now collapses five causally distinct failure modes (jwt_secret length, TLS cert/key pairing, TLS file readability, malformed `SUPABASE_API_*` override, zero port, unreadable/malformed dotenv) into one telemetry fingerprint with only a free-text `message`. Its siblings in this same file (`LegacyStorageMissingApiKeyError`, `LegacyStorageApiKeysNetworkError`) split on `fingerprint_suffix` for exactly this reason. Not blocking, but might be worth a typed `reason` + `fingerprint_suffix` split at some point so these don't all bucket together as "repeats" in telemetry.

apps/cli/src/legacy/commands/seed/buckets/buckets.integration.test.ts
```diff
@@ -1061,6 +1062,210 @@ describe("legacy seed buckets", () => {
...
+      "SUPABASE_API_PORT",
+      "55511",
+      Effect.gen(function* () {
+        const exit = yield* legacySeedBuckets(DEFAULT_FLAGS).pipe(
+          Effect.provide(layer),
+          Effect.exit,
+        );
+        expect(Exit.isSuccess(exit)).toBe(true);
+        expect(requests.length).toBeGreaterThan(0);
+        expect(requests.every((r) => new URL(r.url).port === "55511")).toBe(true);
```
Coly010: Nit: `expect(requests.every((r) => new URL(r.url).port === "55511")).toBe(true)` loses the actual mismatching value on failure (just reports `false !== true`). Something like `expect([...new Set(requests.map((r) => new URL(r.url).port))]).toEqual(["55511"])` would surface the wrong port directly in the assertion failure. Also, if you're touching this file again: a case combining the port override with an `external_url` override (confirming `external_url` still wins), and the `SUPABASE_API_ENABLED=true` direction over a `enabled = false` config (confirming the override can *turn on* port/TLS validation, not just turn it off) aren't covered yet.

apps/cli/src/legacy/shared/legacy-storage-credentials.ts
7ttp: fixed: 397b142b9

apps/cli/src/legacy/shared/legacy-storage-credentials.ts
7ttp: fixed: 397b142b9

apps/cli/src/legacy/shared/legacy-storage-credentials.ts
7ttp: agreed — will file a follow-up

apps/cli/src/legacy/shared/legacy-storage-credentials.errors.ts
7ttp: agreed — will file a follow-up

apps/cli/src/legacy/commands/seed/buckets/buckets.integration.test.ts
7ttp: fixed: 397b142b9

apps/cli/src/legacy/commands/seed/buckets/buckets.integration.test.ts
```diff
@@ -1061,6 +1062,257 @@ describe("legacy seed buckets", () => {
...
+          Effect.exit,
+        );
+        expect(Exit.isSuccess(exit)).toBe(true);
+        expect(requests.length).toBeGreaterThan(0);
+        expect([...new Set(requests.map((r) => new URL(r.url).port))]).toEqual(["55511"]);
+      }),
+    );
+  });
+
+  it.live("honors a SUPABASE_API_PORT set only in supabase/.env", () => {
```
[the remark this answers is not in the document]
7ttp: fixed: ea4123bbe

apps/cli/src/legacy/commands/seed/buckets/buckets.integration.test.ts
```diff
@@ -1061,6 +1062,257 @@ describe("legacy seed buckets", () => {
...
+      },
+      routes: [
+        { method: "GET", match: "/storage/v1/bucket", body: [] },
+        { method: "POST", match: "/storage/v1/bucket", body: { name: "images" } },
+      ],
+    });
+    return Effect.gen(function* () {
+      const exit = yield* legacySeedBuckets(DEFAULT_FLAGS).pipe(Effect.provide(layer), Effect.exit);
+      expect(Exit.isSuccess(exit)).toBe(true);
+      expect(requests.every((r) => r.url.startsWith("https:"))).toBe(true);
```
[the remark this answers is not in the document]
7ttp: fixed: ea4123bbe

apps/cli/src/legacy/shared/legacy-seed-buckets.ts
```diff
@@ -246,7 +252,11 @@ export const legacySeedBucketsRun = Effect.fnUntraced(function* (opts: {
   }
 
   // Build the Storage service-gateway client (local or remote).
-  const credentials = yield* legacyResolveStorageCredentials({ projectRef, config });
+  const credentials = yield* legacyResolveStorageCredentials({
```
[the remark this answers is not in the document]
7ttp: fixed: ea4123bbe

apps/cli/src/legacy/shared/legacy-storage-credentials.ts
```diff
@@ -97,7 +118,13 @@ export const legacyResolveStorageCredentials = Effect.fnUntraced(function* (opts
...
   const path = yield* Path.Path;
-  const baseUrl = resolveLocalBaseUrl(opts.config);
+  const projectEnvValues =
+    opts.projectEnvValues ??
+    (yield* legacyLoadProjectEnv(fs, path, cliSettings.workdir).pipe(
+      Effect.mapError((cause) => new LegacyStorageConfigError({ message: cause.message })),
+    ));
+  const api = yield* resolveLocalApiConfig(opts.config.api, projectEnvValues);
+  const baseUrl = legacyResolveApiExternalUrl(api, legacyGetHostname());
   const apiKey = yield* resolveLocalServiceRoleKey(opts.config.auth);
```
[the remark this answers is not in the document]
7ttp: already agreed as a follow-up with Coly010 above — same class as #6452, but the encrypted: path makes a faithful fix bigger than this PR. filing it separately.

apps/cli/src/legacy/commands/seed/buckets/buckets.integration.test.ts
```diff
@@ -1061,6 +1062,257 @@ describe("legacy seed buckets", () => {
...
+    // the override must not smuggle a zero port into the gateway URL.
+    const { layer, requests } = setupLegacySeedBuckets(tmp.current, {
+      toml: "[api]\nport = 54321\n[storage.buckets.images]\npublic = true\n",
+      files: { "supabase/.env": "SUPABASE_API_PORT=0\n" },
+      routes: [{ method: "GET", match: "/storage/v1/bucket", body: [] }],
+    });
+    return Effect.gen(function* () {
+      const exit = yield* legacySeedBuckets(DEFAULT_FLAGS).pipe(Effect.provide(layer), Effect.exit);
+      expect(Exit.isFailure(exit)).toBe(true);
+      expect(JSON.stringify(exit)).toContain("Missing required field in config: api.port");
```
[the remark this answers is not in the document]
7ttp: this is the established assertion pattern in this file (and the parity harness) — the tests already go through Exit.isFailure, the stringify only backs the message-contains checks. converting just the new lines would fork the idiom mid-file; a file-wide cleanup can happen separately if we want it.

Coly010 (approved):

PR #6478, opened 2026-09-05
test(cli): cover `network-restrictions` get and update (CLI-2288)

7ttp:
## TL;DR
adds live e2e coverage for `network-restrictions get` and `update`, covering the command family

## whats introduced?
- `network-restrictions get`: reads the target project's restrictions and proves the json payload carries `entitlement`, `config` and `status`
- `network-restrictions update`: captures the current allowlist, replaces it with documentation ranges, proves the replacement in its own output and through get, then restore
s...
## ref:
- closes: CLI-2288
- tested here: https://github.com/supabase/cli/actions/runs/33969368307


apps/cli/src/legacy/commands/network-restrictions/update/update.live.test.ts
```diff
@@ -0,0 +1,165 @@
...
+} from "../../../../../tests/helpers/live.ts";
+
+type LiveCli = LiveFixtures["cli"];
+
+// Every subprocess is bounded so the capture, the update, both proofs and the
+// restore fit inside the live testTimeout even when one command hangs: once a
+// test has timed out its fixtures are disposed, so a late restore cannot take
+// effect and the shared project stays locked down.
+const EXIT_TIMEOUT_MS = 60_000;
+const POLL_ATTEMPT_EXIT_TIMEOUT_MS = 20_000;
```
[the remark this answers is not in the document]
7ttp: fixed: c1e0891d

apps/cli/src/legacy/commands/network-restrictions/update/update.live.test.ts
```diff
@@ -0,0 +1,165 @@
...
+      "network-restrictions get proof for network-restrictions update",
+    );
+  } catch (error) {
+    targetError = error;
+  } finally {
+    try {
+      const restored = await cli(updateArgs(baselineCidrs, flags), {
+        exitTimeoutMs: EXIT_TIMEOUT_MS,
+      });
+      requireLiveSuccess(restored, "network-restrictions update restore of the baseline allowlist");
```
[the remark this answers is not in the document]
7ttp: fixed: c1e0891d

apps/cli/src/legacy/commands/network-restrictions/get/get.live.test.ts
```diff
@@ -0,0 +1,20 @@
...
+  const result = await cli([
+    "network-restrictions",
+    "get",
+    ...experimentalProjectLiveFlags(project),
+    "-o",
+    "json",
+  ]);
+  expect(result.exitCode, result.stderr).toBe(0);
+  expect(result.stdout, result.stderr).not.toBe("");
+  expect(JSON.parse(result.stdout), result.stdout).toMatchObject({
```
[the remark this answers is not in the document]
7ttp: fixed: c1e0891d

apps/cli/src/legacy/commands/network-restrictions/update/update.live.test.ts
```diff
@@ -0,0 +1,165 @@
...
+      EXIT_TIMEOUT_MS,
+    )
+  ).cidrs;
+  // An empty allowlist is restrict-all on the platform, so an empty or absent
+  // family is restored to allow-all rather than leaving the shared project
+  // locked down for every later live test that reaches the database directly.
+  const baselineCidrs: AllowedCidrs = {
+    v4: captured.v4.length > 0 ? captured.v4 : ALLOW_ALL_CIDRS.v4,
+    v6: captured.v6.length > 0 ? captured.v6 : ALLOW_ALL_CIDRS.v6,
+  };
```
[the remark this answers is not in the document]
7ttp: fixed: 9cd12b70

apps/cli/src/commands/network-restrictions/update/update.live.test.ts
```diff
@@ -0,0 +1,186 @@
...
+// plus two proof polls that can each run 102s (a 60s deadline that still
+// finishes an in-flight 20s attempt, waits the 2s interval and runs one last
+// 20s attempt), 444s in all, on top of the workspace fixture's own 60s init.
+// Once a test has timed out its fixtures are disposed, so a late restore
+// cannot take effect and the shared project stays locked down.
+const EXIT_TIMEOUT_MS = 60_000;
+const POLL_ATTEMPT_EXIT_TIMEOUT_MS = 20_000;
+const PROOF_TIMEOUT_MS = 60_000;
+const PROOF_INTERVAL_MS = 2_000;
+const LIVE_TIMEOUT_MS = 600_000;
```
Coly010, line 30: [non-blocking] `LIVE_TIMEOUT_MS = 600_000` is 2x the `live` project's suite-wide `testTimeout` (300_000 in `apps/cli/vitest.config.ts`) and half of the entire `Run live e2e` step budget (`timeout-minutes: 20` in `.github/workflows/live-e2e.yml`) that all 41 `*.live.test.ts` files share, on top of project provisioning that can itself poll for 5 minutes.

The arithmetic in the comment above is right (240s of bounded subprocesses + 2x102s of polling + 60s workspace init ≈ 504s), but the consequence isn't stated: if the platform degrades, this one test can burn half the workflow's wall clock and starve the other 40 live files, and the step gets killed rather than any single test failing cleanly.

Worth either trimming the ceiling (e.g. drop the second restore attempt from the budget, or lower `PROOF_TIMEOUT_MS`) or extending the comment to name the 20-minute workflow bound as the real constraint the ceiling has to fit inside.

apps/cli/src/legacy/commands/network-restrictions/update/update.live.test.ts
```diff
@@ -0,0 +1,186 @@
...
+    ...flags,
+  ];
+}
+
+function describeAttempt(attempt: number, result: LiveRun): string {
+  return `\nattempt ${attempt} (exit ${result.exitCode})\nstdout:\n${result.stdout}\nstderr:\n${result.stderr}`;
+}
+
+// A family the platform leaves absent has nothing to restore, so it reads as
+// `[]` like an explicitly empty one.
```
Coly010: [non-blocking] This collapses two genuinely different platform states into one. `V1GetNetworkRestrictionsOutput.config` declares both CIDR families as `Schema.optionalKey`, so the response distinguishes "family absent" (never configured) from "family present but `[]`" (explicitly block-all) — and `?? []` on lines 89-90 throws that distinction away.

Combined with the fallback on line 135, a project that was deliberately locked down to block-all would be *widened to allow-all* by the restore rather than restored. That can't happen today (the live global setup provisions a brand-new project per run — `${liveProjectName()}-${runId}-${randomUUID()}` — so the capture is always the absent case), but the safety of the fallback rests entirely on that precondition, and nothing here says so.

Cheapest fix: keep `undefined` for an absent family in `Posture`, and reserve the `ALLOW_ALL_CIDRS` fallback for "both families absent", letting an explicitly-empty capture restore as-read. Failing that, name the fresh-project precondition in the comment on line 128.

apps/cli/src/legacy/commands/network-restrictions/update/update.live.test.ts
```diff
@@ -0,0 +1,186 @@
...
+// The allow-all sentinels `config.toml` ships as the `db.network_restrictions`
+// defaults; ADR 0022 treats them as the platform's unconfigured state.
+const ALLOW_ALL_CIDRS: AllowedCidrs = { v4: ["0.0.0.0/0"], v6: ["::/0"] };
+
+const NetworkRestrictions = Schema.Struct({
+  config: Schema.Struct({
+    dbAllowedCidrs: Schema.optionalKey(Schema.Array(Schema.String)),
+    dbAllowedCidrsV6: Schema.optionalKey(Schema.Array(Schema.String)),
+  }),
+  status: Schema.Literals(["stored", "applied"]),
```
Coly010: [non-blocking] Both `config` fields being `optionalKey` with no other required key under `config` means `Schema.is(NetworkRestrictions)` accepts a payload where the CIDR keys were renamed or dropped upstream — `readPosture` then reports an empty allowlist rather than failing, and the capture silently degrades to the `ALLOW_ALL_CIDRS` branch.

The two `expectApplied` proofs would catch that downstream via `toEqual`, so it isn't a hole in the assertions, but the *capture* is unguarded. Adding `entitlement: Schema.Literals(["disallowed", "allowed"])` to this struct would at least pin that the payload is the network-restrictions shape and not some other decodable object.

apps/cli/src/commands/network-restrictions/update/update.live.test.ts
```diff
@@ -0,0 +1,186 @@
...
+    } finally {
+      try {
+        // One re-issue covers a restore that failed transiently. The proof runs
+        // whatever the restore reported: it alone shows whether the allowlist
+        // came back, and a restore killed while its request was still in flight
+        // exits non-zero after the platform may already have applied it.
+        const restore = () =>
+          cli(updateArgs(baselineCidrs, flags), { exitTimeoutMs: EXIT_TIMEOUT_MS });
+        const first = await restore();
+        const restored = first.exitCode === 0 ? first : await restore();
```
Coly010, line 172: [non-blocking] When the first restore fails and the second succeeds, `first` is discarded with no record anywhere — `describeAttempt` only ever renders on the failed-twice path. A restore that needs the retry on *every* run would look identical to one that never needs it, so a chronically flaky restore against staging is invisible in the workflow log.

A `console.warn(...)` with `describeAttempt(1, first)` before the retry costs nothing and makes the retry budget observable.

apps/cli/src/legacy/commands/network-restrictions/update/update.live.test.ts
```diff
@@ -0,0 +1,186 @@
...
+  cidrs: AllowedCidrs,
+  label: string,
+): Promise<void> {
+  return expect
+    .poll(() => readPosture(cli, flags, label, POLL_ATTEMPT_EXIT_TIMEOUT_MS), {
+      interval: PROOF_INTERVAL_MS,
+      timeout: PROOF_TIMEOUT_MS,
+      message: label,
+    })
+    .toEqual({ cidrs, applied: true });
```
Coly010: [non-blocking] `toEqual` on the CIDR arrays makes the proof order-sensitive: the platform has to echo the allowlist back in exactly the order it was posted. That's fine for both allowlists in play today (single-element `TEST_CIDRS` / `ALLOW_ALL_CIDRS`), so this is purely forward-looking — but the "restore the capture as read" branch on line 135 is the one that could see a multi-element array, and that's exactly where an ordering difference would fail the *cleanup* proof after a restore that actually worked. Sorting both sides, or `expect.arrayContaining` plus a length check, would make that branch robust if it ever gets real data.

apps/cli/src/legacy/commands/network-restrictions/get/get.live.test.ts
```diff
@@ -0,0 +1,23 @@
...
+  const result = await cli([
+    "network-restrictions",
+    "get",
+    ...experimentalProjectLiveFlags(project),
+    "-o",
+    "json",
+  ]);
+  expect(result.exitCode, result.stderr).toBe(0);
+  expect(requireLiveJson(result, "network-restrictions get"), result.stdout).toMatchObject({
+    entitlement: expect.stringMatching(/^(?:allowed|disallowed)$/u),
```
Coly010: [non-blocking] Two of these three assertions are close to tautological: `entitlement` and `status` are already `Schema.Literals` in `V1GetNetworkRestrictionsOutput` (`packages/api/src/generated/contracts.ts:3858`), so a payload that failed either regex would have failed the generated client's decode first and the command would have exited non-zero — already caught by the `exitCode` check on the line above.

The assertion carrying real live-only signal is `config`, and `expect.any(Object)` is the weakest one here: on the freshly-provisioned suite project `config` is `{}`, so it only proves the key exists (and `expect.any(Object)` also admits arrays and `Date`s). Since the golden path can't assume a populated allowlist, consider at least pinning that the CIDR families, when present, are arrays of strings — that's the part the platform can actually drift on.

apps/cli/tests/helpers/live.ts
```diff
@@ -124,6 +124,20 @@ export function requireLiveSuccess(
...
 
+/** Parse a command's stdout as JSON, failing with both streams when it is not. */
+export function requireLiveJson(
+  result: { readonly stdout: string; readonly stderr: string },
+  command: string,
+): unknown {
+  try {
+    return JSON.parse(result.stdout);
+  } catch {
+    throw new Error(
```
Coly010, line 126: [non-blocking] The underlying `SyntaxError` is dropped. `new Error(msg, { cause: error })` keeps the parse position, which is the difference between "stdout isn't JSON" and "stdout is JSON followed by a stray progress line" when this fires on CI against staging.

Also worth noting for future readers: inside `expectPostgresConfigLiveOverride` there are now two differently-worded errors for adjacent failures — "did not print JSON" from here, and "unexpected postgres-config get payload" from the `Predicate.isObject` guard just below for valid-JSON-but-not-an-object. That's a behaviour-preserving split (the old code funnelled both into the second message), just no longer a single message.

Coly010 (approved):
Solid, unusually careful live coverage. Resource hygiene is the thing I looked hardest at and it holds up: the mutation is wrapped in try/finally, the restore is retried once, the restore is *proved* through `get` rather than trusted, and cleanup failures are aggregated via `throwWithCleanup` instead of swallowed. I verified all four CIDRs (`203.0.113.0/24`, `2001:db8::/32`, `0.0.0.0/0`, `::/0`) survive the local `validateAndPartitionCidrs` private-range check, so the restore can't be rejected client-side and strand the project; the code comments citing RFC 5737/3849, ADR 0022 and the `V1GetNetworkRestrictionsOutput` annotation are all accurate. The `live.ts` change is backward-compatible: `expectPostgresConfigLiveOverride` was already throwing on non-JSON stdout (`config = undefined` -> `!Predicate.isObject` -> throw), so the only difference is the wording of the error, and both call sites (postgres-config `update`/`delete`) keep the same control flow inside and outside `expect.poll`. No secrets, no real project refs, no `.only`/`.skip`, `gh pr checks` is green, and the linked Live E2E run (33969368307) passed on 9cd12b703 — the final content of all three files (the later merge commit doesn't touch them). Findings below are all non-blocking.

apps/cli/src/commands/network-restrictions/update/update.live.test.ts
7ttp: fixed: 34af1c0b

apps/cli/src/legacy/commands/network-restrictions/update/update.live.test.ts
7ttp: fixed: 34af1c0b

apps/cli/src/legacy/commands/network-restrictions/update/update.live.test.ts
7ttp: fixed: 34af1c0b

apps/cli/src/commands/network-restrictions/update/update.live.test.ts
7ttp: fixed: 34af1c0b

apps/cli/src/legacy/commands/network-restrictions/update/update.live.test.ts
7ttp: fixed: 34af1c0b

apps/cli/src/legacy/commands/network-restrictions/get/get.live.test.ts
7ttp: fixed: 34af1c0b

apps/cli/tests/helpers/live.ts
7ttp: fixed: 34af1c0b

PR #6483, opened 2026-09-06
fix(cli): consume `--log-level` value (CLI-2329)

### LINKED ISSUE #6482: supabase sso update fails when using --log-level
### Affected area

Auth

### Supabase CLI version

v2.117.0-beta.21

### Operating system

Linux (Arch)

### Installation method

npm

### Command

```shell
supabase sso update --log-level error 00000000-0000-0000-0000-000000000000 --metadata-url https://example.com
```

### Actual output

```shell
accepts 1 arg(s), received 2

```
Image

### Expected behavior

The issue is that --log-level isn't being recognized as a value-taking global flag, so its value ("error") gets counted as an extra positional argument.

### Steps to reproduce

1.Run sso update with the global flag --log-level error:  supabase sso update --log-level error 00000000-0000-0000-0000-000000000000 --metadata-url https://example.com

2. Notice that it fails with accepts 1 arg(s), received 2.

3.Run the same command with another global flag that takes a value, such as --workdir .:  supabase sso update --workdir . 00000000-0000-0000-0000-000000000000 --metadata-url https://example.com

### Crash report ID

_No response_

```shell

```

### Additional context

_No response_

7ttp:
## TL;DR
fixes `supabase sso update --log-level error <id>` failing with 
`accepts 1 arg(s), received 2` by registering the built-in `--log-level` as value taking in the raw argv scanners..

## whats biting the user?
`--log-level` shows up in every command's help and the parser accepts it, but the raw argv scanners did not know it consumes a value
so `error` was counted as an extra positional and the command refused to run. A typed `--log-level` also killed shell completion for the rest of the line.

## now fixed by:
- registering `log-level` in `PERSISTENT_VALUE_FLAG_NAMES` and `globalFlagsWithValues`, so positional counting consumes its value like every other global
- registering the built-in flags where output format and shell completion resolve, so `--version` output and tab completion keep working around them

also added regression tests for the issue's exact spelling plus the pre-path, inline, completion, and version spellings

## ref: 
- closes CLI-2329
- closes https://github.com/supabase/cli/issues/6482


apps/cli/src/legacy/cli/legacy-complete.ts
```diff
@@ -272,6 +268,11 @@ export function legacyCollectInScopeFlags(
   const ownParams: Array<Param.AnyFlag> = [
     ...globalFlagParamsOf(finalCommand),
     GlobalFlag.Help.flag,
+    // The built-in `--log-level` is part of every resolved command's real
+    // flag set (`GlobalFlag.BuiltIns`, shown in `--help`), so the strict
+    // flag walk must resolve it or a typed `--log-level error` poisons the
+    // whole line like an unknown flag (issue #6482).
+    GlobalFlag.LogLevel.flag,
```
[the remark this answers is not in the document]
7ttp: fixed: 4b620a385

apps/cli/src/shared/cli/run.ts
```diff
@@ -104,6 +104,7 @@ const globalFlagsWithValues = new Set([
   "--network-id",
   "--dns-resolver",
   "--agent",
+  "--log-level",
```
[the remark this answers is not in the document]
7ttp: fixed: 4b620a385

apps/cli/src/legacy/cli/legacy-complete.ts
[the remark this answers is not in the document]
7ttp: fixed: 4b620a385

apps/cli/src/legacy/shared/legacy-db-target-flags.ts
```diff
@@ -97,6 +97,10 @@ export const VALUE_CONSUMING_LONG_FLAGS = new Set([
   "network-id",
   "dns-resolver",
   "agent",
+  // The CLI library's built-in `--log-level` is deliberately not listed: its
+  // value is parser-enforced to a closed enum and can never be flag-shaped,
+  // so no scanner here can mis-consume around it (issue #6482 registered it
+  // only where positional counting depends on it).
```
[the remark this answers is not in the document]
7ttp: fixed: 4b620a385

apps/cli/src/shared/cli/agent-output.ts
```diff
@@ -113,13 +120,19 @@ function isRootValueFlagWithInlineValue(arg: string): boolean {
...
 }
 
 function isRootBooleanFlag(arg: string): boolean {
   return (
-    arg === "--debug" || arg === "--experimental" || arg === "--yes" || arg === "--create-ticket"
+    arg === "--debug" ||
+    arg === "--experimental" ||
+    arg === "--yes" ||
+    arg === "--create-ticket" ||
+    arg === "--wizard"
```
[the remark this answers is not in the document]
7ttp: fixed: 4b620a385

apps/cli/src/shared/cli/cobra-flag-groups.ts
```diff
@@ -103,12 +103,20 @@ export function lastExplicitLongFlagValue(
  * Value-taking long flags registered persistently on the Go root command
  * (`apps/cli-go/cmd/root.go:324-333`: `--workdir`, `--network-id`,
  * `--profile`, `--output`, `--dns-resolver`, `--agent`), plus the TS-only
- * `--output-format` global (`shared/cli/global-flags.ts`) which the TS
- * parser accepts on any subcommand. pflag lets any of these consume the
- * following argv token, so a pflag-faithful scan must know them or it will
- * miscount positionals on perfectly normal invocations like
- * `sso update --workdir . <id>`. Keep in sync with `globalFlagsWithValues`
- * in `shared/cli/run.ts`.
+ * globals the TS parser accepts on any subcommand: `--output-format`
```
Coly010, line 106: [non-blocking] Stale line citation on line 104: that persistent-flag block is now `apps/cli-go/cmd/root.go:337-348` (`flags := rootCmd.PersistentFlags()` at 337, `BindPFlags` at 348), not `324-333`. The drift is pre-existing, but this PR is rewriting this exact doc block, so it is a free fix. Two more of the same nearby: `PERSISTENT_VALUE_FLAG_SHORTHANDS`'s doc (`cmd/root.go:330`, now 344) and `shared/legacy/global-flags.ts:118` (`root.go:344-354`, same 337-348 block).

Alternative worth considering: everything this set guards is TS-parser behaviour, and `apps/cli/CLAUDE.md` asks new comments not to be framed against Go. Describing the set in its own terms and dropping the file:line citation would stop it going stale a fourth time.

apps/cli/src/shared/cli/run.ts
```diff
@@ -104,6 +106,8 @@ const globalFlagsWithValues = new Set([
...
+  "--completions",
```
Coly010: [non-blocking] This PR is the strongest argument yet for consolidating these registries: adding one flag required four hand edits (`globalFlagsWithValues` here, `PERSISTENT_VALUE_FLAG_NAMES`, both `agent-output.ts` predicates, `legacyCollectInScopeFlags`), a deliberate skip in a fifth (`VALUE_CONSUMING_LONG_FLAGS`), and `commands/gen/types/types.handler.ts:159-176` already carried its own `LONG_FLAGS_WITH_VALUES` listing `log-level` and `completions` before this PR. One consumer being ahead of the others is exactly how #6482 slipped through.

`PERSISTENT_VALUE_FLAG_NAMES`'s new doc asserts this set "carries every name here plus `--completions`", but nothing enforces it. Cheap follow-up: export one source of truth and derive each consumer's shape from it, or at minimum add a drift guard asserting this set equals `PERSISTENT_VALUE_FLAG_NAMES` mapped to `--`-prefixed tokens plus `--completions` -- there is precedent in `global-flags.unit.test.ts`, which pins the `LEGACY_GLOBAL_FLAGS` id set the same way. Simplest alternative: just add `completions` to `PERSISTENT_VALUE_FLAG_NAMES` for uniformity; I verified it would be inert, since no argv carrying a parsed `--completions` ever reaches a handler (the action returns at `runWith` step 5).

apps/cli/src/shared/cli/agent-output.ts
```diff
@@ -113,14 +120,53 @@ function isRootValueFlagWithInlineValue(arg: string): boolean {
...
+}
+
+// Inline spellings count for the skipped booleans with ANY value: the Version
+// action is scanned on presence before `--wizard=bogus` ever parses, so even
+// an invalid inline value there still renders the plain version line.
 function isRootBooleanFlag(arg: string): boolean {
-  return (
-    arg === "--debug" || arg === "--experimental" || arg === "--yes" || arg === "--create-ticket"
-  );
+  return ROOT_BOOLEAN_FLAGS.some((name) => isFlagOccurrence(arg, name));
```
Coly010, line 160: [non-blocking] Inline boolean spellings are handled now, but the space-separated boolean literal still is not -- and the real parser does consume it: for a bare boolean flag, `consumeFlagValueWithTokens` takes the next token whenever `asBooleanLiteral` matches (`effect/unstable/cli/internal/parser.ts`, over `Config.TrueValues`/`FalseValues`). So `--wizard false --version` and `--debug 0 --version` print the plain version line, while this scan stops on the operand `false` and returns `json` for a detected agent: the same class of mismatch this PR fixes for `--completions bash --version`. `run.ts`'s `rootFlagTokens`/`firstPositionalIndex` share it (the literal becomes the first positional, so a later `--version` is discounted).

Pre-existing for `--debug`/`--yes`/`--experimental`/`--create-ticket` and obscure enough not to block, but now that `BOOLEAN_FLAG_VALUES` is exactly the literal set the parser accepts, it is a couple of lines: when `isRootBooleanFlag` matched the bare spelling and `args[i+1]` is in `BOOLEAN_FLAG_VALUES`, skip that token too.

apps/cli/src/shared/cli/agent-output.ts
```diff
@@ -113,14 +120,53 @@ function isRootValueFlagWithInlineValue(arg: string): boolean {
...
+  "--wizard",
+];
+
+/** Bare or inline (`--flag` / `--flag=<value>`) occurrence of `name`. */
+function isFlagOccurrence(arg: string, name: string): boolean {
+  return arg === name || arg.startsWith(`${name}=`);
+}
+
+/** Inline values the CLI's boolean primitive accepts (lowercase only). */
+const BOOLEAN_FLAG_VALUES: ReadonlySet<string> = new Set([
```
Coly010, line 134: [non-blocking] Verified this is precisely `Config.TrueValues` + `Config.FalseValues` (`Config.ts:1224-1227`), including the case-sensitivity the comment calls out -- good.

Worth a cross-reference though: `run.ts` answers the same `--version=<value>` question with `PFLAG_BOOL_TRUE` (`1/t/T/TRUE/true/True`) in `hasRootHelpOrVersionFlag`, so the two sets now disagree in both directions -- `--version=t` and `--version=True` (pflag-true, but rejected by this parser, so no action runs) and `--version=yes|on|y` (accepted here, pflag-false). Only the upgrade-notice base-dir/force-fetch choice rides on the `run.ts` one, so nothing user-visible breaks today, but a one-line pointer between the two -- or sharing this set -- would stop the next reader assuming they agree.

apps/cli/src/shared/cli/agent-output.ts
```diff
@@ -147,7 +193,8 @@ function hasRootVersionRequest(args: ReadonlyArray<string>): boolean {
 function hasHelpRequest(args: ReadonlyArray<string>): boolean {
   for (const arg of args) {
     if (arg === "--") return false;
-    if (arg === "--help" || arg === "-h") return true;
+    if (isBooleanActionOccurrence(arg, "--help") || isBooleanActionOccurrence(arg, "-h"))
```
Coly010, line 199: [non-blocking] `hasRootVersionRequest` skips the token a value-taking flag consumes; `hasHelpRequest` still scans raw and only stops at `--`. Now that `--log-level`/`--completions` are value-taking, `--log-level -h` reports a help request even though the real parse fails first (`consumeGlobalFlags` records `missingFlagValue` and `runWith` returns `ShowHelp` at step 4, before any action), so that error renders as plain text instead of the agent JSON envelope.

Same function also misses a clustered `-hv`/`-vd`, which the lexer does split into separate short options (`internal/lexer.ts`, clustered-short branch), so `supabase -hv` under a detected agent JSON-wraps help output. Both pre-existing and low impact; if you want them consistent, `hasHelpRequest` could reuse the same skip loop `hasRootVersionRequest` already has.

apps/cli/src/cli/legacy-complete.ts
```diff
@@ -275,6 +266,16 @@ export function legacyCollectInScopeFlags(
...
+    // The built-ins `--log-level`, `--wizard`, and `--completions` are part
+    // of every resolved command's real flag set (`GlobalFlag.BuiltIns`, shown
+    // in `--help`), so the strict flag walk must resolve them or a typed
+    // `--log-level error` poisons the whole line like an unknown flag
+    // (issue #6482). Unlike `--help`/root `--version`, whose short-circuit
+    // below ends the completion request, these three keep the line
+    // completing normally.
+    GlobalFlag.LogLevel.flag,
+    GlobalFlag.Wizard.flag,
+    GlobalFlag.Completions.flag,
```
Coly010, line 278: [non-blocking] Cosmetic. Of these three, `--wizard` genuinely should keep completing -- wizard mode takes a command path, so offering subcommands after `supabase --wizard ` is useful. `--completions bash` is a print-and-exit action like `--version`, so `__complete --completions bash ""` offering root subcommands reads a little oddly next to the `--help`/root-`--version` short-circuit below. No user impact (the completion script never emits that argv), so this is just a note in case the grouping later looks accidental rather than chosen.

apps/cli/src/docs/legacy-docs-introspection.ts
```diff
@@ -76,11 +76,12 @@ export function legacyFlattenSubcommands(
...
- * and `GlobalFlag.LogLevel` are TS-only framework additions with no Go/cobra
- * equivalent; they are normally only injected via `GlobalFlag.BuiltIns` at
+ * and `GlobalFlag.LogLevel` are only injected via `GlobalFlag.BuiltIns` at
  * parse time (never stored on a command's own `.globalFlags`), so the filter
- * is a defensive guard rather than something that changes today's output —
- * kept explicit so it stays true if that ever changes.
+ * is a defensive guard rather than something that changes today's output.
+ * Unlike shell completion (`legacy-complete.ts`, which resolves the built-ins
+ * because the real parser does — issue #6482), the docs spec deliberately
+ * keeps the Go-shaped flag surface and leaves them out.
```
Coly010: [non-blocking] Two small things about this rewrite.

(a) The filter is provably dead -- the built-ins only ever arrive via `GlobalFlag.BuiltIns` at parse time, which is exactly why the identical guard could be deleted in `legacy-complete.ts` in this same PR. Keeping it here as future-proofing is defensible, but the two files now treat one fact oppositely with only prose tying them together.

(b) "keeps the Go-shaped flag surface" is the sort of new Go-framed justification `apps/cli/CLAUDE.md` asks us to avoid. The substantive question underneath is that generated docs omit three flags `--help` does show -- a docs-surface decision rather than something to settle in this PR. A follow-up issue plus a reason stated in its own terms (for example "the docs spec covers command-owned flags only") would age better.

apps/cli/src/commands/sso/update/update.integration.test.ts
```diff
@@ -671,11 +671,15 @@ describe("legacy sso update integration", () => {
...
-      // sail through to the PUT exactly as before.
+      // sail through to the PUT exactly as before. The two `--log-level`
+      // variants pin issue #6482: the built-in global's value mis-counted as
+      // a positional (post-path), and the pre-path spelling's anchoring.
       const argvVariants: ReadonlyArray<ReadonlyArray<string>> = [
         ["sso", "update", "--workdir", ".", VALID_PROVIDER_ID],
         ["sso", "update", "--output-format", "json", VALID_PROVIDER_ID],
         ["sso", "update", "-o", "json", VALID_PROVIDER_ID],
+        ["sso", "update", "--log-level", "error", VALID_PROVIDER_ID],
+        ["--log-level", "error", "sso", "update", VALID_PROVIDER_ID],
```
Coly010, line 685: [non-blocking] The post-path variant on line 681 is a genuine red-before/green-after guard. This pre-path one is not quite what the comment above claims: unregistered, the anchor walk broke on `error` and fell back to an unanchored scan, and the arity check is gated on `scan.anchored`, so this argv already sailed through to the PUT before the fix. Its real value is "the newly anchored path must not break a valid invocation", which is worth keeping -- the anchoring assertion itself lives at the right level in `cobra-flag-groups.unit.test.ts` (`expect(scan.anchored).toBe(true)`). Suggest softening the comment so a future reader does not trust this line as the anchoring regression guard.

Coly010 (approved):
Reviewed as a TS-shell change: `--log-level`/`--wizard`/`--completions` come from `effect/unstable/cli`'s `GlobalFlag.BuiltIns` and have no cobra counterpart, so Go parity is not a review dimension here. I checked the fix against the Effect parser's own semantics (`internal/parser.ts`, `internal/lexer.ts`, `Command.runWith`'s step order), the shell's tests, and shipped output. The core fix is right and well targeted: registering `log-level` in `PERSISTENT_VALUE_FLAG_NAMES`/`globalFlagsWithValues` fixes the arity miscount at the only positional-counting consumer (`sso update`), all three `pflagArgvScan` call sites spread that same registry so they cannot drift, the `agent-output.ts` action predicates match `runWith`'s real order (built-ins prepended, Help > Version > Wizard > Completions, actions matched on presence before any Setting is parsed), `BOOLEAN_FLAG_VALUES` is exactly `Config.TrueValues` + `Config.FalseValues`, and I separately confirmed the deliberate `VALUE_CONSUMING_LONG_FLAGS` omission is safe (its three consumers are all handler-time, and any argv where `--log-level` would mis-consume fails `consumeGlobalFlags` before a handler runs). CI is fully green.

No blocking findings. 8 non-blocking notes below: a stale line citation this PR is already rewriting, registry sprawl with no drift guard, two remaining fidelity gaps in the pre-parse scanners (space-separated boolean literals, `hasHelpRequest` not skipping consumed values), a boolean-vocabulary split with `run.ts`, and some comment-accuracy nits.

apps/cli/src/shared/cli/cobra-flag-groups.ts
7ttp: fixed: 135fd37ae

apps/cli/src/shared/cli/run.ts
7ttp: fixed: 135fd37ae

apps/cli/src/shared/cli/agent-output.ts
7ttp: fixed: 135fd37ae

apps/cli/src/shared/cli/agent-output.ts
7ttp: fixed: 135fd37ae

apps/cli/src/docs/legacy-docs-introspection.ts
7ttp: fixed: 135fd37ae

apps/cli/src/commands/sso/update/update.integration.test.ts
7ttp: fixed: 135fd37ae

apps/cli/src/shared/cli/agent-output.ts
7ttp: deliberate for now — the walk is a conservative spelling allowlist, and `--log-level -h` actually lands on the exit-1 ShowHelp path that still renders the full help doc as text, so adding the value-skip here would turn that render into a JSON-wrapped help doc for agents, which felt worse than the current text. clustered `-hv` is real but needs the lexer's cluster model inside this walk — lumping both into the registry-consolidation follow-up rather than growing a third hand-rolled scanner here.

apps/cli/src/cli/legacy-complete.ts
7ttp: deliberate — the strict walk has to resolve `--completions` either way, and short-circuiting `--completions bash` would special-case an argv the completion script never emits; keeping the line completing matches how `--wizard`/`--log-level` are treated, and the comment next to the built-ins records the choice so the grouping reads as chosen. happy to revisit if it ever bites.

7ttp: deliberate
 the strict walk has to resolve `--completions` either way, and short-circuiting
 `--completions bash` would special-case an argv the completion script never emits; keeping the line completing matches how `--wizard`/`--log-level` are treated, and the comment next to the 
built-ins records the choice so the grouping reads as chosen. will revisit if it ever bites...

PR #6492, opened 2026-09-07
test(cli): cover `branches` get, update and disable (CLI-2327)

7ttp:
## TL;DR

Adds live e2e coverage for `branches get`, `branches update` and `branches disable`

## whats introduced?

- `branches get`: creates a branch, fetches it by name and asserts the pretty connection table renders
- `branches update`: creates a branch, renames it with `--name --output json`, asserts the confirmation and payload, then proves the new name resolves through `branches get`
- `branches disable`: creates and deletes a branch so branching is enabled with no preview branches left, disables preview branching for the project and asserts the confirmation on stdout

## ref:

- closes: CLI-2327
- passed here: https://github.com/supabase/cli/actions/runs/34112181467


apps/cli/src/commands/branches/update/update.live.test.ts
```diff
@@ -0,0 +1,59 @@
...
+        "branches",
+        "delete",
+        current,
+        "--project-ref",
+        project.ref,
+        "--yes",
+      ]);
+      if (
+        cleanup.exitCode !== 0 &&
+        !/not found|does not exist/i.test(`${cleanup.stdout}\n${cleanup.stderr}`)
```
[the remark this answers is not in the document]
7ttp: fixed: dc8ae793

apps/cli/src/commands/branches/get/get.live.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  const name = `cli-e2e-get-${randomUUID().slice(0, 8)}`;
+  let targetError: unknown;
+  let cleanupError: unknown;
+  try {
+    const created = await cli(["branches", "create", name, "--project-ref", project.ref]);
+    requireLiveSuccess(created, "branches create");
+
+    const result = await cli(["branches", "get", name, "--project-ref", project.ref]);
+    expect(result.exitCode, result.stderr).toBe(0);
+    expect(/HOST.*STATUS/u.test(result.stdout)).toBe(true);
```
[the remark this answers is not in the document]
7ttp: fixed: dc8ae793

apps/cli/src/commands/branches/get/get.live.test.ts
```diff
@@ -0,0 +1,42 @@
...
+        name,
+        "--project-ref",
+        project.ref,
+        "--yes",
+      ]);
+      if (
+        cleanup.exitCode !== 0 &&
+        !/not found|does not exist/i.test(`${cleanup.stdout}\n${cleanup.stderr}`)
+      ) {
+        cleanupError = new Error(
```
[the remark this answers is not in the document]
7ttp: fixed: dc8ae793

apps/cli/src/commands/branches/get/get.live.test.ts
```diff
@@ -0,0 +1,32 @@
...
+  const name = `cli-e2e-get-${randomUUID().slice(0, 8)}`;
+  let targetError: unknown;
+  let cleanupError: unknown;
+  try {
+    const created = await cli(["branches", "create", name, "--project-ref", project.ref]);
+    requireLiveSuccess(created, "branches create");
+
+    const result = await cli(["branches", "get", name, "--project-ref", project.ref]);
+    expect(result.exitCode, result.stderr).toBe(0);
+    expect(/HOST.*STATUS/u.test(result.stdout), result.stderr).toBe(true);
```
Coly010: [non-blocking] Two things here.

1. `/HOST.*STATUS/u` only proves *a* table header rendered — `renderBranchGetTable` always emits those 7 headers (`branches.format.ts:29`) regardless of which branch the detail belongs to. Exit code 0 already proves the name→ref lookup succeeded, so this line adds structural confirmation but no identity confirmation. Cheap upgrade: create with `--output-format json` (as `update.live.test.ts` does), keep `project_ref`, and assert the rendered row belongs to that branch, e.g. `expect(result.stdout).toContain(branchRef)` (the HOST cell is `detail.db_host`, which carries the branch ref).

2. `expect(regex.test(stdout), stderr).toBe(true)` collapses the failure message to "expected false to be true" plus stderr. `expect(result.stdout, result.stderr).toMatch(/HOST.*STATUS/u)` gives the same guarantee and prints the actual stdout that failed to match — a real difference when debugging a live-only failure you can't reproduce locally.

apps/cli/src/commands/branches/update/update.live.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    expect(updated.stderr).toContain("Updated preview branch");
+    expect(JSON.parse(updated.stdout)).toMatchObject({ name: renamed });
+
+    const proof = await cli(["branches", "get", renamed, "--project-ref", project.ref]);
+    expect(proof.exitCode, proof.stderr).toBe(0);
+  } catch (error) {
+    targetError = error;
+  } finally {
+    try {
+      await removeLiveBranch(cli, project, branchRef ?? name);
```
Coly010, line 77: [non-blocking] The `?? name` fallback has a silent-leak shape: after a successful rename the branch is `renamed`, so if `branchRef` is ever `undefined` this deletes `name`, `removeLiveBranch` tolerates the "not found", and the real branch survives the run with no signal. Today it's unreachable — `project_ref` is a required `Schema.String` on `V1CreateABranchOutput` (`packages/api/src/generated/contracts.ts:997`) and the json output layer emits `{...data, message}` flat (`output.layer.ts:458-459`), so the parse on line 28 always yields a ref. Still worth making loud rather than lossy: type the cast as `{ project_ref: string }` and add `expect(branchRef, created.stdout).toBeTruthy()` right after line 28, so a future payload/envelope change fails the test instead of leaking a billable branch.

apps/cli/src/commands/branches/update/update.live.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    const updated = await cli([
+      "branches",
+      "update",
+      name,
+      "--project-ref",
+      project.ref,
+      "--name",
+      renamed,
+      "--output",
```
Coly010, line 40: [non-blocking] Nit: the same test uses `--output-format json` for create (line 24) and `--output json` here. Both are deliberate and both are supported (legacy-shell invariant 6), and `-o json` is specifically the path where `Updated preview branch:` goes to **stderr** while the payload goes to stdout (`update.handler.ts:103-106`) — which is exactly what lines 41-43 assert. But that's three files of reading to reconstruct. A one-line comment ("`-o json` so the confirmation lands on stderr and stdout stays payload-only") would keep the next reader out of the handler.

apps/cli/src/commands/branches/disable/disable.live.test.ts
```diff
@@ -0,0 +1,40 @@
...
+  try {
+    mayExist = true;
+    const created = await cli(["branches", "create", name, "--project-ref", project.ref]);
+    requireLiveSuccess(created, "branches create");
+
+    const removed = await cli(["branches", "delete", name, "--project-ref", project.ref, "--yes"]);
+    if (removed.exitCode === 0) mayExist = false;
+    requireLiveSuccess(removed, "branches delete");
+
+    const disabled = await cli(["branches", "disable", "--project-ref", project.ref]);
```
Coly010, line 34: [non-blocking] Worth a comment: unlike every sibling in this family, the asserted command here is **project-wide** — `DELETE /v1/projects/{ref}/branches` (see `disable/SIDE_EFFECTS.md`) removes *every* preview branch on the shared live project and leaves branching off for the remainder of the run, i.e. it mutates suite-shared state that `create`/`get`/`list`/`update` all depend on, and it is never restored.

I believe this is safe today: the live project sets `fileParallelism: false` + `maxWorkers: 1` (`vitest.config.ts:101-108`), and each sibling test creates its own branch first — which is also what enables branching (the live global setup never enables it, yet `create` works on a fresh project). So the state is self-healing whichever order Vitest picks. But Vitest's file order is not stable, nothing in this file records that invariant, and the repo's cleanup guidance is "exact owned resources only" — so please state it in a comment, e.g. "`branches disable` is project-wide and deletes all preview branches; safe under the serial live suite because every branches test creates its own branch, which re-enables branching."

apps/cli/src/commands/branches/disable/disable.live.test.ts
```diff
@@ -0,0 +1,40 @@
...
+} from "../../../../tests/helpers/live.ts";
+
+test("disables preview branching", async ({ cli, project }) => {
+  const name = `cli-e2e-disable-${randomUUID().slice(0, 8)}`;
+  let mayExist = false;
+  let targetError: unknown;
+  let cleanupError: unknown;
+  try {
+    mayExist = true;
+    const created = await cli(["branches", "create", name, "--project-ref", project.ref]);
```
Coly010, line 26: [non-blocking] The create-then-delete preamble reads as unexplained setup. The PR description has the rationale ("so branching is enabled with no preview branches left") — put that one line in the file, since it's the precondition that makes the assertion on line 27 meaningful rather than a no-op/4xx.

apps/cli/src/commands/branches/disable/disable.live.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    requireLiveSuccess(created, "branches create");
+
+    const removed = await cli(["branches", "delete", name, "--project-ref", project.ref, "--yes"]);
+    if (removed.exitCode === 0) mayExist = false;
+    requireLiveSuccess(removed, "branches delete");
+
+    const disabled = await cli(["branches", "disable", "--project-ref", project.ref]);
+    expect(disabled.exitCode, disabled.stderr).toBe(0);
+    expect(disabled.stdout).toContain(`Disabled preview branching for project: ${project.ref}`);
```
Coly010, line 36: [non-blocking] This asserts the CLI printed the line on stdout, which does match the documented contract (`SIDE_EFFECTS.md`: "written to **stdout**"), so it's a legitimate golden-path assertion. Optional strengthening, in the spirit of the proof step `update.live.test.ts:45` uses: follow with a `branches list` and assert no preview branches remain, so the test also proves the platform actually applied the disable rather than only that the message rendered.

apps/cli/tests/helpers/live.ts
```diff
@@ -146,6 +146,23 @@ export async function removeStorageLiveObject(
...
+/** Exact cleanup for branches live tests by name or ref; deleting an already-removed branch is tolerated. */
+export async function removeLiveBranch(
+  cli: LiveFixtures["cli"],
+  project: LiveProject,
+  branch: string,
+): Promise<void> {
+  const removed = await cli(["branches", "delete", branch, "--project-ref", project.ref, "--yes"]);
+  if (
+    removed.exitCode !== 0 &&
+    !/not found|does not exist/i.test(`${removed.stdout}\n${removed.stderr}`)
```
Coly010: [non-blocking] The extraction itself is correct and byte-identical to the four inlined copies it replaces, and being purely additive it can't regress existing live callers — good.

Since this is now the single place the tolerance lives, it's the right moment to harden it: the already-deleted case surfaces as `unexpected find branch status 404: <body>` (`branches.resolver.ts` → `mapLegacyHttpError`), so tolerance depends on the platform's 404 *body* happening to contain "not found"/"does not exist". If that body wording changes, cleanup throws and turns a passing test red; conversely a genuinely failed cleanup whose message happens to contain "not found" for an unrelated reason gets swallowed and orphans a branch. Matching the status marker instead (e.g. also allow `/status 404\b/`) makes the intent explicit and decouples it from server copy.

Coly010 (approved):
Test-only PR that adds three live golden paths and correctly factors the duplicated branch-delete cleanup into `removeLiveBranch`. Resource hygiene is sound: every test wraps its target in try/finally, cleanup is idempotent ("not found" tolerated), `throwWithCleanup` preserves both failures, and the shared-helper change is purely additive so no existing live caller can regress. CI is green and I found no blocking issues — 6 non-blocking notes below, mostly about assertion strength on `get` and documenting that `branches disable` mutates project-wide state for the whole live run.

apps/cli/src/commands/branches/get/get.live.test.ts
7ttp: fixed: 4f9fc07f

apps/cli/src/commands/branches/update/update.live.test.ts
7ttp: fixed: 4f9fc07f

apps/cli/src/commands/branches/update/update.live.test.ts
7ttp: fixed: 4f9fc07f

apps/cli/src/commands/branches/disable/disable.live.test.ts
7ttp: fixed: 4f9fc07f

apps/cli/src/commands/branches/disable/disable.live.test.ts
7ttp: fixed: 4f9fc07f

apps/cli/tests/helpers/live.ts
7ttp: fixed: 4f9fc07f

apps/cli/src/commands/branches/disable/disable.live.test.ts
7ttp: left this one as is 

 the branch is already deleted before the disable, so an empty `list` afterwards would only re-prove the delete. and

 `list` on a project with branching off isn't a documented shape (spec only has 200/500), so i'd rather not bet the serial run on it. worth adding once the api documents the disabled-state,
 i'll pick it up then..

PR #6494, opened 2026-09-07
test(cli): cover `config push` and `unlink` (CLI-2293, CLI-2326)

7ttp:
## TL;DR

adds live e2e coverage for `config push` and `unlink`, closing the last gaps in the config and link command families

## whats introduced?

- `config push`: declares a single `api.max_rows` value in a sparse config.toml, pushes it, proves convergence 
`config diff`, then restores the captured value with a re-proven push in the same test
- `unlink`: links the suite project for real, unlinks it, and proves the workspace is left with no local link state

## ref:

- closes: CLI-2293 &  CLI-2326
- passed here: https://github.com/supabase/cli/actions/runs/34119883928/job/101735308268


apps/cli/src/commands/config/push/push.live.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    };
+    expect(payload, pushed.stdout).toEqual(expect.objectContaining({ project_ref: project.ref }));
+    expect(
+      (payload.services ?? []).filter((service) => service.status === "updated"),
+      pushed.stdout,
+    ).toEqual([expect.objectContaining({ service: "api", changes: [["api", "max_rows"]] })]);
+    expect(payload.message, pushed.stdout).toContain(`1 property pushed to ${project.ref}.`);
+
+    const proof = await diffMaxRows("config diff proof");
+    expect(proof.entries, proof.stdout).toEqual([]);
```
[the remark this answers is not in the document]
7ttp: fixed: 38c666670

apps/cli/src/commands/config/push/push.live.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    requireLiveSuccess(result, label);
+    let changes: Array<{ path: string[]; remote?: unknown }>;
+    try {
+      ({ changes } = JSON.parse(result.stdout) as {
+        changes: Array<{ path: string[]; remote?: unknown }>;
+      });
+      changes = changes.filter((change) => change.path.join(".") === "api.max_rows");
+    } catch {
+      throw new Error(`${label}: unexpected config diff payload\n${result.stdout}`);
+    }
```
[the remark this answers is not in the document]
7ttp: fixed: 38c666670

apps/cli/src/commands/unlink/unlink.live.test.ts
```diff
@@ -0,0 +1,24 @@
...
+  workspace,
+}) => {
+  const linked = await cli(["link", "--project-ref", project.ref, "--skip-pooler"]);
+  requireLiveSuccess(linked, "link setup for unlink");
+
+  const result = await cli(["unlink"]);
+  expect(result.exitCode, result.stderr).toBe(0);
+  expect(result.stdout).toContain("Finished supabase unlink.");
+  expect(result.stderr).toContain(`Unlinking project: ${project.ref}`);
+  expect(existsSync(join(workspace.path, "supabase", ".temp")), result.stderr).toBe(false);
```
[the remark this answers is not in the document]
7ttp: fair on the letter of the rule, but the point of this one is the real link → unlink round trip — e2e has no platform creds so it can only seed a fake ref, which never proves a real link's written state actually gets cleaned up. setup reaches the platform, assertions stay on unlink. keeping it live

apps/cli/src/commands/config/push/push.live.test.ts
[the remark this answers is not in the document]
7ttp: the stale read precedents are both propagation-style endpoints
 (postgres-config, ssl posture) this diffs against the same v2 config read push itself writes through, and the adjudication couldn't find staleness for this pair either. it's been green on the live runs so far

so i'd rather not add a poll on a hypothesis.

apps/cli/src/commands/config/push/push.live.test.ts
```diff
@@ -0,0 +1,97 @@
...
+        `${label}: unexpected config diff payload (${String(error)})\nstdout:\n${result.stdout}\nstderr:\n${result.stderr}`,
+      );
+    }
+    return { entries: changes, stdout: result.stdout };
+  };
+
+  // No diff entry means the remote already reads the probe value.
+  await writeConfig(777);
+  const capturedEntry = (await diffMaxRows("config diff capture")).entries[0];
+  const captured = capturedEntry === undefined ? 777 : capturedEntry.remote;
```
Coly010, line 50: [non-blocking] The "no diff entry means the remote already reads the probe value" inference has a second, silent solution: no entry is also what you get when the API never returned the `api` block at all (absent, or present-but-empty on a permission-truncated response — see `config.format.ts:32-39` on `scope.missing`). In that case `captured` becomes 777 rather than the project's real value, the restore push writes 777, and the post-push proof at line 79 passes vacuously too (an unread block never produces an entry).

Cheap guard using a field the payload already carries and the sibling test already asserts on (`diff.live.test.ts` reads `payload.scope.present`): have `diffMaxRows` also destructure `scope` and fail `label` unless `scope.present` includes `"api"`. Then "no entry" can only mean convergence.

Non-blocking because the live project is provisioned fresh per run and deleted at teardown (`tests/helpers/live-project.ts:531-566`), so the blast radius is intra-run only.

apps/cli/src/commands/config/push/push.live.test.ts
```diff
@@ -0,0 +1,97 @@
+import { writeFile } from "node:fs/promises";
+import path from "node:path";
+
+import { expect } from "vitest";
+
+import { requireLiveSuccess, test, throwWithCleanup } from "../../../../tests/helpers/live.ts";
+
+// Worst case (two pushes, three diffs) stays below the live testTimeout.
```
Coly010, line 8: [non-blocking] Residual of the bounding fix in 38c6666 — the arithmetic is correct but tight: 2x90s + 3x30s = 270s against the live project's 300s `testTimeout` (`vitest.config.ts:106`), and that sum excludes the `workspace` fixture's own `supabase init` subprocess and the five node-shim startups, all of which run inside the same test task. If the budget is blown, the test is aborted at the vitest timeout and the `finally` restore (lines 82-95) is exactly what gets cut short, leaving `api.max_rows` at the probe value for the rest of the serial run — the failure mode the explicit bounds were added to prevent.

Suggest trimming to something with visible slack, e.g. 75s pushes / 20s diffs = 210s (20s matches `POLL_ATTEMPT_EXIT_TIMEOUT_MS` in `ssl-enforcement/update.live.test.ts`), and updating the comment to state the numeric worst case so the next reader can re-check it without doing the multiplication.

apps/cli/src/commands/unlink/unlink.live.test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+// Golden path only: a real `link`'s workspace state must round-trip through
+// unlink, leaving no local link state. Error paths live in unlink.e2e.test.ts
+// and unlink.integration.test.ts.
+test("unlinks a linked workspace, leaving no local link state", async ({
+  cli,
+  project,
+  workspace,
+}) => {
+  const linked = await cli(["link", "--project-ref", project.ref, "--skip-pooler"]);
```
Coly010, line 16: [non-blocking] Taking your reply to the earlier scope finding at face value (the value here is the real link -> unlink round trip, which e2e cannot seed): nothing in the test actually observes that real state. Lines 20-23 assert exactly what `unlink.e2e.test.ts:14-36` already asserts from a hand-seeded `project-ref`, so if `link`'s write set ever shrank to just `project-ref`, this test would stay green and would silently become a duplicate of the e2e test — the stated justification would evaporate with no signal.

One line before the unlink call makes the distinction real and self-documenting, e.g. assert that `.temp` contains `linked-project.json` (the file `link.live.test.ts` already asserts on) or that `readdirSync(.temp)` has more than one entry. Then the absence assertion at line 23 demonstrably covers state only a real link can produce.

Worth noting in the same breath: the harness sets `SUPABASE_NO_KEYRING=1` (`tests/helpers/cli.ts:330`), so unlink's other side effect — deleting the db-password keyring entry — is not exercised here either, which leaves the `.temp` contents as the only thing this test can uniquely prove.

apps/cli/tests/helpers/cli.ts
```diff
@@ -430,7 +434,7 @@ export function spawnSupabase(
     });
 
     disposeOwnHome();
-    return result;
+    return timedOut ? { ...result, timedOutAfterMs: timeoutMs } : result;
```
Coly010, line 437: [non-blocking] Nice diagnostic, but it is only consulted in the `exitCode !== 0` branch of `requireCliSuccess`, and `runSupabase` forces `exitCode: killedByUntil ? 0 : result.exitCode` (line 556). So a run that matched its `until` pattern and then ignored SIGTERM until the exit bound fired reports plain success and swallows the new signal — the one case where "the harness had to SIGKILL it" is most interesting is the one case it cannot be reported. Consider treating a set `timedOutAfterMs` as a failure in `requireCliSuccess` regardless of the reported exit code (or not zeroing the exit code when `timedOut` is set). No current caller combines `until` with a tight `exitTimeoutMs`, hence non-blocking.

apps/cli/tests/helpers/live.unit.test.ts
```diff
@@ -1,6 +1,24 @@
 import { describe, expect, it } from "vitest";
 
-import { throwWithCleanup } from "./live.ts";
+import { requireLiveSuccess, throwWithCleanup } from "./live.ts";
+
+describe("requireLiveSuccess", () => {
```
Coly010: [non-blocking] Placement nit: the behavior under test (including the new SIGKILL branch) now lives in `tests/helpers/cli.ts:559-577`, but the coverage lives in `live.unit.test.ts` and reaches it through the `requireLiveSuccess` alias re-exported at `live.ts:116`. There is no `tests/helpers/cli.unit.test.ts`, so grepping for `requireCliSuccess` tests finds nothing and a future edit to the cli helper has no obviously-owning test file. Either add `tests/helpers/cli.unit.test.ts` for these three cases or import `requireCliSuccess` directly here, so the `describe` names the symbol it actually covers.

Coly010 (approved):
Test-coverage-only PR; CI is green, `apps/cli` typecheck passes locally, and the 6 tests in `tests/helpers/live.unit.test.ts` pass. The `config push` live test is the strong part of this change — it follows the established `ssl-enforcement/update.live.test.ts` capture/mutate/prove/restore shape, asserts real payload content (`services[api].changes`, the `1 property pushed to <ref>.` message) rather than exit codes, and proves convergence through an independent `config diff` in both directions. No blocking findings: no secrets/refs, no `.only`/`.skip`, the shared-helper change is additive (optional `timedOutAfterMs` + a `requireCliSuccess` alias, all 25 `requireLiveSuccess` call sites keep identical behavior), and remote mutation is restored in a `finally` against a per-run disposable project. Five non-blocking notes below, mostly about making two inferences explicit so the tests cannot pass vacuously.

apps/cli/src/commands/config/push/push.live.test.ts
7ttp: good catch — the old sum actually went over budget once you count the fixture's init (60 + 270 = 330 vs 300), so this was a real hole, not just slack. trimmed to 75s/20s. fixed in: https://github.com/supabase/cli/commit/05caa45f4

apps/cli/src/commands/unlink/unlink.live.test.ts
7ttp: fair — added the linked-project.json existence check before the unlink so the real-write-set claim is pinned, not asserted. fixed in: https://github.com/supabase/cli/commit/05caa45f4

apps/cli/tests/helpers/live.unit.test.ts
7ttp: went with your second option — imported requireCliSuccess directly so the describe names the symbol it covers, without adding a file. fixed in: https://github.com/supabase/cli/commit/05caa45f4

apps/cli/src/commands/config/push/push.live.test.ts
7ttp: traced this one — a declared path whose block the response doesn't carry still produces an entry (class local_only, remote null; diff's SIDE_EFFECTS documents the partial-response rule), so the capture guard throws on it rather than falling through to 777. the no-entry case really is only remote == 777. leaving as is

apps/cli/tests/helpers/cli.ts
7ttp: agreed it's a real corner, but with no caller combining until with a tight bound i'd rather not pick semantics for it speculatively — both options change until behavior in ways someone will eventually depend on. happy to revisit when a caller actually exists

7ttp: agreed it's a real corner, but with no caller combining until with a tight 
bound i'd rather not pick semantics for it speculatively both options change until behavior in ways someone will eventually depend on. will revisit when a caller actually exists

PR #6506, opened 2026-09-07
feat(cli): add experimental stack start

jgoux:
`supabase experimental stack start` creates or resumes a managed stack for the current project and branch, with named-stack and explicit-ID targeting, Docker/native selection, and eager activation or background artifact preparation.

Translate project configuration at the CLI boundary and embed the supervisor and native-process dispatch in the compiled binary. The stack package owns lifecycle and persistent state, and resolves PostgreSQL major versions through its release catalog.


packages/stack/src/model/Compiler.ts
```diff
@@ -447,7 +447,23 @@ const releaseFor = <T>(
...
+  const majorSelector = module.name === "database" && /^\d+$/.test(selector);
+  const resolvedSelector = majorSelector
+    ? (() => {
+        const sameMajor = (version: string | undefined): boolean =>
+          version !== undefined && version.split(".", 1)[0] === selector;
+        if (sameMajor(previousVersion) && previousVersion !== undefined) return previousVersion;
+        if (sameMajor(module.defaultVersion)) return module.defaultVersion;
+        return Object.keys(module.releases).find(
+          (version) => version.includes(".") && sameMajor(version),
+        );
```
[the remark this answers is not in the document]
jgoux: Keeping the current release selection policy. The concrete catalog has one supported PostgreSQL release per major (17.6.1.168 and 15.14.1.168), so the fallback is unambiguous. The resolver deliberately preserves an existing release first, then prefers the catalog default. A newest-release policy is not part of this change and could change that behavior; a generic version comparator is unnecessary for the current catalog.

apps/cli/src/commands/experimental/stack/start/start.handler.ts
```diff
@@ -0,0 +1,228 @@
...
+            "StackLifecycleConflictError",
+            "StackUpgradeRequiredError",
+            "StackRuntimeError",
+            "StackCleanupError",
+            () => ({
+              reason: "lifecycle" as const,
+              suggestion: "Stop the stack before starting it again.",
+            }),
+          ),
+          Match.orElse(() => ({ reason: "unknown" as const })),
```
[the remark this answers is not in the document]
jgoux: Addressed in `3055b18d4`. Added explicit handling for invalid and unsupported persisted state, with state-specific recovery guidance. These no longer fall through to unknown; handler regressions cover the failures.

apps/cli/src/commands/experimental/stack/start/start.handler.ts
```diff
@@ -0,0 +1,228 @@
...
+            "StackNotRunningError",
+            "StackMustBeStoppedError",
+            "StackLifecycleConflictError",
+            "StackUpgradeRequiredError",
+            "StackRuntimeError",
+            "StackCleanupError",
+            () => ({
+              reason: "lifecycle" as const,
+              suggestion: "Stop the stack before starting it again.",
+            }),
```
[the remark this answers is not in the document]
jgoux: Addressed in `3055b18d4`. Separated runtime and cleanup failures from lifecycle conflicts and replaced the unconditional stop-first advice with operational diagnostics and --debug guidance. Their underlying causes are opaque, so they are no longer reported as invalid configuration; the current closed telemetry vocabulary cannot safely label them Docker, network, or internal-bug failures.

apps/cli/src/commands/experimental/stack/stack-config.ts
```diff
@@ -0,0 +1,731 @@
...
+} from "../../../shared/telemetry/error-actionability.ts";
+
+/** A config error suitable for an experimental stack command's user-facing boundary. */
+class LegacyStackConfigError extends Data.TaggedError("LegacyStackConfigError")<{
+  readonly message: string;
+}> {
+  get [ErrorActionabilityId](): CliErrorActionabilityDeclaration {
+    return actionability.invalidConfig;
+  }
+}
```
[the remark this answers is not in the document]
jgoux: Addressed in `3055b18d4`. Exported both error classes while retaining their own actionability declarations.

apps/cli/src/commands/experimental/stack/stack-config.ts
```diff
@@ -0,0 +1,731 @@
...
+        );
+    const shared = yield* read(path.join(root, ".env"));
+    const entries = yield* fs.readDirectory(root).pipe(Effect.orElseSucceed(() => []));
+    const result: Record<string, Readonly<Record<string, Redacted.Redacted<string>>>> = {};
+    for (const entry of entries.filter((name) => !name.startsWith(".") && !name.startsWith("_"))) {
+      if (!/^[A-Za-z0-9_-]+$/u.test(entry)) continue;
+      const info = yield* fs.stat(path.join(root, entry)).pipe(Effect.option);
+      if (Option.isNone(info) || info.value.type !== "Directory") continue;
+      const functionEnv = yield* read(path.join(root, entry, ".env"));
+      result[entry] = { ...shared, ...functionEnv };
```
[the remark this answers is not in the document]
jgoux: Addressed in `3055b18d4`. Added dotenv-key validation that identifies the source file and offending key without exposing its value. Disabled runtime/functions skip unused dotenv input. Integration regressions cover the diagnostics and disabled cases.

apps/cli/src/commands/experimental/stack/stack-config.ts
```diff
@@ -0,0 +1,731 @@
...
+const legacyFunctionRelativePath = (value: string, name: string): string => {
+  if (value.length === 0 || value.startsWith("/")) return value;
+  const normalized = value.replace(/^\.\//, "").replace(/^supabase\//, "");
+  const target = normalized.split("/").filter((part) => part.length > 0);
+  const base = ["functions", name];
+  let common = 0;
+  while (common < base.length && common < target.length && base[common] === target[common])
+    common += 1;
+  return [...base.slice(common).map(() => ".."), ...target.slice(common)].join("/");
+};
```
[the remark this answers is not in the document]
jgoux: Addressed in `3055b18d4`. Added normalized functions-root containment validation before passing paths to the package. Unsupported paths now produce a targeted function/field error instead of silently making the function unavailable; normalized paths that stay inside the root remain valid.

apps/cli/src/commands/experimental/stack/stack-config.ts
```diff
@@ -0,0 +1,731 @@
...
+  "slack_oidc",
+  "spotify",
+  "workos",
+  "zoom",
+] as const;
+
+const legacyStackProjectPath = (value: string): string =>
+  value.length === 0 || value.startsWith("/") || value.startsWith("supabase/")
+    ? value
+    : `supabase/${value.startsWith("./") ? value.slice(2) : value}`;
```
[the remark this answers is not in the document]
jgoux: Addressed in `3055b18d4`. Removed the special treatment of a leading supabase/ component. Relative values resolve beneath the existing project supabase directory, including signing material and function paths. Regression cases cover the nested-prefix behavior.

apps/cli/src/commands/experimental/stack/stack-config.ts
```diff
@@ -0,0 +1,731 @@
...
+    functions,
+  };
+};
+
+const legacyConfigInput = (
+  projectRoot: string,
+  config: CliConfig,
+  document?: Record<string, unknown>,
+  projectEnvValues: Readonly<Record<string, string>> = {},
+) => {
```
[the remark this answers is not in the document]
jgoux: Addressed in `3055b18d4`. The parameter now has a concrete use: normalized function-path translation and containment checks resolve against the selected project root. The old unused argument was removed before those path checks made it necessary.

apps/cli/src/commands/experimental/stack/stack-config.ts
```diff
@@ -0,0 +1,731 @@
...
+  const storage = config.storage;
+  const realtime = config.realtime;
+  const studio = config.studio;
+  const analytics = config.analytics;
+  const mail = config.local_smtp;
+  const pooler = db.pooler;
+  const capability = (enabled: boolean, settings: unknown) =>
+    enabled ? { settings } : { enabled: false as const };
+  return {
+    preparation: "background",
```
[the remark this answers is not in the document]
jgoux: Addressed in `3055b18d4`. Removed the redundant preparation value from the translated config. Start supplies its command policy; other consumers use the package default.

packages/stack/src/supervisor/ingress.integration.test.ts
```diff
@@ -53,6 +53,9 @@ const listenBackend = (server: ReturnType<typeof createHttpServer>) =>
     () => closeServer(server),
   ).pipe(Effect.as(server));
 
+const uniquePort = (stackId: string, offset: number): number =>
+  42_000 + (Number.parseInt(stackId.slice(0, 6), 16) % 7_000) + offset;
```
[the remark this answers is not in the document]
jgoux: Addressed in `3055b18d4`. Replaced the new hash-derived live listener ports with scoped, owned listeners and removed unnecessary public listeners from lifecycle-only tests. No reserve-close-reuse handoff is used; the tests retain ownership through adoption.

apps/cli/src/commands/experimental/stack/stack.shared.ts
```diff
@@ -0,0 +1,132 @@
...
+export const legacyExperimentalStackApiLayer = Layer.succeed(LegacyExperimentalStackApi, {
+  createStack,
+  openStack,
+  inspectStack,
+});
+
+/** Runtime configuration for the first stack command. Later commands reuse this layer. */
+export const legacyExperimentalStackTargetResolverLayer = Layer.effect(
+  LegacyExperimentalStackTargetResolver,
+  Effect.succeed({
```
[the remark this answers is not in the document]
jgoux: Addressed in `3055b18d4`. Changed the constant target-resolver service to Layer.succeed.

apps/cli/src/commands/experimental/stack/stack-config.ts
```diff
@@ -0,0 +1,810 @@
...
+        analytics: storage.analytics,
+        vector: storage.vector,
+      }),
+      functions: capability(
+        config.edge_runtime.enabled,
+        legacyFunctionsSettings(projectRoot, path, config, document, projectEnvValues),
+      ),
+      studio: capability(studio.enabled, {
+        api_url: studio.api_url,
+        openai_api_key: secret(studio.openai_api_key),
```
avallete, line 630: Default `studio.api_url` is `http://127.0.0.1` (no port). Classic start rewrites that with `legacyResolveStudioApiUrl` when the host is empty or equals the local hostname. The stack uses `studio.api_url || apiListenerUrl` for Studio `SUPABASE_URL`, so the default wins and Studio talks to port 80 while `SUPABASE_PUBLIC_URL` uses the allocated API port.

Omit host-only/empty URLs (or reuse `legacyResolveStudioApiUrl`) so auto-allocated ports stay correct. `supabase init` then `experimental stack start --eager` hits this on native.

apps/cli/src/commands/experimental/stack/stack-config.ts
```diff
@@ -0,0 +1,810 @@
...
+    const unwrapped = Redacted.value(value);
+    return typeof unwrapped === "string" ? Redacted.make(unwrapped) : undefined;
+  }
+  return typeof value === "string" ? Redacted.make(value) : undefined;
+};
+
+const parseEnv = (contents: string): Record<string, Redacted.Redacted<string>> =>
+  Object.fromEntries(
+    Object.entries(parseDotenv(contents)).map(([key, value]) => [key, Redacted.make(value)]),
+  );
```
avallete, line 53: `functions serve` reads these same files with `parseDotEnv` (`command-internal/legacy-dotenv.ts`), which expands `$VAR`/`${VAR}` and accepts `KEY: value`. npm `dotenv` does not. Reuse `parseDotEnv`, then wrap values in `Redacted`.

apps/cli/src/commands/experimental/stack/stack-config.ts
```diff
@@ -0,0 +1,810 @@
...
+};
+
+const secret = (value: unknown): Redacted.Redacted<string> | undefined => {
+  if (Redacted.isRedacted(value)) {
+    const unwrapped = Redacted.value(value);
+    return typeof unwrapped === "string" ? Redacted.make(unwrapped) : undefined;
+  }
+  return typeof value === "string" ? Redacted.make(value) : undefined;
+};
```
avallete, line 48: `loadCliConfig` does not decrypt dotenvx `encrypted:` values — that lives in `legacyDecryptSecret` / `legacy-local-config-values`. Wrapping the string as `Redacted` will pass ciphertext into the stack. Decrypt with the existing helper and fail `LegacyStackConfigError` if decryption fails.

apps/cli/src/commands/experimental/stack/stack-config.ts
```diff
@@ -0,0 +1,810 @@
...
+  "slack_oidc",
+  "spotify",
+  "workos",
+  "zoom",
+] as const;
+
+const legacyStackProjectPath = (value: string): string =>
+  value.length === 0 || value.startsWith("/")
+    ? value
+    : `supabase/${value.startsWith("./") ? value.slice(2) : value}`;
```
avallete: Absolute paths are detected with `startsWith("/")`, so a Windows `C:\...` `auth.signing_keys_path` is prefixed with `supabase/`. Use `Path.isAbsolute`.

apps/cli/src/commands/experimental/stack/stack-config.ts
```diff
@@ -0,0 +1,810 @@
...
+    }
+  }
+  return undefined;
+};
+
+/** Loads and translates the effective project config for all experimental stack commands. */
+export const legacyLoadStackConfig = (projectRoot: string): LegacyStackConfigEffect =>
+  Effect.gen(function* () {
+    const path = yield* Path.Path;
+    return yield* legacyLoadLocalProjectContext(
```
avallete: This uses decoded `CliConfig` and never the AutomaticEnv overlays in `legacy-local-config-values`. Overrides such as `SUPABASE_AUTH_ENABLED=false` that `supabase start` honors are ignored. Reuse or factor that overlay, or document that experimental start is file-only.

apps/cli/src/commands/experimental/stack/start/start.handler.ts
```diff
@@ -0,0 +1,241 @@
...
+    ? Effect.fail(
+        new LegacyExperimentalStackTargetFlagsError({
+          message: "--stack and --stack-id cannot be used together",
+        }),
+      )
+    : Effect.void;
+
+export const legacyExperimentalStackStart = Effect.fn("legacy.experimental.stack.start")(function* (
+  flags: LegacyExperimentalStackStartFlags,
+) {
```
avallete: `apps/cli/AGENTS.md` requires `Effect.ensuring(telemetryState.flush)` on every handler (`start`, `db start`, and experimental workers all do this). Without it, `~/.supabase/telemetry.json` / session rotation never update. Same `ensuring` as `db start` is enough; skip linked-project cache if this command has no project ref.

apps/cli/src/commands/experimental/stack/stack-config.ts
```diff
@@ -0,0 +1,810 @@
...
+          : legacyReadFunctionEnvironments(
+              projectRoot,
+              new Set(
+                Object.entries(context.config.functions)
+                  .filter(([, functionConfig]) => functionConfig.enabled === false)
+                  .map(([name]) => name),
+              ),
+              context.config.edge_runtime.enabled === false,
+            ).pipe(
+              Effect.mapError((cause) => new LegacyStackConfigError({ message: String(cause) })),
```
avallete: `mapError` with `String(cause)` re-wraps an already-constructed `LegacyStackConfigError` (e.g. invalid dotenv keys), so the user can see `LegacyStackConfigError: LegacyStackConfigError: …`. Map platform errors only and pass this class through.

apps/cli/src/commands/experimental/stack/start/start.integration.test.ts
```diff
@@ -0,0 +1,587 @@
...
+    legacyValidateExperimentalStackStartTarget({
+      stack: Option.some("feature-a"),
+      stackId: Option.some("a".repeat(64)),
+    }).pipe(
+      Effect.flip,
+      Effect.tap((failure) =>
+        Effect.sync(() => expect(failure.message).toContain("cannot be used together")),
+      ),
+    ),
+  );
```
avallete: This isolated check only asserts the message. The handler scenario at lines 491–530 already covers the same copy and that resolve/create do not run. Safe to drop.

avallete (approved):
Approving. Dispatch, targeting, and error classification look good for an experimental first command.

Inline comments are follow-ups on config translation (Studio default URL, functions `.env` parsing, `encrypted:` secrets, `SUPABASE_*` overlays, Windows signing-key paths, telemetry flush). None of those should block landing this.

On the Depthfirst thread: `db.network_restrictions` / `ssl_enforcement` / `vault` are accepted by the stack schema but nothing in the stack runtime reads them today, same as local `supabase start`. Optional to forward for definition fidelity; not a current local access-control hole.

apps/cli/src/commands/experimental/stack/stack-config.ts
avallete: This might be on purpose though, just need documenting / error handling.

apps/cli/src/commands/experimental/stack/stack-config.ts
avallete: This might be on purpose though, just need documenting / error handling.

apps/cli/src/commands/experimental/stack/stack-config.ts
```diff
@@ -0,0 +1,810 @@
...
+  const analytics = config.analytics;
+  const mail = config.local_smtp;
+  const pooler = db.pooler;
+  const capability = (enabled: boolean, settings: unknown) =>
+    enabled ? { settings } : { enabled: false as const };
+  return {
+    capabilities: {
+      database: {
+        version: String(db.major_version),
+        settings: { health_timeout: db.health_timeout, settings: db.settings },
```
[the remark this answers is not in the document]
jgoux: The access-control regression claim does not hold for the current runtime. These fields occur only in the database schema/defaults (and vault serialization coverage); no workload/runtime code consumes them. Forwarding them would not enforce CIDRs or TLS. Keeping the translation scoped to supported runtime settings and documenting this limitation in the command's SIDE_EFFECTS.md, including vault, as discussed in the human review.

Handled in d81788e54.

apps/cli/src/commands/experimental/stack/stack-config.ts
jgoux: Fixed: the default host-only Studio API URL is omitted at translation, allowing the stack runtime to use the allocated API listener port. Explicit custom URLs remain unchanged. Added a loader regression for the default config.

Handled in d81788e54.

apps/cli/src/commands/experimental/stack/stack-config.ts
jgoux: Fixed: function dotenv files now use the shared parseDotEnv parser. Regression coverage exercises variable expansion and colon assignments. Parser failures become typed, file-specific configuration errors without exposing dotenv contents.

Handled in d81788e54.

apps/cli/src/commands/experimental/stack/stack-config.ts
jgoux: Taking the explicit unsupported-contract option from your follow-up: translated encrypted: secrets now fail with a targeted configuration error before reaching the stack, including function dotenv values. The diagnostic identifies the field without exposing ciphertext. Documented that experimental stack start does not perform dotenvx decryption.

Handled in d81788e54.

apps/cli/src/commands/experimental/stack/stack-config.ts
jgoux: Fixed: signing-key path translation now uses the injected Path.isAbsolute API. Added a Windows-path regression using the real win32 path implementation.

Handled in d81788e54.

apps/cli/src/commands/experimental/stack/stack-config.ts
jgoux: Taking the documented file-based option from your follow-up. Command help and SIDE_EFFECTS.md now state that config fields come from supabase/config.toml and explicit env(NAME) references; automatic SUPABASE_* section overrides are not applied. Runtime settings such as SUPABASE_HOME continue to work.

Handled in d81788e54.

apps/cli/src/commands/experimental/stack/start/start.handler.ts
jgoux: Fixed: the handler now ensures telemetryState.flush on success, runtime failure, and early flag errors. Added the telemetry layer to the production stack command composition. Integration assertions cover those outcomes, and a compiled CLI smoke check verifies telemetry.json is written on a flag error.

Handled in d81788e54.

apps/cli/src/commands/experimental/stack/stack-config.ts
jgoux: Fixed: existing LegacyStackConfigError instances pass through unchanged; only other filesystem failures are mapped. Malformed dotenv errors also omit source contents, with regression coverage for the clean diagnostic.

Handled in d81788e54.

apps/cli/src/commands/experimental/stack/start/start.integration.test.ts
jgoux: Removed the redundant isolated helper test and made the now-internal helper private. The handler scenario retains the message assertion and verifies that neither resolution nor creation runs; it now also checks telemetry flushing.

Handled in d81788e54.

Coly010:
I merged `develop` into this branch to resolve the conflicts caused by #6525 (the legacy-prefix rename, which touched ~1400 files). No functional changes to this PR's own work — the merge was mechanical: renamed identifiers/imports to match the post-rename names (e.g. `legacyExperimentalStackCommand` → `experimentalStackCommand`, `LegacyCliSettings` → `CommandSettings`, `legacy-cli-settings.layer.ts` → `command-settings.layer.ts`, etc.), and dropped the removed `entrypoint: "legacy"` e2e test option.

While resolving this, a new commit (d81788e5, "address experimental stack start review feedback") landed on this branch mid-resolution, so I merged that in too rather than force-pushing over it — its content is unchanged, just the same rename applied consistently (e.g. `findEncryptedSecret`, the `Path`-aware `stackProjectPath` fix, and the `TelemetryState`/`Effect.ensuring(telemetryState.flush)` wiring in `stack.command.ts`/`start.handler.ts` are all preserved).

Verified: `pnpm check:all` passes (the only failure is the pre-existing `@supabase/cli-go#lint:check` gosec warnings, unrelated to this branch — no Go files touched), and the full `apps/cli` unit + integration suite passes (9204 tests). Please double-check the diff when you get a chance.

PR #6516, opened 2026-09-08
feat(cli): select local stack commands by project backend

jgoux:
Expose the new local runtime as `supabase stack start|stop|status|list|logs|prepare|restart`, replacing `supabase experimental stack`.

Projects can set `[experimental] stack = true` in `supabase/config.toml` to make top-level `start`, `stop`, and `status` use the corresponding new commands. Backend selection happens before argument parsing and completion, so each implementation retains its own flags. Explicit `supabase stack` commands always use the new backend. `SUPABASE_EXPERIMENTAL_STACK=1` or `0` overrides the config setting for top-level aliases; unset or empty values fall back to config, and other values are rejected. Without an environment override, absent or false configuration preserves the legacy aliases.

The backends retain separate state and databases. Switching the flag does not migrate or reuse legacy data. The flag is local CLI configuration and is excluded from hosted project configuration. Other local-stack command families remain unchanged.

Routing follows the same working-directory and TOML-file selection as the lifecycle handlers, including both shell completion modes. Command telemetry retains the invoked command path and a unique run identifier.

New routing code stays within the Effect-linted stack command directory. This follows the command implementations through #6512.

### DIFF
--- a/apps/cli/docs/stack-commands.md
+++ b/apps/cli/docs/stack-commands.md
@@ -23,16 +23,6 @@ The top-level `supabase start`, `supabase stop`, and `supabase status` commands 
 stack = true
 ```
 
-The equivalent in `supabase/config.json` is:
-
-```json
-{
-  "experimental": {
-    "stack": true
-  }
-}
-```
-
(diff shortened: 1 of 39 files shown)


jgoux:
The feature flag is on `develop` so every new commands should already rely on it.

PR #6517, opened 2026-09-08
feat(cli): add stack command options and destruction

jgoux:
Add service selection, bulk stopping, environment export, and explicit destruction to the new local stack commands. These options also apply to the config-selected top-level start, stop, and status aliases.

- `start --exclude` accepts service names, comma-separated or repeated. It leaves the project file unchanged and records the effective start configuration in stack state.
- `stop --all` retains data and attempts every readable registered stack. Unreadable entries are skipped with warnings; skipped entries and stop failures produce a nonzero result after the remaining stacks are processed.
- `status --env` exports connection variables as dotenv or a JSON variable map, with `--override-name` for application-specific names. Ordinary status does not reveal credentials. Database-only stacks can retrieve database credentials with API credentials omitted when Auth is disabled.
- `stack destroy` permanently removes one selected stack after confirmation, with `--yes` for unattended execution.

This follows #6516. The legacy implementations and database data remain separate.

### DIFF
--- a/apps/cli/docs/stack-commands.md
+++ b/apps/cli/docs/stack-commands.md
@@ -47,7 +47,9 @@ project’s configured services. `--eager` waits for enabled services to become 
 
 `supabase stack stop --all` stops every stack in the new backend’s registry and
 preserves data. It cannot be combined with `--stack` or `--stack-id`. Failures
-are reported after attempting the other stacks.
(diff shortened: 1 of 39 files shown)


jgoux:
Superseeded by https://github.com/supabase/cli/pull/6576

PR #6521, opened 2026-09-08
ci(repo): unbreak dependabot Dockerfile bumps, move stack pins onto the slim-services feed

avallete:
## Summary

Dependabot's docker updates on `apps/cli-go/pkg/config/templates/Dockerfile` have been landing red — see #6502 (realtime) and #6503 (postgres). Two independent causes, plus the automation the second one leaves behind.

**1. A unit test asserted the exact current pins while reading them from the live manifest.**

`slim-images.unit.test.ts` had a `maps current docker.io pins onto the published slim tags` case that fed `dockerfileServiceImageRaw(alias)` in and asserted a spelled-out version for `pg`, `supavisor`, `realtime`, and `storage`. Any bump of those four failed the unit suite by construction.

The assertions that must track the manifest — which slim repository each alias maps to — already live in the `it.each` above and slice the tag off before comparing, so they stay. Live-tag validation also survives more strongly elsewhere: `start.slim-images.e2e.test.ts` translates the current manifest pins and actually pulls them, so an unpublished translated tag fails CI on the registry rather than on a hand-typed string. The version-bearing block only encoded the per-service tag-prefix scheme, so it is replaced with fixed pins covering the same scheme, plus the uppercase-`V` normalization arm in `slimTagForService` that had no coverage.

**2. `sync-stack-service-versions.yml` targeted files that no longer exist.**

It ran `pnpm sync:versions` in `packages/stack` and committed `packages/stack/src/ServiceCatalog.ts`. #6440 deleted both the script and that file, so the workflow would have hard-failed on the next dependabot Dockerfile PR. It survived #6502/#6503 only because those were opened about two hours before #6440 merged.

**3. The stack's pins now ride the feed that can actually maintain them.**

Deleting that workflow leaves `WorkloadCatalog.ts` — `ServiceCatalog.ts`'s replacement — with no automation, so this adds it. The catalog pins each workload to an exact slim-services artifact release: a version **plus its `ghcr.io/supabase/cli` image digest**, per ADR 0017, which makes the artifact release the boundary for service startup defaults. Dependabot owns the Dockerfile and structurally cannot own this table — it resolves registry tags and never produces a `sha256:` digest, which is why repointing the old workflow was not an option.

slim-services already sends this repo a `mirror-slim-image` `repository_dispatch` per release carrying `service`/`version`/`digest`, to drive the ECR mirror. `sync-stack-workload-catalog.yml` subscribes to that same dispatch and opens a PR pinning the release. It is deliberately a separate workflow from `mirror-slim-image.yml`: that mirror runs against the sender's 15-minute verification poll, and a catalog PR must never delay it or turn its run red.

## Reviewer notes

Only two source values change per release. `artifactFor` derives `releaseTag`, `assetName`, and every download URL from `service` + `version`, and `releases` is derived from `defaultVersion` plus the container image, so rewriting the `native(...)` positional version and image is the whole change.

**postgres is the one service carrying two supported release lines** (17.x and 15.x via `additionalReleases`), so the plan picks its target by release line — a 15.x release moves the additional entry and can never overwrite the 17.x default. That is the main correctness risk here and it has a dedicated test. A release on a line the catalog does not carry, a service it does not model, and a re-dispatch of an already-pinned release are all successful no-ops, so the sender's retry path does not open duplicate PRs.

The dispatch payload arrives with whatever authority holds the dispatch token, so `sync-workload-catalog.ts` revalidates `service`/`version`/`digest` against the same patterns `mirror-slim-image.yml` uses rather than trusting the workflow. Those patterns are what stop a version or digest breaking out of the TypeScript string literals it writes into.

`planCatalogUpdate` is pure and covered by 21 `bun:test` cases in `.github/scripts/`, which `github-scripts-ci.yml` already tests and type-checks. One of them sweeps every service the real catalog models, so a newly modelled workload is covered without editing a fixture list.

`WorkloadCatalog.ts` also gains a provenance comment: it and the Dockerfile list overlapping service versions and are meant to diverge, so the next reader needs to know not to "reconcile" them.

## Linked issue

No Linear ticket — repository maintenance prompted by the two failing dependabot PRs above.

https://claude.ai/code/session_01PWXmTVTiQCwuyjZ8Mfw9NL


avallete:
/ai-review

Re-requesting: the earlier review ran on `524cba4` and only saw the test change and the workflow deletion. `b170805` added the replacement automation — `.github/scripts/sync-workload-catalog.ts`, its suite, and `sync-stack-workload-catalog.yml` — which is the substantially riskier half of this PR and has not been reviewed.

Worth particular attention:

- **Release-line targeting.** postgres carries two supported lines (17.x default, 15.x in `additionalReleases`). A 15.x release must move the additional entry and must never overwrite the 17.x default. `planCatalogUpdate` decides this from the leading numeric component.
- **Untrusted payload.** The `mirror-slim-image` dispatch arrives with whatever authority holds the dispatch token. `version` and `digest` are written into TypeScript string literals, so the validation patterns are the thing standing between a hostile payload and arbitrary source injection into `WorkloadCatalog.ts`.
- **Regex source rewriting.** The `native(...)` positional args and `additionalReleases` entries are matched textually. `postgres`/`postgrest` prefix collision is the obvious trap.

---

.github/scripts/sync-workload-catalog.ts
```diff
@@ -0,0 +1,248 @@
...
+ * with SLIM_SERVICE / SLIM_VERSION / SLIM_DIGEST set from the payload.
+ *
+ * Exit codes: 0 the sync ran (whether or not it changed anything), 1 invalid
+ * payload or tool failure. A service the catalog does not model, and a release
+ * line it does not carry, are both successful no-ops — slim-services publishes
+ * for consumers beyond this catalog.
+ *
+ * `planCatalogUpdate` is pure and unit-tested in `sync-workload-catalog.test.ts`;
+ * `main()` wires up the real filesystem.
+ */
```
avallete, line 14: Way too long comment

.github/workflows/sync-stack-workload-catalog.yml
```diff
@@ -0,0 +1,138 @@
...
+# This is deliberately a separate workflow from `mirror-slim-image.yml` rather
+# than another job in it: that mirror runs against the sender's 15-minute
+# verification poll, and a catalog PR failing must never delay it or turn its
+# run red. Both workflows subscribe to the same dispatch type and run
+# independently.
+#
+# Opens a PR rather than pushing: a service version pin is a reviewable change
+# and `develop` is protected. Re-dispatches are no-ops (the script exits
+# without touching the file when the pin already matches), so the sender's
+# retry path does not produce duplicate PRs.
```
avallete: Same too long

apps/cli/src/shared/services/slim-images.unit.test.ts
```diff
@@ -43,19 +43,21 @@ describe("toSlimImage", () => {
...
-    );
-    expect(toSlimImage("supavisor", dockerfileServiceImageRaw("supavisor"))).toBe(
-      "ghcr.io/supabase/cli/pooler:v2.9.12",
-    );
-    expect(toSlimImage("realtime", dockerfileServiceImageRaw("realtime"))).toBe(
+  // Tag-scheme assertions use fixed pins rather than the Dockerfile manifest:
+  // dependabot bumps that manifest, and an expectation spelling out the current
+  // pin would turn every bump into a failing test. The `it.each` above keeps the
+  // live manifest covered for the part that must track it — the repository each
+  // alias maps to.
```
avallete: Be terse

packages/stack/src/model/WorkloadCatalog.ts
```diff
@@ -52,7 +52,21 @@ const native = (
...
+ * release the boundary for service startup defaults, so a version here carries
+ * its `ghcr.io/supabase/cli` image digest and moves independently of the
+ * Dockerfile pin for the same service. The two deliberately diverge; do not
+ * "reconcile" them.
+ *
+ * Maintained by `.github/workflows/sync-stack-workload-catalog.yml`, off the
+ * `mirror-slim-image` dispatch slim-services sends on each release. Dependabot
+ * owns the Dockerfile and cannot own this table: it resolves registry tags and
+ * never produces a `sha256:` digest.
+ */
```
avallete, line 61: be terse

.github/scripts/sync-workload-catalog.ts
avallete: Cut 43 lines to 13 in `3953876`. Kept only the non-inferable bits: dependabot can't own these pins because they carry digests, and the payload is untrusted. Also trimmed the inline doc blocks in the same file.

---

.github/workflows/sync-stack-workload-catalog.yml
avallete: Header cut 26 lines to 11 in `3953876`, plus the three in-body comments and the generated PR body.

---

apps/cli/src/shared/services/slim-images.unit.test.ts
avallete: Down to 3 lines in `3953876`.

---

packages/stack/src/model/WorkloadCatalog.ts
avallete: 15 lines to 5 in `3953876`. Kept the "deliberately diverge, don't reconcile" warning — that's the trap, since this file and the Dockerfile list overlapping versions.

---

avallete:
## Red checks here are runner loss, not this PR

Seven jobs across three runs on this branch have failed with an identical, measurable signature that is distinguishable from a test failure:

- **Terminated at exactly 600 seconds**, to the second, from their own start times.
- **Empty `steps` array** in the jobs API — not a failed step, *no step records at all*, not even `Set up job`.
- **Logs 404**, while logs for successful jobs in the same run download fine.

A job that fails a test assertion always records the failing step and uploads logs. These never executed a test body. There is no `timeout-minutes` on these jobs in `test.yml`, so the 600s wall is not coming from the workflow.

| Run | Job | Window |
|---|---|---|
| 5426 | `Check code quality` | 13:35:07→13:45:07 |
| 5426 | CLI e2e shard 3/3 | 13:34:33→13:44:33 |
| 5426 | CLI e2e shard 1/3 | 13:36:44→13:46:44 |
| 5426 | stack e2e (container) | 13:34:35→13:44:35 |
| — | `Build CLI artifacts` (32vcpu) | 13:36:38→13:46:38 |
| 5429 | `Run integration tests` | 14:22:03→14:32:03 |
| 5429 | CLI e2e shard 2/3 | 14:22:40→14:32:40 |

Three independent cross-checks say it is not the diff:

1. **Matrix siblings pass on the same commit.** On `3953876`, e2e shards 1/3 and 3/3 and `Check code quality` all completed green with full step logs while shard 2/3 died at 600s.
2. **Every casualty in run 5426 passed on a single re-run**, with no code change — `Check code quality` in 2m24s and `Build CLI artifacts` in 1m25s, having previously burned the full ten doing nothing.
3. **The current commit is comment-only.** `3953876` changes nothing but comment text; `.github/scripts/**` is in no build path, and `apps/cli` declares no dependency on `@supabase/stack`.

It spans three Blacksmith runner groups (4vcpu/14, 8vcpu/15, 32vcpu/16), so it is not one bad pool.

**What I am doing:** re-running the failed jobs once the current run finishes (`rerun_failed_jobs` needs a completed run). I already spent a re-run on run 5426 and said a second failure would be real — I am revisiting that deliberately, because these are new failures on a new commit from an ongoing external incident, and the zero-step-record evidence says no test ever ran. If a job ever fails here *with* step records and logs, that is a real failure and I will treat it as mine.

Worth a look from someone with Blacksmith access if it keeps up.

---

Coly010 (approved):

PR #6529, opened 2026-09-08
fix(stack): acquire ports before committing reservations

jgoux:
Port allocation could change an existing automatic assignment when an earlier listener was enabled, or let a fresh candidate collide with an exact or retained private assignment. Validate all retained and exact claims first, then acquire public listeners and temporary private reservations inside one registry transaction. Commit assignments once acquisition succeeds and let scopes release unsuccessful attempts, removing the probe/replan/rollback path.

Use a shared, randomized port pool with bounded retries for occupied or inaccessible candidates. Allocate native listener resources separately for each Effect evaluation, avoid duplicate internal API listeners when a public wildcard already covers the address, and preflight only private bindings still requested by the candidate plan. Document the reservation guarantees and add a focused Bun CI matrix for macOS, Linux, and Windows.


packages/stack/src/state/PortCoordinator.ts
```diff
@@ -188,306 +129,282 @@ const readAuthoritativeStates = (options: PortCoordinatorOptions) =>
...
+        }
+        const exactAssignments = new Map<PortField, HostPortAssignment>();
+        for (const field of fields) {
+          const intent = listenerIntents[field];
+          if (!intent.enabled || intent.port === "automatic") continue;
+          if (!validPort(intent.port)) return yield* unavailable(intent.port, field);
+          const foreign = foreignConflict(intent.port, field);
+          if (foreign !== undefined) return yield* foreign;
+          const duplicate = claim(intent.port, field);
+          if (duplicate !== undefined) return yield* duplicate;
```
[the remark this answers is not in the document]
jgoux: Keeping the documented sticky-port contract: retained public and private assignments are hard constraints. A conflicting exact public request fails explicitly without moving the retained private port; only fresh automatic assignments can try another candidate. Silently reallocating the private assignment would change that policy, so I am not applying this suggestion.

packages/stack/src/state/ports.integration.test.ts
```diff
@@ -34,910 +42,777 @@ const disabledIntents = (): ListenerIntents => ({
...
-        expect(
-          ((yield* store.read(stackId))?.ports ?? []).some(
-            (assignment) => assignment.port === occupiedPort,
-          ),
-        ).toBe(true);
+        expect(Exit.isFailure(result)).toBe(true);
+        expect((yield* store.read(a))?.privatePorts).toEqual([assignment]);
       }),
     ),
   );
```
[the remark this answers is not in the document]
jgoux: Restored this at the persistence boundary in state-store.integration.test.ts. The test initializes valid state, attempts to replace it with overlapping public/private ports, checks the typed overlap error, and verifies that the previous document remains unchanged.

packages/stack/src/state/PortCoordinator.ts
```diff
@@ -188,306 +129,282 @@ const readAuthoritativeStates = (options: PortCoordinatorOptions) =>
...
+        const crypto = yield* Crypto.Crypto;
+        const randomStart = yield* crypto.randomIntBetween(0, PORT_POOL_SIZE - 1);
+        let offset = 0;
+        const allocateFresh = <A>(
+          field: string,
+          bind: (port: number) => Effect.Effect<A, PortUnavailableError, Scope.Scope>,
+        ): Effect.Effect<
+          { readonly port: number; readonly value: A },
+          PortAllocationError | PortUnavailableError,
+          Scope.Scope
```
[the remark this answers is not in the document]
jgoux: Both current call sites explicitly provide the correct owner: attemptScope for public listeners and privateScope for temporary private reservations. The failure and interruption tests verify cleanup while the caller-provided scope remains open. There is no current ownership defect here; I am keeping the local helper and the explicit scope provision at its call sites rather than changing the interface for a hypothetical future omission.

docs/adr/0017-simplified-managed-stack-architecture.md
```diff
@@ -63,16 +63,37 @@ first removes exact stack-owned remnants and creates fresh resources;
...
 composition preflights configuration without trying to bind ports that the
 current Supervisor intentionally owns. Cold recovery never adopts those
-resources. If an authoritative bind loses a race for a fresh automatic public
-assignment, planning retries with a bounded exclusion set; exact assignments
-and previously persisted sticky automatic ports remain hard failures and are
-never silently moved.
+resources. An actual bound wildcard may cover the internal API and avoid a
+duplicate listener or gateway adoption; IPv6 wildcards explicitly use dualstack.
+A remaining separate internal bridge bind can still fail once its topology is
+discovered. The reservation policy does not
```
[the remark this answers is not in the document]
jgoux: Reflowed the added port paragraphs to the document's wrapping style. Verified that the word sequence is unchanged; this is formatting only.

packages/stack/src/state/PortCoordinator.ts
```diff
@@ -188,306 +129,282 @@ const readAuthoritativeStates = (options: PortCoordinatorOptions) =>
...
+                  ports: assignments,
+                  privatePorts: privateAssignments,
+                };
+                yield* options.store.replaceUnlocked(stackId, next);
+                yield* Scope.close(privateScope, Exit.void);
+                return { assignments: byField, privateAssignments, hostListeners: listeners };
+              }).pipe(
+                Effect.onExit((exit) =>
+                  Exit.isSuccess(exit) ? Effect.void : Scope.close(attemptScope, exit),
+                ),
```
avallete: `replaceUnlocked` sits inside `restore()`, so an interrupt after this write and before a successful return takes the `onExit` failure path, closes `attemptScope`, and leaves ports persisted with no listeners.

The interrupt test only cancels a waiting bind, not this post-persist window. Next start usually rebinds retained ports if they are still free; the bad case is another process taking them first.

Worth moving the persist (and the success return) into the uninterruptible region, with binds staying under `restore()`.

packages/stack/src/state/ports.integration.test.ts
```diff
@@ -34,910 +42,777 @@ const disabledIntents = (): ListenerIntents => ({
...
-        );
-        expect(result.assignments.api?.port).toBe(55440);
-        expect((yield* store.read(stackId))?.desiredLifecycle).toBe("running");
+        const value = identity(root, "stopped");
+        const id = yield* deriveStackId(value);
+        const before = state(id, value, [{ field: "api", port: 24_001, intent: "automatic" }]);
+        yield* store.initialize(id, { ...before, desiredLifecycle: "stopped" });
+        const coordinator = makePortCoordinator(coordinatorOptions(store, root));
+        const result = yield* coordinator.acquire(id, intents(), []).pipe(Effect.exit);
+        expect(result).toHaveProperty("_tag", "Failure");
```
avallete: This inspects Effect's `Exit` representation. The rest of this file already uses `Exit.isFailure`.

packages/stack/src/runtime/ProductionRuntime.ts
```diff
@@ -807,12 +808,20 @@ export const makeProductionRuntime = (
...
             );
           }
-          for (const assignment of input.state.privatePorts)
+          const requestedPrivate = new Set(
+            privateBindingIntentsFor(input.plan).map(
+              (binding) => `${binding.workloadId}\u0000${binding.binding}`,
+            ),
+          );
+          for (const assignment of input.state.privatePorts) {
+            if (!requestedPrivate.has(`${assignment.workloadId}\u0000${assignment.binding}`))
```
avallete: Same `workloadId\u0000binding` format as `bindingKey` in `PortCoordinator.ts`. A mismatch here would silently skip or keep the wrong private preflight. Export one helper and reuse it.

packages/stack/src/state/ports.integration.test.ts
```diff
@@ -34,910 +42,777 @@ const disabledIntents = (): ListenerIntents => ({
...
+          .pipe(Effect.provideService(Crypto.Crypto, deterministic));
+        expect(result.assignments.database?.port).toBe(20_000);
+        expect(result.assignments.api?.port).not.toBe(20_000);
       }),
     ),
   );
 
-  it.live("keeps exact and sticky automatic assignments after bind failure", () =>
-    withPlatform(
+  it.live("retries retryable fresh bind failures beyond the old sixteen-attempt limit", () =>
```
avallete: This name encodes a deleted 16-attempt cap, and `attempts === 18` is an implementation count. The 64-exhaustion case already locks the real budget. Fold "can succeed after many EADDRINUSE" into that test, or drop this one.

avallete (approved):
Acquire-then-commit looks correct: retained/exact claims first, binds under the registry lock, commit after success, scopes roll failed attempts back. Sticky exact vs retained private matches the ADR.

Non-blocking notes inline: persist sits in the interruptible region, two `_tag` Exit assertions, duplicated private binding keys, and a leftover 16-attempt test. None of these should block merge.

packages/stack/src/state/PortCoordinator.ts
jgoux: Fixed in fc3c256f0. Binding remains interruptible; persistence, temporary private-scope teardown, and the successful reservation return now run inside the mask. Failure cleanup closes only the attempt scope. Added a regression that pauses after the real state write and requests interruption: it fails on the previous code because the committed listener closes, and passes now with the listener held until the caller closes its scope.

packages/stack/src/state/ports.integration.test.ts
jgoux: Fixed both assertions to use Exit.isFailure in fc3c256f0.

packages/stack/src/runtime/ProductionRuntime.ts
jgoux: Added privateBindingKey alongside PrivatePortAssignment in StackState.ts and reused it in the coordinator, runtime preflight, and persisted-state duplicate validation. This keeps the shared identity format with its domain type without coupling runtime preflight to the coordinator. Included in fc3c256f0.

packages/stack/src/state/ports.integration.test.ts
jgoux: Removed the obsolete sixteen-attempt test in fc3c256f0. The 64-failure exhaustion test retains budget coverage, and the existing partial-acquisition retry scenario still verifies successful retry without rebinding earlier listeners.

PR #6544, opened 2026-09-09
feat(cli): select stack start and stop with feature flag

jgoux:
Expose the new local runtime as `supabase stack start|stop`. Projects can opt top-level `start` and `stop` into that runtime with `[experimental] stack = true` in `supabase/config.toml`, or override it with `SUPABASE_EXPERIMENTAL_STACK=1|0`. Selection happens before argument parsing and completion, preserving each implementation’s flags and the invoked command’s telemetry identity. Unreadable or invalid routing config defaults to the legacy command tree; explicit environment overrides take precedence, and each command retains its own execution-time config validation.

The shared experimental flag resolver is reusable for the later workers migration. `status` keeps its existing implementation and `experimental workers` remains available. The two stack backends retain separate state and databases; enabling the flag does not migrate existing data. The flag is excluded from hosted project configuration and accepted as a TypeScript-owned field by the Go sidecar.

Extracts the start/stop portion of #6516 onto `develop` after #6507, so it can merge independently of the remaining stack commands.

Regenerates the published configuration schemas, including previously stale auth fields already present in the source schemas. Refreshes stale Go config fixtures to match the current serializer and schema.


packages/config/src/experimental.ts
```diff
@@ -32,6 +32,12 @@ const inspectRule = Schema.Struct({
 }).pipe(Schema.withDecodingDefaultKey(Effect.succeed({})));
 
 export const experimental = Schema.Struct({
+  stack: Schema.optionalKey(
+    Schema.Boolean.annotate({
+      description: "Use the new local stack backend for top-level start and stop commands.",
+      tags,
+    }),
+  ),
```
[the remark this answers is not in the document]
jgoux: Fixed in 76e83b480. The Go config model now accepts experimental.stack as a non-emitted TypeScript-owned field. Regression coverage loads true/false from both base and remote config and verifies Eject omits the key.

apps/cli/src/command-internal/experimental-feature.ts
```diff
@@ -0,0 +1,36 @@
...
+  if (override === undefined || override === "") {
+    return input.configValue.pipe(Effect.map((value) => value === true));
+  }
+  if (override === "1") return Effect.succeed(true);
+  if (override === "0") return Effect.succeed(false);
+  return Effect.fail(
+    new ExperimentalFeatureFlagError({
+      envName,
+      message: `${envName} must be 0 or 1 when set`,
+    }),
```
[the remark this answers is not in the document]
jgoux: Keeping the documented feature-flag contract: 1 enables, 0 disables, and empty/unset falls back to project config. Other spellings fail with an actionable error. This is an explicit switch for the new feature-flag mechanism, rather than a general config boolean override.

apps/cli/src/cli/root.ts
```diff
@@ -59,111 +68,124 @@ import {
...
+      backupsCommand,
+      bootstrapCommand,
+      branchesCommand,
+      completionCommand,
+      configCommand,
+      dbCommand,
+      domainsCommand,
+      encryptionCommand,
+      experimentalCommand,
+      stackCommand,
```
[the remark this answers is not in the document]
jgoux: Fixed in 76e83b480: stack is now registered between sso and start, preserving declaration order for help and documentation.

apps/cli/src/commands/experimental/stack/stack-backend.ts
```diff
@@ -0,0 +1,153 @@
...
+    const completePath =
+      commandPath[0] === "__complete" || commandPath[0] === "__completeNoDesc"
+        ? commandPath.slice(1)
+        : commandPath;
+    const command = completePath[0];
+
+    // The explicit namespace is always backed by the stack runtime and does
+    // not need a project config or environment lookup to select it.
+    if (command === "stack") return "stack";
+    if (command !== "start" && command !== "stop") return "legacy";
```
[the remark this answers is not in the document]
jgoux: Keeping fail-closed backend selection for help so malformed routing config cannot silently show the wrong implementation’s flags. 76e83b480 adds an actionable recovery suggestion: SUPABASE_EXPERIMENTAL_STACK=0 for legacy start/stop, or the explicit supabase stack namespace.

apps/cli/src/cli/main.ts
```diff
@@ -1,13 +1,34 @@
...
+const selectedRoot = Exit.isSuccess(backendExit)
+  ? rootCommandForBackend(backendExit.value)
+  : rootCommand;
+const completionRoot = Exit.isSuccess(backendExit) ? selectedRoot : undefined;
+
+if (!(await tryComplete(defaultCompleteDeps(completionRoot)))) {
+  await runCli(selectedRoot, {
     analyticsLayer: analyticsLayer,
     afterSuccess: upgradeNoticeHook,
+    ...(Exit.isFailure(backendExit) ? { beforeParse: Effect.failCause(backendExit.cause) } : {}),
```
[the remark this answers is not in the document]
jgoux: Fixed in 76e83b480. The completion interceptor receives the original routing failure and renders it on stderr before exiting 1; stdout remains reserved for the completion protocol, including JSON output mode. Added coverage for completion failures and real entrypoint failure rendering. Selected start-backend flag coverage also verifies that the alias exposes the new runtime flags.

apps/cli/src/cli/main.ts
```diff
@@ -1,13 +1,34 @@
...
+);
+
+const backendExit = await Effect.runPromiseExit(
+  resolveStackBackend({ args, cwd: process.cwd(), env: process.env }).pipe(
+    Effect.provide(BunServices.layer),
+  ),
+);
+const selectedRoot = Exit.isSuccess(backendExit)
+  ? rootCommandForBackend(backendExit.value)
+  : rootCommand;
```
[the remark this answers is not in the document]
jgoux: Fixed in 76e83b480 by reusing the existing legacy root for legacy invocations. The stack tree is constructed only when selected; no memoization layer is needed.

apps/cli/src/commands/experimental/stack/stack-backend.ts
```diff
@@ -0,0 +1,153 @@
...
+
+const parseConfig = (path: string, content: string): Effect.Effect<unknown, StackRoutingError> =>
+  Effect.try({
+    try: () => SmolToml.parse(content),
+    catch: (cause) =>
+      new StackRoutingError({
+        message: `Unable to read ${path}: ${String(cause)}`,
+        cause,
+      }),
+  });
```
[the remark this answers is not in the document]
jgoux: Fixed in 76e83b480: TOML syntax failures now say Unable to parse; filesystem failures retain Unable to read.

apps/cli/src/docs/docs-spec.tables.ts
```diff
@@ -139,6 +140,8 @@ export const DOCS_DEFAULT_OVERRIDES: Readonly<Record<string, string>> = {
   "supabase output": "pretty",
   "supabase output-format": "text",
   "supabase profile": "supabase",
+  "supabase-stack-start runtime": "auto",
+  "supabase-stack-start preparation": "background",
```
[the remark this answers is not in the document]
jgoux: Fixed in 76e83b480: the stack defaults follow seed-buckets, with preparation before runtime.

apps/cli/src/cli/root.ts
```diff
@@ -59,111 +68,124 @@ import {
...
-        const agent = yield* AgentFlag;
-        const cliArgs = yield* CliArgs;
+const stackStartAliasCommand = experimentalStackStartCommand.pipe(
+  Command.provide(commandRuntimeLayer(["start"])),
+  Command.provide(experimentalStackRuntimeLayer),
+);
+export const stackStopAliasCommand = experimentalStackStopCommand.pipe(
+  Command.provide(commandRuntimeLayer(["stop"])),
+  Command.provide(experimentalStackRuntimeLayer),
+);
```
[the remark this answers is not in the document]
jgoux: Keeping the invoked command identity as designed. Backend-specific rollout measurement is not a requirement of this extraction, and the suggestion is conditional on that need. A separate instrumentation change can define the event property if rollout analysis requires it.

apps/cli/src/commands/experimental/stack/stack-command-telemetry.integration.test.ts
```diff
@@ -0,0 +1,84 @@
...
+} from "../../../../tests/helpers/mocks.ts";
+import { CliArgs } from "../../../shared/cli/cli-args.service.ts";
+import { DebugFlag, ProfileFlag, WorkdirFlag } from "../../../command-internal/global-flags.ts";
+import {
+  EventCommandExecuted,
+  PropCommand,
+  PropCommandRunId,
+} from "../../../shared/telemetry/event-catalog.ts";
+import { textCliOutputFormatter } from "../../../shared/output/text-formatter.ts";
+import { stackCommand } from "../../../commands/experimental/stack/stack.command.ts";
```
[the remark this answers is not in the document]
jgoux: Fixed in 76e83b480: the test uses the direct sibling import ./stack.command.ts.

apps/docs/public/cli/config.schema.json
```diff
@@ -979,6 +993,47 @@
               },
               "additionalProperties": false
             },
+            "figma": {
```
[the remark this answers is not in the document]
jgoux: The PR description now explicitly calls out the previously stale auth fields included by schema regeneration. Keeping generated artifacts synchronized with their current source avoids manually pruning generator output.

apps/cli/src/commands/experimental/stack/stack-backend.ts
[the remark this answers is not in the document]
jgoux: Fixed the reproduced cursor-token issue in 76e83b480: both completion protocols exclude the final token before backend selection and workdir lookup. Root candidates remain available with malformed config or an invalid flag environment value, while completed start/stop command paths select the configured backend. The synthetic help-path concern does not currently change observable candidates for these leaf commands, so no additional help-routing special case was added.

apps/cli/src/cli/root.ts
```diff
@@ -59,111 +68,124 @@ import {
...
+      postgresConfigCommand,
+      projectsCommand,
+      secretsCommand,
+      seedCommand,
+      servicesCommand,
+      snippetsCommand,
+      sslEnforcementCommand,
+      ssoCommand,
+      stackCommand,
+      backend === "stack" ? stackStartAliasCommand : startCommand,
```
Coly010, line 116: Legacy `start` has `-x/--exclude`, `--ignore-health-check`, `--preview`; the stack alias (`experimentalStackStartCommand`) has none of them. Same on line 116 for `stop`'s `--project-id/--backup/--no-backup/--all`. With `experimental.stack = true`, running any of those produces a bare `Unrecognized flag: --all in command supabase stop` — nothing tells the user a backend switch is why the flag disappeared, and `--all` (the only machine-wide Docker cleanup) has no equivalent on the stack side at all.

Could the alias wrapping thread the selected backend into the parse-error path (e.g. via `run.ts`'s `CliErrorSuggestionContext`) so an unrecognized flag that's known to belong to the *other* backend gets a suggestion like "`--all` belongs to the legacy backend; set `SUPABASE_EXPERIMENTAL_STACK=0`"? Right now this is a guaranteed Monday-morning CI failure for the first team that adopts the flag, with no self-service fix in the error itself.

apps/cli/src/cli/main.ts
```diff
@@ -1,13 +1,42 @@
...
+    defaultCompleteDeps(
+      completionRoot,
+      Exit.isFailure(backendExit) ? backendExit.cause : undefined,
+    ),
+  ))
+) {
+  await runCli(selectedRoot, {
     analyticsLayer: analyticsLayer,
     afterSuccess: upgradeNoticeHook,
+    ...(Exit.isFailure(backendExit) ? { beforeParse: Effect.failCause(backendExit.cause) } : {}),
```
Coly010, line 40: When `resolveStackBackend` fails (e.g. a malformed `config.toml`), this unconditionally fails `beforeParse` — including for `supabase start --help`/`supabase stop --help`. Help should never depend on project config being valid; right now a broken `[experimental]` block breaks `--help` outright instead of showing the flag list.

Worth special-casing a help request the same way `hasRootVersionFlag` already short-circuits `--version` to legacy in `stack-backend.ts`, so `--help`/`-h` skips routing entirely.

apps/cli/src/commands/experimental/stack/stack-backend.ts
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  get [ErrorActionabilityId](): CliErrorActionabilityDeclaration {
+    return actionability.invalidConfig;
+  }
+}
+
+const stackRoutingSchema = Schema.Struct({
+  experimental: Schema.optionalKey(
+    Schema.Struct({ stack: CliConfigSchema.fields.experimental.to.fields.stack }),
```
Coly010, line 30: This schema only decodes the top-level `experimental.stack`. `apps/cli-go/pkg/config/config_test.go`'s new "remote true"/"remote false" cases assert that `[remotes.prod.experimental] stack = true` resolves on the Go side, and the published schema mirrors `stack` under `remotes.*` too — but this router never looks at `remotes`, so setting it there is a silent no-op on the TS side. Either honor remotes here, or the Go remote-override tests (and the schema) are asserting semantics this router doesn't actually implement.

apps/cli/src/commands/experimental/stack/stack-backend.ts
```diff
@@ -0,0 +1,164 @@
...
+      const path = yield* Path.Path;
+      const explicitWorkdir = firstExplicitLongFlagValue(routingArgs, "workdir");
+      const resolvedWorkdir = yield* resolveWorkdir(
+        explicitWorkdir === undefined ? Option.none() : Option.some(explicitWorkdir),
+        input.env["SUPABASE_WORKDIR"],
+        input.cwd,
+        (filePath) => fs.exists(filePath).pipe(Effect.orElseSucceed(() => false)),
+        path,
+      );
+      const configPath = path.join(resolvedWorkdir.workdir, "supabase", "config.toml");
```
Coly010, line 108: This always joins `config.toml`, but `findConfigInRoot` (packages/config) prefers `config.json` when both exist. Two divergences follow: a JSON-only project can never opt into the stack backend (no warning — `stack-commands.md` documents this as expected, but a silent no-op is still a silent no-op), and a project with a live `config.json` plus a stale leftover `config.toml` containing `stack = true` will route to the stack backend even though `config.json` is the config that's actually in effect. `packages/config/src/io.unit.test.ts`'s new test asserts `experimental.stack` loads from `config.json`, which this code path doesn't honor at all.

Routing through the real loader (or at least warning when a `config.json` exists alongside the `stack` key in `config.toml`) would close this gap.

apps/cli/src/commands/experimental/stack/stack-backend.ts
```diff
@@ -0,0 +1,164 @@
...
+): string | undefined => {
+  for (const { token, index } of rootFlagTokens(args)) {
+    if (token === `--${flagName}`) return args[index + 1];
+    if (token.startsWith(`--${flagName}=`)) return token.slice(flagName.length + 3);
+  }
+  return undefined;
+};
+
+/** Extracts command path tokens while honoring optional separated boolean values. */
+const extractRoutingCommandPath = (args: ReadonlyArray<string>): ReadonlyArray<string> => {
```
Coly010: This is now the 4th independent argv command-path scanner in the codebase (alongside `extractCommandPath`/`rootFlagTokens` in `shared/cli/run.ts` and the version-flag scanner in `agent-output.ts`), and it disagrees with the others on short flag clusters: `rootFlagTokens`, imported two lines above in this same file, already has `shortClusterConsumesNextToken` handling that this loop doesn't reuse. Concretely, `supabase -yo json start` with `stack = true` resolves `commandPath[0] === "json"` here (since `-yo`'s value-consuming behavior isn't modeled), so `command` never becomes `"start"` and this silently routes to legacy — no error, just the wrong backend running. `run.ts` already carries comments flagging this scanner family as a "DELIBERATE MODEL SPLIT" needing consolidation; this adds a fourth divergent copy rather than building on `rootFlagTokens` directly.

apps/cli/src/command-internal/experimental-feature.ts
```diff
@@ -0,0 +1,36 @@
...
+  readonly feature: string;
+  readonly configValue: Effect.Effect<boolean | undefined, E, R>;
+  readonly env: Readonly<Record<string, string | undefined>>;
+}): Effect.Effect<boolean, E | ExperimentalFeatureFlagError, R> => {
+  const envName = `SUPABASE_EXPERIMENTAL_${input.feature.toUpperCase()}`;
+  const override = input.env[envName];
+  if (override === undefined || override === "") {
+    return input.configValue.pipe(Effect.map((value) => value === true));
+  }
+  if (override === "1") return Effect.succeed(true);
```
Coly010, line 28: Only `"1"`/`"0"` are accepted — `SUPABASE_EXPERIMENTAL_STACK=true` (the spelling `config.toml`'s own `stack = true` trains users to reach for, and what Go's `strconv.ParseBool`-based env handling elsewhere accepts) hard-fails with "must be 0 or 1 when set." Would it make sense to accept the same boolean set the rest of the CLI does (`run.ts`'s `PFLAG_BOOL_TRUE`/`agent-output.ts`'s `BOOLEAN_FLAG_VALUES`), and keep the error only for genuinely unparseable values?

apps/cli/src/commands/experimental/stack/stack.command.ts
```diff
@@ -1,18 +1,30 @@
...
+
+const stackStartCommand = experimentalStackStartCommand.pipe(
+  Command.provide(commandRuntimeLayer(["stack", "start"])),
+);
+const stackStopCommand = experimentalStackStopCommand.pipe(
+  Command.provide(commandRuntimeLayer(["stack", "stop"])),
+);
+
+export const stackCommand = Command.make("stack").pipe(
+  Command.withDescription("Manage a local Supabase stack with the new backend."),
```
Coly010: `stackCommand` is registered unconditionally in `root.ts` (not `Command.unlisted`, unlike `experimentalCommand`), and this description carries none of the "excluded from the CLI's compatibility promise" language `experimentalCommand` uses for everything else under this directory. Combined with `docs-spec.tables.ts` tagging `supabase-stack` as a plain `local-dev` doc entry with no experimental marker, a user reading the published CLI reference has no signal that this command and the backends it selects between aren't data-compatible with each other — which the PR description says they aren't. Worth either adding the instability disclaimer here and to `DOCS_EXPERIMENTAL`, or keeping this unlisted until the runtime is committed to.

apps/cli/src/docs/docs-spec.tables.ts
```diff
@@ -67,6 +67,7 @@ export const DOCS_TAGS: Readonly<Record<string, ReadonlyArray<string>>> = {
   "supabase-snippets": ["management-api"],
   "supabase-ssl-enforcement": ["management-api"],
   "supabase-sso": ["management-api"],
+  "supabase-stack": ["local-dev"],
```
Coly010, line 71: `supabase-stack` is tagged like any other stable `local-dev` command here, but is absent from `DOCS_EXPERIMENTAL` below (line 82) even though its source lives entirely under `commands/experimental/` and every symbol is `experimentalStack*`. Should this be added to `DOCS_EXPERIMENTAL` (or its start/stop leaves specifically) to match how the rest of the experimental surface is documented?

apps/cli/src/commands/stop/SIDE_EFFECTS.md
```diff
@@ -1,5 +1,10 @@
 # `supabase stop`
 
+This document describes the legacy backend. With `SUPABASE_EXPERIMENTAL_STACK=1`, or
+`[experimental] stack = true` when the environment override is unset or empty, `supabase stop`
```
Coly010, line 4: The "Files Read" table further down (still says `config.toml` is "skipped entirely when `--project-id` or `--all` is set") doesn't get a row for what this new blurb implies: the pre-parse routing step (`stack-backend.ts`'s `resolveStackBackend`, wired in from `main.ts`) now reads `config.toml` for *every* `stop` invocation, including `--all`/`--project-id`, before the command even dispatches — so that "skipped entirely" claim is no longer true. Concretely, a malformed `config.toml` now breaks `supabase stop --all` (verified — exit 1 `StackRoutingError`), which previously succeeded on develop precisely because `--all` bypassed config entirely. That's the one escape hatch you reach for when config is broken, so worth confirming this is intended rather than an oversight of the routing addition, and updating the table if so.

Coly010 (changes requested):
Ran this through parallel architecture, implementation, and DX reviews. The pre-parse backend-selection mechanism itself (lazy config reads, typed routing errors, completion-cursor handling, preserved telemetry identity) is well-built. The problems are all in what happens once someone actually flips `experimental.stack = true` — inline comments below cover the specific spots, and this summarizes what doesn't map to a single line:

**No signposting when the backend switches.** `supabase start`/`stop` with the flag on produce a completely different output shape (a `Stack <id> / Runtime / Lifecycle` block instead of the usual API URL / anon key table) with nothing printed anywhere saying "this is the experimental stack backend" or "your legacy stack's data is untouched." A user flipping the flag has no way to tell this from data loss.

**The rest of local dev doesn't follow the flag.** `status`, `db reset`, and `db diff` stay bound to the legacy Docker stack regardless of `experimental.stack`, so with the flag on, `supabase start` (new stack) followed by `supabase status` (legacy) reports nothing running. None of this is documented in `docs/stack-commands.md`, which only mentions `status` in passing.

**`stack stop` doesn't know about a legacy stack from the same project.** If a legacy stack is running when the flag gets flipped on, `supabase stop` (now routed to the stack backend) exits 0 having found nothing to do, while the legacy containers keep the port bound — so the next `supabase start` fails with a `PortUnavailableError` that never mentions the other backend is the actual cause.

**Docs ordering.** `docs/stack-commands.md` buries the "backends don't share data" caveat at the bottom in defensive phrasing ("does not import, copy, seed from, or reuse") instead of leading with it, and references a command path (`supabase experimental stack`) that only ever existed in an unreleased beta — likely to confuse rather than help.

**Scope creep.** The 14 `apps/cli-go/pkg/config/testdata/*.diff` fixture updates plus the `db_test.go`/`updater_test.go` changes fix pre-existing drift (a `db_allowed_cidrs`→`allowed_cidrs` rename, a swapped diff direction, a deleted gock stub) unrelated to stack routing — worth splitting into its own PR so a reviewer of a start/stop feature flag isn't also on the hook for auth/network-restrictions fixture correctness.

Given the number of user-facing regressions once the flag is actually used, I'd want at minimum the flag-stripping diagnostic, the `--help`/malformed-config interaction, and the config-loader bypass (comments below) resolved before merge.

apps/cli/src/cli/root.ts
jgoux: The feature flag intentionally replaces the top-level commands with the new backend’s command definitions, including their supported arguments and flags. We are keeping normal parser errors rather than adding cross-backend compatibility hints. The remaining local-stack commands will migrate in follow-up PRs before this flag is promoted.

apps/cli/src/cli/main.ts
jgoux: Addressed in 529b25422: unreadable, malformed, or invalid routing config now defaults to legacy when selecting the command tree. Valid config still selects backend-specific help; explicit environment overrides remain authoritative. Each selected command retains its own execution-time config validation. Invalid environment override values still produce an error.

apps/cli/src/commands/experimental/stack/stack-backend.ts
jgoux: Local lifecycle commands do not select a remote project, so remote blocks do not control their backend. The Go tests cover accepting the shared field when Go loads an explicitly selected remote, not local routing semantics. Keeping the local flag in the base TOML config; selecting remote overlays here would introduce new behavior beyond this extraction.

jgoux:
Thanks for the detailed review. This PR is the foundation for the new local stack: status and the remaining local-stack commands will migrate in follow-up PRs before the feature flag is promoted. Backend-specific arguments and flags are intentional, and each backend continues to own its own resources. We are not adding temporary cross-backend compatibility behavior or migration banners for this intermediate state.

This revision fixes short-cluster routing, defaults unresolved config-based routing to legacy, and clarifies experimental status and configuration reads. Local TOML-only loading, remote-selection behavior, and the 1|0 environment contract remain unchanged for the reasons in the inline replies.

The fixture-only Go repairs are retained so the full nested config module suite passes; they align stale fixtures with existing serializers and do not change runtime behavior.

apps/cli/src/commands/experimental/stack/stack-backend.ts
jgoux: Both legacy start and the new stack runtime call loadLocalProjectContext, which explicitly uses tomlOnly: true. The pre-parse router therefore follows its actual consumers; the generic config loader’s JSON preference does not apply to these local commands. Keeping this existing local-loading policy and its documentation. JSON support would need to change routing and execution together.

apps/cli/src/commands/experimental/stack/stack-backend.ts
jgoux: Fixed in 529b25422. Routing now shares short-cluster consumption logic and skips the output value correctly. Coverage includes the reported -yo spelling at the scanner level and the valid -ho json start cluster, which exposes the selected backend’s help. Separated boolean values and the completion-cursor boundary remain covered.

apps/cli/src/command-internal/experimental-feature.ts
jgoux: Keeping the explicit feature-flag contract agreed for this mechanism: 1 enables, 0 disables, empty/unset falls back to config. Invalid environment values remain errors. The new backend’s opt-in interface does not need to mirror all legacy boolean parsing conventions.

apps/cli/src/commands/experimental/stack/stack.command.ts
jgoux: Addressed the instability wording in 529b25422 while keeping the command listed. DOCS_EXPERIMENTAL is not a maturity marker: it adds a required --experimental flag, which this command does not require. Its description now communicates the experimental compatibility status without introducing a new gate.

apps/cli/src/docs/docs-spec.tables.ts
jgoux: DOCS_EXPERIMENTAL represents runtime commands that require --experimental and injects that required flag into generated docs. Adding stack there would document an incorrect requirement. 529b25422 adds experimental/unstable wording to the command description instead, which is consumed by documentation generation.

apps/cli/src/commands/stop/SIDE_EFFECTS.md
jgoux: Addressed in 529b25422. Routing may read config before dispatch, but an unreadable/malformed routing config now falls back to legacy. The legacy handler can then preserve its existing --all/--project-id config bypass. Updated the side-effect documentation to distinguish routing reads from handler reads.

Coly010 (approved):

PR #6589, opened 2026-09-13
fix(cli): handle piped `config pull` input (CLI-2425)

### LINKED ISSUE #6585: supabase config pull crashes with Service not found: supabase/runtime/Stdin when stdin is piped into the command.
### Affected area

Other

### Supabase CLI version

v2.117.0

### Operating system

Arch Linux

### Installation method

npm

### Command

```shell
echo y | supabase config pull
```

### Actual output

```shell
Apply 1 change(s) to supabase/config.toml? [Y/n] Service not found: supabase/runtime/Stdin
```

### Expected behavior

supabase config pull should be able to read input from a pipe and continue normally. Passing y should apply the change, while n should decline it.

This matches the behavior described in supabase config pull --help: even in a non-interactive run, the prompt should be written to stderr and one line should be read from piped stdin, with y/n honored and the default used otherwise. --dry-run can be used to preview the change first.

### Steps to reproduce

1. In a linked Supabase project where remote config differs from local supabase/config.toml:
2. Run echo y | supabase config pull (or pipe n, without --yes).
3. The command immediately panics with Service not found: supabase/runtime/Stdin.

### Crash report ID

_No response_

```shell

```

### Additional context

While comparing how the legacy commands handle piped input, I noticed that config pull is the only one missing stdinLayer.

In apps/cli/src/legacy/commands/config/pull/pull.command.ts, Command.provide passes legacyManagementApiRuntimeLayer(["config", "pull"]) without merging in stdinLayer.

The other commands that use legacyPromptYesNo—such as config push, branches create, functions deploy, and secrets unset—all merge stdinLayer, which is why they handle piped input correctly.

7ttp:
## TL;DR

Prevents `supabase config pull` from crashing when confirmation input is piped
the command could reach `promptYesNo` without providing `stdinLayer`

which is now fixed by:
Providing `stdinLayer` at the command boundary and covering the piped input path with a binary regression test.

## Ref:

- closes https://github.com/supabase/cli/issues/6585

### DIFF
--- a/apps/cli/AGENTS.md
+++ b/apps/cli/AGENTS.md
@@ -68,9 +68,10 @@ Every applicable command must preserve these invariants:
    [profile loader's schema](src/command-internal/profile-load.ts). Both the
    [E2E harness](../../packages/cli-test-helpers/src/harness.ts) and
    [live fixture](tests/helpers/live.ts) depend on file-path mode.
-5. Sibling layers in `Layer.mergeAll` each receive required services explicitly. Any production
-   path that can reach `promptYesNo` provides `stdinLayer`; handler-only tests do not prove this
-   wiring. Production layer changes require a CLI build and a targeted binary smoke test.
+5. Sibling layers in `Layer.mergeAll` each receive required services explicitly. Every production
+   effect graph retaining `promptYesNo`'s `Stdin` requirement provides `stdinLayer`, even when
+   runtime guards avoid its non-TTY branch. Paths that can enter that branch require a CLI build
+   and a targeted piped-input binary test; handler-only tests do not prove this wiring.
 6. Honor both output flags. `-o`/`--output` takes precedence over
    `--output-format`; a new command may reject `--output` with guidance to use
    `--output-format`, as established by `config diff` and `config pull`.
--- a/apps/cli/AGENTS.md
+++ b/apps/cli/AGENTS.md
@@ -68,8 +68,9 @@ Every applicable command must preserve these invariants:
    [profile loader's schema](src/command-internal/profile-load.ts). Both the
    [E2E harness](../../packages/cli-test-helpers/src/harness.ts) and
    [live fixture](tests/helpers/live.ts) depend on file-path mode.
-5. Sibling layers in `Layer.mergeAll` each receive required services explicitly. Production layer
-   changes require a CLI build and a targeted binary smoke test.
+5. Sibling layers in `Layer.mergeAll` each receive required services explicitly. Any production
+   path that can reach `promptYesNo` provides `stdinLayer`; handler-only tests do not prove this
+   wiring. Production layer changes require a CLI build and a targeted binary smoke test.
 6. Honor both output flags. `-o`/`--output` takes precedence over
    `--output-format`; a new command may reject `--output` with guidance to use
    `--output-format`, as established by `config diff` and `config pull`.
--- a/apps/cli/src/commands/config/pull/pull.command.ts
+++ b/apps/cli/src/commands/config/pull/pull.command.ts
@@ -1,9 +1,10 @@
-import { Option } from "effect";
+import { Layer, Option } from "effect";
 import type * as CliCommand from "effect/unstable/cli/Command";
 import { Command, Flag } from "effect/unstable/cli";
 
 import { PROJECT_REF_PATTERN } from "../../../config/project-ref.service.ts";
 import { withJsonErrorHandling } from "../../../shared/output/json-error-handling.ts";
+import { stdinLayer } from "../../../shared/runtime/stdin.layer.ts";
 import { GLOBAL_OUTPUT_FORMATS } from "../../../command-internal/global-flags.ts";
 import { managementApiRuntimeLayer } from "../../../command-internal/management-api-runtime.layer.ts";
 import { withCommandTelemetry } from "../../../telemetry/command-telemetry.ts";
@@ -75,5 +76,5 @@ export const configPullCommand = Command.make("pull", config).pipe(
     },
   ]),
   Command.withHandler(configPullHandler),
-  Command.provide(managementApiRuntimeLayer(["config", "pull"])),
+  Command.provide(Layer.mergeAll(managementApiRuntimeLayer(["config", "pull"]), stdinLayer)),
 );
--- a/apps/cli/src/commands/config/pull/pull.e2e.test.ts
+++ b/apps/cli/src/commands/config/pull/pull.e2e.test.ts
@@ -81,7 +81,9 @@ describe("config pull CLI surface", () => {
       );
 
       expect(exitCode, `${stdout}\n${stderr}`).toBe(0);
-      expect(stderr).toContain(`Apply 1 change(s) to supabase/config.toml? [Y/n] n\n`);
+      expect(stderr).toContain(
+        `Apply 1 change(s) to ${join("supabase", "config.toml")}? [Y/n] n\n`,
+      );
       expect(stdout).toContain("not written (declined)");
       expect(await readFile(configPath, "utf8")).toBe(before);
     } finally {
--- a/apps/cli/src/commands/config/pull/pull.e2e.test.ts
+++ b/apps/cli/src/commands/config/pull/pull.e2e.test.ts
@@ -1,13 +1,15 @@
-import { mkdtemp, rm } from "node:fs/promises";
+import { mkdir, mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { describe, expect, test } from "vitest";
 
 import { runSupabase } from "../../../../tests/helpers/cli.ts";
+import { v2ProjectConfigResponse } from "../../../../tests/helpers/config-fixtures.ts";
 
 // A fake-but-well-formed token bypasses the eager SUPABASE_ACCESS_TOKEN check, so the run
 // reaches this command's own handler instead of failing generically first.
 const TEST_TOKEN = "sbp_" + "a".repeat(40);
+const TEST_REF = "abcdefghijklmnopqrst";
 
 describe("config pull CLI surface", () => {
   test("plain `config pull` parses — no boolean flag is accidentally required", async () => {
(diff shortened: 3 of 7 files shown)


apps/cli/src/commands/config/pull/pull.command.ts
```diff
@@ -75,5 +76,5 @@ export const configPullCommand = Command.make("pull", config).pipe(
     },
   ]),
   Command.withHandler(configPullHandler),
-  Command.provide(managementApiRuntimeLayer(["config", "pull"])),
+  Command.provide(Layer.mergeAll(managementApiRuntimeLayer(["config", "pull"]), stdinLayer)),
```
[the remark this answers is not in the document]
7ttp: fixed: https://github.com/supabase/cli/commit/e1af5d87b75d1cf39696822e7e0e89d48cedbe77

apps/cli/AGENTS.md
```diff
@@ -68,8 +68,9 @@ Every applicable command must preserve these invariants:
    [profile loader's schema](src/command-internal/profile-load.ts). Both the
    [E2E harness](../../packages/cli-test-helpers/src/harness.ts) and
    [live fixture](tests/helpers/live.ts) depend on file-path mode.
-5. Sibling layers in `Layer.mergeAll` each receive required services explicitly. Production layer
-   changes require a CLI build and a targeted binary smoke test.
+5. Sibling layers in `Layer.mergeAll` each receive required services explicitly. Any production
+   path that can reach `promptYesNo` provides `stdinLayer`; handler-only tests do not prove this
+   wiring. Production layer changes require a CLI build and a targeted binary smoke test.
```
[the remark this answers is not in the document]
7ttp: fixed: https://github.com/supabase/cli/commit/e1af5d87b75d1cf39696822e7e0e89d48cedbe77

apps/cli/src/commands/config/pull/pull.e2e.test.ts
```diff
@@ -33,4 +35,58 @@ describe("config pull CLI surface", () => {
...
+          env: {
+            SUPABASE_ACCESS_TOKEN: TEST_TOKEN,
+            SUPABASE_PROFILE: profilePath,
+            SUPABASE_WORKDIR: cwd,
+          },
+        },
+      );
+
+      expect(exitCode, `${stdout}\n${stderr}`).toBe(0);
+      expect(stderr).toContain(`Apply 1 change(s) to supabase/config.toml? [Y/n] n\n`);
```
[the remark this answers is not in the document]
7ttp: fixed: https://github.com/supabase/cli/commit/e1af5d87b75d1cf39696822e7e0e89d48cedbe77

apps/cli/src/commands/experimental/stack/destroy/destroy.e2e.test.ts
```diff
@@ -0,0 +1,13 @@
+import { describe, expect, test } from "vitest";
+
+import { runSupabase } from "../../../../../tests/helpers/cli.ts";
+
+describe("stack destroy CLI surface", () => {
+  test("rejects an invalid stack id through the compiled command wiring", () =>
+    runSupabase(["--experimental", "stack", "destroy", "--yes", "--stack-id", "invalid"]).then(
```
Coly010: **Blocker:** this passes on the PR head but fails on head + develop. #6591 (merged 2026-09-14 07:58Z, ~20h after this PR's green run) made `stack` routing depend on `SUPABASE_EXPERIMENTAL_STACK=1` / `[experimental] stack = true` instead of the subcommand name, and `--experimental` never gated it (that's the Go passthrough flag). Merged locally, the child gets the root help dump and `Unknown subcommand "stack"`, so this will fail in the merge queue.

Verified that this passes on the merged tree:

```ts
runSupabase(["--experimental", "stack", "destroy", "--yes", "--stack-id", "invalid"], {
  cwd: /* temp dir */,
  env: { SUPABASE_EXPERIMENTAL_STACK: "1" },
})
```

(same as `stack/start/start.e2e.test.ts:195`). Please also pin `cwd` to a temp dir: without it the child inherits `apps/cli` and the gate's config discovery searches upward, so an ambient `supabase/config.toml` above the repo changes the routing decision.

Also worth being precise in the test name about what it proves. `Command.provide` builds the layer eagerly, but `Stdin` is never yielded on this path (`validateStackId` fails first, `--yes` short-circuits `promptYesNo`, and destroy's own non-TTY guard makes the read branch unreachable by design). It proves "the command constructs with the extra layer", not "piped input works". Fine as a smoke test, or drop it — the tsc gate is what covers destroy now.

apps/cli/src/shared/cli/run.ts
```diff
@@ -90,7 +89,6 @@ export type AllowedRunCliServices =
   | TelemetryRuntime
   | Tty
   | CommandPlatformApiFactory
```
Coly010: **Should-fix (here or immediate follow-up):** `CommandPlatformApiFactory` is the same latent defect as the `Stdin` entry you just removed. The root provide chain never supplies it (only the type import in this file) and it isn't a die-stub in `fallbackCommandLayer` either. Deleting this line fails tsc at `root.ts:91` — `startCommand` leaks it through `resolveStorageCredentials` (`storage-credentials.ts:84`). Unreachable today because `start` passes `projectRef: ""`, i.e. exactly the `Stdin`-on-`start` situation this PR papers over with `stdinLayer`. Suggest closing it the same way: a dying stand-in in `fallbackCommandLayer` (the established pattern for `Credentials`/`CliConfigStore`) or provide the layer in `startRuntimeLayer`, then drop the entry so the allow-list is exactly "what the root provides".

While here: the doc comment above this union (lines 69-73) reads as if listing a service is what makes an unprovided one fail at the boundary; it's the opposite. One sentence like "only services `cliProgramFor`/`runCli` actually provide (or die-stub in `fallbackCommandLayer`) may appear here" would stop the next #6303.

apps/cli/AGENTS.md
```diff
@@ -68,8 +68,10 @@ Every applicable command must preserve these invariants:
    [profile loader's schema](src/command-internal/profile-load.ts). Both the
    [E2E harness](../../packages/cli-test-helpers/src/harness.ts) and
    [live fixture](tests/helpers/live.ts) depend on file-path mode.
-5. Sibling layers in `Layer.mergeAll` each receive required services explicitly. Production layer
-   changes require a CLI build and a targeted binary smoke test.
+5. Sibling layers in `Layer.mergeAll` each receive required services explicitly. Every production
+   effect graph retaining `promptYesNo`'s `Stdin` requirement provides `stdinLayer`, even when
+   runtime guards avoid its non-TTY branch. Paths that can enter that branch require a CLI build
+   and a targeted piped-input binary test; handler-only tests do not prove this wiring.
```
Coly010: **Should-fix:** two things. (a) This drops the general sentence "Production layer changes require a CLI build and a targeted binary smoke test" and narrows the rule to `promptYesNo` paths — keep the general sentence. (b) It never names the mechanism. What actually prevents this bug class is `AllowedRunCliServices`/`CliRootCommand` in `src/shared/cli/run.ts`, enforced by the `CliRootCommand` annotation at `src/cli/root.ts:91`. Say so, and add: "never widen `AllowedRunCliServices` to silence a leaked requirement — provide the layer in the command". Otherwise the next person to hit that tsc error will do what #6303 did.

apps/cli/src/commands/config/pull/pull.command.ts
Coly010, line 81: **Nit:** every peer that provides `stdinLayer` carries a one-line "why" (`projects/delete/delete.command.ts:33`, `secrets/unset/unset.command.ts:40`, `functions/deploy/deploy.command.ts:78`, `branches/create/create.command.ts:101`, `init/init.command.ts:49`). Worth adding here and on the `destroy`/`start` provides — especially those two, where `Stdin` is provably unreachable and the provide otherwise reads as dead wiring. `start` could also fold `stdinLayer` into `startRuntimeLayer`'s `mergeAll` to match pull/push; the chained provides are semantically equivalent, so purely cosmetic.

apps/cli/src/commands/config/pull/pull.e2e.test.ts
```diff
@@ -33,4 +35,60 @@ describe("config pull CLI surface", () => {
       await rm(cwd, { recursive: true, force: true });
     }
   });
+
+  test("reads a piped confirmation through the production command layer", async () => {
```
Coly010, line 39: **Optional:** piped `n` is the right primary case (default is Y, so `n` proves the line was actually read rather than defaulted). If you want the reported repro covered too, a piped `y` case asserting `1 change written.` is cheap; #6586 had both.

Coly010 (changes requested):
The fix is right and I verified it end to end on head + develop: with `pull.command.ts` reverted, `tsc` fails at `src/cli/root.ts:91` (`Stdin` not assignable to `AllowedRunCliServices`) and the new binary e2e reproduces the exact `Service not found: supabase/runtime/Stdin` from #6585; with the fix both pull e2e tests pass and `pnpm check:all` is green. Removing `Stdin` from the allow-list is the real fix — that entry came in incidentally with #6303, which is what let #6438 ship `config pull` without `stdinLayer`.

Requesting changes for one blocker (the destroy e2e fails once #6591 is merged in) plus two should-fixes in the same spirit as the allow-list cleanup; details inline.

One process note: @mandar1045 posted the identical one-line fix plus integration tests in #6586 two days earlier, and it auto-closed for lacking `open-for-contribution`. A `Co-authored-by` trailer or a thanks on #6585 would be a good look.

mandar1045:
Thanks @Coly010 for catching that and calling it out! Really appreciate it.
Glad to see the fix going in. Happy to be added as a co-author:
Co-authored-by: Mandar Joshi <mandarjoshi1045@gmail.com>

apps/cli/src/commands/experimental/stack/destroy/destroy.e2e.test.ts
7ttp: fixed: e11a0a810

apps/cli/src/shared/cli/run.ts
7ttp: fixed: e11a0a810

apps/cli/AGENTS.md
7ttp: fixed: e11a0a810

apps/cli/src/commands/config/pull/pull.command.ts
7ttp: fixed: e11a0a810

apps/cli/src/commands/config/pull/pull.e2e.test.ts
7ttp: sticking with just the piped n case for now,
the `y `write path is pinned in the integration tests. if that ever lets a regression through i'll pick up the binary case then

Coly010 (approved):

PR #6592, opened 2026-09-14
feat(cli): add stack list

jgoux:
Adds `supabase stack list` to the existing opt-in stack command family. It lists persisted managed stacks across projects and worktrees, including stopped and unconfigured stacks, with stable ordering and text or structured output.

Readable entries appear in a table with compact IDs. Each unreadable registry entry is included with its full ID and diagnostic; JSON and stream-JSON retain full IDs and persisted metadata. Listing succeeds when the registry can be enumerated, including when individual entries are corrupted or use an unsupported format. Missing state remnants are ignored.

Uses the feature gating already on `develop` and reads registry state without contacting stack owners or changing stack lifecycle. Desired lifecycle represents persisted intent; `supabase stack status` provides live state. The new error tag is `ExperimentalStackListError`; the superseded implementation never shipped.

Supersedes #6509


apps/cli/src/commands/experimental/stack/list/list.errors.ts
```diff
@@ -0,0 +1,17 @@
+import { Data } from "effect";
+import {
+  actionability,
+  type CliErrorActionabilityDeclaration,
+  ErrorActionabilityId,
+} from "../../../../shared/telemetry/error-actionability.ts";
+
+export class StackCommandListError extends Data.TaggedError("LegacyExperimentalStackListError")<{
```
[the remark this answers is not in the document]
jgoux: The tag is intentionally retained from #6509, as recorded in the PR description. It identifies this command's errors and does not select or classify the legacy backend. Keeping that explicit carryover decision; no tag rename.

apps/cli/src/commands/experimental/stack/list/list.handler.ts
```diff
@@ -0,0 +1,90 @@
...
+            ...(error.suggestion === undefined ? {} : { suggestion: error.suggestion }),
+            cause: error,
+          }),
+      ),
+    );
+    const api = yield* StackApi;
+    const result = yield* api.discoverStacks().pipe(Effect.mapError(mapStackError));
+    const stacks = result.stacks.map(entry).sort(compareEntries);
+    const firstError = result.errors[0];
+    if (firstError !== undefined) return yield* mapStackError(firstError.error);
```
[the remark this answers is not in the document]
jgoux: Addressed in 5a4ba5d05. The inventory now includes readable stacks and every unreadable entry. Unreadable entries carry their registry ID, readable: false, and an error code/message; readable entries carry readable: true and their persisted metadata. Per-entry corruption no longer fails listing. Registry-root discovery errors remain fatal. Mixed and unreadable-only registries are covered, including real-registry binary checks in text, JSON, and stream-JSON.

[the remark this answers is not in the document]
jgoux: Addressed in 5a4ba5d05. Every discovery issue is now retained as its own unreadable stack entry, ordered by ID, with its diagnostic available in text and structured output. The first-error-only path has been removed.

apps/cli/src/commands/experimental/stack/list/SIDE_EFFECTS.md
```diff
@@ -0,0 +1,23 @@
...
+
+Entries are sorted by project root, stack name, and id. Text output includes the
+identity, project root, branch context, runtime, and desired lifecycle. Structured
+output returns the same fields under `stacks`. The legacy `-o/--output` flag is
+rejected; use `--output-format json`. Discovery is fail-whole-list when a
+persisted entry is corrupt or unsupported.
+
+The command reads `${SUPABASE_HOME ?? $HOME/.supabase}/managed/stacks/<stack-id>/`.
+Exit code 0 indicates success; exit code 1 indicates a registry read error or
+rejected legacy output flag.
```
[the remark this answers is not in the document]
jgoux: Addressed in 5a4ba5d05. The contract now documents shared profile/workdir reads and failures, telemetry.json persistence, command-event capture and consent behavior, plus the exhaustive inventory output and exit semantics. Verified that an explicit missing profile fails before listing and telemetry state is written even when event delivery is disabled.

apps/cli/src/commands/experimental/stack/list/list.command.ts
```diff
@@ -0,0 +1,12 @@
+import { Command } from "effect/unstable/cli";
+import { withJsonErrorHandling } from "../../../../shared/output/json-error-handling.ts";
+import { withCommandTelemetry } from "../../../../telemetry/command-telemetry.ts";
+import { stackList } from "./list.handler.ts";
+
+export const stackListCommand = Command.make("list").pipe(
+  Command.withDescription("List persisted managed local Supabase stacks."),
+  Command.withShortDescription("List managed local stacks"),
```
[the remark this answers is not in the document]
jgoux: Leaving help examples as optional polish for this update. This command has no local flags; global help documents output selection, and the stack command guide includes the invocation and structured-output options. This does not require a behavior change.

apps/cli/src/commands/experimental/stack/list/list.integration.test.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  it.live("rejects the legacy output flag", () => {
+    const fixture = setup({ outputFlag: true });
+    return stackList().pipe(
+      Effect.flip,
+      Effect.tap((error) =>
+        Effect.sync(() => {
+          expect(error).toBeInstanceOf(StackCommandListError);
+          expect(error.reason).toBe("flags");
+          expect(fixture.telemetry.flushed).toBe(true);
```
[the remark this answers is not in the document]
jgoux: Added a command-level integration case in 5a4ba5d05 that pins the canonical stack list telemetry identity. Failure and telemetry-flush behavior remain covered. Direct assertions of the fixed actionability getter mapping would repeat the implementation, so those were not added.

apps/cli/src/commands/experimental/stack/list/list.command.ts
```diff
@@ -0,0 +1,12 @@
...
+  Command.withHandler(() =>
+    stackList().pipe(withCommandTelemetry({ flags: {} }), withJsonErrorHandling),
```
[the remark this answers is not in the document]
jgoux: Keeping the existing form. Both absent flags and an empty object produce an empty handler-flag map and the same global-flag fallback in the telemetry wrapper. There is no behavioral difference to address.

apps/cli/src/commands/experimental/stack/list/list.errors.ts
Coly010: **Rename to `ExperimentalStackListError`.** The retain-from-#6509 rationale does not hold: #6509 was closed unmerged, so no PostHog fingerprint exists to preserve. #6525 (merged 2026-09-08) removed the `Legacy` prefix repo-wide, and this is now the only `Legacy*` tag left in `error-tags.txt`.

It is also user-visible, not just telemetry. A registry-root failure emits this in `--output-format json`:

```json
{"_tag":"Error","error":{"code":"LegacyExperimentalStackListError","message":"BadResource: FileSystem.readDirectory (...)","suggestion":"..."}}
```

The AGENTS.md preserve-on-rename rule protects shipped fingerprints. This one has not shipped, so now is the free moment.

apps/cli/src/commands/experimental/stack/list/list.handler.ts
```diff
@@ -0,0 +1,108 @@
...
+    ...(index === 0 ? [] : [""]),
+    ...(stack.readable
+      ? [
+          `${stack.name} (${stack.id})`,
+          `  Project: ${stack.project_root}`,
+          `  Branch: ${stack.branch_context}`,
+          `  Runtime: ${renderRuntime(stack.runtime)}`,
+          `  Desired lifecycle: ${stack.desired_lifecycle}`,
+        ]
+      : [`Unreadable stack (${stack.id})`, `  Error: ${stack.error.message}`]),
```
Coly010: **The ID is printed twice for unreadable entries.** `discoverStacks` already enriches every per-entry error message to `Failed to read managed stack <id>: …` (see `enrichStackDiscoveryError`), so the real text output is:

```
Unreadable stack (cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc)
  Error: Failed to read managed stack cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc: Unable to parse state document: SchemaError(Expected a valid JSON string)
```

Either render the underlying cause on the `Error:` line (the enriched errors carry `cause`), or drop the ID from the header. The integration fixtures use bare messages like `"corrupt state"`, which is why this did not show up in tests.

apps/cli/src/commands/experimental/stack/list/list.integration.test.ts
```diff
@@ -0,0 +1,238 @@
...
+      BunServices.layer,
+    ),
+  };
+};
+
+describe("stack list", () => {
+  it.live("sorts persisted stacks and includes stopped and unconfigured lifecycles", () => {
+    const fixture = setup({
+      stacks: [
+        descriptor("e", "/work/z", "zeta", "stopped"),
```
Coly010: **The `project_root` sort key is untested.** Every conflicting fixture shares `/work/a`, and `/work/z`/`zeta` already sorts last by name alone, so deleting the project comparison in `compareEntries` leaves this test green. Add something like `descriptor("d", "/work/z", "aaa", "stopped")` and assert it renders after the `/work/a` entries.

apps/cli/src/commands/experimental/stack/list/list.integration.test.ts
```diff
@@ -0,0 +1,238 @@
...
+  });
+
+  it.live("renders every healthy stack and every discovery issue", () => {
+    const healthy = descriptor("a", "/work", "main", "running");
+    const fixture = setup({
+      stacks: [healthy],
+      errors: [
+        {
+          id: StackIdSchema.make("d".repeat(64)),
+          error: new StackStateInvalidError({ message: "corrupt state" }),
```
Coly010: **Fixtures do not match real discovery errors.** Real `errors[].error.message` always has the enriched `Failed to read managed stack <id>: …` shape, but these fixtures (and the ones at L142 and L167) use bare strings, so the tests cannot observe production text. That is how the duplicated ID in `render` slipped through.

Cheapest fix: build the fixtures with the enriched shape. A single test against the real `stackApiLayer` with a temp `SUPABASE_HOME` (seed via `createStack`, then hand-write one corrupt `state.json`) would cover it more honestly.

apps/cli/src/commands/experimental/stack/list/list.integration.test.ts
```diff
@@ -0,0 +1,238 @@
...
+          expect(fixture.output.stdoutText).not.toContain("Project:");
+          expect(fixture.output.stdoutText).not.toContain("No managed stacks found.");
+        }),
+      ),
+      Effect.provide(fixture.layer),
+    );
+  });
+
+  it.live("emits the complete discriminated inventory in json and stream-json", () =>
+    Effect.forEach(["json", "stream-json"] as const, (outputFormat) => {
```
Coly010: **This loop asserts the same thing twice.** `mockOutput.success` pushes to `messages` regardless of `format`, so the `stream-json` iteration is identical to the `json` one, and the "one `result` NDJSON event" claim in SIDE_EFFECTS.md stays unverified. Either assert on real output events for stream-json or drop the parametrization.

apps/cli/src/commands/experimental/stack/list/list.handler.ts
```diff
@@ -0,0 +1,108 @@
...
+  });
+
+const renderRuntime = (runtime: StackRuntime): string =>
+  Match.value(runtime).pipe(
+    Match.when({ kind: "native" }, () => "native"),
+    Match.when({ kind: "container" }, ({ engine }) => `container (${engine})`),
+    Match.exhaustive,
+  );
+
+const render = (stacks: ReadonlyArray<StackEntry>): string => {
```
Coly010, line 66: **Non-blocking, before the gate comes off:** every other list command (`projects`, `branches`, `functions list`, `migration list`, `backups list`, the experimental `compute list`, …) renders through `renderGlamourTable`. This is the only one using vertical detail blocks, which `compute.format.ts` explicitly reserves for single-item output. With 64-char IDs that is six lines per stack. A `NAME | PROJECT | BRANCH | RUNTIME | DESIRED | ID` table would match the family; full IDs stay in JSON.

apps/cli/src/commands/experimental/stack/list/SIDE_EFFECTS.md
```diff
@@ -0,0 +1,101 @@
...
+| Value of `--profile`, `SUPABASE_PROFILE`, or persisted profile        | YAML profile    | By shared `CommandSettings` when the selected profile is not a built-in name; an unreadable, malformed, or invalid explicit profile fails the command.                                                                        |
+| `<workdir>/supabase/config.toml`                                      | existence probe | By shared `CommandSettings` while resolving a default workdir from its ancestor directories. The list handler does not perform a full project config load.                                                                    |
+| `<SUPABASE_HOME or ~/.supabase>/managed/stacks/`                      | directory       | Registry enumeration. Entries that are not stack IDs are ignored.                                                                                                                                                             |
```
Coly010: **Non-blocking:** `discoverStacks` also skips ID-shaped directories whose `state.json` is missing (`isMissingStateRemnantError` → `Effect.void`), so a half-destroyed stack appears in neither array. Verified with an empty `<id>/` directory. Worth a row here, since the contract currently only says non-ID entries are ignored and decode failures become unreadable entries.

apps/cli/docs/supabase-home.md
```diff
@@ -106,8 +106,10 @@ supervisor and are not written to the document.
   report only a live owner as running.
 - `supabase stop` asks the owner to stop, waits for `stopped` and control release, and keeps the
   document by default. `--no-backup` removes it after deterministic cleanup.
-- `supabase stack list` enumerates healthy managed documents and checks control ownership for live
-  status; `--project-dir` scopes the result to one checkout.
+- `supabase stack list` enumerates all persisted managed stacks, including stopped and unconfigured
```
Coly010: **Non-blocking, please confirm intentional:** the previous text documented checkout scoping (`--project-dir`) and live-owner status; this drops both. `discoverStacks` already accepts `{ projectRoot }`, so scoping is a one-flag addition. Without owner contact, `Desired lifecycle: running` read from disk still says `running` after a reboot. A hint pointing at `stack status` for live state would head off the "list says running, status says stopped" question.

Coly010 (changes requested):
Ran the command from source against a seeded registry (three readable stacks, one corrupt `state.json`, one ID-shaped directory with no state file, one non-ID directory), an empty registry, and a registry root that is a file. Behavior matches the SIDE_EFFECTS contract, and the touched test files pass locally.

Requesting changes for three things: the `Legacy` error tag (it never shipped, so there is nothing to preserve, and it is user-visible in JSON errors), a duplicated ID in the unreadable-entry text output, and test-fidelity gaps that let that slip through. Three further inline notes are non-blocking and can land before the experimental gate comes off.

apps/cli/src/commands/experimental/stack/list/list.errors.ts
jgoux: Addressed in 64b16a346. Agreed; the earlier retention rationale was incorrect because #6509 never shipped. Renamed the tag to ExperimentalStackListError, updated the telemetry fixture, and removed that rationale from the PR description. The tag stability tests and binary JSON-error check pass.

apps/cli/src/commands/experimental/stack/list/list.handler.ts
jgoux: Addressed in 64b16a346. Fixed. Unreadable entries now appear in a diagnostic section using the enriched discovery message, with no additional ID prefix. Real-registry integration and binary smoke checks assert each unreadable full ID appears once.

apps/cli/src/commands/experimental/stack/list/list.integration.test.ts
jgoux: Addressed in 64b16a346. Added a /work/z stack named aaa, opposing the /work/a names, and asserted the complete project/name/ID ordering plus text ordering. Removing the project comparison now changes the expected result.

apps/cli/src/commands/experimental/stack/list/list.integration.test.ts
jgoux: Addressed in 64b16a346. Replaced the mock discovery-error cases with a real stackApiLayer integration flow. It seeds a healthy stack via createStack under an isolated SUPABASE_HOME, then writes corrupt and unsupported state documents and adds missing-state and non-ID directories. Assertions cover actual diagnostics, exhaustive inventory, skipped remnants, and unreadable-only output.

apps/cli/src/commands/experimental/stack/list/list.integration.test.ts
jgoux: Addressed in 64b16a346. Replaced the duplicated mock-output loop with the production JSON and stream-JSON Output layers over captured Stdio. The test parses actual output and checks one NDJSON result event, its complete payload, full IDs, and empty stderr.

apps/cli/src/commands/experimental/stack/list/list.handler.ts
jgoux: Addressed in 64b16a346. Addressed now: readable stacks use renderGlamourTable with NAME, PROJECT, BRANCH, RUNTIME, DESIRED, and ID columns. Text IDs use eight-character prefixes; structured output preserves full IDs. Unreadable entries retain their full diagnostic below the table.

apps/cli/src/commands/experimental/stack/list/SIDE_EFFECTS.md
jgoux: Addressed in 64b16a346. Documented missing state files as ignored remnants in SIDE_EFFECTS.md and the command guides. The real-registry integration test includes an empty ID-shaped directory and verifies it is absent from the inventory.

apps/cli/docs/supabase-home.md
jgoux: Addressed in 64b16a346. Confirmed intentional: this command is a global persisted inventory without checkout scoping or owner contact. Added explicit guidance that desired running does not prove a live owner exists, and a pointer to supabase stack status for live state. No new scoping flag was added.

Coly010 (approved):

PR #6593, opened 2026-09-14
feat(cli): add stack logs

jgoux:
Adds `supabase stack logs` behind the `experimental.stack` feature flag. Read logs for the current project or select a stack with `--stack` or `--stack-id`, filter by `--service`, limit retained history with `--tail`, and follow new entries with `--follow` / `-f`.

Supports text, bounded JSON, and streaming JSON output. Interrupting a log follower leaves the stack running, and retained logs remain readable after the stack stops.

Supersedes #6510.


apps/cli/src/commands/experimental/stack/logs/logs.handler.ts
```diff
@@ -0,0 +1,159 @@
...
+  );
+  return new StackCommandLogsError({
+    ...classification,
+    message: error.message,
+    cause: error,
+  });
+};
+
+const renderEntry = (entry: StackLogEntry) =>
+  `${entry.timestamp} ${entry.source}/${entry.stream}: ${entry.message}\n`;
```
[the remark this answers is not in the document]
jgoux: Addressed in becc4871e. Stack text output now uses the shared Compute control-sequence sanitizer. Coverage exercises OSC, CSI, carriage returns, and C0 controls in retained and followed text, while JSON and stream-json preserve the original messages.

apps/cli/src/commands/experimental/stack/logs/logs.handler.ts
```diff
@@ -0,0 +1,159 @@
...
+      cursor: batch.cursor,
+    };
+    yield* stack.followLogs(followQuery).pipe(
+      Stream.mapError(logsError),
+      Stream.runForEach((entry) =>
+        output.format === "stream-json"
+          ? output.event(eventForEntry(entry, "live"))
+          : output.raw(renderEntry(entry)),
+      ),
+    );
```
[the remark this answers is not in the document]
jgoux: Confirmed as a shared, pre-existing telemetry limitation: external fiber interruption runs finalizers but skips the wrapper's trailing cli_command_executed capture. Per the agreed scope, this is deferred to a shared telemetry follow-up; adding a stack-specific signal path would leave the underlying issue elsewhere. The command contract now documents the limitation. SIGINT still exits 130 and leaves the managed stack running.

apps/cli/src/commands/experimental/stack/logs/SIDE_EFFECTS.md
```diff
@@ -0,0 +1,34 @@
...
+This command reads retained logs from the managed stack identified by the current project,
+an optional `--stack` name, or `--stack-id`. It calls the public `@supabase/stack` logs
+and followLogs APIs without starting the stack, stopping it, or changing its owner lifecycle.
+
+## Files read and written
+
+The stack package reads its normal durable state under `<SUPABASE_HOME or ~/.supabase>` and
+the selected stack's persisted log state. The CLI reads the current workdir and experimental
+stack feature setting through its normal settings resolution. This command writes no project
+files, stack state, credentials, or runtime resources.
```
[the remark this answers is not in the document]
jgoux: Addressed in becc4871e. The side-effect contract now specifies feature gating, telemetry.json, the text format and sanitization, bounded JSON shapes, and the stream-json event/result envelope. It also distinguishes state flushing from the command event that interruption may skip.

apps/cli/src/commands/experimental/stack/logs/logs.handler.ts
```diff
@@ -0,0 +1,159 @@
...
+              ...(Option.isSome(flags.stack) ? { name: flags.stack.value } : {}),
+            })
+            .pipe(Effect.mapError(logsError))
+        : Option.some({ id });
+    if (Option.isNone(targetOption)) {
+      if (Option.isSome(flags.stack))
+        return yield* new StackCommandLogsError({
+          reason: "flags",
+          message: `No managed stack named "${flags.stack.value}" was found for this project.`,
+        });
```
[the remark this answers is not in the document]
jgoux: Addressed in becc4871e. A missing named stack now suggests choosing an existing --stack name or omitting --stack for the current project, matching stack stop.

apps/cli/src/commands/experimental/stack/logs/logs.command.ts
```diff
@@ -0,0 +1,63 @@
...
+      (value) => value >= 0 && value <= MAX_TAIL,
+      (value) => `Expected --tail between 0 and ${MAX_TAIL}, got ${value}`,
+    ),
+    Flag.withDescription(
+      "Number of retained log entries to print. Use 0 with --follow to skip retained history.",
+    ),
+    Flag.withDefault(100),
+  ),
+  follow: Flag.boolean("follow").pipe(
+    Flag.withAlias("f"),
```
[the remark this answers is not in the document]
jgoux: Already fixed in 3b252ce41. The telemetry scanner determines known aliases' value consumption from their canonical flag, so boolean -f no longer consumes --service or --tail. The regression was verified failing on the original scanner and passing with the fix; existing value-taking aliases remain covered.

apps/cli/src/commands/experimental/stack/logs/logs.integration.test.ts
```diff
@@ -0,0 +1,495 @@
...
+import { OutputFlag } from "../../../../command-internal/global-flags.ts";
+import {
+  ErrorActionabilityId,
+  actionability,
+} from "../../../../shared/telemetry/error-actionability.ts";
+import { StackApi } from "../stack.shared.ts";
+import { stackLogs } from "./logs.handler.ts";
+import { StackCommandLogsError } from "./logs.errors.ts";
+import { stackLogsCommand } from "./logs.command.ts";
+import { mockTelemetryStateLayer } from "../../../../../tests/helpers/command-mocks.ts";
```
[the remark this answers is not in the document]
jgoux: Addressed in becc4871e. Both command-mocks helpers now share one import declaration.

jgoux:
The additional documentation finding in the review summary is also addressed in becc4871e: the maintained stack command reference now includes stack logs, target selection, history/follow behavior, and structured output examples. All inline findings have been handled; the pre-existing shared interruption telemetry limitation is deferred as explained in its thread.

apps/cli/src/commands/experimental/stack/logs/logs.handler.ts
```diff
@@ -0,0 +1,163 @@
...
+      tail: flags.tail,
+    };
+    const batch = yield* stack.logs(query).pipe(Effect.mapError(logsError));
+    const emitEntries = (source: "history" | "live", entries: ReadonlyArray<StackLogEntry>) =>
+      output.format === "stream-json"
+        ? Effect.forEach(entries, (entry) => output.event(eventForEntry(entry, source)), {
+            discard: true,
+          })
+        : Effect.forEach(entries, (entry) => output.raw(renderEntry(entry)), { discard: true });
+    if (!flags.follow) {
```
Coly010, line 143: Bounded stream-json read has no terminal `result`. logs.handler.ts:139 branches on `json` only, so stream-json emits `log-entry` events and nothing else, while the not-found path at line 121 does emit a `result`. A quiet found stack prints zero bytes and exits 0. `compute logs` does the opposite and emits one `result` for every bounded machine read. Pick one convention for both logs commands.

apps/cli/src/commands/experimental/stack/logs/logs.command.ts
```diff
@@ -0,0 +1,63 @@
+import { Command, Flag } from "effect/unstable/cli";
+import type * as CliCommand from "effect/unstable/cli/Command";
+import { CAPABILITY_NAMES } from "@supabase/stack/effect";
+import { withJsonErrorHandling } from "../../../../shared/output/json-error-handling.ts";
+import { withCommandTelemetry } from "../../../../telemetry/command-telemetry.ts";
+import { stackLogs } from "./logs.handler.ts";
+
+const MAX_TAIL = 10_000;
```
Coly010: `--tail` ceiling is ten times what retention can return. MAX_TAIL is 10,000 but the log store keeps 1,000 entries or 1 MiB and production never overrides that. `compute logs` caps at 1,000.

apps/cli/src/commands/experimental/stack/logs/logs.handler.ts
```diff
@@ -0,0 +1,163 @@
...
+    message: error.message,
+    ...(error.suggestion === undefined ? {} : { suggestion: error.suggestion }),
+    cause: error,
+  });
+
+const logsError = (
+  error: StackDiscoveryError | OpenStackError | ApiStackLogsError,
+): StackCommandLogsError => {
+  const classification = Match.value(error).pipe(
+    Match.tag("StackNotFoundError", "InvalidStackIdentityError", () => ({
```
Coly010: Unknown `--stack-id` gets a suggestion-less error. Reproduced live: "Stack state was not found" plus the generic `--debug` trailer. `stack status` gives a real suggestion for the same error, and the `--stack <name>` path in this handler already does too.

apps/cli/src/shared/output/strip-control-sequences.ts
```diff
@@ -0,0 +1,17 @@
+/* oxlint-disable no-control-regex -- these patterns intentionally match terminal control bytes */
+const OSC_SEQUENCE = /\u001b\][^\u0007\u001b]*(?:\u0007|\u001b\\)?/gu;
+const CSI_SEQUENCE = /\u001b\[[0-?]*[ -/]*[@-~]/gu;
+const ESCAPE_SEQUENCE = /\u001b[@-Z\\-_]/gu;
+const CRLF_PAIR = /\u000d\u000a/gu;
+const C0_CONTROLS = /[\u0000-\u0008\u000b-\u001f\u007f]/gu;
```
Coly010, line 6: The hoisted sanitizer ignores 8-bit C1 controls. I confirmed against the shipped function that U+009B CSI, U+009D OSC and U+0085 NEL pass through untouched, and the test's `not.toContain("\u001b")` passes while the bypass is live. Pre-existing in compute logs, but this PR makes it shared infra. Three regex lines and one assertion fix it. Only terminals honouring UTF-8 C1 are affected, so this is defence in depth rather than a broad exploit.

apps/cli/src/commands/experimental/stack/logs/logs.command.ts
```diff
@@ -0,0 +1,63 @@
...
+    Flag.withDescription(
+      "Select an existing stack by name (defaults to the current project stack).",
+    ),
+    Flag.optional,
+  ),
+  stackId: Flag.string("stack-id").pipe(
+    Flag.withDescription("Read logs from an existing stack by id."),
+    Flag.optional,
+  ),
+  service: Flag.choice("service", CAPABILITY_NAMES).pipe(
```
Coly010, line 21: `--service` hides the line you're looking for. The package drops supervisor and gateway entries whenever a capability filter is set, so `--service database` filters out "database capability failed to start". It's also single-valued although the package accepts an array. Needs a decision or at least documentation.

apps/cli/src/commands/experimental/stack/logs/logs.command.ts
```diff
@@ -0,0 +1,63 @@
...
+      description: "Print the latest database logs",
+    },
+    {
+      command: "supabase stack logs --follow --output-format stream-json",
+      description: "Stream new stack logs as structured events",
+    },
+  ]),
+  Command.withHandler((flags) =>
+    stackLogs(flags).pipe(
+      withCommandTelemetry({ flags, config, aliases: { f: "follow" } }),
```
Coly010, line 61: `compute logs` still has the `-f` telemetry bug this PR fixed for stack logs. One-line `aliases: { f: "follow" }`.

Coly010 (changes requested):
Should fix before merge

apps/cli/src/commands/experimental/stack/logs/logs.handler.ts
jgoux: Addressed in 00328c46a. Both logs commands now use one result for bounded machine reads. Stack stream-json returns the same bounded payload in a result envelope, including a found stack with no entries; follow mode retains history/live log-entry events. Updated contract documentation and coverage.

apps/cli/src/commands/experimental/stack/logs/logs.command.ts
jgoux: Addressed in 00328c46a. The CLI now accepts --tail from 0 to 1,000, matching production retention and Compute. Documentation explains that the 1 MiB retention limit can reduce available history further; boundary coverage checks 1,000 and 1,001.

apps/cli/src/commands/experimental/stack/logs/logs.handler.ts
jgoux: Addressed in 00328c46a. Missing stacks now suggest choosing an existing stack with --stack or --stack-id, or omitting both for the current project. This also covers a stack disappearing after name/default resolution, without suggesting creation of a different stack.

apps/cli/src/shared/output/strip-control-sequences.ts
jgoux: Addressed in 00328c46a. The shared text sanitizer now handles C1 CSI/OSC sequences and remaining C1 controls, including NEL. Coverage verifies the resulting text and preservation of original messages in structured output.

apps/cli/src/commands/experimental/stack/logs/logs.command.ts
jgoux: Addressed through help and documentation in 00328c46a. --service remains a single workload-service filter and excludes supervisor/gateway entries, including startup diagnostics; omit it to see all sources. We are keeping multi-service selection outside this command scope.

apps/cli/src/commands/experimental/stack/logs/logs.command.ts
jgoux: Addressed in 00328c46a. Compute logs now declares f as the follow telemetry alias. The regression runs the real command with isolated credentials/project/home and verifies canonical follow and kind flags. Removing the alias makes it fail; restoring it passes.

Coly010 (approved):

PR #6594, opened 2026-09-14
fix(cli): report functions serve outcomes from real container state (CLI-2426)

Prashansa-K:
## TL;DR

`supabase functions serve` reported almost every session as a failure, and almost none of those failures carried a usable classification. This makes the command report what actually happened.

Refs [CLI-2426](https://linear.app/supabase/issue/CLI-2426)

## What was wrong

**Outcomes that were not failures were reported as failures.** A container that stopped on its own with exit `0` was returned as `Effect.fail`. A container torn down by a supervisor — `supabase stop` in another terminal stops every project container — exited `143` and was charged to the internal-bug counter. Both are ordinary ways for a serve session to end.

**Decisions were made from the wrong source.** `streamContainerLogs` judged whether the Docker daemon was reachable, and whether a stream failure was retriable, by substring-matching the tail of the container's own logs. That text is arbitrary user function output: a function logging "Cannot connect to the Docker daemon" earned a Docker suggestion, and one logging "no such container" could turn a fatal stream failure into an endless retry. A private `inspectContainerState` shadowed an existing shared helper, parsed a different wire format, and failed with bare `Error`s.

**A clean `docker logs -f` exit was read as a stopped container.** The code checked only `.State.ExitCode`, which is also `0` while a container runs, so a daemon that closed the stream underneath a healthy runtime got it torn down and reported as a graceful stop.

**The check that runs on every invocation carried no classification.** `assertLocalDbRunning` failed with bare `Error`s. Those have no tag and no actionability declaration, so they fall through to `unknown` — which is most of what this command's failures looked like in telemetry.

**The shutdown race is real, but only on Windows.** The Effect spawner defaults to `detached` off Windows and `spawnContainerCli` never overrides it, so the `docker logs -f` child runs in its own session and never receives the terminal's SIGINT — the two arms of the race cannot become ready together there. On Windows the default flips and `CTRL_C_EVENT` reaches every console-attached process, so the tie-break does matter. An earlier revision of this description had that backwards.

## What changed

**Container state decides the outcome.** `streamContainerLogs` inspects `Running` and `ExitCode` together through the shared `inspectContainerState`, and re-attaches when the container is still alive. A container that has already been removed ends the session instead of surfacing raw Docker text.

**Ordinary endings end ordinarily.** Exit `0` and the supervisor-teardown codes (`129`/`130`/`131`/`143`) end the session successfully, each with its own message so scrollback can tell "I stopped it" from "the runtime walked out". Real crash signals still fail as `runtime_crash`. `137` keeps its existing retry.

**Classification comes from Docker's own output.** `daemonDown` and the retry/fatal decision are derived from a follow-up `docker container inspect`, never from the log tail. No container log text is emitted to telemetry.

**The failure paths carry tagged identities** — including the local-database pre-flight, which is the check every invocation runs: a missing stack suggests `supabase start`, an unreachable daemon suggests starting Docker. A daemon that dies mid-session now says "start Docker" rather than "install Docker Desktop", which is what it said before despite having just created and started a container.

**Re-attaches are bounded.** They resume with `--since` instead of replaying the container's whole log history, are capped, and run in their own scope so child handles stop accumulating for the session.

**Shutdown is latched once for the loop's lifetime**, so a signal arriving between races is not dropped, and the grace period guards both the startup race and the log-stream race — a cold image pull is where a Windows user is most likely to reach for Ctrl-C. Both timers are injectable, so their tests do not race a real clock.

**New `runtime_crash` category.** A runtime we launch and supervise can die on its own; `panic` (the CLI threw) and `impossible_state` (unreachable branch) don't describe that, so these were filed as `unknown`. Note this makes the internal-bug counter-metric go **up**. That reading is worse and correct: those failures were always ours, just mis-filed.

## Known gaps, deliberately not here

- **Exit `137` is still swallowed** by a retry loop. The container has no `--restart` policy. Filed separately.
- **`gen types` has the same untagged local-stack pre-flight failure.** Filed separately.
- **`time_to_ready_ms`** for `functions serve` — next PR.


apps/cli/src/shared/functions/serve.ts
```diff
@@ -1315,19 +1325,33 @@ const streamContainerLogs = Effect.fnUntraced(function* (containerId: string) {
...
         continue;
       }
-      return yield* Effect.fail(new Error(`error running container: exit ${containerExitCode}`));
+      return yield* Effect.fail(
+        new EdgeRuntimeContainerCrashedError({
+          message: `error running container: exit ${containerExitCode}`,
+          containerId,
+          exitCode: containerExitCode,
+        }),
+      );
```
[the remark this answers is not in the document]
Prashansa-K: Fixed in 9103be50f, with a narrower split than suggested.

Routing every `128+N` exit away from `runtime_crash` would have over-corrected: SIGSEGV (139) and SIGABRT (134) are also signal-derived, but they mean the runtime died of its own fault, which is genuinely our bug. So only the supervisor-teardown signals move — 129 SIGHUP, 130 SIGINT, 131 SIGQUIT, 143 SIGTERM — and they reuse the existing `actionability.cancelled` preset rather than adding a category. Crash signals stay `runtime_crash`. 137 is untouched; it is still retried before this error is constructed.

The command still fails on an externally stopped container; only the classification changed.

apps/cli/src/commands/functions/serve/serve.integration.test.ts
```diff
@@ -2037,6 +2045,179 @@ describe("functions serve integration", () => {
...
+        // failing exit — the log-stream failure and the SIGINT delivery are
+        // exactly concurrent, with nothing to break the tie except the
+        // shutdown-race grace period.
+        const childSpawner = mockDockerLogSpawner([
+          {
+            exitCode: 1,
+            stderr: "docker logs killed by signal",
+            onSpawn: () => processControl.signal("SIGINT"),
+          },
+        ]);
```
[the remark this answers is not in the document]
Prashansa-K: Confirmed and fixed. Added `exits cleanly when SIGINT arrives shortly after the docker-logs stream has already failed`, where the mock defers the signal ~15ms instead of offering it from `onSpawn`, so it becomes observable strictly after the failure.

Verified your diagnosis rather than assuming it: with `shutdownSignalGracePeriod` temporarily set to zero, the new test fails and the original Ctrl-C test still passes — so the original did indeed pass for the wrong reason, and the new one genuinely defends the 50ms window. Restored to 50ms; the constant stays module-private, so that negative control is not committed.

Caveat: it is a real-clock `it.live` test with a 15ms delay against a 50ms budget. Stable over 5 consecutive local runs, but a badly starved CI runner could in principle exceed it.

apps/cli/src/shared/functions/serve.ts
```diff
@@ -122,6 +124,9 @@ const ignoredDirNames = new Set([
 ]);
 const dockerLogRetryDelay = Duration.millis(400);
 const dockerLogDiagnosticTailLength = 4_096;
+// Ctrl-C hits this process and the un-detached `docker logs -f` child at once, so a shutdown
```
Coly010: **should-fix** — This premise is inverted. The spawner defaults to `detached: true` off Windows (`@effect/platform-node-shared` `buildSpawnOptions`: `options.detached ?? platform !== "win32"`) and `spawnContainerCli` doesn't override it, so `docker logs -f` runs in its own session and never sees the tty's SIGINT. Verified on Bun 1.3.14: the child's pgid differs from the CLI's. `go-proxy.layer.ts:180-183` documents the same default and sets `detached: false` for exactly this reason.

The only platform where this tie-break does anything is Windows, where `CTRL_C_EVENT` reaches every console-attached process. Reword the comment to that — and see the top-level comment for what the telemetry says the macOS/Linux unknowns actually are.

apps/cli/src/shared/functions/serve.ts
```diff
@@ -1916,9 +1947,16 @@ export const serveFunctions = Effect.fn("functions.serve")(function* (
...
   const loop = Effect.gen(function* () {
+    // Hoisted to the loop's lifetime (not per-race) so a signal arriving
+    // between races isn't dropped while nothing is listening for it.
+    const shutdownRequested = yield* Deferred.make<void>();
+    yield* processControl
+      .awaitSignal()
+      .pipe(Effect.andThen(Deferred.succeed(shutdownRequested, void 0)), Effect.forkScoped);
+
     for (;;) {
       const startOutcome = yield* Effect.raceFirst(
```
Coly010: **should-fix** — If the grace period stays (for Windows), it belongs on this race too. On Windows the `docker pull`/`create`/`cp`/`start` children also receive `CTRL_C_EVENT`; `docker-image-resolve.ts:216-221` turns a failed pull effect into an immediate `failCause`; and this `raceFirst` has no `catch`. Same coin flip, on the phase where a user is most likely to reach for Ctrl-C (a cold image pull). A small `withShutdownGrace(self, shutdownRequested)` helper wrapping both races keeps them consistent.

apps/cli/src/shared/functions/serve.ts
```diff
@@ -1258,10 +1263,15 @@ function appendDiagnosticTail(existing: string, text: string) {
     : combined.slice(combined.length - dockerLogDiagnosticTailLength);
 }
 
-const inspectContainerExitCode = Effect.fnUntraced(function* (containerId: string) {
+interface ContainerState {
+  readonly running: boolean;
+  readonly exitCode: number;
+}
+
+const inspectContainerState = Effect.fnUntraced(function* (containerId: string) {
```
Coly010: **should-fix** — `command-internal/docker-lifecycle.ts:194` already exports an `inspectContainerState(spawner, containerId)`: `docker container inspect --format {{json .State}}`, returns `{ running, status, health }`, fails with the tagged `DockerLifecycleInspectError` (which carries `daemonDown`). This same-named private copy parses a different wire format and fails with bare `Error`s (`:1283`, `:1290`) — i.e. `error:unknown`, on a path this PR makes hotter.

Concrete case: `supabase stop` in another terminal removes the container between the `docker logs` exit and this inspect → `No such object` → red raw text, exit 1, `unknown`. Add `exitCode` to `parseContainerState` there and call it; treat "no such container" as end-of-session rather than a failure.

apps/cli/src/shared/functions/serve.ts
```diff
@@ -1313,21 +1324,41 @@ const streamContainerLogs = Effect.fnUntraced(function* (containerId: string) {
     );
 
     if (exitCode === 0) {
-      const containerExitCode = yield* inspectContainerExitCode(containerId);
-      if (containerExitCode === 0) {
-        return yield* Effect.fail(new Error(`container exited gracefully: ${containerId}`));
+      const containerState = yield* inspectContainerState(containerId);
+      if (containerState.running) {
```
Coly010: **should-fix** — The re-attach re-spawns `docker logs -f --timestamps <id>` (`:1302`) with no `--since`/`--tail`, so every iteration replays the container's entire log history, and nothing bounds the loop — a daemon that keeps EOF'ing the stream produces a 2.5 Hz full-log flood. Each spawn also leaves its handle finalizer in the loop-lifetime scope (`Effect.scoped(loop)` at `:2003`; `streamContainerLogs` has no scope of its own), so handles accumulate for the session.

Pre-existing shape via the 137 branch, but this adds the common-path entry. Suggest `--since <last forwarded timestamp>` on re-attach, a consecutive-reattach cap that fails with a real error (`lost the Edge Runtime log stream N times; container <id> is still running`), and `Effect.scoped` around each attach.

apps/cli/src/shared/functions/serve.errors.ts
```diff
@@ -0,0 +1,65 @@
...
+export class EdgeRuntimeContainerCrashedError extends Data.TaggedError(
+  "EdgeRuntimeContainerCrashedError",
+)<{
+  readonly message: string;
+  readonly containerId: string;
+  readonly exitCode: number;
+}> {
+  get [ErrorActionabilityId](): CliErrorActionabilityDeclaration {
+    if (externalTerminationExitCodes.has(this.exitCode)) {
+      return { ...actionability.cancelled, fingerprint_suffix: "cancelled" };
```
Coly010: **should-fix (decision)** — 129/130/131/143 are now `user_cancelled` for telemetry, but the command still fails. `supabase stop` in another tab — a first-party workflow, `docker-remove-all.ts:130-146` stops every project container — prints `error running container: exit 143` in red, exit 1, plus the generic `--debug` hint that `output.layer.ts:380-387` falls back to because there's no `suggestion` on this error. Meanwhile exit 0 gets `Stopped serving` + exit 0.

Either (a) route these codes through the same `return` as exit 0 and print a distinct `Edge Runtime container stopped (exit 143)` — nothing to telemeter — or (b) keep failing but don't file it under `user_cancelled`, whose other populations are fiber interrupts and prompt declines and which is excluded from failure KPIs. I lean (a). Either way `containerId` is already on the error; use it in the message.

apps/cli/src/shared/functions/serve.errors.ts
```diff
@@ -0,0 +1,65 @@
...
+  get [ErrorActionabilityId](): CliErrorActionabilityDeclaration {
+    if (this.daemonDown) {
+      return { ...actionability.dockerNotRunning, fingerprint_suffix: "docker_not_running" };
+    }
+    return actionability.unknown;
+  }
+
+  /** Keeps the user-visible remediation in sync with the `dockerNotRunning` actionability above. */
+  get suggestion(): string | undefined {
+    return this.daemonDown ? SUGGEST_DOCKER_INSTALL : undefined;
```
Coly010, line 105: **should-fix** — `SUGGEST_DOCKER_INSTALL` reads *"Docker Desktop is a prerequisite for local development. Follow the official docs to install…"*. On this path Docker was running a moment ago: it created and started the container. "Install" is the wrong remediation for a daemon that died mid-session, and a missing binary can't reach here (that fails at spawn as `ContainerRuntimeNotFoundError`).

It's the only Docker constant in `docker-suggest.ts`. Add a `SUGGEST_DOCKER_START` (*"Docker is no longer reachable. Start Docker, then rerun `supabase functions serve`."*) and leave the install text on the pre-flight check where a missing binary is actually possible.

apps/cli/src/shared/functions/serve.ts
```diff
@@ -1313,21 +1324,41 @@ const streamContainerLogs = Effect.fnUntraced(function* (containerId: string) {
...
+        // `docker logs -f` exiting 0 doesn't mean the container stopped —
+        // the daemon can close the stream while it keeps running; re-attach.
+        yield* Effect.sleep(dockerLogRetryDelay);
+        continue;
       }
-      if (containerExitCode === 137) {
+      if (containerState.exitCode === 0) {
```
Coly010: **should-fix** — This outcome and Ctrl-C both print the identical `Stopped serving supabase/functions`, so scrollback can't distinguish "I stopped it" from "the runtime walked out". Exit 0 is defensible (compose precedent), but in a `functions serve &` CI step a runtime that exits 0 on its own now looks green and the test step fails later with connection-refused.

Cheap fix: a distinct line (`Edge Runtime exited (code 0). Stopped serving …`) and a SIDE_EFFECTS note about the CI implication.

apps/cli/src/commands/functions/serve/SIDE_EFFECTS.md
```diff
@@ -64,6 +64,7 @@ back to local keys. No scheme/host validation is performed on the discovered URL
 | Code | Condition                                                                                                                                                                          |
 | ---- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `0`  | clean shutdown after `SIGINT`, `SIGTERM`, or stdin close                                                                                                                           |
```
Coly010: **should-fix (docs)** — Pre-existing, but this PR rewrites this table so it should be true when it merges: stdin close is only wired in `ProcessControl.awaitShutdown` (`process-control.layer.ts:42-64`); serve uses `awaitSignal()` (`serve.ts:1954`) and `selfManagedSignalProgram` (`run.ts:674-680`) adds nothing. Either latch stdin close into the same `Deferred` or drop the claim.

Two more while you're here: line 118 says a crash "terminates the command with a non-zero exit", but a 137 exit never terminates — the retry loop at `serve.ts:1339` spins until CLI-2427 lands; and the grace period is itself an exit-code rule (a stream/container failure within 50 ms of a signal now reports `0`), which belongs in this table.

apps/cli/src/commands/functions/serve/serve.integration.test.ts
```diff
@@ -2037,6 +2045,295 @@ describe("functions serve integration", () => {
...
+      await writeFunctionFile("hello", "index.ts", 'Deno.serve(() => new Response("hello"))\n');
+      await writeFunctionFile("hello", "deno.json", '{"imports":{}}\n');
+    }
+
+    it.live(
+      "exits cleanly on SIGINT even when the docker-logs stream fails around the same time",
+      () => {
+        deployMockState.runHandler = baseDockerRunHandler();
+        const processControl = mockQueuedProcessControl();
+        // The same Ctrl-C reaches both the CLI and the un-detached
```
Coly010: **should-fix** — Same inverted premise as `serve.ts:127`: the real child is detached on macOS/Linux, so "reaches both" only describes Windows. Keep the test — it pins the tie-break logic — but say what it models.

apps/cli/src/shared/functions/serve.ts
```diff
@@ -1313,21 +1324,41 @@ const streamContainerLogs = Effect.fnUntraced(function* (containerId: string) {
...
     const trimmedStderr = stderrText.trim();
     if (!isRetriableDockerLogsError(trimmedStderr)) {
       return yield* Effect.fail(
-        new Error(trimmedStderr.length > 0 ? trimmedStderr : `docker logs exited with ${exitCode}`),
+        new DockerLogsStreamError({
+          message: trimmedStderr.length > 0 ? trimmedStderr : `docker logs exited with ${exitCode}`,
+          containerId,
+          exitCode,
+          stderr: trimmedStderr,
+          daemonDown: isDockerDaemonUnreachable(trimmedStderr),
```
Coly010: **nit** — `stderrText` here is the last 4 KB of the *container's* stderr (docker logs demuxes it onto the child's stderr) plus whatever the docker client appended, so `daemonDown` is decided by substring-matching the user's own function output. An edge function that logs "Cannot connect to the Docker daemon" gets a Docker suggestion; one that logs "no such container" flips a fatal stream failure into an infinite retry via `isRetriableDockerLogsError`. Pre-existing for the retriable check; new for `daemonDown`. A cleaner discriminant is the failure of a follow-up inspect, which is docker-only output.

apps/cli/src/commands/functions/serve/serve.integration.test.ts
```diff
@@ -2037,6 +2045,295 @@ describe("functions serve integration", () => {
...
+        const processControl = mockQueuedProcessControl();
+        // Delays SIGINT past the log-stream failure so only the shutdown
+        // grace period, not a same-tick race, can produce a clean exit.
+        const childSpawner = mockDockerLogSpawner([
+          {
+            exitCode: 1,
+            stderr: "docker logs killed by signal",
+            onSpawn: () => {
+              Effect.runFork(
+                Effect.sleep(Duration.millis(15)).pipe(
```
Coly010, line 2197: **nit** — 15 ms signal vs 50 ms grace on a real clock, and the signal path is three scheduler hops (timer → `Queue.offer` → `awaitSignal` fiber → `Deferred.succeed` → `Deferred.await`), each through `setImmediate`. I couldn't make it flake — 8 unloaded runs plus 6 under 2× CPU saturation, all green — so not blocking. But since this is the only test that actually exercises the grace, making `shutdownSignalGracePeriod` injectable and setting it to ~2 s here removes the wall-clock dependency, and the same lever kills the real 400 ms sleep in the re-attach test.

apps/cli/src/shared/functions/serve.ts
```diff
@@ -1916,9 +1947,16 @@ export const serveFunctions = Effect.fn("functions.serve")(function* (
   });
 
   const loop = Effect.gen(function* () {
+    // Hoisted to the loop's lifetime (not per-race) so a signal arriving
+    // between races isn't dropped while nothing is listening for it.
+    const shutdownRequested = yield* Deferred.make<void>();
+    yield* processControl
+      .awaitSignal()
+      .pipe(Effect.andThen(Deferred.succeed(shutdownRequested, void 0)), Effect.forkScoped);
```
Coly010: **nit** — `Effect.forkScoped({ startImmediately: true })`: the `process.once` registration inside `awaitSignal` is deferred until the forked fiber is first scheduled, so there's still a one-turn window with no listener. Free to close.

apps/cli/src/shared/telemetry/error-actionability.ts
```diff
@@ -382,6 +384,19 @@ export const actionability = {
     has_suggestion: true,
     suggestion_type: CliSuggestionType.RerunDebug,
   },
+  /**
+   * A runtime the CLI launched and supervises died on its own — the process
```
Coly010: **nit** — As written ("a runtime the CLI launched and supervises died on its own") this also describes `DbDumpRunError` / `DbPullDumpError` / `MigrationSquashDumpError`, which emit the identical `error running container: exit N` and are correctly `dbConnection`. Narrow it to long-running runtimes whose failure modes we own, as opposed to short-lived tool containers whose exit reflects the user's data or connection.

Coly010 (changes requested):
**Verdict: Request changes.** The mechanical pieces here are good and I want most of them merged — the hoisted `Deferred`, the tagged errors, exit 0 on a clean container exit, `daemonDown`. What I can't approve is the root-cause story: it only holds on Windows, and the telemetry the PR cites points at a different bucket that this PR leaves untouched.

### 1. The `docker logs -f` child is not in the CLI's process group on macOS/Linux

The PR body and `serve.ts:127` say `spawnContainerCli` never passes `detached`, so the child shares the process group and one Ctrl-C hits both arms. The spawner's default is the opposite:

```js
// @effect/platform-node-shared/dist/internal/nodeChildProcessSpawner.js
const detached = options.detached ?? platform !== "win32";
```

Verified empirically on Bun 1.3.14 through `BunServices.layer` with the same options object `streamContainerLogs` passes (`{ stdin: "ignore", stdout: "pipe", stderr: "pipe", extendEnv: true }`):

```
{"via":"effect-spawner-default","childPgid":"5118","parentPgid":"5091","sameGroup":false}
{"via":"effect-spawner-detached-false","childPgid":"5091","parentPgid":"5091","sameGroup":true}
```

A detached child is in its own session, so the tty's SIGINT never reaches it. The race this PR arbitrates with a 50 ms sleep cannot occur on macOS/Linux — and the codebase already knows this: `go-proxy.layer.ts:180-183` sets `detached: false` precisely so Ctrl-C *does* reach the Go child. On Windows the default flips to `detached: false` and `CTRL_C_EVENT` goes to every console-attached process, so there the race is real.

### 2. What the 30-day telemetry actually shows for `functions serve`

`cli_command_executed`, current versions only (2.116.0 + 2.117.0), os × outcome:

| os | outcome | runs | devices | p50 duration | p90 | in a TTY |
|---|---|---|---|---|---|---|
| darwin | success | 41,446 | 4,454 | 41.9 s | 2,105 s | 12% |
| darwin | `error:unknown` | 40,811 | 2,792 | **0.0 s** | 48.5 s | 4% |
| linux | `error:unknown` | 406,963 | 11,059 | **0.1 s** | 0.3 s | 0.2% |
| linux | success | 32,322 | 17,924 | 24.7 s | 504 s | 4% |
| win32 | `error:unknown` | 11,639 | 3,432 | **141.6 s** | 12,394 s | 22% |
| win32 | success | 345 | 111 | 279.8 s | 7,391 s | 99% |

- On macOS/Linux the unknown failures die in well under a second and almost never in a TTY. Those are **startup** failures, not Ctrl-C. The only sub-second path that runs on every invocation and fails with a bare `Error` is `assertLocalDbRunning` — `serve.ts:1388` (`supabase start is not running.`) and `:1400-1401` (`failed to inspect service: …`, including the daemon-down branch). That is most likely the bucket CLI-2426 is actually about, and this PR doesn't tag it. (Inferred from timing — the fingerprint carries no message — but two tagged errors there would settle it in a week of data.)
- The fingerprint a signal-killed `docker logs` child would leave (`tag:PlatformError:platform_error` — `exitCode` fails on `code === null`) appears **34 times** on darwin in 30 days, against ~88k `error:unknown`.
- Windows is the one platform whose unknowns are long-lived sessions, consistent with the race as described. It is ~3% of runs.

### 3. What this means for the PR

Keep the mechanism; fix the explanation and close the gap the data points at:

1. `serve.ts:127`, `serve.integration.test.ts:2086` and the PR body should describe a **Windows console-signal tie-break**. If the 50 ms grace stays for Windows (harmless elsewhere), apply it to the startup race too — a Windows user sits on Ctrl-C during `docker pull`, not during log streaming.
2. Tag the two `assertLocalDbRunning` failures (a `user_actionable` "run `supabase start`" and the existing `dockerNotRunning` preset) — here or in a follow-up opened before this merges. Without them the KPI-1 goal in CLI-2426 doesn't move.
3. The inline items below: the duplicated `inspectContainerState`, the re-attach log replay, the 143/`cancelled` inconsistency, the wrong Docker suggestion text, and the SIDE_EFFECTS drift.

Stress-tested the two timing tests locally: 8 unloaded runs plus 6 under 2× CPU saturation, all green. Nice work on the `Deferred` hoist — that was a real dropped-signal bug.

apps/cli/src/shared/functions/serve.ts
Prashansa-K: Confirmed and reworded. `NodeChildProcessSpawner` does default `detached` to `process.platform !== "win32"`, and `spawnContainerCli` never overrides it, so the child is in its own session on macOS/Linux and the race cannot occur there. The comment now describes the Windows `CTRL_C_EVENT` tie-break, and the PR title and description have been rewritten — they asserted the inverted premise too.

apps/cli/src/shared/functions/serve.ts
Prashansa-K: Done. Extracted `withShutdownGrace(self, shutdownRequested, gracePeriod, onShutdown)` and applied it to both the startup race and the log-stream race. Added an integration test for the startup half specifically, since that was the case you called out; a zero-grace negative control confirms both grace tests fail without the window rather than passing incidentally.

apps/cli/src/shared/functions/serve.ts
Prashansa-K: Fixed. The private copy is gone; `streamContainerLogs` calls the shared `inspectContainerState`, and `parseContainerState` now also returns `exitCode`. All five existing callers only destructure `running`/`status`/`health`, so none needed changes. Your concrete case is covered by a test: a container removed between the log stream ending and the inspect now ends the session instead of printing raw Docker text and exiting 1.

apps/cli/src/shared/functions/serve.ts
Prashansa-K: All three done. Re-attaches pass `--since <last forwarded timestamp>`, are capped at 5 consecutive attempts with no new output, and each attach is wrapped in `Effect.scoped` so handles no longer accumulate for the session. Exceeding the cap fails with a tagged `EdgeRuntimeLogStreamLostError` classified `external_service`/`network`. One wart left in: `--since` is inclusive, so a re-attach can repeat a single already-printed line. That seemed better than adding timestamp-dedup logic for a path that only runs on a dropped stream.

apps/cli/src/shared/functions/serve.errors.ts
Prashansa-K: Took option (a). 129/130/131/143 now end the session successfully with their own message naming the exit code, and nothing is telemetered for them. `EdgeRuntimeContainerCrashedError` lost its `cancelled` branch and always declares `runtime_crash`; real crash signals (132/134/135/136/139) still fail through it, so this doesn't launder genuine crashes into the success path. 137 keeps its retry. Container id is in the crash message.

apps/cli/src/shared/functions/serve.ts
Prashansa-K: Done. The container-exited outcomes now print distinct lines (`Edge Runtime exited (code 0) …`, `Edge Runtime container stopped (exit N) …`, `Edge Runtime container is no longer available …`) rather than reusing the user-initiated `Stopped serving …`. `streamContainerLogs` returns the reason rather than `void` so the loop can render it. SIDE_EFFECTS carries the CI implication you described.

apps/cli/src/commands/functions/serve/SIDE_EFFECTS.md
Prashansa-K: All three addressed. The stdin-close claim was wrong and is dropped: `awaitShutdown` is the only thing that latches stdin, serve uses `awaitSignal()`, and `awaitShutdown` turns out to have no production caller at all — only two test mocks satisfying the interface shape. The 137 line is corrected (retried, not terminating). The grace is now its own exit-code row, and covers both races. The table also gained rows for supervisor teardown and the container-gone case.

apps/cli/src/commands/functions/serve/serve.integration.test.ts
Prashansa-K: Corrected, and the intent moved into the test name rather than the comment. Kept the test — it still pins the tie-break.

apps/cli/src/shared/functions/serve.ts
Prashansa-K: Fixed, and this turned out to be more than a nit in one direction: `daemonDown` now comes from the follow-up inspect's own failure, and so does the retry/fatal decision, so `isRetriableDockerLogsError` is deleted. The "function logs 'no such container' and gets an infinite retry" case you describe is closed along with the `daemonDown` one. Confirmed separately that neither `stderr` nor `message` reaches telemetry — the payload is only the sanitized classification fields.

apps/cli/src/commands/functions/serve/serve.integration.test.ts
Prashansa-K: Made both `shutdownSignalGracePeriod` and `dockerLogRetryDelay` injectable through the existing internal dependencies object, so no new exported knob. The grace test now injects 2s, and the re-attach tests inject a near-zero retry delay — that removed roughly 2.8s of real sleeping from the suite. Production defaults are unchanged.

apps/cli/src/shared/functions/serve.ts
Prashansa-K: Applied. Checked it against the v4 source first: `forkUnsafe` calls `child.evaluate(effect)` synchronously instead of scheduling when `immediate` is set, so it does close the window before `process.once` registers.

apps/cli/src/shared/telemetry/error-actionability.ts
Prashansa-K: Narrowed. Your premise checked out — all three of `DbDumpRunError`, `DbPullDumpError`, and `MigrationSquashDumpError` map to `dbConnection` and emit the same `error running container: exit N`. The doc now scopes `runtimeCrash` to long-running supervised runtimes and names short-lived tool containers as explicitly out of scope.

apps/cli/src/shared/functions/serve.errors.ts
Prashansa-K: Agreed, and added `SUGGEST_DOCKER_START`. Your reasoning checked out: `ContainerRuntimeNotFoundError` is raised at spawn, upstream of anywhere `DockerLogsStreamError` is constructed, so a missing binary genuinely cannot reach this path. Worth noting this also fixes a disagreement — the telemetry declaration already said `suggestion_type: start_docker` while the text on screen said "install". The pre-flight check keeps `SUGGEST_DOCKER_INSTALL`, where a missing binary is possible.

Coly010 (approved):

PR #6596, opened 2026-09-14
fix(api): expand deepObject query parameters into per-key pairs

SaxonF:
Object-valued query parameters in the v2 spec are `style: deepObject`, which puts one `param[key]=value` pair on the wire per entry. The client serialized them as a JSON blob, so a caller passing `page: { size: 100, after: cursor }` reached the server with no page size and no cursor — a paginated walk silently returned the first default-sized page forever, with no error to notice.

`buildRequest` now expands every object-valued query parameter into its own `param[key]=value` pairs. Arrays keep the repeated-key form `normalizeUrlValue` already handles, and entries whose value is `undefined` are dropped rather than sent empty.

## How to try it

```sh
pnpm --filter @supabase/api run test:unit
```

The added cases in `src/internal/client.unit.test.ts` cover expansion, array handling, and dropped entries.

## Notes

- The expansion applies to all object-valued query parameters rather than a per-parameter style recorded in the contract, because every object-valued query parameter in the spec is `deepObject`. If that stops being true, the contract needs to carry the style and this needs to read it.
- Also unblocks `page` / `filter` on `v2-list-organization-members`, `v2-list-organization-projects`, and `v2-list-organization-github-connections`, which the README documented as unusable. The README is updated accordingly.


avallete (approved):
Approve. `page[size]` / `page[after]` works. A pull of 104 notebooks (page size 100) took two list calls and wrote every file.

Nit, not blocking: the deepObject rationale is repeated in `client.ts`, the unit test, and the README. One of those is enough.

PR #6597, opened 2026-09-14
refactor(cli): extract the shared resource output policy

SaxonF:
Pure refactor, no behavior change. Prepares the ground for the `notebooks` command family, which needs exactly the same `-o`/`--output` policy that `compute` already had.

## What changed

`compute.output.ts` had grown a generic output policy for structured resources that have no command-specific encoding:

- which `-o` values answer with a payload (an allowlist, so a format the global flag learns later does not silently start emitting TOML),
- how that payload is encoded (`json` / `yaml` / `toml` through the generic encoders),
- an up-front refusal of `-o env`, whose flat `KEY=value` shape cannot hold a structured payload.

None of that is compute-specific, so it moves to `command-internal/resource-output.ts` as `resourceOutput(EnvNotSupportedError)`, parameterized by the command family's own env-not-supported error so each family keeps its own error tag and actionability. `compute.output.ts` now builds its three exports from it and keeps only `computeRenderFormat` and its retry-suggestion helper.

## Reviewer notes

The extraction dropped the long explanatory comments that lived on the compute versions. Worth deciding as part of review whether the allowlist rationale in particular should be carried onto the shared helper, since it now governs two command families rather than one.


SaxonF:
Closing: dropped from the notebooks stack to keep that work isolated.

Upstream #6554 replaced `experimental/workers` with `experimental/compute` while this stack was in flight, so this extraction would now have to edit a command family that already ships, purely to make room for `notebooks`. That is more review surface and more risk than the feature itself warrants.

Instead, `notebooks.output.ts` in #6598 carries its own copy of the policy, with a comment recording that it duplicates `compute.output.ts` and that the two are expected to agree. `compute` is left exactly as it is on `develop`.

Worth doing as a standalone refactor once a third command family needs the same policy — at that point it can be reviewed on its own rather than inside a feature PR.

PR #6598, opened 2026-09-14
feat(cli): add supabase notebooks pull

SaxonF:
Adds the `supabase notebooks` command family and its first subcommand, `notebooks pull`. Notebooks live in `supabase/notebooks/<name>.json`, one committed JSON file each, alongside `supabase/functions/` and `supabase/compute/`.

```sh
supabase notebooks pull                    # write notebooks missing locally
supabase notebooks pull <notebook-id>      # replace one local copy
```

## How to try it

```sh
supabase link --project-ref <ref>
supabase notebooks pull
```

The notebook id for the single-notebook form is the UUID in the notebook's dashboard URL.

Notebooks are behind a dashboard feature preview; switching it on lets you create and edit them in the dashboard, which is the quickest way to set up states worth pulling — several notebooks at once, charts and log cells, a notebook renamed or deleted after a pull, or two notebooks sharing a name.

## Notes

- **Identity is the file name, not the uuid.** A checkout is shared through git, where a uuid in a filename is unreadable. Names the filesystem cannot hold are reported and skipped rather than sanitised, since a sanitised name would push back later as a rename of somebody's notebook. Portable-filename aliases (`Sales` vs `sales`, NFC vs NFD) are refused before anything is written.
- **Duplicate names are refused, not guessed at.** Worth knowing for review: the dashboard names new notebooks "New Notebook", so a project with a few unrenamed ones fails every command that lists notebooks until they are renamed.
- **Local files naming no project notebook** get three answers — leave them, create them in the project, delete them locally — because such a file is either one somebody deleted in the dashboard or one somebody added locally, and the lists do not say which. An unattended run (non-TTY, machine output format, cancelled prompt) reports the divergence and changes nothing.
- Files are published through a temp file plus a hard link, so an interrupted pull leaves no partial file. The list walk follows the server's `links.next` cursor and fails closed on a missing or repeated cursor. `pull/SIDE_EFFECTS.md` records the full side-effect surface.


avallete (approved):
Approve. Staging pull passed: by id, unsafe names skipped on a broad pull and rejected on an explicit pull, `Sales` vs `sales` refused, duplicate names refused, and pagination past 100.

Nits, not blocking:

- `-o pretty --output-format json` prints JSON. `emitNotebooksMachineOutput` ignores `pretty`, then the handler follows `output.format`. `compute.output.ts` treats `-o pretty` as text. Worth aligning if a third caller shows up, not a problem for this command.
- `notebooks.output.ts` copies `compute.output.ts`. The header says to wait for a third caller. Fine to keep here.
- `tests/helpers/notebooks.ts` rebuilds the platform mock that `mockCommandPlatformApi` already provides.

PR #6599, opened 2026-09-14
feat(cli): add supabase notebooks push

SaxonF:
Adds `supabase notebooks push`, the mirror of `notebooks pull` from the previous PR in this stack. Every file in `supabase/notebooks/` is written to the project notebook of that name, or to a new one when the project has none.

```sh
supabase notebooks push                # push the whole directory
supabase notebooks push <name>         # push one notebook
```

## How to try it

```sh
supabase notebooks pull
# edit a file in supabase/notebooks/, or add a new one
supabase notebooks push <name>
supabase notebooks pull <notebook-id>  # see what the project stored
```

Notebooks are behind a dashboard feature preview; switching it on lets you create and edit them in the dashboard, so both sides can be moved independently — edit the same notebook in each place, set `favorite` in the dashboard and leave it out of the file, or add a notebook the directory does not have to reach the reconciliation prompt.

## Notes

- **Read everything before writing anything.** All selected files are read and decoded before the first upload, so a directory holding one unreadable notebook fails without having half-pushed the rest. The reconciliation plan is resolved before the first upload for the same reason.
- **Absent keys stay absent.** Keys a file leaves out are omitted from the request rather than sent as explicit nulls, which is how `favorite` keeps whatever the dashboard set on a notebook whose file never mentions it.
- **Cell identities survive.** Cell `id`s written out on pull are echoed on push, so a cell keeps its identity across an update instead of being replaced by a copy. The API rejects a cell carrying an id the notebook does not have.
- **Reconciliation points the other way.** Project notebooks naming no local file get the three answers `pull` offers, mirrored. Passing a notebook name pushes just that one and reconciles nothing — and because it only checks that one name for duplicates, it still works in a project where other notebooks share a name.
- `push/SIDE_EFFECTS.md` records the side-effect surface.


avallete (approved):
Approve. Staging push passed: create in Studio, pull, edit the file, push, and Studio showed the pushed markdown. Omitting `favorite` or `description` left the stored value. A broken file and a missing name failed before any upload. Unattended `-o json` kept remote-only notebooks. The interactive copy/delete prompt was not driven on a TTY; the integration tests cover those answers.

Nits, not blocking:

- Same `-o pretty --output-format json` case as pull: after the emitter declines `pretty`, the handler follows `output.format` and prints JSON.
- After the user chooses to write remote-only notebooks locally, one name that cannot be a file (`reports/weekly`) fails the whole push and uploads nothing. Pull skips and counts that name.
- `notebooks push sales.json` says there is no notebook named `sales.json`. Stripping a trailing `.json` would match the filename people pass.

PR #6511, opened 2026-09-07
feat(cli): add experimental stack prepare

jgoux:
`supabase experimental stack prepare` downloads the resources needed by all enabled capabilities or a selected subset, without starting services. Resolve project configuration and stack targeting at the CLI boundary, and delegate preparation and dependency selection to the stack package.

This PR builds on experimental stack logs (#6510).

PR #6601, opened 2026-09-14
feat(cli): add stack prepare

jgoux:
Adds `supabase stack prepare` to download artifacts for all enabled capabilities or a selected subset and their dependencies without starting services. It creates and registers the target stack when absent, waits for preparation to complete, and reports failures using the current stack targeting, configuration, output, and telemetry conventions.

Supersedes #6511


apps/cli/src/commands/experimental/stack/prepare/SIDE_EFFECTS.md
```diff
@@ -0,0 +1,63 @@
...
+`--stack` and `--stack-id` are mutually exclusive. `--runtime` defaults to `auto` for new stacks;
+existing stacks reuse their persisted runtime, and an explicit runtime that differs from an existing
+stack fails. When no project configuration exists, the stack package's default settings are used.
+With no `--capability`, all enabled capabilities are
+prepared. Repeated `--capability` values are passed to the package, which includes dependencies and
+rejects unknown or disabled capabilities. The legacy `-o/--output` flag is rejected; use
+`--output-format`.
+
+Text output lists the stack ID and each prepared capability, version, and outcome (`cached`,
+`downloaded`, or `pulled`). JSON and stream-JSON output return `{ id, capabilities }`. Errors use
```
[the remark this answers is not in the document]
jgoux: Fixed in 5c9b8a089. SIDE_EFFECTS.md now documents the empty `message` field in JSON and the complete stream-JSON result envelope, matching the output layers and integration assertions.

apps/cli/src/command-internal/db-target-flags.ts
```diff
@@ -104,6 +104,7 @@ export const VALUE_CONSUMING_LONG_FLAGS = new Set([
   "remove-domains",
   "role",
   "runtime",
+  "capability",
```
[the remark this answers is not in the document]
jgoux: Moved `capability` into the existing experimental stack flag section in 72d15cf88. This is an organizational change; membership and parsing behavior are unchanged.

apps/cli/src/commands/experimental/stack/prepare/prepare.command.ts
```diff
@@ -0,0 +1,41 @@
...
+    Flag.optional,
+  ),
+  runtime: Flag.choice("runtime", ["auto", "docker", "native"] as const).pipe(
+    Flag.withDescription("Runtime to use for a new stack."),
+    Flag.withDefault("auto" as const),
+  ),
+  capability: Flag.atMost(
+    Flag.choice("capability", CAPABILITY_NAMES),
+    CAPABILITY_NAMES.length,
+  ).pipe(Flag.withDescription("Capability to prepare (repeatable).")),
```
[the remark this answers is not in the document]
jgoux: Keeping the existing occurrence bound. It permits every distinct capability in one invocation, and the package's Set-based dependency selection already prevents duplicate preparation. Repeating the same value beyond that bound is rejected by the parser; adding another deduplication layer in the CLI would duplicate package ownership without addressing a demonstrated workflow requirement.

apps/cli/src/commands/experimental/stack/prepare/prepare.errors.ts
```diff
@@ -0,0 +1,85 @@
...
+            "StackVersionUnsupportedError",
+            "InvalidStackIdentityError",
+            "InvalidProjectRootError",
+            "StackStateInvalidError",
+            "StackStateFormatUnsupportedError",
+            () => ({ reason: "invalid-config" as const }),
+          ),
+          Match.tag("StackNotFoundError", "StackRuntimeMismatchError", () => ({
+            reason: "flags" as const,
+          })),
```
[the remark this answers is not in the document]
jgoux: Fixed in 5c9b8a089. `StackRuntimeMismatchError` now has its own branch with guidance to omit `--runtime` or choose another stack name. The handler integration test covers this named-stack failure path.

apps/cli/src/commands/experimental/stack/prepare/prepare.handler.ts
```diff
@@ -0,0 +1,100 @@
...
+    const result = yield* stack
+      .prepare({
+        config,
+        ...(flags.capability.length === 0 ? {} : { capabilities: flags.capability }),
+      })
+      .pipe(
+        Effect.tapError((error) => task.fail(error.message)),
+        Effect.tap(() => task.clear()),
+        Effect.mapError(stackPrepareError),
+      );
```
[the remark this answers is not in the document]
jgoux: Keeping aggregate progress for this command: one task indicates ongoing preparation, and the final result lists capability outcomes. `onProgress` is an optional package hook; artifact-level updates would be a separate UX enhancement, not a missing completion or failure signal. Cancellation settlement is addressed separately in 72d15cf88.

apps/cli/src/commands/experimental/stack/prepare/prepare.handler.ts
```diff
@@ -0,0 +1,100 @@
...
+            .pipe(Effect.mapError(stackPrepareError));
+    const task = yield* output.task("Preparing local Supabase stack...");
+    const result = yield* stack
+      .prepare({
+        config,
+        ...(flags.capability.length === 0 ? {} : { capabilities: flags.capability }),
+      })
+      .pipe(
+        Effect.tapError((error) => task.fail(error.message)),
+        Effect.tap(() => task.clear()),
```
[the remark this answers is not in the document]
jgoux: Confirmed and fixed in 72d15cf88. One exit-aware finalizer now clears the task on success, reports the original message on typed failure, and cancels it on interruption or defects while preserving the original cause. The cancellation regression failed before the fix and passes afterward; defect coverage also verifies task settlement and the preserved defect value. This command now owns that cleanup even though the sibling start command uses the older pattern.

jgoux:
Addressed the AI review findings across 5c9b8a089 and 72d15cf88. The command guide now includes `stack prepare` (the finding outside the original diff), the machine-output documentation and runtime-mismatch guidance are corrected, and preparation now settles its output task on interruption and defects. The flag entry also sits with the other stack flags.

Kept the existing bounded capability selection and aggregate progress display, with the reasoning in their inline replies. All inline review conversations have been answered and resolved.

apps/cli/src/commands/experimental/stack/prepare/prepare.handler.ts
```diff
@@ -0,0 +1,117 @@
...
+            reason: "invalid-config",
+            message: error.message,
+            cause: error,
+          }),
+      ),
+    );
+    const disabledCapability = flags.capability.find((capability) => {
+      const configured = config.capabilities?.[capability];
+      return configured !== undefined && "enabled" in configured && configured.enabled === false;
+    });
```
Coly010: **Blocking.** This pre-check only inspects the capabilities named directly on `--capability`. The package rejects the full dependency closure, not just the roots (`packages/stack/src/public/EffectStack.ts`, `dependencyClosure`). `studio` depends on `rest` and `analytics` (`packages/stack/src/model/capabilities/studio.ts`), so with `[analytics]\nenabled = false` and `--capability studio`, this check passes, a stack gets created, and the package then fails mid-preparation with `StackPreparationError("Capability analytics is disabled")` — which `prepare.errors.ts` maps to an external-network error (see comment there).

Suggest fixing this at the source: have the package raise `InvalidStackConfigError` for a disabled-capability rejection (in the closure expansion, not just `StackPreparationError`) — that tag is already in `PREPARE_STACK_ERROR_TAGS` and already maps to `invalid-config` in `prepare.errors.ts`, so it's a one-line package change with no CLI change needed, and it fixes both the direct and transitive case with one rule instead of two. Please add a regression test for the transitive case (root enabled, dependency disabled) either way — the current test only covers naming the disabled capability directly.

apps/cli/src/commands/experimental/stack/prepare/prepare.errors.ts
```diff
@@ -0,0 +1,90 @@
...
+          Match.tag("ContainerPullError", () => ({
+            reason: "registry" as const,
+            suggestion:
+              "Check registry connectivity and image availability, then retry the command.",
+          })),
+          Match.tag("StackPreparationError", "ArtifactIntegrityError", () => ({
+            reason: "artifact" as const,
+            suggestion:
+              "Retry the stack preparation with --debug if the artifact cannot be prepared.",
```
Coly010, line 60: This tag is the package's catch-all preparation-failure wrapper, so it currently absorbs both "a real download/extract failure" and "you asked for (or transitively depend on) a disabled capability" — see the linked comment on `prepare.handler.ts`. If the package fix (raising `InvalidStackConfigError` for the disabled-capability case) isn't feasible in this PR, an interim fix here would be to branch on whether the error carries capability/config-shaped detail before defaulting to `reason: "artifact"`, so a config problem doesn't get a "retry the network operation" suggestion or land in the external-failure telemetry bucket.

Separately (non-blocking): this suggestion ("Ensure the selected container engine is running...") for `ContainerEngineError` dropped "installed" compared to `start.handler.ts`'s wording for the same tag, which matters since that error's message is often a raw `spawn docker ENOENT` for a missing binary, not just a stopped daemon. And `StackStateInvalidError`/`StackStateFormatUnsupportedError` get no suggestion here vs. targeted ones in `start.handler.ts` — worth aligning, or better, hoisting this classifier into `stack.shared.ts` since it's now duplicated with drift across five sibling command files.

apps/cli/src/commands/experimental/stack/prepare/prepare.handler.ts
```diff
@@ -0,0 +1,117 @@
...
+    const result = yield* stack
+      .prepare({
+        config,
+        ...(flags.capability.length === 0 ? {} : { capabilities: flags.capability }),
+      })
+      .pipe(
+        Effect.onExit((exit) =>
+          Exit.isSuccess(exit)
+            ? task.clear()
+            : Option.match(Cause.findErrorOption(exit.cause), {
```
Coly010, line 94: `Cause.findErrorOption` returns `None` for both interruption *and* a genuine defect (e.g. `Effect.die`), so a real bug/panic inside `@supabase/stack` currently settles the spinner as `task.cancel()` ("canceled") rather than `task.fail()`. The added test ("settles the task on a preparation defect...") locks this in as intended behavior, but it reads backwards to a user — "canceled" implies they did something, when actually the tool crashed.

Suggest branching on `Cause.hasInterrupts(exit.cause)` explicitly instead of `Cause.findErrorOption`, so only a real interruption maps to `cancel()` and a defect maps to `fail()`. Non-blocking relative to the misclassification above, but worth fixing in the same pass since it's a small, contained change and the test needs to flip either way.

apps/cli/src/commands/experimental/stack/prepare/SIDE_EFFECTS.md
```diff
@@ -0,0 +1,65 @@
...
+## Exit codes
+
+- `0` when preparation completes.
+- `1` for invalid targets/configuration, unavailable runtime or registry, preparation failures, or
+  rejected output flags.
+
+## Telemetry
+
+The command emits the standard `cli_command_executed` event through `withCommandTelemetry` and
+flushes telemetry state after successful and failed handler execution. It emits no custom event.
```
Coly010, line 70: Non-blocking, but this file is the command's compatibility contract so worth tightening:

- "A managed stack descriptor/state record when the current or named target does not exist" is the *only* place this is disclosed. The command description (`prepare.command.ts`) and `docs/stack-commands.md` both describe `prepare` purely as "download artifacts without starting services," with no hint that running it against a project with no existing stack silently registers one (it then shows up in `supabase stack list` and needs `supabase stack destroy` to remove). Worth a line in the user-facing docs/description, not just here.
- Exit codes are missing `130` for interruption — every sibling that documents interruption (`status`, `stop`, `logs`) lists it, and this PR specifically added interruption handling.
- Line 10 says `config.toml` **or** `config.json` is read by the shared loader; `loadStackConfig` actually passes `tomlOnly: true`, so `config.json` isn't read here (that's only true for the backend-detection gate other commands use).

apps/cli/src/commands/experimental/stack/prepare/prepare.command.ts
Coly010, line 21: Non-blocking nit: `--exclude` on `stack start` covers the same capability domain via a comma-separated `stringSliceFlag`, so `--capability rest,auth` is the natural thing to try here and it fails (this flag is repeat-only). `Flag.atMost(..., CAPABILITY_NAMES.length)` also bounds *occurrences* rather than distinct values, so `--capability rest` repeated 11 times trips an "at most 10" error instead of just deduping. Not asking for a change here specifically, but the family now has two different idioms (opt-out CSV vs. opt-in repeat-only) over the same enum — worth a deliberate decision recorded somewhere rather than drift.

Coly010 (changes requested):
Ran a 4-agent review pass (architecture, engineering, security, DX) over this PR. All four independently converged on the same root-cause bug, so I'm flagging it as blocking; everything else below is non-blocking commentary bundled into this one review.

**Blocking:** a capability disabled only via dependency (not the one named on `--capability`) gets misclassified as an external/network failure instead of a config error — see inline comments on `prepare.handler.ts` and `prepare.errors.ts`. Concrete repro: `studio` depends on `rest` + `analytics`; with `[analytics]\nenabled = false` in config, `supabase stack prepare --capability studio` passes the CLI's shallow pre-check, creates a stack, then fails deep in `@supabase/stack` with `StackPreparationError`, which this file maps to `reason: "artifact"` → external-network actionability and a "retry with `--debug`" suggestion that doesn't apply to a config typo. It also miscounts the error in the external-failure telemetry bucket. The direct case (naming the disabled capability itself) is handled correctly — only the transitive/dependency case is wrong.

Security review otherwise **approved** — no exploitable issue; `--stack-id`/`--capability` validation, telemetry redaction, and artifact/registry trust delegation are all sound.

Thanks for adding the `Effect.onExit` task-settlement handling here — it correctly settles the spinner on interruption, which the existing `stack start`/`stop` handlers still don't do. One correctness note on it below, but the overall approach is worth backporting to the siblings once fixed.

apps/cli/src/commands/experimental/stack/prepare/prepare.handler.ts
jgoux: Moved disabled-capability rejection to `InvalidStackConfigError` in the package and removed the duplicate CLI precheck. Added public API regressions for a directly disabled capability and enabled Studio with disabled Analytics, asserting the typed config error, no artifact progress, unchanged state, and no owner acquisition. One detail from verification: the reported transitive scenario already fails in `createExecutionPlan` with `InvalidStackConfigError`, before reaching the preparation closure. I reproduced that with the compiled CLI for both Analytics and REST. The package change makes the direct rejection consistent and keeps validation under one owner.

Addressed in d6485ef49.

apps/cli/src/commands/experimental/stack/prepare/prepare.errors.ts
jgoux: Applied the source-level fix: disabled selections now raise `InvalidStackConfigError`, which maps to invalid-config actionability without inspecting messages or error details. The handler regression checks that mapping; actual artifact failures retain their existing classification. Also restored the installed-engine guidance and aligned the invalid/unsupported persisted-state recovery suggestions with start. Kept this scoped to prepare rather than refactoring all sibling classifiers.

Addressed in d6485ef49.

apps/cli/src/commands/experimental/stack/prepare/prepare.handler.ts
jgoux: Fixed: only a pure interruption cancels the task; a defect fails it. Typed failures retain their diagnostic message, including causes that also contain interruption. The regression now expects failure for a defect and verifies that the original defect propagates; interruption cleanup remains covered.

Addressed in d6485ef49.

apps/cli/src/commands/experimental/stack/prepare/SIDE_EFFECTS.md
jgoux: Updated the command description and user guide to disclose creation/registration of absent stacks, including visibility in stack list and removal with stack destroy. The side-effect contract now documents exit 130 and distinguishes TOML-only command settings from the feature gate that may inspect JSON.

Addressed in d6485ef49.

apps/cli/src/commands/experimental/stack/prepare/prepare.command.ts
jgoux: Kept the repeat-only enum choice intentionally: each occurrence is one validated capability, with at most ten occurrences; the package deduplicates the dependency closure. Documented that contract in the guide and side-effect contract, with repeated-flag examples and an explicit note that CSV is not accepted. This preserves the current parser behavior, including the occurrence limit for duplicates. We can consider family-wide flag consistency separately.

Addressed in d6485ef49.

jgoux:
Addressed the review in d6485ef49: moved disabled-selection validation into the package, corrected defect task feedback, aligned recovery suggestions, and clarified registration, configuration, interruption, and flag behavior. The transitive disabled-dependency cases were already rejected as configuration errors during plan compilation; regressions now cover both direct and transitive cases. All five inline conversations have individual replies.

Independent Fable review reached consensus with no remaining actionable findings. Local verification passed: 49 package and 196 CLI stack integration tests, type/lint/format checks, the CLI build, and compiled-command checks for direct and transitive disabled selections.

Coly010 (approved):

PR #6605, opened 2026-09-14
refactor(cli): cover `login`, `logout`, `feedback` and `test` with effect lint (CLI-2430)

7ttp:
## TL;DR

brings the account and a few smol command families under the effect lint, coverage goes from 381 to 424 files.

## whats introduced?

effect lint applied to `login`, `logout`, `feedback` and `test`:

- one allow list entry per family in `.oxlintrc.effect.json`
- the only handler logic touched is `login`: the `SUPABASE_PROFILE` read stays a live `process.env` read behind a short alias, now pinned by a test; everything else is `Effect.fail` unwraps to yieldable errors, behavior identical
- cause assertions move to `Cause.pretty`, env tests onto the shared `withEnvVar` helper, and the profile persistence tests unset `SUPABASE_HOME` so an ambient value cannot write outside the temp root
- redaction scans keep their full json surface and gain a positive anchor, and the `feedback delete` gates pin their exact remediation text
- e2e files move onto the effect native harness with a shared scoped temp home, the live round trip teardown moves under `Effect.ensuring`, and `test new` moves onto the `FileSystem`/`Path` services

## ref:

- closes: CLI-2430


apps/cli/src/commands/logout/logout.e2e.test.ts
```diff
@@ -1,69 +1,77 @@
...
-      expect(exitCode).toBe(0);
-      expect(stderr).toContain("You were not logged in, nothing to do.");
-    },
+    () =>
+      Effect.gen(function* () {
+        const home = yield* tempHomeScoped;
+        const { exitCode, stderr } = yield* runSupabaseEffect(["logout", "--yes"], {
+          home: home.dir,
+          // The runner may export a real token; not-logged-in needs it absent.
+          env: { HOME: home.dir, SUPABASE_ACCESS_TOKEN: undefined },
```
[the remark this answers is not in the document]
7ttp: fixed: 524f9de89

apps/cli/src/commands/feedback/feedback.layers.unit.test.ts
```diff
@@ -18,57 +18,67 @@ function recordingLogger() {
...
+  it.live("defaults the logged method to GET when the request carries none", () =>
+    Effect.gen(function* () {
+      const { logger, httpLines } = recordingLogger();
+      const inner = recordingInnerFetch();
+      const fetch = feedbackFetch({ dnsResolver: "native", logger, innerFetch: inner.fetch });
 
-    await fetch("https://feedback.supabase.co/rest/v1/interfaces_feedback");
+      yield* Effect.promise((signal) =>
+        fetch("https://feedback.supabase.co/rest/v1/interfaces_feedback", { signal }),
+      );
```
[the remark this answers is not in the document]
7ttp: fixed: 524f9de89

apps/cli/src/commands/login/login.handler.ts
```diff
@@ -45,7 +45,11 @@ export const login = Effect.fn("login")(function* (flags: LoginFlags) {
     onNone: () => undefined,
     onSome: ({ args }) => lastExplicitLongFlagValue(args, [], "profile"),
   });
-  const envProfile = process.env["SUPABASE_PROFILE"];
+  // Read the live `process.env` proxy at run time, like the resolver in command-settings: a
+  // `Config` read was tried and rejected (its env snapshot is case-sensitive, breaking Windows
+  // parity). The alias satisfies `process-env-in-effect`, which flags only direct reads.
+  const env = process.env;
+  const envProfile = env["SUPABASE_PROFILE"];
```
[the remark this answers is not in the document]
7ttp: the alias is deliberate: config was tried first and reverted, its env snapshot is case sensitive and broke windows parity. and disables are off the table for this rollout

real fix is widening CommandSettings to carry the raw token, i'll pick that up when src/config comes under coverage

jgoux: **[P2] Remove the intentional environment lint bypass**

This alias preserves the direct ambient environment dependency while hiding it from `process-env-in-effect`, so adding this handler to the lint allow list does not complete its Effect migration. The Windows concern applies to the default snapshot provider; it does not require bypassing the rule. Please expose the raw profile through the existing settings owner or provide a live environment-backed Config provider at the composition boundary, then remove the alias and its explanatory comment. Preserve the raw profile token, runtime lookup, explicit-flag precedence, and Windows case-insensitive environment behavior.

apps/cli/tests/helpers/cli.ts
```diff
@@ -146,6 +146,21 @@ export function makeTempHome() {
...
+  readonly cause: unknown;
+}> {}
+
+/** Scoped temp home for Effect-native e2e tests; the scope owns disposal. */
+export const tempHomeScoped = Effect.acquireRelease(
+  Effect.try({
+    try: () => makeTempHome(),
+    catch: (cause) => new TempHomeSetupError({ message: "temp home setup failed", cause }),
+  }),
+  (owned) => Effect.sync(() => owned[Symbol.dispose]()),
```
jgoux: **[P3] Handle fallible temp-home disposal explicitly**

`owned[Symbol.dispose]()` calls `rmSync`, which can throw, so this `Effect.sync` turns an operational cleanup failure into a defect. I reproduced this by creating an unreadable directory containing a file inside the owned temp home: scope closure failed with `ENOTEMPTY`. The same helper file already represents disposal failures with `CliHomeDisposeError`. Please use an explicit cleanup error policy; a `withTempHome(use)` helper built on `Effect.acquireUseRelease` can propagate that typed error. A bare `Effect.try` replacement here is insufficient because `acquireRelease` requires an infallible release effect. This is a test-helper concern, with no demonstrated production regression.

jgoux (changes requested):
Please address the Effect lint bypass in login and make the new temp-home cleanup failure policy explicit. The two inline findings concern Effect migration discipline; no production behavior regression was demonstrated.

apps/cli/src/commands/login/login.handler.ts
7ttp: fixed: d8df9717e

the raw value now rides on `CommandSettings`, captured by the same live `process.env`
 read the resolver already does, so windows casing is untouched and the read happens exactly once. only in-process env writer is `applyProjectEnv`, allowlisted to `SUPABASE_INTERNAL_IMAGE_REGISTRY`, so build-time capture cant drift from handler time, and the resolved profile and the persisted token can no longer disagree...

apps/cli/tests/helpers/cli.ts
7ttp: fixed: b8e6d2d5c

went with your withTempHome shape on acquireUseRelease, release failures map into the existing CliHomeDisposeError. same bracket withEnvVar already uses

jgoux (approved):

PR #6613, opened 2026-09-15
feat(stack): model lazy service lifecycle with explicit states

jgoux:
Represent stack commands, service startup, idle retirement, and cleanup recovery with explicit lifecycle states and owned operations. Requests share startup, wait for an in-progress retirement before waking a service, and cannot restart services after an explicit stop. Failed cleanup preserves resource ownership for recovery. Status reports the recovery reason and required command; HTTP responses distinguish recovery-required failures.

Lazy REST, Auth, Realtime, Studio, and pooler services stop after 60 seconds without activity, with per-service timeout overrides through the Effect API. The CLI’s `--eager` option activates all enabled capabilities and disables idle retirement. Active requests, streams, sockets, and dependants keep services running; data and listeners survive retirement.

On macOS, native cleanup verifies that a process group has no live members before accepting `EPERM` for an exited group. Live groups and failed, inconclusive, or timed-out inspection still fail cleanup.

Supersedes #6579


packages/stack/src/supervisor/StatusProjection.ts
```diff
@@ -12,86 +8,75 @@ import {
...
+        control === undefined
+          ? configured === undefined || !configured.enabled
+            ? "disabled"
+            : "stopped"
+          : snapshot.stack._tag === "starting" &&
+              snapshot.stack.prior._tag === "stopped" &&
+              control._tag === "starting" &&
+              control.completion._tag === "workload"
+            ? "starting"
+            : publicCapabilityState(control);
```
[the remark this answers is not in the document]
jgoux: Removed the redundant cold-start branch in 944fa8bec. Capability status now delegates directly to publicCapabilityState, which already projects starting. The full stack integration suite passes.

packages/stack/src/supervisor/SupervisorState.ts
```diff
@@ -0,0 +1,175 @@
...
+      readonly completion: LifecycleCompletion;
+      readonly prior: StableStackState | RecoveryState;
+    };
+
+type TransitionState =
+  | Extract<StackControlState, { readonly _tag: "starting" }>
+  | Extract<StackControlState, { readonly _tag: "stopping" }>
+  | Extract<StackControlState, { readonly _tag: "destroying" }>;
+
+export type CommandAdmission =
```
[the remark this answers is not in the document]
jgoux: Addressed in 944fa8bec: LifecycleCompletion, CommandAdmission, and EndpointState are module-local. Supervisor imports the shared LifecycleKind instead of redeclaring it. Type and unused-code checks pass.

packages/stack/src/supervisor/Supervisor.ts
```diff
@@ -274,21 +392,677 @@ export const makeSupervisor = (
...
+                case "dormant":
+                case "starting":
+                case "ready":
+                case "stopping":
+                case "cleanup-failed":
+                  return { ...state, traffic: Math.max(0, count - 1) };
+                default:
+                  return state;
+              }
+            });
```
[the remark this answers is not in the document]
jgoux: Verified that gateway tracking acquires a lease before activation rejects a cleanup-failed session, so a rejected request could decrement an existing count without incrementing it. Added cleanup-failed to acquisition in 944fa8bec, making acquisition and release symmetric. The recovery fence and stale-session protection remain intact; the full stack integration suite passes.

packages/stack/src/gateway/ActivityTracker.ts
```diff
@@ -0,0 +1,28 @@
...
+  readonly track: <A, E>(
+    capability: CapabilityName,
+    effect: Effect.Effect<A, E>,
+  ) => Effect.Effect<A, E>;
+}
+
+export interface GatewayActivityCallbacks {
+  readonly begin: (capability: CapabilityName) => Effect.Effect<unknown>;
+  readonly end: (capability: CapabilityName, lease: unknown) => Effect.Effect<void>;
+}
```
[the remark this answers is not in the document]
jgoux: Parameterized GatewayActivityCallbacks and makeGatewayActivity with the lease type in 944fa8bec. endTraffic now accepts TrafficLease directly, and the runtime shape check and unused token are removed. Effect.acquireUseRelease owns one release per acquired lifetime; sessionId still rejects releases from an earlier session. Type checks and stack integration tests pass.

packages/stack/src/public/whole-stack.e2e.test.ts
```diff
@@ -1587,6 +1780,197 @@ describe("managed Supabase stack whole-stack E2E", () => {
...
+        );
+        expect(secondRows).toEqual([{ id: 1, payload: marker }]);
+        await expectOwnedWorkloads(mode, stack.id, ["database:database", "rest:rest"]);
+        const restartedRest = await stack.status();
+        expect(capabilityState(restartedRest, "rest")).toBe("ready");
+        expect(endpoint(restartedRest, "api").port).toBe(api.port);
+        expect(await supervisorPid(stack.id)).toBe(initialSupervisorPid);
+      },
+    );
+    test.skipIf(mode.runtime.kind === "container")(
```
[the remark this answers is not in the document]
jgoux: Removed the container skip in 944fa8bec. There was no concrete runtime limitation requiring it. Ran this exact test with an isolated stack home in Docker mode: it passes and verifies live HTTP, WebSocket, and TCP connections settle during stop. The test is now enabled for both runtimes.

docs/adr/0017-simplified-managed-stack-architecture.md
```diff
@@ -268,7 +285,7 @@ Tests follow consumed boundaries:
 - stack integration covers identity, sticky ports, durable lifecycle,
   ownership, stale-owner recovery, and interrupted cleanup;
 - supervisor integration covers detached ownership, RPC, stop, and
-  destroy; and
+  destroy; dedicated idle-stop integration covers retirement and wake-up; and
```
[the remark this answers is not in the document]
jgoux: Updated the ADR in 944fa8bec to describe retirement and wake-up as part of supervisor integration coverage, matching the actual test organization. Formatting checks pass.

packages/stack/src/supervisor/SessionLauncher.ts
```diff
@@ -72,128 +96,166 @@ export const makeSessionLauncher = (options: {
...
-    const stop = Effect.suspend(() => Ref.get(session).pipe(Effect.flatMap(cleanup)));
+        ),
+      );
+    const stop = Effect.suspend(() => Ref.get(session).pipe(Effect.flatMap(cleanupOrFail)));
+    const stopCapabilities = (
+      capabilities: ReadonlySet<import("../public/Capability.ts").CapabilityName>,
+    ) =>
+      Ref.get(session).pipe(
+        Effect.flatMap((entries) =>
+          cleanupOrFail(entries.filter(({ workload }) => capabilities.has(workload.capability))),
```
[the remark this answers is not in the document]
jgoux: The native cleanup risk is real, but both this ledger rule and the swallowed process-kill failure exist at the develop merge-base. The proposed launcher-only change is insufficient: NativeRuntime.cleanup also removes its resource record after closing the scope, so retaining the launcher's key would not make that process retryable. Successful container removal can also prove absence after an earlier stop failure. I am deferring this to a separate native removal-contract fix that retains and reports unproven process cleanup, rather than adding an ineffective ledger guard here.

packages/stack/src/supervisor/Supervisor.ts
```diff
@@ -348,120 +1144,209 @@ export const makeSupervisor = (
...
+              if (Predicate.isTagged(admitted, "rejected"))
+                return yield* new StackLifecycleConflictError({
+                  stackId: options.stackId,
+                  message:
+                    admitted.reason === "stop-required"
+                      ? "Exact runtime cleanup is required; retry stop before starting"
+                      : admitted.reason === "destroy-required"
+                        ? "Destructive cleanup is required; retry destroy before starting"
+                        : `Lifecycle operation ${kind} is already in progress`,
+                });
```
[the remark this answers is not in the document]
jgoux: Fixed in [327f3de26](https://github.com/supabase/cli/commit/327f3de26842fcba863b0c1bb3590268b467758a): destroy-required rejection now says to retry destroy before proceeding, so the guidance is accurate for both start and stop.

packages/stack/src/supervisor/Supervisor.ts
```diff
@@ -518,188 +1407,406 @@ export const makeSupervisor = (
...
+          )
+          .pipe(
+            Effect.onExit((result) =>
+              settleOwner(
+                Predicate.isTagged(owner, "endpoint")
+                  ? { ...owner, result: Exit.map(result, (value) => value.endpoint) }
+                  : { ...owner, result },
+              ),
+            ),
+          );
```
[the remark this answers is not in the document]
jgoux: Fixed in [327f3de26](https://github.com/supabase/cli/commit/327f3de26842fcba863b0c1bb3590268b467758a): activation settlement now completes while the owner still holds execution, before a queued stop selects capability cleanup. Acquisition interruption still completes its waiters without releasing an unacquired permit. The final review also exposed the inverse ordering: a stop can acquire a released semaphore before an existing activation waiter resumes. A matching activation that subsequently fails on a stopped stack now leaves its capability stopped, rather than restoring dormant. Verified the semaphore ordering with an executed reproduction and the latter transition with a test that fails against the previous implementation. The in-flight activation/queued-stop integration flow also passes; it is smoke coverage, not a deterministic reproduction of the original race.

packages/stack/src/supervisor/Supervisor.ts
```diff
@@ -348,120 +1144,209 @@ export const makeSupervisor = (
...
+          const cause = Exit.isFailure(closed)
+            ? Cause.combine(launchCause, closed.cause)
+            : launchCause;
+          const workloadCleanup = mapSessionCleanup(launch.cleanup);
+          return {
+            _tag: "failed",
+            cause,
+            cleanup: Exit.isFailure(closed)
+              ? { _tag: "unproven", cause: closed.cause }
+              : workloadCleanup,
```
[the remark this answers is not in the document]
jgoux: Fixed in [327f3de26](https://github.com/supabase/cli/commit/327f3de26842fcba863b0c1bb3590268b467758a): launch failure paths combine ingress and workload cleanup evidence instead of replacing the workload cause when ingress closing also fails. The primary failure retains the relevant cleanup diagnostics without adding an already-included workload failure again.

packages/stack/src/supervisor/StatusProjection.ts
```diff
@@ -12,86 +8,75 @@ import {
...
       return {
         name,
         activation:
-          definition?.capabilities[name].activation ?? (name === "database" ? "eager" : "lazy"),
+          configured?.activation ?? (name === "database" ? ("eager" as const) : ("lazy" as const)),
         state: capability,
-        ...(error === undefined ? {} : { error }),
+        ...(cleanupError === undefined && observedFailure?.error === undefined
+          ? {}
+          : { error: cleanupError ?? observedFailure?.error }),
```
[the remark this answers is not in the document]
jgoux: Fixed in [327f3de26](https://github.com/supabase/cli/commit/327f3de26842fcba863b0c1bb3590268b467758a): status collects all failed workload diagnostics, preserving workload identifiers for multiple failures. Regression coverage includes both multiple nonempty errors and an earlier failed workload without error text.

packages/stack/src/supervisor/Supervisor.ts
```diff
@@ -518,188 +1407,406 @@ export const makeSupervisor = (
...
+            cleanup: { _tag: "unproven", cause: started.cause },
+            durable: "unsafe",
+          } satisfies CommandResult;
+        }
+        if (Predicate.isTagged(started.value, "failed")) {
+          yield* settleStartupFailures(
+            started.value.cause,
+            started.value.cleanup,
+            started.value.durable,
+          );
```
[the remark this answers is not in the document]
jgoux: Retained both calls because they cover different timing and failure paths. startOperation settles startup workload waiters before releasing execution when the controller returns a failed result. The outer lifecycle finalizer also covers early errors, defects, and interruption that bypass that result branch. The settlement only consumes matching starting/workload states, so a second invocation cannot overwrite a completed state or complete a waiter twice. Removing either call would need a broader lifetime change; it is not a safe redundancy-only deletion.

packages/stack/src/supervisor/Supervisor.ts
```diff
@@ -518,188 +1407,406 @@ export const makeSupervisor = (
...
+                    }),
+                  ),
+                ),
+                Match.when({ _tag: "cleanup-failed" }, () =>
+                  reject(`Capability ${capability} cleanup failed; retry stop before activating`),
+                ),
+                Match.when({ _tag: "ready" }, (state) =>
+                  Effect.gen(function* () {
+                    if ((yield* currentPhase()) !== "running")
+                      return yield* reject(`Capability ${capability} is already transitioning`);
```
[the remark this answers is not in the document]
jgoux: Verified that ensureActivationPhaseAllowed already requires running under the same admission permit. Removed the redundant phase check and its unreachable error in [327f3de26](https://github.com/supabase/cli/commit/327f3de26842fcba863b0c1bb3590268b467758a).

packages/stack/src/supervisor/OperationSettlement.ts
```diff
@@ -0,0 +1,378 @@
...
+    completionCause === undefined ? commandResultExit(operation) : Exit.failCause(completionCause);
+  const next: StackControlState = Match.value(current).pipe(
+    Match.tag("start-recovery", (state) => ({
+      _tag: "stop-required" as const,
+      cause: Match.value(operation).pipe(
+        Match.tag("succeeded", () => state.cause),
+        Match.tag("failed", (failure) => Cause.combine(state.cause, failure.cause)),
+        Match.exhaustive,
+      ),
+    })),
```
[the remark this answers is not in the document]
jgoux: Retained this as a readability tradeoff rather than a correctness fix. Both calculations use the same exhaustive cases, and stale completions are gated before either calculation. Directly returning Exit values from Match widens to Effect with the installed Effect version; that alternative was reproduced and checked in the prior review. I am retaining the current settlement structure in this correctness-focused pass rather than reshaping it for optional deduplication.

packages/stack/README.md
```diff
@@ -24,10 +24,40 @@ capability is lazy. Starting the stack therefore launches only
...
+service running, while open WebSocket or TCP connections do. Configure a different positive
+timeout, or disable traffic stopping for a capability, with `idleTimeoutSeconds`:
+
+```ts
+await stack.start({
+  config: {
+    capabilities: {
+      rest: { idleTimeoutSeconds: 120 },
+      realtime: { idleTimeoutSeconds: false },
+    },
```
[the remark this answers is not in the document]
jgoux: Documented in [327f3de26](https://github.com/supabase/cli/commit/327f3de26842fcba863b0c1bb3590268b467758a): changing idleTimeoutSeconds on a running stack requires stop(), followed by start() with the updated configuration. This matches the existing stopped-time definition validation.

jgoux:
Review follow-up is in [327f3de26](https://github.com/supabase/cli/commit/327f3de26842fcba863b0c1bb3590268b467758a). Addressed the lifecycle ordering, cleanup diagnostics, multi-workload status, rejection wording, redundant guard, and idle-timeout documentation findings. The summary-only empty-message finding is also fixed: maintenance cleanup errors now include the workload identifier and underlying reason.

The final cross-check additionally caught a late rejected activation restoring dormant state after stop; that transition now remains stopped. Two optional settlement simplifications were retained for the reasons recorded in their threads. The native process-removal/ownership issue predates this PR and needs a separate runtime removal-contract fix; the launcher-only suggestion would not preserve retryability.

All review conversations have been answered and resolved. Final independent review reached consensus with no outstanding in-scope findings. Validation: 774 stack integration tests, five settlement tests, 27 CLI status tests, and four targeted native E2E tests passed, along with type, lint, format, and unused-code checks.

packages/stack/src/state/StackState.ts
```diff
@@ -118,7 +118,13 @@ const hasCompleteDefaults = (value: unknown, defaults: unknown): boolean => {
   return true;
 };
 
-const capabilityKeys = ["enabled", "activation", "version", "settings"] as const;
+const capabilityKeys = [
+  "enabled",
+  "activation",
+  "idleTimeoutSeconds",
```
Coly010, line 124: **Blocker: every existing stack becomes undecodable, and un-destroyable, after upgrade.**

`hasExactKeys(capability, capabilityKeys)` now *requires* `idleTimeoutSeconds`, but `STACK_STATE_FORMAT` is still `supabase-stack-state-v1` and `normalizeDurableState` in `StackStateStore.ts` has no entry for the new key. Any `state.json` written by current develop fails `isDefinitionShape` and surfaces as the generic:

```
Invalid persisted stack state: SchemaError(Materialized definition shape is invalid at ["definition"])
```

I reproduced this with a stack I started on develop earlier today (`rest` keys on disk: `activation, enabled, settings, version`). The blast radius is wider than one stack: `PortCoordinator.readAuthoritativeStates` reads *every* sibling state before allocating ports, so a single stale stack makes `stack start` fail for brand-new stacks, and `stop`/`destroy` on the old stack fail too. Recovery is `rm -rf` on the state dir plus hand-reaping processes/containers. The whole-stack e2e suite also fails locally on any machine that has ever run `stack start` on develop.

Cheapest fix: default an absent `idleTimeoutSeconds` to `false` in `normalizeDurableState` (it already has a per-capability rewrite loop). Alternatively bump the format tag so `StackStateFormatUnsupportedError` fires and give it an actionable suggestion.

packages/stack/src/supervisor/OperationSettlement.ts
```diff
@@ -0,0 +1,380 @@
...
+    }
+    if (Exit.isSuccess(owner.result))
+      return {
+        snapshot,
+        notification: { _tag: "retirement", completion: owner.completion, result: Exit.void },
+        reconcile: "none",
+      };
+    return {
+      snapshot: {
+        ...stopRecoverySnapshot(snapshot, owner.result.cause),
```
Coly010, line 245: **Blocker: a failed background retirement takes the whole stack offline with no visible reason.**

On retirement cleanup failure this calls `stopRecoverySnapshot`, which flips `running` → `stop-required`. `ensureActivationPhaseAllowed` runs *before* the per-capability match in `activate`, so every capability, including healthy ready/eager ones, is rejected with `Exact runtime cleanup is required; retry stop before activating`. `HttpGateway.mapFailure` collapses that to a bare `503 {"error":"Service unavailable"}`, and the text `stack status` renderer never prints the capabilities array or its `error` field, so the user sees `Lifecycle: stopping` on a stack they never stopped. The only breadcrumb is on the internal log stream.

Fail-closed cleanup is documented design and I agree with it. What's new is the trigger: before this PR the only path here was a user-initiated `stop`; now it fires from an unattended 60s timer per lazy capability, and the macOS EPERM path has a realistic trigger (slow/failed `/bin/ps` spawn or the 2s timeout).

Suggest either (a) for the *retirement* owner specifically, mark only the affected capability `cleanup-failed` and leave the stack `running`, or (b) keep the escalation but surface failed capabilities + the recovery command in status text and distinguish the 503 body for "fenced, run `stack stop && stack start`" vs "waking".

packages/stack/src/supervisor/Supervisor.ts
```diff
@@ -348,120 +1162,214 @@ export const makeSupervisor = (
...
+                      yield* settleCapabilityCleanup(Exit.failCause(operation.cause));
+                    yield* settleOwner({
+                      _tag: "lifecycle",
+                      completion: deferred,
+                      result: operation,
+                    });
+                  }),
+                ),
+              );
+              yield* FiberSet.run(ownedFibers, owner, { startImmediately: true });
```
Coly010: **Should fix: `FiberSet.run` on a closed set never runs the effect, so the Deferred is never settled.**

Effect's `FiberSet.run` short-circuits when the set is closed (`return Effect.sync(constInterruptedFiber)`), so `owner` is never executed and the `onExit`/`settleOwner` that is the sole settler of `owned` never registers. If the supervisor scope closes without `signalShutdown` having run (teardown, crash, interrupt), `ensureAcceptingOperations` still passes, this fork is dropped, and `Deferred.await(owned)` on the next line blocks forever. Same pattern at L1603 for `activate`, where later callers for that capability join the same stuck deferred.

Masked today by finalizer ordering (the gateway's FiberSet closes after the supervisor's and interrupts the waiters), which is accidental rather than structural. Fix: inspect the returned fiber (already-interrupted on a closed set) and settle with a `StackLifecycleConflictError`, or race `Deferred.await(deferred)` against `Fiber.await(fiber)`.

packages/stack/src/public/Config.ts
```diff
@@ -39,6 +39,9 @@ const optionalCapability = <S extends Schema.Top>(settings: S) =>
     Schema.Struct({
       enabled: Schema.optionalKey(Schema.Literal(true)),
       activation: Schema.optionalKey(ActivationModeSchema),
+      idleTimeoutSeconds: Schema.optionalKey(
+        Schema.Union([Schema.Finite.check(Schema.isGreaterThan(0)), Schema.Literal(false)]),
```
Coly010, line 55: **Should fix: the public type is wider than the validator, and the eager case is inconsistent.**

`optionalCapability()` adds `idleTimeoutSeconds` to all nine optional capabilities, but `validateIdleTimeouts` in `Compiler.ts` rejects a number for storage/functions/mail/analytics at runtime. So `{ storage: { idleTimeoutSeconds: 120 } }` type-checks and then fails with `InvalidStackConfigError`. Meanwhile `{ activation: "eager", idleTimeoutSeconds: 120 }` is silently coerced to `false` rather than rejected.

Suggest a `retirableCapability(settings)` schema variant applied only to the five supporting modules (which makes `validateIdleTimeouts` unnecessary), and either reject or document the eager combination. Related: `defaultIdleTimeoutSeconds?: number | false` conflates "unsupported" with "supported but off by default", so a module can't opt into retirement as off-by-default later.

apps/cli/docs/stack-commands.md
```diff
@@ -164,6 +164,13 @@ including `--workdir` and `SUPABASE_WORKDIR`, and prefers JSON when both files e
 
 ## Service selection and shutdown
 
+Lazy REST, Auth, Realtime, Studio, and pooler services stop after 60 seconds without traffic. The
+active traffic lease keeps a capability running. Use `supabase stack start --eager` to keep these
+services running instead of applying idle stopping. A request arriving while a capability is stopping waits
```
Coly010: **Should fix: `--eager` is the only CLI-visible opt-out, and it's a sledgehammer.**

`idleTimeoutSeconds` has zero references outside `packages/stack`; there's no config.toml key or flag. So the only way a CLI user can stop Realtime from retiring every 60s is `--eager`, which also boots Studio, Storage, Functions, Analytics and the pooler up front on every `stack start`.

Also, `start/SIDE_EFFECTS.md` still describes `--eager` only as "requests enabled capabilities be activated before the command returns"; its new effect (disables idle stopping for the session) isn't recorded in the flag's own contract doc.

Suggest documenting the second effect in `start/SIDE_EFFECTS.md`, and either adding a CLI-visible per-service override or stating plainly here that per-service timeouts are Effect-API-only for now. Minor: "active traffic lease" and "fence" are supervisor vocabulary; this doc reads better in terms of commands and observable behaviour.

packages/stack/src/model/capabilities/realtime.ts
```diff
@@ -25,6 +25,7 @@ export const RealtimeModule: CapabilityModule<RealtimeSettings> = {
   },
   defaultEnabled: true,
   defaultActivation: "lazy",
+  defaultIdleTimeoutSeconds: 60,
```
Coly010, line 28: **Should fix: 60s is aggressive for realtime and pooler (same line in `pooler.ts`).**

There's no activation-side deadline on the wake path, so the first request after retirement waits on the readiness budget (30s container, 2min native). For REST/Auth/Studio that's a tolerable cold-start stall. For realtime, `realtime-js`/phoenix reconnects on a ~10s timer, so the first reconnect after retirement will often time out client-side while the container is still booting, and on upgrade failure the socket is destroyed with no close frame. For the pooler, a pooled app that idles >60s between bursts eats a Supavisor cold start behind a silent TCP hang each time, and on failure libpq reports `server closed the connection unexpectedly`.

Suggest a longer default for these two (or `false` for pooler), and ideally a bounded activation deadline with a distinguishable 503/504 body so clients can tell "retry shortly" from "broken".

packages/stack/src/supervisor/Supervisor.ts
```diff
@@ -479,13 +1387,16 @@ export const makeSupervisor = (
         for (const result of [ingress, launched, driver])
           if (Exit.isFailure(result)) cause = Cause.combine(cause, result.cause);
         if (cause.reasons.length > 0) {
-          yield* Ref.set(cleanupProven, false);
+          yield* settleCapabilityCleanup(Exit.failCause(cause));
```
Coly010, line 1422: **Should fix: a stack-wide cleanup failure is attributed to every stopping capability.**

`SessionLauncher` now computes precise per-workload `SessionCleanupOutcome`s, but `cleanupOrFail` flattens them to a single `SessionCleanupError`, `cleanupRuntime` merges that with the ingress/driver causes, and this line marks *every* `stopping` capability `cleanup-failed`. Scenario: `stack stop` where every workload was removed and proven but `runtime.ingress.close` fails; all capabilities report `state: "failed"` with the ingress error and the stack goes to `stop-required`. Recovery still works because the ledger was pruned per entry, but it's a misdiagnosis and contradicts the ADR 0017 amendment ("unrelated healthy capabilities retain their observations"). Threading the per-entry outcomes through to settlement would fix this.

packages/stack/src/supervisor/Supervisor.ts
```diff
@@ -274,21 +416,640 @@ export const makeSupervisor = (
...
+                _tag: "workload",
+                deferred: completion,
+              }),
+            );
+          }
+          yield* Ref.set(machine, { ...snapshot, capabilities });
+          return claimed;
+        }),
+      );
+    const restoreActivationFailureInAdmission = (
```
Coly010, line 537: **Should fix (fine to defer): the reducer extraction stopped at terminal settlement.**

`OperationSettlement.ts` establishes exactly the right pattern: pure `(snapshot, owner) → { snapshot, notification, reconcile }`, unit-tested as a reducer. But this ~80-line function, plus `setReadySetInAdmission`, `claimWorkloads`, `promoteActivationSet`, `settleStartupFailures`, `enter/settleCapabilityCleanup`, `begin/endTraffic`, are the same shape and stayed inline, which is why `Supervisor.ts` is now 2053 lines with 20 `Ref.set/update(machine, …)` sites. This function and `settleActivationOwner` both decide the fate of a `starting` capability on failure with slightly different rules, in two files; that's the divergence class one of the earlier threads already caught once.

Also `SupervisorState.command` (a pure 3×8 admission table) and `CapabilityState`'s transitions have no direct unit tests, only indirect coverage through the 4400-line integration file. Happy for this to be a follow-up, but worth recording.

Coly010 (changes requested):
Really strong core: the terminal-settlement reducer, the traffic lease acquired before activation and disarmed in the same critical section, whole-connection leases for WebSocket/TCP, health checks that neither wake nor pin, and the fail-closed `/bin/ps` gate are all the right shape. Locally: unit + integration (774, run twice, no flakes) + CLI status suites pass, `pnpm check:all` clean.

Requesting changes for two issues outside the state machine itself, both reproduced or traced end to end, plus a handful of should-fixes. Details inline.

**Blockers**
1. Persisted `state.json` from current develop no longer decodes (exact-key check gained `idleTimeoutSeconds`, no format bump, no normalizer entry). Reproduced on my machine; because the port coordinator reads every sibling state, one stale stack blocks `stack start` for *new* stacks and `destroy` of the old one.
2. A failed *background* idle retirement escalates the whole stack to `stop-required`, so every capability 503s with a generic body and `stack status` text gives no reason.

**Should fix**
- `FiberSet.run` on a closed set drops the owner effect, leaving settlement Deferreds unsettled forever.
- `idleTimeoutSeconds` is typed on all optional capabilities but rejected at runtime for four of them; a number on an eager capability is silently discarded.
- No CLI/config surface for the knob; `--eager` is documented as the only opt-out and its contract doc doesn't mention the new effect.
- 60s default is aggressive for realtime and pooler.
- Stack-wide cleanup failure is attributed to every stopping capability, contradicting the ADR amendment.
- `Supervisor.ts` is 2053 lines with most transitions still inline; fine to defer, but flagging.

packages/stack/src/state/StackState.ts
jgoux: Fixed in [c2ebab63b](https://github.com/supabase/cli/commit/c2ebab63b2ff216209e1a64f699e8aa72743be8a). The existing durable-state normalizer now supplies `idleTimeoutSeconds: false` only when the field is absent, preserving the previous behavior and keeping explicitly invalid values invalid. Regressions cover reading a develop-era definition and allocating/persisting ports for a sibling stack while that old state exists. The format identifier stays unchanged, so cleanup remains available instead of merely producing a different decoding error. Older saved definitions retain idle stopping as disabled; restart preserves that saved policy, while stop/start with project configuration adopts the current defaults. The upgrade behavior and possible effective-default drift are now documented.

packages/stack/src/supervisor/OperationSettlement.ts
jgoux: Fixed using option (b) in [c2ebab63b](https://github.com/supabase/cli/commit/c2ebab63b2ff216209e1a64f699e8aa72743be8a). We retain the fail-closed recovery gate, but status now exposes the recovery operation and its cause, and CLI text prints capability diagnostics plus the stop/start commands addressed to that stack ID. HTTP activation failures caused by this gate return `STACK_RECOVERY_REQUIRED` with the required operation and safe guidance. The gateway does not reflect arbitrary cleanup/backend error details; those remain in status. Recovery rejection before a WebSocket upgrade now returns HTTP 503 as well. Regressions cover failed-retirement status, ingress propagation, HTTP bodies, and upgrade rejection. The upgrade rejection closes the socket after flushing the response, including peers that leave their side open.

packages/stack/src/supervisor/Supervisor.ts
jgoux: Fixed in [c2ebab63b](https://github.com/supabase/cli/commit/c2ebab63b2ff216209e1a64f699e8aa72743be8a). Both lifecycle and activation submissions inspect the returned owner's immediate completion; an already-interrupted owner is settled with `StackLifecycleConflictError` through the same terminal-settlement reducer. Settlement runs in the admission section already held by submission, so it cannot re-acquire that permit or leave the machine in an in-progress state. This uses the single existing completion Deferred, without adding a second waiter/latch. Tests exercise calls retained beyond owner-scope closure as well as interruption of an existing owner.

packages/stack/src/public/Config.ts
jgoux: Fixed in [c2ebab63b](https://github.com/supabase/cli/commit/c2ebab63b2ff216209e1a64f699e8aa72743be8a) with the suggested capability-specific schema. Only REST/Auth/Realtime/Studio/pooler accept `idleTimeoutSeconds`; the redundant compiler validation is removed. Eager activation continues to win over an explicit timeout, and the README now states that its effective timeout becomes `false`. I have not added separate support/default metadata yet: there is no current supporting module with an off-by-default policy, and removing the validator removes the present coupling that rejected such a future default.

apps/cli/docs/stack-commands.md
jgoux: Addressed through the documentation alternative in [c2ebab63b](https://github.com/supabase/cli/commit/c2ebab63b2ff216209e1a64f699e8aa72743be8a). CLI docs now state plainly that per-service timeout values are Effect-API-only. `--eager` help and its SIDE_EFFECTS contract describe both activating all enabled capabilities and disabling idle stops. The docs explain active requests, open WebSocket/TCP connections, and idle HTTP keep-alive in user terms. A per-service CLI/config surface remains separate work rather than adding an unplanned configuration contract to this PR.

packages/stack/src/model/capabilities/realtime.ts
jgoux: The cold-start concern for a later new connection is valid, but I am keeping the defaults unchanged in this PR pending measurements. An open WebSocket or TCP pooler connection already holds the service active even with no traffic, so an existing pooled connection does not retire between bursts. The realtime reconnect schedule is distinct from a handshake deadline, and the runtime readiness budgets are upper bounds rather than measured cold-start latency. [c2ebab63b](https://github.com/supabase/cli/commit/c2ebab63b2ff216209e1a64f699e8aa72743be8a) improves the recovery-required response, including HTTP 503 before a WebSocket upgrade; a WebSocket close frame would not be valid before that upgrade. Default tuning and a separate activation deadline are recorded for follow-up with actual native/container first-connect measurements. TCP failures remain protocol-neutral closure; this PR does not claim a new distinguishable TCP error protocol.

packages/stack/src/supervisor/Supervisor.ts
jgoux: The shared-cleanup attribution scenario is accurate. [c2ebab63b](https://github.com/supabase/cli/commit/c2ebab63b2ff216209e1a64f699e8aa72743be8a) adds a separate root recovery operation/reason to public status and clarifies the existing operation-level contract in the README/ADR: participating capability `failed` states describe incomplete cleanup, not necessarily failed workload processes. An explicit stop includes all stopping capabilities; the statement about unrelated healthy observations applies to capabilities outside that cleanup operation. The exact remaining resource ledger still controls retries. I have retained that state model here. `SessionCleanupOutcome` is an aggregate result at this boundary, so precise per-workload public status requires carrying additional results into settlement and defining shared ingress/driver failure separately. That finer-grained status change is recorded alongside the reducer follow-up rather than adding another partial accounting path in this PR.

packages/stack/src/supervisor/Supervisor.ts
jgoux: Agreed; I am keeping this for a separate follow-up PR. The follow-up will consolidate the remaining related transition decisions (including starting-capability failure rules) behind cohesive pure reducers and add focused command/capability transition tests. I want that work to reduce duplicated decisions and replace redundant coverage, rather than add a mechanical assertion for every table cell. This PR preserves the existing terminal settlement ordering and resource ownership safeguards.

Coly010 (approved):

PR #6615, opened 2026-09-15
fix(cli): report a killed edge runtime container instead of a lost log stream

Prashansa-K:
## TL;DR

An edge runtime container that gets killed during `supabase functions serve` — most often by the out-of-memory killer — is retried until the re-attach cap gives up, and then reported as a lost log stream against a container the message claims is still running. This reports the kill instead, and separates a memory-limit kill from one the CLI cannot attribute.

Refs [CLI-2427](https://linear.app/supabase/issue/CLI-2427) — *`functions serve` misreports a killed edge runtime container as a lost log stream (exit 137)*. This is the `137` gap [#6594](https://github.com/supabase/cli/pull/6594) left open.

## What was wrong

**The retry can never succeed.** `streamContainerLogs` treats container exit `137` as retriable, but the container is created with no `--restart` policy, so once it is killed it stays dead. `docker logs -f` is respawned, exits, the container is inspected, `137` comes back.

**The resulting message says the opposite of what happened.** After the cap the command fails with `lost the Edge Runtime log stream 5 times; container … is still running`. The container is not running — that is the one fact the preceding `docker container inspect` established.

**The classification sends the user the wrong way.** `EdgeRuntimeLogStreamLostError` declares `external_service` / `network` with a "rerun with `--debug`" remediation. An out-of-memory kill is not a network problem, and that remediation leads nowhere.

The retry also contradicts both of its neighbours: `apps/cli-go/internal/utils/docker.go` maps `137` to `ErrContainerKilled` with no retry, and `db diff` treats `137` as a genuine failure.

## What changed

`137` falls through to `EdgeRuntimeContainerCrashedError`, which now carries `State.OOMKilled` from the `docker container inspect` the code already makes. That call uses `--format "{{json .State}}"`, so this is one more field read in `parseContainerState` — no format string changes.

| Container state | `error_kind` / `error_category` | Suggestion |
| --- | --- | --- |
| `137`, `OOMKilled: true` | `user_actionable` / `resource_limit` | raise the container runtime's memory allocation |
| `137`, `OOMKilled: false` | `unknown` / `unknown` | none |
| `129`/`130`/`131`/`143` | unchanged — the session ends successfully | — |
| any other non-zero | unchanged — `internal_bug` / `runtime_crash` | — |

`resource_limit` is a new `user_actionable` category. Both `137` rows get their own fingerprint suffix, so they stay separable from each other and from the untagged bulk of their categories.

## Two calls worth reviewing

**An out-of-memory kill is not `internal_bug` / `runtime_crash`.** It is Docker's memory allocation or the user's own function, and the user can act on it. Filing it against the internal-bug counter would be wrong in the direction this work exists to correct.

**A non-OOM `137` is `unknown`, not `user_cancelled`.** Docker sets `OOMKilled` reliably only for container-limit OOMs; a host or VM out-of-memory kill on macOS lands here with `OOMKilled: false`. Calling that "the user cancelled" asserts something we do not know. `unknown` with a named fingerprint keeps it separable so it can be reclassified once there is data on which cause dominates — at the cost of knowingly adding to the `unknown` bucket.

## Behaviour change to be aware of

A session whose container is killed now fails immediately rather than after five re-attaches, and fails with a different tagged error than before. A `functions serve &` CI step that was passing through the re-attach window will now fail at the kill.


Prashansa-K:
Working on the third filed minor issue `When docker logs -f exits non-zero and inspection finds a stopped OOM-killed container, the code reports a generic log-stream error instead of the container crash and memory remediation.` in a separate PR, so as to not make this one too broad.

apps/cli/src/command-internal/docker-suggest.ts
```diff
@@ -8,6 +8,10 @@ export const SUGGEST_DOCKER_INSTALL =
 export const SUGGEST_DOCKER_START =
   "Docker is no longer reachable. Start Docker, then rerun `supabase functions serve`.";
 
+/** Remediation hint shown when the OOM killer kills a container. */
+export const SUGGEST_CONTAINER_MEMORY_LIMIT =
+  "The container was killed for using too much memory. Raise its memory limit, or free up memory on the host, and try again.";
```
Coly010: Names both remedies that are actually reachable, leads with the one that fits most setups, and drops the implication of a container-level limit the CLI never sets. Phrased as "your container runtime" rather than Docker Desktop alone, since `serve` also runs under podman, Colima, and OrbStack.

```suggestion
export const SUGGEST_CONTAINER_MEMORY_LIMIT =
  "The container ran out of memory and was killed. Raise the memory allocated to your container runtime (Docker Desktop: Settings > Resources), or free up memory on the host, and try again.";
```

Coly010 (changes requested):
Request changes on one user-facing string.

`SUGGEST_CONTAINER_MEMORY_LIMIT` tells the user to "raise its memory limit", but `functions serve` never sets one. The `docker create` argv at `apps/cli/src/shared/functions/serve.ts:1902` passes `--ulimit nofile` and no memory flag, and there is no `--memory` argument or memory key anywhere in `apps/cli/src` or `packages/config/src`.

So `State.OOMKilled: true` here can only mean the container's cgroup inherited its parent's limit and exhausted it — the host's RAM on Linux, the Docker Desktop / Colima / OrbStack VM's allocation on macOS and Windows. There is no per-container limit for the user to raise: `config.toml` has no key for it, `serve` has no flag for it, and `docker update --memory` evaporates on the next restart since the container is recreated each time.

That matters more than usual in this PR. The whole point of the new `user_actionable` / `resource_limit` classification is to replace a remediation that sent users somewhere useless (`external_service` / `network` plus "rerun with `--debug`"). As written, the replacement sends them hunting for a config key that does not exist, and the remedy that actually fits most machines — raising the container runtime's own memory allocation — is not mentioned at all.

Suggestion inline. I have left the constant's name alone: `SUGGEST_CONTAINER_OUT_OF_MEMORY` would describe the event rather than a configured limit, but that rename also touches `serve.errors.ts` and both test files, so a single-file suggestion cannot carry it. Your call whether it is worth doing here.

Scoped to this one ask.

Coly010 (approved):

PR #6616, opened 2026-09-15
chore(api): refresh production contracts and compute routes

johnstonmatt:
Refresh `@supabase/api` from the production OpenAPI documents and switch the CLI's compute commands to the deployed `/compute` routes, operation names, and resource types.

Remove the temporary compute bridge and obsolete hostname/storage corrections, and retarget surviving overrides to the renamed response schemas. Update CLI schema imports, handle missing hostname SSL details and null storage migration versions, and align the internal project-config mirror with the current storage response. Webhook exclusions remain because their OpenAPI 3.0 response schemas contain `const` keywords unsupported by the converter.

Supersedes #6622.


packages/api/scripts/openapi-overrides.json
```diff
@@ -723,5 +723,819 @@
       "description": "Postgres engine version. If not provided, the latest version will be used."
     },
     "$comment": "CLI-2180: the public spec deliberately hides this field (upstream marks it deprecated/null) even though POST /v1/projects accepts it; enum mirrors CreateBranchBody.postgres_engine in the same spec."
+  },
+  {
+    "op": "add",
+    "path": "/components/schemas/V2ComputeInstanceResponse_Output",
```
jgoux: **[P2] Remove the bridge now that production exposes compute**

The block's comment says these additions will overwrite identical values once production catches up, but `addJsonPointerValue` throws whenever the key already exists (`download-openapi.ts:124–125`). Production now exposes the compute routes. Applying just this new block to a fresh production v2 spec with the PR's downloader fails immediately with `JSON pointer "/components/schemas/V2ComputeInstanceResponse_Output" cannot be added.`

The temporary compute override block can now be deleted and the client regenerated from production. Keeping it will block the hourly sync at generation, before a refresh PR can be opened.

jgoux: Applied in 05b842008. The temporary compute bridge is removed, the client is regenerated from production, and the remaining overrides target the current response schemas.

jgoux (approved):

PR #6646, opened 2026-09-16
chore(stack): centralize supervisor lifecycle transitions

jgoux:
Supervisor lifecycle decisions were spread across orchestration and settlement helpers, making ownership and completion ordering difficult to follow. Consolidate command admission, workload claims, startup outcomes, cleanup, traffic leases, and idle retirement into local pure transitions, with one ordered path for committing state, reconciling timers, and notifying callers.

Keep resource operations in Effect orchestration, share the active lifecycle projection, and replace the separate settlement module and cleanup fallback with owner-fenced transitions and local cleanup ownership. Register activation settlement before restoring interruption after workload claims, and settle lifecycle outcomes before releasing execution permits. Preserve closed-owner recovery, existing diagnostics, and saved-state behavior.

Follow-up to #6613.


packages/stack/src/supervisor/Supervisor.ts
```diff
@@ -1400,37 +961,44 @@ export const makeSupervisor = (
...
-        }
-        yield* settleCapabilityCleanup(Exit.void);
-        yield* completeDormantCleanup();
-        if (!destroy) {
-          yield* setRootSet(new Set());
-        } else {
-          yield* launcher.clear;
+      Effect.uninterruptibleMask((restore) =>
+        Effect.gen(function* () {
+          const admissionPermit = yield* restore(admission.take(1)).pipe(Effect.exit);
```
Coly010: **Blocker: an interrupt during this acquisition leaks the admission permit and wedges the supervisor.**

`Semaphore.take` is `suspend(() => { taken += n; return succeed(n) })`. The run loop may insert a scheduler yield (`MaxOpsBeforeYield`, default 2048 ops) between that increment and the `Effect.exit` continuation here. Because the take is under `restore`, the fiber is interruptible during that yield: an interrupt replaces the pending success with the interrupt cause, `Exit.isFailure(admissionPermit)` is true, this branch returns without releasing, and `admission.taken` stays at 1 for the rest of the process. Every `*InAdmission` call then blocks, several of them uninterruptibly (line 995 is inside this mask), so `Scope.close` on the owner never completes while `status` keeps answering. The base used `admission.withPermit(...)` here, which does `taken += n` and installs the `onExit` release in one uninterruptible step and restores interruptibility only around the wait and the body.

Reproduced against the pinned rc.112: scanning `MaxOpsBeforeYield` 2..10 × interrupt offsets 0..40, this shape leaked the permit in 2 of 369 runs (e.g. `maxOps=3,k=6`, `maxOps=9,k=0`); `withPermit` leaked in 0 of 369.

<details><summary>Reproduction (run from <code>packages/stack</code> so <code>effect</code> resolves to rc.112)</summary>

```ts
import { Effect, Exit, Fiber, Semaphore, Scheduler } from "effect";

// Same shape as this function: interruptible take, release registered afterwards.
const takeRelease = (sem: Semaphore.Semaphore) =>
  Effect.uninterruptibleMask((restore) =>
    Effect.gen(function* () {
      const acquired = yield* restore(sem.take(1)).pipe(Effect.exit);
      if (Exit.isFailure(acquired)) return yield* Effect.failCause(acquired.cause);
      return yield* Effect.ensuring(restore(Effect.void), sem.release(1));
    }),
  );

const attempt = (maxOps: number, k: number) =>
  Effect.gen(function* () {
    const sem = Semaphore.makeUnsafe(1);
    const fiber = yield* Effect.forkChild(
      takeRelease(sem).pipe(Effect.provideService(Scheduler.MaxOpsBeforeYield, maxOps)),
      { startImmediately: true },
    );
    for (let i = 0; i < k; i++) yield* Effect.yieldNow;
    yield* Fiber.interrupt(fiber);
    yield* Fiber.await(fiber);
    return (sem as unknown as { taken: number }).taken; // 1 means the permit leaked
  });
```

Swap `takeRelease(sem)` for `Effect.uninterruptibleMask((restore) => restore(sem.withPermit(Effect.void)))` and no combination leaks.
</details>

Fix keeps the interruptible wait and the masked commit:

```ts
const handles = yield* restore(
  admission.withPermit(Effect.uninterruptible(enterCapabilityCleanupInAdmission())),
);
```

The same shape is at line 773 (`execution`, new in this PR) and line 1055 (`execution`, from #6613). Worth fixing all three in one pass.

packages/stack/src/supervisor/Supervisor.ts
```diff
@@ -1190,43 +739,55 @@ export const makeSupervisor = (
...
+                  return operation;
+                });
+              const owner = Effect.uninterruptibleMask((restore) =>
+                Effect.gen(function* () {
+                  const cancelled = yield* restore(cancelIdleTimers).pipe(Effect.exit);
+                  if (Exit.isFailure(cancelled)) {
+                    yield* finish(Exit.failCause(cancelled.cause));
+                    return yield* Effect.failCause(cancelled.cause);
+                  }
+                  const acquired = yield* restore(execution.take(1)).pipe(Effect.exit);
```
Coly010: **Blocker, same class as line 966: a lost `execution.take` here leaks the permit.** The next `start` is then admitted, flips the stack to `starting`, and blocks on `execution.take(1)` forever with its completion Deferred never settled.

The settle-before-release ordering this PR introduces survives with `withPermit`, because it releases in `onExit` after the body:

```ts
return yield* restore(
  execution.withPermit(
    Effect.uninterruptibleMask((inner) =>
      Effect.gen(function* () {
        const result = yield* inner(effect).pipe(Effect.exit);
        const operation = yield* finish(result); // still before the release
        return yield* joinExit(
          Predicate.isTagged(operation, "failed")
            ? Exit.failCause(operation.cause)
            : Exit.succeed(operation),
        );
      }),
    ),
  ),
);
```

and the explicit `Effect.ensuring(..., execution.release(1))` goes away.

packages/stack/src/supervisor/SupervisorTransitions.unit.test.ts
```diff
@@ -0,0 +1,399 @@
...
+      expect(settlement.snapshot.stack).toEqual({ _tag: "running" });
+      const notification = settlement.notifications[0];
+      if (notification !== undefined && Predicate.isTagged("activation")(notification)) {
+        expect(notification.completion).toBe(completion);
+        expect(notification.result).toEqual(Exit.failCause(cause));
+      } else expect.fail("expected activation notification");
+    }),
+  );
+
+  it.effect("does not restore claims into a newer session", () =>
```
Coly010, line 268: **Should fix: this test returns before the guard it is named after.**

`ready()` defaults `endpoint` to `unresolved`, so `matchesActivationOwner` is false for the `endpoint` owner and `settleActivationTerminal` returns at line 901 without ever reading `claims`. Delete the whole claim loop (911–939), including the `current.sessionId === entry.prior.sessionId` check, and this still passes; `toBe(snapshot)` is satisfied by the early return. The other two staleness tests (179, 210) also go stale via the state *tag*, so replacing both Deferred-identity comparisons in `matchesActivationOwner` with `true` passes the whole suite.

Suggest giving the owner a matching `resolving` endpoint, putting the claim on a different capability under a newer session, and asserting that capability is untouched; and turning one of the three into a genuine Deferred-mismatch case. Separately, I could not find a reachable path to the session guard (a session reset needs `execution`, which the activation holds from claim to settlement), so dropping the guard instead of testing it is also defensible.

packages/stack/src/supervisor/SupervisorTransitions.ts
```diff
@@ -0,0 +1,1229 @@
...
+      },
+    ],
+    reconcile: "none",
+  };
+};
+
+const commandResultExit = (operation: CommandResult): Exit.Exit<void, StackError> =>
+  Predicate.isTagged(operation, "failed") ? Exit.failCause(operation.cause) : Exit.void;
+
+const settleStartingFailure = (
```
Coly010, line 1107: **Should fix: the rule this follow-up was chartered to consolidate has no focused test.**

Your #6613 reply committed to "focused command/capability transition tests" for the starting-capability failure rules. `admitLifecycle`, `activationGate` and `admitActivation` got them; `settleStartingFailure` here and `settleStartingCapabilities` (855) are only reached indirectly from integration tests, and both `settleLifecycleOwner` unit tests pass `succeeded`. Four rows cover it: `starting(prior: running)` × {proven, unproven} and `starting(prior: stopped)` × {proven + durable `stopped`, proven + durable `unsafe`}, asserting `snapshot.stack`, the capability dispositions, and that each workload Deferred appears in `notifications`.

While there: `settleRetirementOwner` has one of four outcomes covered. `Exit.succeed(false)` (declined) versus `Exit.succeed(true)` (retired) is the branch where dropping `.value` would demote a live `ready` capability to `dormant`, and nothing catches that today.

packages/stack/src/supervisor/SupervisorTransitions.ts
```diff
@@ -0,0 +1,1229 @@
...
+        _tag: "endpoint",
+        completion: owner.endpoint,
+        result: !failed ? Exit.succeed(outcome.value.endpoint) : Exit.failCause(outcome.cause),
+      }
+    : {
+        _tag: "activation",
+        completion: owner.completion,
+        result: !failed ? Exit.succeed(outcome.value) : Exit.failCause(outcome.cause),
+      };
+  if (!ownerMatches) return { snapshot, notifications: [notification], reconcile: "all-ready" };
```
Coly010, line 908: **Decide deliberately: this early return also skips the unproven → `stop-required` escalation.**

The base applied `stopRecoverySnapshot` whenever cleanup was `unproven`, independent of owner identity (`restoreActivationFailureInAdmission` fenced per claim entry, not per owner), so unproven evidence about the runtime always reached the recovery gate. Now a stale owner with unproven cleanup leaves the stack `running`. I agree with the AI review that this is unreachable today, because the activation holds `execution` from claim to settlement, but the unit test at 210 ("ignores a stale unproven activation failure without claims") pins the fail-open shape as intended. Either hoist the escalation (and the per-entry-fenced claim loop) above this check, or say in the test why stale unproven evidence is safe to discard.

Coly010 (changes requested):
The consolidation delivers what the #6613 follow-up promised: every machine write now goes through one `applyTransitionInAdmission` (commit → reconcile timers → notify), the two rules that decided a `starting` capability's fate on failure are one function, cleanup settlement is fenced by handle identity instead of relying on the `submitLifecycle` fallback, and `activationGate` drops the two `ensureActivationPhaseAllowed` branches that `activeCommand()` had already made unreachable. I traced the eight behaviours that landed during #6613's review rounds (closed-owner settlement, settle-before-release, stopped-stack activation failure, combined cleanup evidence, symmetric cleanup-failed leases, the retirement recovery gate, timer fencing, durable-state ordering) and all survive. Locally: unit 13/13; supervisor, lifecycle, status-projection and observability integration 145/145. The other local integration failures are the known listener-environment issue on my machine, and CI is green.

Requesting changes for one issue, reproduced against the pinned Effect: the hand-rolled `restore(take(1))` … `release(1)` protocol that replaced `withPermit` on both semaphores can leak a permit under interruption, and on `admission` that wedges the whole supervisor. Details and a runnable reproduction inline.

**Blocker**
- `restore(sem.take(1)).pipe(Effect.exit)` followed by a later `release(1)` is not interrupt-safe in rc.112 (`Supervisor.ts:966` on `admission`, `:773` on `execution`; `:1055` has the same shape from #6613). `withPermit` is, and it still allows settle-before-release.

**Should fix**
- "does not restore claims into a newer session" returns before the guard it is named after (the AI review's MINOR is right), and none of the three staleness tests exercise a Deferred-identity mismatch.
- The starting-capability failure rule (`settleStartingFailure` × `settleStartingCapabilities`) and three of the four `settleRetirementOwner` outcomes have no focused unit test, which was the chartered purpose of this follow-up.
- `docs/adr/0017-simplified-managed-stack-architecture.md:308` (not in this diff, so no inline anchor): the §Consequences paragraph still says only terminal owner settlement is the pure step and that workload-readiness, rollback and cleanup notifications "remain separate". After this PR every transition returns `{ snapshot, notifications[], reconcile }` and `applyTransitionInAdmission` is the single commit → reconcile → notify path for all of them, intermediate notifications included. One paragraph stating the general contract keeps the ADR honest.
- `SupervisorTransitions.ts:901`: the `!ownerMatches` short-circuit now also skips the unproven→`stop-required` escalation the base applied unconditionally. Unreachable today, but a unit test now pins the fail-open shape, so accept or reject it deliberately.

**On the AI review:** the codex MAJOR (activation waits for admission while masked) is a bounded delay, not a hang; your new test at `supervisor.integration.test.ts:1850` drives exactly that interleaving and shows scope close completing. The real shutdown-hang risk is the permit leak above, which it missed. The MINOR is confirmed. The nits are fine to take or leave.

packages/stack/src/supervisor/SupervisorTransitions.unit.test.ts
```diff
@@ -0,0 +1,399 @@
...
+          _tag: "claimed",
+          claimed: [{ name: "rest", completion, prior }],
+          affected: new Set(["rest"]),
+        },
+        { _tag: "failed", cause, cleanup: { _tag: "proven" } },
+      );
+      expect(settlement.snapshot).toBe(snapshot);
+      expect(settlement.snapshot.capabilities.get("rest")).toBe(current);
+    }),
+  );
```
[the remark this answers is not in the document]
jgoux: Updated the test to use a matching resolving endpoint owner and a separate claimed capability from a newer session. It now reaches claim restoration and checks that the newer capability is preserved.

Addressed in 2706ea914.

packages/stack/src/supervisor/SupervisorTransitions.ts
```diff
@@ -0,0 +1,1229 @@
...
+      mutated = true;
+    }),
+    Match.exhaustive,
+  );
+  const changed = mutated || (failed && Predicate.isTagged(cleanup, "unproven"));
+  let next = !changed
+    ? snapshot
+    : failed && Predicate.isTagged(cleanup, "unproven")
+      ? stopRecoverySnapshot({ ...snapshot, capabilities }, cleanup.cause)
+      : { ...snapshot, capabilities };
```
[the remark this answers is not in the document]
jgoux: Removed the redundant mutation/change flags and simplified the terminal snapshot calculation.

Addressed in 2706ea914.

packages/stack/src/supervisor/Supervisor.ts
```diff
@@ -443,324 +415,106 @@ export const makeSupervisor = (
...
-        yield* Ref.set(machine, { ...snapshot, capabilities });
+        const handles = new Map<CapabilityName, CleanupHandle>();
+        for (const name of CAPABILITY_NAMES)
+          handles.set(name, {
+            operation: Symbol("cleanup"),
+            completion: yield* Deferred.make<Exit.Exit<void, StackError>, never>(),
+          });
+        const transition = transitionEnterCapabilityCleanup(yield* Ref.get(machine), handles);
+        yield* applyTransitionInAdmission(transition);
+        return handles;
```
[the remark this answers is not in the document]
jgoux: Cleanup now reads the admitted snapshot first and allocates handles only for ready or cleanup-failed capabilities.

Addressed in 2706ea914.

packages/stack/src/supervisor/Supervisor.ts
```diff
@@ -1639,145 +1213,68 @@ export const makeSupervisor = (
...
-                Match.exhaustive,
+              const endpoint = yield* Deferred.make<EndpointExit, never>();
+              const activation = yield* Deferred.make<ActivationExit, never>();
+              const decision = admitActivation(
+                yield* Ref.get(machine),
+                capability,
+                options.stackId,
+                endpoint,
+                activation,
+                Symbol("activation"),
```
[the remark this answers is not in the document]
jgoux: Admission now returns a pure owner plan. Orchestration allocates the endpoint Deferred only for endpoint ownership, or the activation Deferred and operation symbol only for activation ownership. Respond/rejected decisions allocate neither, and classification remains centralized.

Addressed in 2706ea914.

packages/stack/src/supervisor/SupervisorTransitions.ts
```diff
@@ -0,0 +1,1229 @@
...
+  readonly notifications: ReadonlyArray<TransitionNotification>;
+  readonly reconcile: "all-ready" | "none";
+}>;
+
+type TransitionState =
+  | Extract<StackControlState, { readonly _tag: "starting" }>
+  | Extract<StackControlState, { readonly _tag: "stopping" }>
+  | Extract<StackControlState, { readonly _tag: "destroying" }>;
+
+export type LifecycleAdmission =
```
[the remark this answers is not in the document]
jgoux: Removed the six unused type exports.

Addressed in 2706ea914.

packages/stack/src/supervisor/SupervisorTransitions.ts
```diff
@@ -0,0 +1,1229 @@
+import { Cause, Deferred, Exit, Fiber, Match, Predicate } from "effect";
+import type { ActivationResult } from "../gateway/Gateway.ts";
+import type { StackId } from "../public/StackId.ts";
+import {
+  GatewayActivationError,
+  StackLifecycleConflictError,
+  StackNotRunningError,
+  type GatewayActivationError as GatewayActivationErrorType,
```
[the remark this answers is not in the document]
jgoux: Removed the aliased type import and used GatewayActivationError directly in type positions.

Addressed in 2706ea914.

packages/stack/src/supervisor/SupervisorTransitions.ts
```diff
@@ -0,0 +1,1229 @@
...
+  | { readonly _tag: "respond"; readonly token: ActivationToken }
+  | {
+      readonly _tag: "endpoint-owner";
+      readonly owner: Extract<ActivationOwner, { readonly _tag: "endpoint" }>;
+      readonly snapshot: SnapshotTransition;
+    }
+  | {
+      readonly _tag: "activation-owner";
+      readonly owner: Extract<ActivationOwner, { readonly _tag: "activation" }>;
+      readonly snapshot: SnapshotTransition;
```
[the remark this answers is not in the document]
jgoux: Renamed the decision field to transition. It now describes the pure transition plan completed with the allocated owner handles.

Addressed in 2706ea914.

packages/stack/src/supervisor/supervisor.integration.test.ts
```diff
@@ -175,25 +182,42 @@ const makeFixture = (
...
+                  const armed = yield* Ref.modify(fixtureOptions.shutdownReadArmed!, (value) => [
+                    value,
+                    false,
+                  ]);
+                  if (armed) {
+                    yield* Deferred.succeed(fixtureOptions.shutdownReadStarted!, undefined);
+                    yield* Deferred.await(fixtureOptions.shutdownReadGate!);
+                  }
+                  return yield* persistedStore.read(stackId);
+                }),
```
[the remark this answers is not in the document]
jgoux: Migrated the remaining shutdown-read test to the shared read-gate queue and removed the legacy fixture options. Gates preserve the explicitly selected before-read or after-read ordering.

Addressed in 2706ea914.

packages/stack/src/supervisor/Supervisor.ts
```diff
@@ -1445,190 +1013,196 @@ export const makeSupervisor = (
...
+                    { _tag: "none" },
+                    {
+                      _tag: "failed",
+                      cause: prepared.cause,
+                      cleanup: { _tag: "proven" },
+                    },
+                  );
+                  return yield* Effect.failCause(prepared.cause);
+                }
+                const claimed = yield* claimWorkloads(prepared.value.selected).pipe(Effect.exit);
```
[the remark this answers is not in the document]
jgoux: Activation admission waits now run interruptibly, with claim registration masked inside the admitted section and terminal settlement before execution release. The regression checks cancellation and the resulting capability state. The later human review also identified a distinct permit-leak risk; the manual acquisition/release sequences have been replaced by native withPermit.

Addressed in 2706ea914.

packages/stack/src/supervisor/Supervisor.ts
jgoux: Reproduced the permit leak against the pinned Effect version using an observable permit-availability check. Replaced the manual admission take/release with native withPermit. Cleanup handles are registered inside the masked body before interruption can resume, so cancellation cannot lose cleanup ownership.

Addressed in 2706ea914.

packages/stack/src/supervisor/Supervisor.ts
jgoux: Both lifecycle and activation execution paths now use native withPermit. Once the body starts, terminal settlement finishes before the permit is released; cancellation while waiting to enter separately settles the admitted owner.

Addressed in 2706ea914.

packages/stack/src/supervisor/SupervisorTransitions.unit.test.ts
jgoux: The session-guard test now uses a matching resolving owner and a different claimed capability with a newer session. Endpoint and activation staleness coverage also use genuine Deferred identity mismatches. The existing session guard remains.

Addressed in 2706ea914.

packages/stack/src/supervisor/SupervisorTransitions.ts
jgoux: Added focused starting-failure cases for the four requested cleanup/durable-state combinations, asserting stack and capability dispositions plus completion notifications. Retirement coverage now includes declined, retired, failure, and stale ownership outcomes.

Addressed in 2706ea914.

packages/stack/src/supervisor/SupervisorTransitions.ts
jgoux: Kept the owner fence and documented the invariant in the test, using the alternative you suggested. Live activation holds execution from claiming through terminal settlement, so owner/session replacement cannot occur during that interval. A mismatched owner therefore represents stale or already-settled evidence and must not impose recovery on its replacement. The test now uses a genuine activation Deferred mismatch.

Addressed in 2706ea914.

jgoux:
Addressed the review feedback in 2706ea914, including the ADR request from the review summary. The ADR now describes the common snapshot, notification, and reconciliation contract for terminal and intermediate transitions.

All inline conversations have replies and are resolved. The affected transition and integration suites, type checking, formatting, and both lint configurations pass. The queued-activation regression also fails under the incorrect cleanup classification and passes after restoring the fix. Independent follow-up review found no remaining actionable findings.

Coly010 (approved):

PR #6647, opened 2026-09-16
fix(cli): honour start seeding consent (CLI-2443)

7ttp:
## TL;DR

fixes `supabase start` deleting vector buckets without consent during fresh volume seeding.

## whats hurting?

both `seedBucketsRun` call sites in `start.handler.ts` hardcode `yes: true`
 force answering the prune prompt whose default is no. any vector bucket missing from `config.toml` is deleted whenever the vector store outlives the local volume, silently and irreversibly, and `--yes`/`SUPABASE_YES` are dead for start seeding..

## sol:

drop the hardcoded value at both call sites so consent resolves from `--yes` and `SUPABASE_YES` exactly like 
`db reset`, with integration tests pinning the declined default and the consented prune..

## ref:
- closes: CLI-2443


apps/cli/src/commands/start/start.integration.test.ts
```diff
@@ -2499,6 +2555,52 @@ content_path = "./supabase/templates/custom_notice.html"
...
+          route: freshVolumeRoute(defaultRoute()),
+          httpClientLayer: http.layer,
+        });
+        // A truthy ambient `SUPABASE_YES` would auto-confirm the prune and void this pin.
+        return withEnvVar(
+          "SUPABASE_YES",
+          undefined,
+          Effect.gen(function* () {
+            yield* start(flags({ exclude: ["edge-runtime"] }));
+            // Proves the vector-seed block ran, so the zero-deletions assertion isn't vacuous.
```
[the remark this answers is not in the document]
7ttp: fixed: c113742c9

apps/cli/src/commands/start/start.handler.ts
```diff
@@ -1787,7 +1786,6 @@ export const start = Effect.fn("start")(function* (flags: StartFlags) {
             projectRef: "",
             emitSummary: false,
             interactive: false,
```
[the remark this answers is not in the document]
7ttp: fair point, tho this is the shared prompt contract rather than start's: `db reset` ships the same read and both SIDE_EFFECTS docs pin it. the right fix lives in `promptYesNo` for every caller at once (skip the read when stdin is a tty under `interactive: false`, no pipe can exist there). out of scope here, i'll pick it up as a followup.

apps/cli/src/commands/start/start.command.ts
```diff
@@ -75,8 +75,8 @@ const startRuntimeLayer = Layer.mergeAll(
   dbConnectionLayer,
   httpClient,
   platformApiFactory,
-  // `stdinLayer` satisfies `promptYesNo`'s `Stdin` requirement (seed-buckets runs with
-  // `yes: true`), so `start` never reads a piped line at runtime — type requirements only.
+  // `stdinLayer` feeds `promptYesNo`'s bounded stdin read for the fresh-volume bucket-seed
+  // overwrite/prune confirmations (consent resolves from `--yes`/`SUPABASE_YES`).
   stdinLayer,
```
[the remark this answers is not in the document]
7ttp: this one isn't constructible docker free: start's only prompt sits behind a healthy stack, unlike `db reset` whose piped e2e fires before any docker dial. the wiring the rule guards is already covered: Stdin provision is compile enforced via the `CliRootCommand` annotation, and the same prompt + stdin path is binary tested through `db reset`'s piped e2e. the start specific coverage only fits the live suite, so parking it as a later followup rather than faking it in e2e here.

apps/cli/src/commands/start/start.handler.ts
```diff
@@ -1754,7 +1754,6 @@ export const start = Effect.fn("start")(function* (flags: StartFlags) {
                   projectRef: "",
                   emitSummary: false,
                   interactive: false,
```
Coly010, line 1756: This call site — the `--ignore-health-check` storage-only recheck-and-seed branch — got the same behaviour change as the main path below, but nothing pins it.

I verified this rather than eyeballing it. Restoring `yes: true` **here only** and running the whole file:

```
Test Files  1 passed (1)
     Tests  171 passed (171)
```

Every start integration test stays green. Restoring it at *both* sites is what makes `keeps a vector bucket missing from config.toml` fail, so the new coverage is really pinning the main path alone.

That asymmetry matters more than it looks: `--ignore-health-check` is the flag CI reaches for, and CI is the population most exposed to CLI-2443 in the first place. A regression that reintroduced the hardcode on just this branch would ship silently.

Ask: one more case driving the downgrade branch (bulk health check fails with an unhealthy-start error, `isFreshVolume`, Storage recheck healthy) asserting no `DeleteVectorBucket` without consent. `freshVolumeRoute` plus an unhealthy bulk-check route should get you most of the way.

apps/cli/src/commands/start/start.command.ts
```diff
@@ -75,8 +75,8 @@ const startRuntimeLayer = Layer.mergeAll(
   dbConnectionLayer,
   httpClient,
   platformApiFactory,
-  // `stdinLayer` satisfies `promptYesNo`'s `Stdin` requirement (seed-buckets runs with
-  // `yes: true`), so `start` never reads a piped line at runtime — type requirements only.
+  // `stdinLayer` feeds `promptYesNo`'s bounded stdin read for the fresh-volume bucket-seed
```
Coly010: The comment this replaces made a guarantee — *"`start` never reads a piped line at runtime — type requirements only"* — and the PR removes the guarantee along with the comment. Documenting the new behaviour honestly is the right instinct, but the behaviour itself is a new side effect on the most-run command in the CLI, and it is observable.

I reproduced it. One bounded `readLine(100)` over fd 0 (the exact call `promptYesNo` makes at `prompt-yes-no.ts:52`), invoked from inside a script piped to `bash`:

```
$ printf 'echo LINE_A\n<probe>\necho LINE_B\necho LINE_C\necho LINE_D\n' | bash
LINE_A
[probe] readLine -> "echo LINE_B"
```

`LINE_C` and `LINE_D` never ran. The read took `echo LINE_B` as the prompt's answer (unparseable, so the default applied) and the 64 KB stream read pulled the remainder out of the pipe, where it was discarded at scope close. Any `curl … | bash` onboarding script, `bash < setup.sh`, or `while read` loop around `supabase start` loses the rest of its input, with exit 0 and no error.

The trigger is narrow — text output mode *and* a confirmation actually firing, which on a fresh volume means only the stale-vector-bucket case, since `storage.buckets` lives in the DB volume being recreated. That is the same population CLI-2443 is about, which is exactly why it is worth closing now rather than as a follow-up.

Also worth flagging: the fix sketched on the earlier review thread (skip the read when stdin is a TTY under `interactive: false`) does not help here. This case *is* a pipe; that is the whole mechanism.

Simplest resolution, and I think the better design anyway: `start` should not ask. It has already decided — prune only on explicit consent — so pass that decision down instead of routing it through a prompt that reads stdin nobody intends it to read. A dedicated "never prune" policy on `seedBucketsRun` would drop the stdin read, the 100 ms-per-bucket stall, and this failure mode in one go, and leave deliberate pruning where it belongs in `supabase seed buckets`.

apps/cli/src/commands/start/SIDE_EFFECTS.md
```diff
@@ -78,6 +78,11 @@ Docker healthcheck — mirroring PostgREST's own probe shape.
 Runs only when `isFreshVolume && Storage started`, after the bulk
 health check genuinely succeeds, right before the `cli_stack_started` telemetry capture. A
 seeding failure rolls back the whole `start` run, same as any other post-bring-up failure.
+Seeding is non-interactive. In text mode its overwrite/prune confirmations print their
+label, read one bounded stdin line, and fall back to their defaults (overwrite → yes,
+prune → no); in machine modes they take those defaults silently. Either way
+`--yes`/`SUPABASE_YES` auto-confirms, so a vector bucket missing from `config.toml`
+is never pruned without explicit consent.
```
Coly010: The bucket is correctly kept now — but nothing tells the user that, or what to do about it.

In text mode they get the question and a blank answer, then `start` moves on to its success banner:

```
Bucket stale-vec not found in supabase/config.toml. Do you want to prune it? [y/N] 
```

No "keeping it" line (`seed-buckets.ts` just `continue`s on a declined prune), and `interactive: false` forces the non-TTY branch even for a human sitting at a terminal, so they get roughly 100 ms to answer a question that is then answered for them. In `--output-format json`/`stream-json` there is no signal whatsoever: the prompt is silent and `start` passes `emitSummary: false`, so `vector_pruned`/`vector_skipped` never surface either. Local Storage diverges from `config.toml` with zero evidence in machine-readable output.

There is also no discoverable path to prune on purpose. Nothing in `start`'s output, `start --help`, or `seed buckets`' own description says that is where pruning lives.

Ask: replace the question with a decision, printed in every output mode — something like `Keeping vector bucket stale-vec: not declared in supabase/config.toml. Run supabase seed buckets to prune.` That resolves this, the stdin read, and the phantom-prompt UX together.

Related, in this same file: the `### --output-format text` section at :271 enumerates `start`'s stderr lines carefully and does not mention the two new prompt lines, which are newly user-visible whenever `--yes` is absent. Worth adding there too. And the paragraph here omits the one fact `reset-local-database.ts:305-308` states explicitly — that a parsed `y`/`n` answer *wins* over the default, and that the read happens even when stdin is a TTY. That is the surprising half of the contract; `db reset`'s own SIDE_EFFECTS spells it out, this one doesn't.

Coly010 (changes requested):
## Review

The diagnosis is right and the fix is the correct shape. I checked the premise rather than taking it on trust: `--yes` really is a root global flag (`global-flags.ts:75`, attached at `cli/root.ts:202`), so both `supabase start --yes` and `SUPABASE_YES` reach `resolveYesWithProjectEnv`, and the new SIDE_EFFECTS wording matches the print-and-scan contract #6643 already established for `db reset`. I also mutation-verified the declined-prune test actually bites: restoring `yes: true` fails it with `expected [ '{"vectorBucketName":"stale-vec"}' ] to have a length of +0`. CI is green across every shard.

Three things to resolve before this merges. Details inline.

1. **Only one of the two fixed call sites is pinned.** Restoring `yes: true` on the `--ignore-health-check` branch alone leaves all 171 start integration tests green — and that is the branch CI reaches.
2. **`start` now consumes the caller's stdin.** This PR deletes a comment that guaranteed the opposite, and the consequence is a real, reproducible swallowed-input bug for anyone running `start` from a piped script. Worth noting the TTY-gating follow-up proposed on the earlier review thread would not cover this case.
3. **Declining prints no outcome and no next step**, and in machine output modes prints nothing at all.

None of these are objections to the direction — the default must be "don't delete", and this gets that right.

apps/cli/src/commands/start/start.handler.ts
7ttp: fixed: 647c6f84

mutation checked both ways: breaking that call site alone turns the new test red, breaking the main one alone leaves it green.

apps/cli/src/commands/start/start.command.ts
7ttp: fixed: 647c6f84

went with never ask rather than never prune, so `--yes` still prunes exactly as it does today but nothing reads fd 0 on any path. promptYesNo already short circuits on consent before it touches stdin, so the resolved outcome matches the old default in every case except the one you reproduced. there was a third site too, the bucket overwrite prompt, same mechanism and reachable on the recheck branch.

`db reset` has the same swallow since it also passes interactive false. felt like a bigger picture thing that shouldnt ride in here, so id rather do that one on its own.

apps/cli/src/commands/start/SIDE_EFFECTS.md
7ttp: fixed: 647c6f84

theres a keeping line now that names the bucket and points at `supabase seed buckets`, and it uses the raw writer so it shows in json and stream json too, not just text. side effects covers it in both the seeding section and the text mode list, and the phantom prompt is gone entirely.

two bits i left: the structured vector_skipped still doesnt surface because start passes emitSummary false, and neither `start --help` nor seed buckets own description mentions pruning. both widen the output contract more than this fix should, so id pick them up separately.

apps/cli/src/commands/start/start.handler.ts
7ttp: fixed: 647c6f84

Coly010 (approved):

PR #6540, opened 2026-09-09
feat(cli): use stack EphemeralPostgres for schema-tooling shadows

avallete:
## Summary

When `[experimental].stack` / `SUPABASE_EXPERIMENTAL_STACK=1` is on:

- **`supabase stack` is gated on the flag**, including help and completion. Top-level `start`/`stop`/`status` become stack aliases. Invalid project config leaves the feature off; invalid env overrides are still errors. Flag off keeps legacy Compose start/stop/status and Docker SQL-template shadows.
- **`--local` uses the project stack.** `db`, `migration`, `test db`, `gen types`, and `inspect` resolve `--local` through stack credentials instead of Compose names (`supabase_db_*`, `db:5432`). `gen types --local` and `inspect db … --local` do not start a stack. Linked and explicit `--db-url` stay URL / Management API targets for engine and runtime selection (`connType`). A `--db-url` that matches `config.toml` host and port is still `isLocal` for dump's tool-container host rewrite. Functions and storage stay on their existing backends. Top-level `pull` uses the same stack shadow as `db pull`.
- **Schema tooling uses `EphemeralPostgres`.** `db diff`, `db pull`, declarative generate/sync, and `migration squash` provision throwaway shadows through `@supabase/stack`. The cluster is a scoped second instance (native host PGDATA or a named volume). Snapshots are runtime-kind specific (`stack-shadow-baseline-*.tar`, same keep-3 / 2-day TTL as the legacy namespace). The runtime marker carries `snapshotKey` on the CLI cache path; a keyless marker is rejected when a key is expected. Archive flags are portable (GNU tar and Darwin bsdtar). Ephemeral env files are `0600`; the cache dir is `0700`. Working state is `$SUPABASE_HOME/managed/ephemeral-postgres/<identity>/`.
- **The stack backend requires in-process pg-delta.** `--use-migra`, `--use-pgadmin`, `--use-pg-schema`, and `db pull --diff-engine migra` are rejected for every stack runtime because the shadow is always `EphemeralPostgres`, including `--linked` / `--db-url`.
- **Live start is Compose-faithful.** First create (`unconfigured`): schema init, overlay (webhooks, grants, vault, `roles.sql`), then migrate-and-seed. Existing cluster: webhooks only — no second overlay or migrate-and-seed. A failed first setup leaves the engine up and does not retry on the next start; recover with `db reset`. Disabled capabilities keep nested pins (`version`, settings).
[...]

PR #6652, opened 2026-09-16
feat(cli): generate types natively against the selected stack (CLI-2366)

Coly010:
## Summary

`gen types` no longer spawns a `postgres-meta` container. Introspection and code generation run
in-process through [`@supabase/postgrest-typegen`](https://github.com/supabase/sdk/tree/main/packages/postgrest-typegen)
over a direct PostgreSQL connection.

This is what `--local` on the managed stack needs: the stack exposes no pg-meta capability, and it
supports a native runtime with no container engine at all, so the old path required a Docker daemon
to generate types for a database that isn't in Docker. #6540 already resolved the stack's real
endpoint here; this removes the container that resolution still fed.

Generation goes through the shared `DbConnection` seam that `db query` uses, so there is no second
driver path — TLS mode, DoH resolution, pooler fallback and connect-error classification are the
ones already in place.

**TLS.** The SSL probe is gone. Linked, project-ref and preview-branch targets are known Supabase
hosts and pin the bundled CA with `sslmode=require`, preserving the stricter posture `gen types`
already had. `--db-url` honours the DSN's own `sslmode`/`sslrootcert` and otherwise pins only for
known Supabase hosts; `--local` stays plaintext. `PgConnInput` gains an optional inline-CA field so
pinning needs no temp file.

**`--query-timeout`** now maps to a server-enforced `statement_timeout` rather than a container env
var, so a stalled introspection is aborted by the server rather than only client-side.

### Behaviour changes

**`--network-id` is rejected on this command.** In-process generation cannot join a Docker network.
Both argv positions are covered, including the persistent form before the command path.

**A `--db-url` pointing at a plain-TCP server now needs `?sslmode=disable`.** The shared connection
layer intentionally does not downgrade from `prefer` to plaintext, and `gen types` no longer probes.
This matches every other `--db-url` command; a connect-error suggestion names the fix.

**Generated output**, measured against `postgres-meta` v0.99.0 on the same schema:

| Language | Result |
| --- | --- |
| Go | byte-identical |
| Swift | byte-identical |
| TypeScript | a `NOT NULL` jsonb column is `NonNullable<Json>` rather than `Json` |
| Python | jsonb columns are `JsonValue` rather than `Json[Any]`; `NotRequired`/`TypeAlias` import from `typing_extensions` |

Both differences are upstream corrections: the old TypeScript type admitted `null` for a `NOT NULL`
column, and pydantic's `Json[Any]` expects an unparsed JSON *string* rather than a decoded row
value. `--linked`/`--project-id` TypeScript is generated server-side by the Management API and is
unaffected, so it can differ from local output on these jsonb cases until the hosted service adopts
the same engine. Recorded in the command's `SIDE_EFFECTS.md`.

### Credit

Supersedes #6404 by @avallete, which first proved this out and worked through the
`bun build --compile` problems with `oxfmt` — the optional-plugin externals and the statically
embedded napi binding here are that PR's design. Restarted from `develop` because #6525 moved every
path it touched.

## Linked issue

Resolves [CLI-2366](https://linear.app/supabase/issue/CLI-2366). Also covers
[CLI-2279](https://linear.app/supabase/issue/CLI-2279) (no GitHub issue).

## Checklist


apps/cli/src/commands/gen/types/types.handler.ts
```diff
@@ -297,6 +285,61 @@ export const genTypes = Effect.fn("gen.types")(function* (flags: GenTypesFlags)
...
 
+  /**
+   * Sets a session-level `statement_timeout` and connect timeout from `--query-timeout` on
+   * every generate attempt — the server-side `statement_timeout` is the real guard, unlike the
+   * pg-meta container's own env-var timeouts it replaces.
+   */
+  const withQueryTimeout = (conn: PgConnInput): PgConnInput => ({
+    ...conn,
+    runtimeParams: { ...conn.runtimeParams, statement_timeout: String(queryTimeoutSeconds * 1000) },
+    connectTimeoutSeconds: queryTimeoutSeconds,
```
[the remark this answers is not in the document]
Coly010: Fixed in 5b15cdd91 — and it was worse than described. `statement_timeout=0` *disables* the bound in PostgreSQL rather than making it immediate, while `connectTimeoutSeconds: 0` becomes `Duration.seconds(0)`, so `--query-timeout 400ms` could never connect at all.

The seconds rounding only existed to satisfy pg-meta's integer-seconds env contract, which is gone, so rather than rejecting or clamping sub-second values the parser now returns milliseconds (`parseQueryTimeoutMillis`). `statement_timeout` gets the exact millisecond value and the connect timeout is `max(1, ceil(ms/1000))`. An explicit `0s` still disables the bound and now leaves the driver's own connect default in place instead of forcing 0.

Regression tests cover `400ms` and `0s`; verified end to end against a live database.

apps/cli/src/commands/gen/types/types.oxfmt.ts
```diff
@@ -0,0 +1,141 @@
...
+          "@oxfmt/binding-linux-arm64-musl",
+        );
+      }
+      return loadOxfmtBinding(
+        () => require("@oxfmt/binding-linux-arm64-gnu"),
+        "@oxfmt/binding-linux-arm64-gnu",
+      );
+    }
+    if (process.arch === "x64") {
+      if (typeof SUPABASE_LIBC !== "undefined" && SUPABASE_LIBC === "musl") {
```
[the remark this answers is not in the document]
Coly010: Fixed in 5b15cdd91. Added a runtime libc probe (absence of `process.report.getReport().header.glibcVersionRuntime`, in a try/catch defaulting to glibc) used only when the `SUPABASE_LIBC` define is absent, so source runs and `build:binary` output both resolve correctly on musl. The per-platform bare `require()` literals are untouched so Bun still embeds the right `.node` statically.

apps/cli/src/commands/gen/types/SIDE_EFFECTS.md
```diff
@@ -12,17 +16,14 @@ When `[experimental].stack` is on, `--local` resolves the project stack through
...
 | `<workdir>{/supabase}/.env*`                      | dotenv     | `--local`; resolves the same nested environment overrides as the CLI                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     |
```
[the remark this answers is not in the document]
Coly010: Fixed in 5b15cdd91. Documented the `[db]`-subtree config read and the nested dotenv read that now happen on every `--db-url` resolution, the conditional libpq service-file read, the `PG*` fallbacks, and the malformed-config failure mode.

apps/cli/tsconfig.types.json
```diff
@@ -0,0 +1,16 @@
+{
+  // Type-check-only overlay. `@supabase/postgrest-typegen`'s `bun` exports condition points at
+  // its unbuilt `src/*.ts`, which does not satisfy this workspace's stricter compiler options;
+  // its published `dist/*.d.ts` describes the same API and does. The pin cannot live in
+  // `tsconfig.json` because Bun honours `paths` at runtime too, and would then load a
+  // declaration file instead of the implementation. `types.generator.integration.test.ts`
+  // exercises the Bun-resolved runtime API so the two views cannot drift unnoticed.
```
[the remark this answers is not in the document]
Coly010: Fixed in 5b15cdd91 — my error, the comment was written before the test existed and named the wrong suffix.

apps/cli/src/commands/gen/types/types.generator.layer.ts
```diff
@@ -0,0 +1,78 @@
...
+          const session = yield* dbConn.connect(input.conn, {
+            isLocal: input.isLocal,
+            dnsResolver: input.dnsResolver,
+          });
+          // `session.query` needs no services, but running it through the current fiber's
+          // context (rather than a bare detached `Effect.runPromise`) keeps the Promise bridge
+          // anchored to this generator effect instead of a disconnected top-level runtime.
+          const runQuery = Effect.runPromiseWith(yield* Effect.context<never>());
+          const queryable: Queryable = {
+            query: (sql) => runQuery(session.query(sql)).then((rows) => ({ rows: [...rows] })),
```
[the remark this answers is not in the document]
Coly010: Fixed in 5b15cdd91. `Effect.tryPromise` already supplies an interruption-linked `AbortSignal`, so forwarding it into each `runQuery` call links the query fiber to the parent instead of leaving it detached. Combined with the timeout fix above, a stalled introspection is now bounded both by interruption and by a server-enforced `statement_timeout`.

apps/cli/src/command-internal/connect-errors.ts
```diff
@@ -369,6 +369,11 @@ export function connectSuggestion(
   ) {
     return SUGGEST_ENV_VAR;
   }
+  // An unset `sslmode` negotiates TLS and fails rather than downgrading, so a server without
+  // TLS needs the caller to opt into plaintext explicitly.
+  if (text.includes(SERVER_REFUSED_SSL) || text.includes("server refused TLS connection")) {
+    return "This server does not accept TLS. Append `?sslmode=disable` to the connection string to connect in plaintext.";
```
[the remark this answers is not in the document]
Coly010: Fixed in 5b15cdd91 — reworded to say to set `sslmode=disable` on the connection string, without prescribing `?`.

apps/cli/src/commands/gen/types/types.integration.test.ts
```diff
@@ -962,47 +715,180 @@ describe("gen types", () => {
...
+      projectId: Option.some(VALID_REF),
+      projectTypes: "ok",
+    });
+
+    return Effect.gen(function* () {
+      yield* genTypes(defaultFlags()).pipe(Effect.provide(layer));
+    });
+  });
+
+  // --- Flag mutex groups and argv-scan precedence -----------------------------------------
```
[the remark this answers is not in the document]
Coly010: Fixed in 5b15cdd91 — the banners are now nested `describe` blocks.

apps/cli/src/commands/gen/types/types.handler.ts
```diff
@@ -603,6 +493,19 @@ export const genTypes = Effect.fn("gen.types")(function* (flags: GenTypesFlags)
...
+    // container. It is a persistent flag, so a pre-command occurrence (`supabase --network-id
+    // net gen types ...`) lands in `prePathOccurrences`, not `occurrences` — check both.
+    if (occurrences.has("network-id") || scan.prePathOccurrences.has("network-id")) {
+      return yield* Effect.fail(
+        new GenTypesNetworkIdUnsupportedError({
+          message:
+            "gen types now generates types in-process and cannot join a Docker network via " +
+            "--network-id; use a host-reachable --db-url instead.",
+        }),
+      );
```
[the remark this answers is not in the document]
Coly010: Not taking this one — warning and ignoring was considered and deliberately rejected.

The flag cannot be honoured any more: generation runs in-process on the host, so there is no container to attach to a Docker network. That leaves three options, and warn-and-ignore is the worst of them. For a database that is genuinely only routable inside a Docker network, ignoring the flag means the connection fails anyway — the user just gets a generic connection error instead of being told why, after we explicitly acknowledged the flag they passed. For a database that is host-reachable, the flag was already doing nothing meaningful.

Failing tells the user exactly what changed and what to do instead (use a host-reachable `--db-url`). The alternative that would preserve the capability is keeping a container-backed path solely for `--network-id`, which means retaining pg-meta image resolution and the whole container code path indefinitely for one flag — that was weighed and rejected.

Worth noting the flag remains globally registered and functional for the commands that still shell out to containers; this rejection is scoped to `gen types`. The behaviour change is called out in the PR description and in the command SIDE_EFFECTS.md.

apps/cli/src/commands/gen/types/types.handler.ts
```diff
@@ -297,6 +285,67 @@ export const genTypes = Effect.fn("gen.types")(function* (flags: GenTypesFlags)
...
+      const types =
+        input.poolerFallback === undefined
+          ? yield* attempt(input.conn)
+          : yield* runWithPoolerFallback({
+              run: attempt(input.conn),
+              retry: attempt,
+              directHost: input.poolerFallback.directHost,
+              eligible: input.poolerFallback.eligible,
+              resolveFallback: input.poolerFallback.resolve,
+              classifyError: isIPv6ConnectivityErrorCause,
```
jgoux: [P2] Preserve network-error details before classifying pooler fallback

The new `attempt` exposes `DbConnectError` from `toConnectError`, which drops the driver cause/code/address. As a result, direct-host failures with `EHOSTUNREACH` or `EADDRNOTAVAIL` on an IPv6 address, or `ENOTFOUND`, no longer trigger the pooler retry: the remaining message matcher handles Node `ENETUNREACH` but not these forms. The former SSL-probe path retained the cause. I reproduced the classifier changing from true to false across the current error mapping for all three errors, and confirmed a real failed `DbConnection.connect` returns an `ENOTFOUND` message without a cause that classifies false. This breaks non-TypeScript linked/project/branch generation on affected IPv4-only networks despite a usable pooler. Preserve the classification information through the connection boundary and cover fallback with an error shaped by the actual connection layer.

apps/cli/scripts/bundle-externals.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Optional prettier plugins that `oxfmt`'s dist lazily `import()`s for non-TypeScript file
+ * types. They are never installed — `gen types` only formats generated TypeScript, through the
+ * statically embedded binding in `src/commands/gen/types/types.oxfmt.ts` — but `bun build`
+ * still resolves every analyzable dynamic import, so each must be marked external.
+ */
+export const OXFMT_OPTIONAL_PLUGIN_EXTERNALS = [
```
jgoux, line 7: [P2] Apply the formatter externals to the local-release compiler too

`pnpm cli-release` compiles this same entrypoint through the separate shell command in `tools/release/local-release.ts:193`, which does not use this list. With the new typegen dependency, that compile step fails resolving oxfmt optional imports such as `@prettier/plugin-oxc` and `prettier-plugin-astro`, preventing a local release. I reproduced this with its macOS ARM64 compile invocation; adding only these seven `--external` flags makes the same invocation succeed. Wire the shared externals into that build path as well.

jgoux (commented):
Two verified regressions remain after independent review and cross-review.

apps/cli/src/commands/gen/types/types.handler.ts
```diff
@@ -726,20 +607,30 @@ export const genTypes = Effect.fn("gen.types")(function* (flags: GenTypesFlags)
...
+        resolved.conn.sslmode === undefined &&
+        resolved.conn.sslrootcert === undefined &&
+        (isDirectDbHost(resolved.conn.host, cliSettings.projectHost) ||
+          isPoolerHost(resolved.conn.host, cliSettings.poolerHost))
+          ? pinSupabaseTls(resolved.conn)
+          : resolved.conn;
+
+      yield* runGenerate({
+        conn,
+        isLocal: resolved.isLocal,
```
avallete, line 642: **issue (blocking):** Honor explicit `--db-url` TLS when the target is classified local

`--db-url` is documented as honouring the DSN’s own `sslmode` / `sslrootcert`. This path still passes `isLocal: resolved.isLocal` into `runGenerate`, and `sslConfigsFor` returns `[false]` whenever `isLocal` is true — before it reads `sslmode`.

A TLS tunnel on `127.0.0.1:54322` with `sslmode=require` or `verify-full` is therefore forced to plaintext. Dogfood on local Postgres matched that: a loopback `--db-url` connected without `sslmode=disable` and never produced a “server refused TLS” hint, because local classification skipped TLS entirely.

Do not force plaintext on `--db-url` when the DSN set `sslmode` / `sslrootcert`. Default loopback-without-ssl can stay plaintext.

avallete (dismissed):
**issue (blocking):** Three regressions remain on `5b15cdd91`

In-process `--local` generation dogfoods clean against the stack-engine matrix: compose, docker eager, docker lazy, and native all exit 0; native still works with Docker off `PATH` (no pg-meta container); `--network-id` rejects; `--query-timeout 400ms` succeeds; staging `--linked` TypeScript (Management API) and Go (direct connect) succeed.

The remaining blockers are inline: @jgoux's two [P2]s (pooler classification dropping driver `cause`/`code`/`address`, and `pnpm cli-release` missing the oxfmt externals) plus one more — `--db-url` classified as local silently ignores explicit `sslmode`.

apps/cli/src/commands/gen/types/types.handler.ts
avallete: **agreement (blocking):** Independently confirmed.

`toConnectError` keeps only `message` / `suggestion` / `retryable`, so `isIPv6ConnectivityErrorCause` never sees `code`/`address`. After wrapping, `ENOTFOUND` becomes `hostname resolving error (getaddrinfo ENOTFOUND …)` (no match on `isIPv6ConnectivityError`), and IPv6 `EHOSTUNREACH` / `EADDRNOTAVAIL` lose `address`. Node `ENETUNREACH` can still match via the message regex; the others cannot.

Handler integration tests retry on a fake `GenTypesGenerationError` message, not a real `DbConnectError` from the connection layer. TypeScript `--linked`/`--project-id` is unaffected (Management API).

apps/cli/scripts/bundle-externals.ts
avallete: **agreement (blocking):** Independently confirmed.

`apps/cli/scripts/build.ts` and `build-binary.ts` apply `OXFMT_OPTIONAL_PLUGIN_EXTERNALS`; `tools/release/local-release.ts:193` still compiles `apps/cli/src/main.ts` with a raw `bun build --compile` and no `--external` list. With the new typegen → oxfmt dependency, that path fails resolving optional plugins such as `@prettier/plugin-oxc`.

Wire the shared list into the local-release compile.

apps/cli/src/commands/gen/types/types.handler.ts
Coly010: Good catch, and correct — this was a real regression. Confirmed the mechanism: `DbConnectError` carries only `message`/`suggestion`/`retryable`, and `toConnectError` drops the driver error, so `isIPv6ConnectivityErrorCause` had nothing structured left to read. Its text fallback matches Node's `ENETUNREACH <ipv6>:<port>` form but not `ENOTFOUND`, nor `EHOSTUNREACH`/`EADDRNOTAVAIL` against an IPv6 address.

Fixed in cee59507c. Rather than attaching the raw driver error to a domain error, the boundary classifies while the driver error is still in scope and carries the verdict on a new `DbConnectError.ipv6Unreachable` — the same shape as the existing `retryable`, which `isDialFailure` already computes there. `classifyError` reads that for a `DbConnectError` and still falls back to `isIPv6ConnectivityErrorCause` for other errors.

Also took the second half of your point about test shape. The old fallback tests only passed because they fabricated errors that happened to classify true. They now build failures through the real `toConnectError`, and there is a new `ENOTFOUND` direct-host test that fails without this change: that error renders as `hostname resolving error (getaddrinfo ENOTFOUND …)`, which matches none of the text patterns, so the old code classified false and never retried.

Checked the other `runWithPoolerFallback` callers — `db-pull-run.ts` and `db/dump/dump.handler.ts` go through `commands/db/shared/pooler-fallback.ts`, which only sets `classifyResult`, so they are unaffected.

apps/cli/scripts/bundle-externals.ts
Coly010: Fixed in cee59507c — I had wired the externals into `build.ts` and `build-binary.ts` and missed this third compile path entirely.

`bundle-externals.ts` now also exports `oxfmtExternalArgs` (the same list rendered as `--external=<name>` flags, since this call site shells out rather than using the `Bun.build()` object API), and `local-release.ts` spreads that in — one source of truth for all three builds rather than a copied list.

Reproduced your finding and the fix with the darwin-arm64 invocation: without the flags the compile reports one `Could not resolve` per oxfmt optional plugin; with them, all of those disappear. Note the compile still cannot complete in my worktree for an unrelated pre-existing reason (`@libpg-query/parser`, a transitive dep of `plpgsql-deparser`, is absent from `node_modules` even after a frozen-lockfile install) — that reproduces identically with and without this change, so I have left it alone.

apps/cli/src/commands/gen/types/types.handler.ts
Coly010: Confirmed and fixed in 967878c85 — thanks, this one was well hidden.

Traced it: `sslConfigsFor` early-returns `[false]` on `isLocal` before any `sslmode` handling, and the `caCert` derivation is gated on `!isLocal` too, so a local-classified target never even loaded the root cert. An explicit `verify-full` was therefore downgraded to plaintext *and* silently unverified.

Fixed at the acquisition boundary rather than inside `sslConfigsFor`: the connection is treated as TLS-exempt only when it is local **and** names no `sslmode`/`sslrootcert`/inline CA. `sslConfigsFor`'s contract is untouched — it now just receives the caller's exemption decision instead of the raw classification — and the same condition opens the CA gate. No other `isLocal` use site moved (connect-timeout default, suggestions, step-down, DoH all unchanged).

Checked the blast radius before touching shared code: `sslmode`/`sslrootcert` are only ever populated by DSN parsing (`parseUrlConnectionString`/`parseKeywordValueDsn`), never by the toml-derived `connType: "local"` path, so ordinary `--local` connections cannot be affected. Loopback with a bare DSN still connects in plaintext, as you asked.

Coverage is behavioural rather than config-shape only: a fake Postgres server asserts whether an SSLRequest actually goes on the wire — TLS attempted for `sslmode=require` on loopback, no SSLRequest for `disable` or for a bare DSN, and a separate test proves the CA gate opened by pointing `sslrootcert` at a missing file and asserting the read is now attempted.

Also worth flagging since it cuts against something I wrote earlier in this PR: your observation that a loopback `--db-url` never produced the “server refused TLS” hint is right, and it means my claim that `--db-url` against a plain-TCP server always needs `?sslmode=disable` only held for non-local targets. SIDE_EFFECTS.md now says so explicitly.

Your two other blockers (@jgoux's P2s) were fixed in cee59507c — replies on those threads.

Coly010:
All three blockers from @avallete's review are now addressed, plus the earlier AI-review round.

| Finding | Status | Commit |
| --- | --- | --- |
| @jgoux — pooler classification dropped the driver `cause`/`code`/`address` | fixed | `cee59507c` |
| @jgoux — `pnpm cli-release` missing the oxfmt externals | fixed | `cee59507c` |
| @avallete — `--db-url` classified local silently ignored explicit `sslmode` | fixed | `967878c85` |

Two of these were real regressions introduced by routing generation through `DbConnection`, and both were invisible to my own tests — the pooler tests only passed because they fabricated errors that happened to classify true. They now build failures through the real `toConnectError`, and the TLS coverage asserts whether an SSLRequest actually reaches the wire rather than inspecting config shape.

@avallete — thank you for the dogfooding matrix; that is far better evidence than anything I could produce locally, particularly native-with-Docker-off-`PATH`, which is the case this change exists for.

One item from the AI round remains open by choice rather than oversight: both bots wanted `--network-id` to warn-and-ignore instead of failing. I kept the hard error, reasoning on that thread — for a network-only database, ignoring the flag fails the connection anyway but with a worse message, and preserving the capability would mean keeping the whole container path alive for one flag. Happy to revisit if either of you disagrees.

Also unchanged and worth a second opinion before merge: the jsonb output deltas in the description. Go and Swift stay byte-identical; TypeScript and Python shift on jsonb columns because the library corrected two upstream bugs. `--linked` TypeScript is served by hosted postgres-meta and so will disagree with local output until that service adopts the same engine.

jgoux (approved):

apps/cli/src/command-internal/db-connection.sql-pg.layer.ts
```diff
@@ -395,6 +396,19 @@ export interface ClientCert {
...
+ * Whether the DSN itself asked for TLS behavior: `--db-url`'s `sslmode`/`sslrootcert` are honored
+ * even against a target classified local (e.g. a TLS tunnel on the loopback stack), so `isLocal`
+ * alone must not force plaintext when one of these is set.
+ */
+export function tlsExplicitlyRequested(cfg: PgConnInput): boolean {
+  return (
+    cfg.sslmode !== undefined ||
+    (cfg.sslrootcert?.length ?? 0) > 0 ||
+    (cfg.sslrootcertInline?.length ?? 0) > 0
+  );
```
avallete, line 409: **issue (blocking):** this is a regression of local `--db-url` (and every other `DbConnection` command), not just a too-eager helper.

`967878c85` was meant to honor an *explicit* TLS demand on a target classified local — a TLS tunnel on `127.0.0.1` with `sslmode=require` / `verify-full` / `sslrootcert`. That part is right: before this commit, `sslConfigsFor` returned `[false]` whenever `isLocal` was true, so the DSN’s TLS settings never reached the wire.

The predicate as written treats **any** defined `sslmode` as that demand:

```ts
cfg.sslmode !== undefined || sslrootcert || sslrootcertInline
```

That is not the same as “the DSN asked for TLS”. Two realistic inputs now lift the exemption and then hit TLS-only dialing (`sslConfigsFor` maps `prefer` / unset to TLS with no plaintext fallback, `isLocal && !explicitTls` at the acquisition site):

1. **`?sslmode=prefer` on a loopback DSN.** libpq `prefer` means try TLS, then plaintext. This CLI has never implemented that fallback — `prefer` is TLS-only — but local classification previously forced plaintext *before* that mapping ran. After this commit, `prefer` is “explicit TLS” and the local Postgres that does not speak SSL fails with a handshake error instead of connecting.
2. **`PGSSLMODE` (or a service-file `sslmode`) with no `sslmode` in the DSN.** `parseConnectionString` copies those into `cfg.sslmode` when the URL omits it (`db-config.parse.ts`, URL and keyword parsers). A developer with `PGSSLMODE=prefer` (or `require`) in the environment, then `gen types --db-url postgresql://postgres:postgres@127.0.0.1:54322/postgres`, now fails against the local stack. The same path is used by `db query --db-url` and anything else that goes through `acquirePgPoolConnection`.

That is a behavior change from `5b15cdd91` / pre-exemption-fix: a classified-local DSN with no TLS keys in the URL connected in plaintext. It is also wider than the bug that was filed — a missing `sslmode=require` on a *TLS* loopback tunnel. `sslmode=disable` happens to still work only because `sslConfigsFor` special-cases it after the exemption is already gone.

Dogfood on this SHA confirmed the intended tunnel case: compose and native loopback with a *bare* DSN and with `sslmode=disable` still connect; `sslmode=require` against plaintext Postgres now fail-closes (SSLRequest on the wire). The hole is the `prefer` / env-filled `sslmode` case, which this helper cannot distinguish from `require`.

**Smallest fix:** treat TLS as explicitly requested only for `require` / `verify-ca` / `verify-full`, or when `sslrootcert` / an inline CA is set. Leave `prefer` / `allow` / unset / env-filled `prefer` on a local target as plaintext. That keeps the TLS-tunnel fix and restores the local `--db-url` that used to work.

avallete (changes requested):
**issue (blocking):** `tlsExplicitlyRequested` is a regression on local `--db-url`

The follow-up in `967878c85` correctly stopped ignoring `sslmode=require` / `verify-full` on a loopback TLS tunnel. The predicate it added is too broad: any populated `sslmode` lifts the local plaintext exemption, including libpq `prefer` and a `PGSSLMODE` fill-in the DSN never set. That makes a previously working local `--db-url` fail TLS against plaintext Postgres, and it is on the shared `DbConnection` seam so `db query` / dump / pull are in the blast radius too.

Inline: how to reproduce and the smallest fix.

apps/cli/src/command-internal/db-connection.sql-pg.layer.ts
Coly010: You are right, and this landed before I saw the review — follow-up in #6664.

Your diagnosis is exact, and the `PGSSLMODE` half is the part I got wrong twice: I noted the ambient-env risk while designing the predicate, convinced myself it was acceptable because libpq would also honour it, and missed that `prefer` is the common ambient value and that this CLI maps `prefer` to TLS-only with no fallback. Treating "has an sslmode" as "demands TLS" conflated a fallback description with a demand.

#6664 narrows it to `require` / `verify-ca` / `verify-full`, or a supplied `sslrootcert` / inline CA — your smallest fix.

Reproduced the regression against a plaintext Postgres on the local-classified port, before and after, same database:

| Connection | merged `develop` | #6664 |
| --- | --- | --- |
| bare loopback DSN | connects | connects |
| `PGSSLMODE=prefer` | **fails** | connects |
| `PGSSLMODE=allow` | **fails** | connects |
| `?sslmode=prefer` | **fails** | connects |
| `?sslmode=disable` | connects | connects |
| `?sslmode=require` | fails closed | fails closed |
| `PGSSLMODE=verify-full` | fails closed | fails closed |

The tunnel case from the previous round is unchanged — an explicit demand still puts an SSLRequest on the wire against a loopback target, and a root cert still opens the CA gate.

On test coverage: the unit tests I shipped last round actively asserted the broad behaviour (including `sslmode=disable` counting as a demand), so they encoded the bug rather than catching it. Those assertions are corrected, and there is now a test that runs the real DSN parser with a `PGSSLMODE` env fill-in into the predicate, so the env-sourced case is covered end to end instead of reasoned about.

Coly010:
Merged before @avallete's second review landed — the regression it identifies is real and is fixed in #6664.

`tlsExplicitlyRequested` treated any populated `sslmode` as a TLS demand. `prefer`/`allow` describe a fallback this CLI does not implement, and `sslmode` is also filled from `PGSSLMODE` and libpq service files, so an ambient `PGSSLMODE=prefer` broke local `--db-url` across `db query`, dump and pull as well as `gen types`. #6664 narrows the predicate to `require`/`verify-ca`/`verify-full` or a supplied root cert, with a before/after matrix against a plaintext local Postgres.

PR #6660, opened 2026-09-17
refactor(cli): cover `unlink` with effect lint (CLI-2453)

7ttp:
## TL;DR

brings the `supabase unlink` command family under the effect lint

## whats introduced?

effect lint applied to `unlink`:

- one allow list entry in `.oxlintrc.effect.json`
- the handler's three failures yield their tagged error directly instead of wrapping it in `Effect.fail`
- integration tests move onto the `FileSystem`/`Path` services, with the project ref fixture written inside the effect
- cause assertions encode through a schema codec instead of `JSON.stringify`, keeping the full serialized surface
- e2e tests run through the effect native harness with a scoped temp dir per test, live tests through the effect fixture

## ref:

- closes: CLI-2453


apps/cli/src/commands/unlink/unlink.handler.ts
```diff
@@ -65,14 +63,12 @@ export const unlink = Effect.fn("unlink")(function* () {
...
       }
       const message = collected.map((e) => e.message).join("\n");
-      return yield* Effect.fail(
-        first._tag === "UnlinkTempRemovalError"
-          ? new UnlinkTempRemovalError({ message })
-          : new CredentialDeleteError({ message }),
-      );
+      return yield* first._tag === "UnlinkTempRemovalError"
+        ? new UnlinkTempRemovalError({ message })
+        : new CredentialDeleteError({ message });
```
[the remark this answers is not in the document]
7ttp: fixed: 3b0dcfe9

apps/cli/src/commands/unlink/unlink.integration.test.ts
```diff
@@ -21,6 +18,8 @@ import { unlink } from "./unlink.handler.ts";
 
 const tempRoot = useTempWorkdir("supabase-unlink-int-");
 
+const jsonText = Schema.encodeEffect(Schema.fromJsonString(Schema.Unknown));
```
Coly010: This satisfies `effecttsgo/prefer-schema-over-json`, but `Schema.fromJsonString(Schema.Unknown)` validates nothing and its encode is identity — it's `JSON.stringify` with codec ceremony, and it's the only place in the repo running a `Cause` through a schema codec this way.

Every other lint-covered family (23 of them, 279 call sites) solved this same lint rule with `Cause.pretty` instead. I checked empirically that `Cause.pretty(exit.cause)` satisfies all four assertions in this file verbatim, including the joined-message branch (renders `Tag: message` per failure, `\n`-joined). There's also direct precedent for the exact tag asserted here: `encryption/get-root-key/get-root-key.integration.test.ts:130` and `ssl-enforcement/update/update.integration.test.ts:412` both do `expect(Cause.pretty(exit.cause)).toContain("...")`.

Swapping to `Cause.pretty` here would remove the lint pressure at its root rather than routing around it with a no-op codec, and keeps this file from becoming the pattern the next ported command copies:

```ts
expect(Cause.pretty(exit.cause)).toContain("ProjectRefNotLinkedError");
```

Coly010 (commented):
Ran an automated engineer + architect review on this. Engineer review approved outright (verified the Effect.fail/yield* and Predicate.isTagged changes are runtime-identical, no coverage regression). Architect review flagged one real issue below, otherwise approved.

apps/cli/src/commands/unlink/unlink.integration.test.ts
7ttp: fixed: 13b34c96

Coly010 (approved):

PR #6662, opened 2026-09-17
refactor(cli): cover `seed` with effect lint (CLI-2452)

7ttp:
## TL;DR

brings the `supabase seed` command family under the effect lint

## whats introduced?

effect lint applied to `seed`:

- one allow list entry in `.oxlintrc.effect.json`
- remote object keys build through injected `Path` services instead of the `node:path` posix namespace, so the os native half and the posix half stay explicit rather than riding on one module
- the shared upload call site in `command-internal` passes both services, built once per call rather than once per file
- the mutually exclusive flags failure yields its tagged error directly
- the buckets integration suite writes its fixtures inside the effect, and the ambient env pin moved from a `beforeEach` onto the shared sandbox helper
- cause assertions moved to `Cause.pretty`, and the two that read a `suggestion` field use typed predicates instead, since `Cause.pretty` drops payload fields
- e2e tests run through the effect native harness with a scoped temp project dir per test

## ref:

- closes: CLI-2452


apps/cli/src/commands/seed/buckets/buckets.upload.ts
```diff
@@ -12,14 +12,21 @@ import * as nodePath from "node:path";
...
  */
-export function bucketObjectKey(bucketName: string, objectsPath: string, filePath: string): string {
-  const relPath = nodePath.relative(objectsPath, filePath);
+export function bucketObjectKey(
+  path: Path.Path,
+  posixPath: Path.Path,
+  bucketName: string,
+  objectsPath: string,
+  filePath: string,
+): string {
```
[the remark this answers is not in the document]
7ttp: fixed: 7fec70dc

apps/cli/src/command-internal/seed-buckets.ts
```diff
@@ -481,6 +482,7 @@ const uploadObjects = Effect.fnUntraced(function* (
   bucketsConfig: BucketsConfig,
   summary: SeedSummary,
 ) {
+  const posixPath = yield* Effect.provide(Path.Path, BunPath.layerPosix);
```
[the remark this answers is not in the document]
7ttp: fixed: 7fec70dc

apps/cli/src/command-internal/seed-buckets.ts
```diff
@@ -475,6 +477,7 @@ const handleVectorError = Effect.fnUntraced(function* (
 const uploadObjects = Effect.fnUntraced(function* (
   fs: FileSystem.FileSystem,
   path: Path.Path,
+  posixPath: Path.Path,
```
Coly010: `uploadObjects` takes `path: Path.Path` and `posixPath: Path.Path` as two adjacent, identically-typed positionals (lines 479-480), called as `uploadObjects(fs, path, posixPath, ...)` at line 283. Swapping the two typechecks and passes every test, because the unit-test job only runs on Linux, where both resolve to the same POSIX behavior — the bug would only surface on Windows, silently corrupting remote object keys (leaking backslashes into storage paths).

The PR's own second commit ("pass seed path services as one argument") already fixed this exact hazard in `bucketObjectKey` by grouping the two services into `{ path, posixPath }`. Worth applying the same grouping here so the call site names both services instead of relying on position:

```ts
const uploadObjects = Effect.fnUntraced(function* (
  fs: FileSystem.FileSystem,
  paths: { path: Path.Path; posixPath: Path.Path },
  output: typeof Output.Service,
  ...
```

apps/cli/src/commands/seed/buckets/buckets.upload.unit.test.ts
```diff
@@ -1,21 +1,34 @@
+import { BunPath } from "@effect/platform-bun";
 import { describe, expect, it } from "@effect/vitest";
+import { Effect, Path } from "effect";
 
 import { bucketObjectKey } from "./buckets.upload.ts";
 
+const path = Effect.runSync(Effect.provide(Path.Path, BunPath.layer));
+const posixPath = Effect.runSync(Effect.provide(Path.Path, BunPath.layerPosix));
```
Coly010, line 8: `BunPath.layer` resolves to the host-platform (POSIX on CI) implementation, so all four cases in this file exercise `posix → posix` — none of them actually cover the `path.sep → posixPath.sep` translation this refactor makes explicit, and none would catch the argument-swap hazard flagged on `uploadObjects`/`bucketObjectKey`. `BunPath.layerWin32` is already used for exactly this purpose elsewhere in the repo (e.g. `command-internal/glob.unit.test.ts`) — one added case with a win32 `path` would pin the contract:

```ts
const win32Path = Effect.runSync(Effect.provide(Path.Path, BunPath.layerWin32));

it("normalises win32 separators in the remote key", () => {
  expect(
    bucketObjectKey({ path: win32Path, posixPath }, "docs", "C:\\assets", "C:\\assets\\sub\\b.txt"),
  ).toBe("docs/sub/b.txt");
});
```

.oxlintrc.effect.json
```diff
@@ -27,6 +27,7 @@
     "!apps/cli/src/commands/projects/**",
     "!apps/cli/src/commands/pull/**",
     "!apps/cli/src/commands/secrets/**",
+    "!apps/cli/src/commands/seed/**",
```
Coly010, line 30: Non-blocking: this allow-list entry covers `commands/seed/**`, but the actual bucket-seeding logic (path resolution, upload orchestration) lives in `command-internal/seed-buckets.ts`, which is outside it and shared with `db reset`/`start`. That's why this PR had to touch an "uncovered" file (`Effect.provide(Path.Path, BunPath.layerPosix)`, the widened `uploadObjects` call) to make the covered files compliant. Pre-existing ownership mismatch, not introduced here — just flagging so "seed is covered" isn't read as covering the whole feature.

Coly010 (commented):
Ran an automated engineer + architect + security review on this. Security review came back clean. Two engineer/architect review agents independently converged on the same real issue below, plus a non-blocking note.

apps/cli/src/command-internal/seed-buckets.ts
7ttp: fixed: df0444b4

apps/cli/src/commands/seed/buckets/buckets.upload.unit.test.ts
7ttp: fixed: df0444b4

.oxlintrc.effect.json
7ttp: yeah thats fair. the entry is the family dir only so it isnt claiming the shared helper, and `seed-buckets.ts` being outside it is exactly why this pr had to reach into an uncovered file.

`command-internal` is tracked as its own area in the rollout and that whole dir gets picked up on its own, `seed-buckets.ts` included. leaving this one open since theres nothing to change here and its worth keeping visible.

Coly010 (approved):

PR #6667, opened 2026-09-17
chore(stack): introduce service instances and migrate CLI consumers

jgoux:
Stack services now have durable instance identities and a shared lifecycle API. Each instance owns its configuration, concrete dependencies, and runtime resources, so Functions and schema-tooling databases can operate independently within the same stack. Database instances also expose snapshot and restore operations.

Replace the custom stack shadow-database implementation with service instances across the CLI's shadow, cache, and reset flows. Functions serve reuses the managed Functions instance and leaves it available after CLI exit. Internal workflows use Effect, with one public Promise facade for non-Effect consumers.

Supersedes #6579


jgoux:
Superseeded by https://github.com/supabase/cli/pull/6684

PR #6670, opened 2026-09-18
fix(cli): honour config push consent (CLI-2450)

### LINKED ISSUE #6657: `config push` applies remote changes despite an explicit piped "n" when agent mode is auto-detected
### Affected area
`config push`: confirmation prompt handling in agent mode

### Supabase CLI version
2.109.1

### Operating system
macOS (arm64)

### Installation method
Homebrew

### Command
```
echo n | supabase config push --project-ref <ref>
```
Run from an AI coding agent's shell (non-TTY). `--agent` left at its default, `auto`.

### Actual output
The auth diff is printed, then the change is **applied** even though "n" was piped:
```
{"project_ref":"<ref>","services":[...,{"service":"auth","status":"updated"},...]}
```
`GET /v1/projects/<ref>/config/auth` confirms the new values. In our case the local `config.toml` was stale, so the push lowered hosted security settings: minimum password length 12 → 6, TOTP MFA enroll/verify disabled, `site_url` set to `http://127.0.0.1:3000`, and rate limits relaxed. We restored them from a snapshot.

### Expected behavior
The piped "n" is honored and the push aborts. In agent/non-text mode, a prompt guarding a remote write should never default to proceeding without `--yes`.

### Steps to reproduce
1. Have a `supabase/config.toml` whose `[auth]` section differs from a hosted project (for example, the file `supabase init` generates).
2. From an environment the CLI detects as an agent (so `--agent auto` resolves to agent mode), run `echo n | supabase config push --project-ref <ref>`.
3. Observe `"auth":"updated"` and the changed remote config.

### Likely cause
In `apps/cli/src/command-internal/prompt-yes-no.ts` (default branch), `promptYesNo` returns `defaultValue` when `output.format !== "text"`, before stdin is read. When agent mode is detected, the output format is non-text. `config push` defaults to "yes", so the explicit answer is never read. The command help says a non-interactive run "defaults to proceeding" when stdin carries no answer; here an answer was piped and ignored.

### Suggestions
- Honor a parseable piped answer regardless of output format.
- For prompts that guard remote writes, default to **no** in non-interactive/agent mode unless `--yes` is passed.
- Ship `supabase config diff` in a release. The `config push` help on the default branch references it, but 2.109.1 does not have it, so there is no read-only preview today.

7ttp:
## TL;DR

Fixes `config push` applying remote changes after a piped `n` in agent mode.

## whats wrong?

JSON and stream JSON confirmations return the default before reading stdin. With `config push` defaulting to yes, an explicit decline is ignored.

## now fixed by

Honor piped answers in machine output.

> [!WARNING]
>     **Breaking change:** 
unattended `config push` now skips updates without affirmative input or `--yes`/`SUPABASE_YES`
 text mode explains skipped confirmations. Interactive text confirmations retain their yes default...

## ref:

- closes https://github.com/supabase/cli/issues/6657
related: https://github.com/supabase/cli/pull/6647


apps/cli/src/commands/config/push/push.handler.ts
```diff
@@ -286,17 +286,17 @@ export const configPush = Effect.fn("config.push")(function* (flags: ConfigPushF
             : { kind: "uuid" };
       }
     }
+    const tty = yield* Tty;
+    const confirm = (label: string, defaultValue: boolean) =>
+      !yes && tty.stdinIsTty && !output.interactive
+        ? Effect.succeed(defaultValue)
+        : promptYesNo(output, yes, label, defaultValue);
```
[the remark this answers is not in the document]
7ttp: fixed: d8a001f

apps/cli/src/command-internal/prompt-yes-no.ts
```diff
@@ -40,19 +41,23 @@ export const promptYesNo = Effect.fnUntraced(function* (
...
+    if (output.format === "text") {
+      yield* output.raw(`${label} [${choices}] `, "stderr");
+    }
     const stdin = yield* Stdin;
     const line = yield* stdin.readLine(NON_TTY_TIMEOUT_MILLIS);
     const input = Option.getOrElse(line, () => "");
-    yield* output.raw(`${input}\n`, "stderr");
+    if (output.format === "text") {
+      yield* output.raw(`${input}\n`, "stderr");
+    }
```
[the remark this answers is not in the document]
7ttp: fixed: d8a001f

apps/cli/src/commands/config/push/push.e2e.test.ts
```diff
@@ -41,4 +42,75 @@ describe("supabase config push", () => {
...
+              SUPABASE_WORKDIR: cwd,
+              SUPABASE_YES: undefined,
+              CODEX_SANDBOX: "1",
+            },
+          },
+        );
+        expect(exitCode, `${stdout}\n${stderr}`).toBe(0);
+        expect(writes).toEqual([]);
+        expect(stdout).toContain('"status":"skipped"');
+        expect(stderr).toContain("api.max_rows [update]");
```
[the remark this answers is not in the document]
7ttp: fixed: d8a001f

apps/cli/src/commands/config/push/SIDE_EFFECTS.md
```diff
@@ -174,9 +174,8 @@ with --yes) [y/N] ` (bare ref, no quotes, when the name is unknown). An EXPLICIT
 prints, and the push proceeds immediately. Declining fails the command
 (`ConfigPushCancelledError`, exit `1`) — the rendered text is `context canceled`
 (`Output.fail`'s standard text-mode rendering, no `--debug` hint) — before any further
-network call (not even the cost-matrix fetch). Unlike this command's other
-confirmations, this gate's default is **no**: a non-TTY run with no piped answer, or
-`--output-format json`/`stream-json`, declines (and fails) rather than proceeding,
+network call (not even the cost-matrix fetch). This gate's default is **no**: a non-TTY run with no affirmative piped answer, or
+a machine-mode TTY run, declines (and fails) rather than proceeding,
 unless `--yes`/`SUPABASE_YES` is set. A plain-project target never shows this prompt.
```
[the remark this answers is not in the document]
7ttp: fixed: d8a001f

apps/cli/src/command-internal/prompt-yes-no.ts
```diff
@@ -40,19 +41,23 @@ export const promptYesNo = Effect.fnUntraced(function* (
     yield* output.raw(`${label} [${choices}] y\n`, "stderr");
```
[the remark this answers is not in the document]
7ttp: fixed: d8a001f

apps/cli/src/commands/config/push/push.handler.ts
```diff
@@ -309,14 +309,22 @@ export const configPush = Effect.fn("config.push")(function* (flags: ConfigPushF
...
     const keep = (name: string) =>
       Effect.gen(function* () {
         const item = cost.get(name);
         const title =
           item === undefined
             ? `Do you want to push ${name} config to remote?`
             : `Enabling ${item.name} will cost you ${item.price}. Keep it enabled?`;
-        return yield* promptYesNo(output, yes, title, true);
+        const confirmed = yield* confirm(title, defaultProceed);
+        if (!confirmed && output.format === "text" && tty.stdinIsTty && !output.interactive) {
```
Coly010: This is the ask from the review body. The condition restricts the hint to `text && stdinIsTty && !interactive` — i.e. run from a terminal with stdout redirected. That's the one non-interactive shape CI *doesn't* have: a CI run has a non-TTY **stdin**, so it falls through to the read path, gets an empty line, skips, and prints nothing beyond the `[y/N] ` echo, at exit 0.

Since that's precisely the shape whose behaviour this PR changes (it used to push), it's the shape that most needs to say so. Widening it to any non-affirmative non-interactive skip would do it, e.g.:

```ts
if (!confirmed && output.format === "text" && !yes) {
  yield* output.raw(
    tty.stdinIsTty && !output.interactive
      ? `Skipped ${name}: confirmation unavailable with redirected output. Pass --yes (or set SUPABASE_YES) to approve.\n`
      : `Skipped ${name}: no confirmation received. Pass --yes (or set SUPABASE_YES) to approve.\n`,
    "stderr",
  );
}
```

Worth checking it doesn't fire on an interactive `n` (there the user just answered, so a hint is noise) — `tty.stdinIsTty && output.interactive` is the case to exclude if it does.

apps/cli/src/commands/config/push/push.handler.ts
```diff
@@ -309,14 +309,22 @@ export const configPush = Effect.fn("config.push")(function* (flags: ConfigPushF
     const cost = yield* getCostMatrix(ref);
 
     // `promptYesNo` scans piped stdin on a non-TTY before falling back to the default.
+    const defaultProceed = tty.stdinIsTty && output.interactive && output.format === "text";
```
Coly010: Minor side effect of making the default TTY-dependent: it also drives the `[Y/n]` / `[y/N]` hint under `--yes`, where the default is irrelevant. A `--yes` run now echoes `[y/N] y` non-interactively but `[Y/n] y` on a TTY — the same invocation, two renderings. The existing `--yes` test only covers the TTY shape, so the non-TTY string is unpinned.

Passing `true` as the prompt default when `yes` is set (or deriving `choices` from the answer inside `promptYesNo`) would keep the echo stable regardless of shape.

apps/cli/src/command-internal/prompt-yes-no.ts
```diff
@@ -34,18 +35,22 @@ export const promptYesNo = Effect.fnUntraced(function* (
...
     return true;
   }
-  if (output.format !== "text") {
+  if (output.format !== "text" && !options.readMachineStdin) {
     return defaultValue;
   }
   const tty = yield* Tty;
-  // `interactive: false` still prints the label and reads one line instead of silently
-  // returning the default — it uses the same non-TTY read path below.
+  if (output.format !== "text" && (!interactive || tty.stdinIsTty)) {
```
Coly010, line 49: This return is unreachable from the only caller that opts in. `config push`'s `confirm` pre-check (`!yes && tty.stdinIsTty && !output.interactive`) already catches machine-mode + TTY stdin before it gets here — `output.interactive` is always `false` in machine mode — and it always passes `interactive: true`, so `!interactive` never fires either. `seed-buckets` does pass a dynamic `interactive`, but it doesn't set `readMachineStdin`, so the first early return catches it.

Not harmful, but it means the machine-mode policy is now expressed in two places with slightly different shapes, and the next command to adopt `readMachineStdin` can get the handler-side pre-check subtly wrong. Letting the helper own the whole matrix — `readMachineStdin` implying "read piped answers, otherwise return the default" — would leave callers with just the flag. If you'd rather keep the guard as defence-in-depth, a unit test in `prompt-yes-no.unit.test.ts` pinning it would stop it from silently rotting.

apps/cli/src/commands/config/push/SIDE_EFFECTS.md
```diff
@@ -218,9 +217,13 @@ own; `[group-write]` is an undeclared companion the endpoint required alongside
 declared change, sent at its schema default because the read didn't report the
 project's current value for it. Secret values never appear in output. Every block ends
 on a blank line. Experimental prints `Enabling webhooks for project: <ref>`. The
-per-service confirmations are unchanged from before CLI-2168: they render
-`<title> [Y/n] ` (or `<title> [Y/n] y` when `--yes`) and still exit **0** on decline —
-only the branch confirmation gate described above fails.
+per-service confirmations render `<title> [Y/n] ` in interactive text output and
+`<title> [y/N] ` for piped text input. `--yes` echoes `y` after the applicable
```
Coly010: Two small precision points against the rest of this file's style, which pins exact strings:

1. `--yes echoes `y` after the applicable prompt` leaves the bracket hint unspecified, and it does vary — `[Y/n] y` on a TTY, `[y/N] y` non-interactively (see the note on `defaultProceed`). Worth pinning both, or pinning one once the echo is made shape-independent.
2. `render `<title> [Y/n] ` in interactive text output` describes the clack path, which actually renders a clack confirm widget rather than that literal string. This predates the PR, but since you're rewriting the paragraph it'd be a good moment to fix.

Coly010 (approved):
Approving — the core fix is right and well fenced. `promptYesNo` returning the default before reading stdin was exactly the cause named in #6657, and gating the new read behind an opt-in `readMachineStdin` keeps the blast radius to this command; the new `projects delete` test asserting a piped `y` still doesn't authorize a deletion is a good guard on that.

Worth stating explicitly since it's easy to mistake for scope creep: flipping the non-interactive no-answer default from yes to no was **requested by the reporter** ("default to **no** in non-interactive/agent mode unless `--yes` is passed"), and the old help text already described the previous behaviour as a footgun. So that part has a mandate.

I verified the matrix with throwaway probes rather than by reading alone.

## One ask before merge

**Text-mode non-interactive skips are silent and still exit 0.** Probing the CI shape — text output, non-TTY stdin, no piped answer, no `--yes` — gives:

```
api.max_rows [update]
  local:  2000
  remote: 1000

Do you want to push api config to remote? [y/N] 
Remote DB config is up to date.
```

Zero PATCHes, exit code **0**, no summary line and no `Note:`. The only signal is a prompt echo with an empty answer, sitting between "up to date" lines. A pipeline that runs `supabase config push` today pushes; after this it pushes nothing and stays green.

Machine mode is fine — it emits `"Nothing was pushed. 1 service was skipped at the prompt."` — and the branch gate is fine too, since `Output.fail` does render its `Pass --yes (or set SUPABASE_YES)…` suggestion in text mode. The gap is only that the helpful hint you added is gated to the one shape CI never has; see the inline note. A one-line condition change covers it.

I'd also call this out as a breaking change in the PR body so it reaches the release notes.

## Smaller things

- The same bug class is still next door: `secrets unset` (default `true`, and destructive) and `db-push-core` (default `true`, ×3) all still ignore a piped `n` in machine mode. `db reset` and `storage rm` already default to `false`, so they're safe. #6657 asked for the class rather than just this command — are those tracked separately?
- `prompt-yes-no.unit.test.ts` still only covers `parseYesNo`, so none of the new matrix has direct unit coverage; it's all exercised indirectly through this command's integration tests.

Nothing else blocking — the SIDE_EFFECTS rewrite is accurate on every claim I checked, and the e2e test remembering to unset `SUPABASE_YES` is a nice touch.

apps/cli/src/commands/config/push/push.handler.ts
7ttp: fixed: 81853d7

apps/cli/src/commands/config/push/push.handler.ts
7ttp: fixed: 81853d7

apps/cli/src/command-internal/prompt-yes-no.ts
7ttp: fixed: 81853d7
kept the guard and added direct tests.

apps/cli/src/commands/config/push/SIDE_EFFECTS.md
7ttp: fixed: 81853d7

PR #6691, opened 2026-09-21
fix(stack): log target failures and handle disconnections quietly

avallete:
## TL;DR

- **A failed lazy wake was invisible.** It surfaced only as a `502` or a dead socket, with the cause discarded and `error=undefined` in `stack status`. The proxy now logs the route and cause, and the failure reaches the service observation.
- **Ordinary client disconnects stay quiet.** A client that goes away is its own typed error, not a proxy failure, so normal traffic adds no error logs.
- **`stack port tests (windows-latest)` could not allocate a port.** The auto scan walked contiguously and died inside a Windows reserved block. It now walks `20000..32767` with stride `257` per [ADR 0017](docs/adr/0017-simplified-managed-stack-architecture.md), so a reserved block up to 12511 of the 12768 ports is escaped within the bounded attempts.
- **Port allocation is reproducible.** The scan start derives from the stack rather than a random draw, so a stack reassigns the same port and a future failure repeats instead of depending on the draw.

## Details

Two independent stack failures, both surfaced while investigating the 2026-09-21 `Test` breakage.

### A failed lazy wake was unobservable

A wake that failed reached clients only as a dead connection. `HttpProxy` discarded the cause in both arms — the request path answered `502` and the upgrade path destroyed the socket — and `Service.startAt` runs `definition.prepare` (where native artifact preparation happens) before it touches the observation, so a preparation failure left `error` undefined. Every wake failure was therefore indistinguishable from a flake.

Both proxy arms now log the route id and the underlying cause before responding, and a failed wake preparation reaches the service observation, so `stack status` and the reopened-stack diagnostics name the cause instead of printing `error=undefined`.

The two halves cover different topologies. A detached owner is spawned with `stdout`/`stderr` ignored, so the proxy log lines land only where the runtime shares the caller's process; the recorded observation is what names the cause through `stack status` and the e2e diagnostics in the detached topology. Retaining a detached owner's output is a separate pre-existing gap.

Client exits stay quiet, and are told apart from upstream faults by type rather than by message: a client that disconnects while the target is waking, goes away mid-response, or resets an established upgrade fails with `HttpProxyDisconnected`, while upstream faults keep failing with `HttpProxyError` and log. Both paths settle before destroying their sockets, because destroying a partially received upstream response emits `aborted` synchronously and would otherwise resettle a disconnect as a proxy failure. A preparation failure is recorded only while the service is still the stopped, registered attempt that prepared it, so a concurrent start that already relaunched keeps its own state.

### `stack port tests (windows-latest)` could not allocate an auto port

The auto scan walked contiguously from a random start over `20000..49999` and gave up after 64 consecutive bind failures. Windows publishes reserved TCP ranges as contiguous blocks commonly wider than 64 ports, so a start landing inside one failed with `No public port is available` while tens of thousands of ports were free. That range also overlapped the Linux default ephemeral range.

The scan follows ADR 0017 again: `20000..32767` with stride `257`, which is co-prime with the span and so traverses every candidate once across a full scan. Within the ADR's 64-attempt bound, a contiguous reserved block up to 12511 of the 12768 ports is always escaped, worst case 49 failures, verified over every start and block offset. Wider blocks can still exhaust the bound — that limit is the ADR's, and the stride does not remove it — but they are far beyond the width of a published excluded range.

The one deliberate divergence from the ADR is the scan start, which derives from the stack's project root, id, and listener key rather than a random draw. Allocation is reproducible: a stack reassigns the same port across runs, separate checkouts stay apart, and a future failure repeats instead of depending on the draw. The ADR records this.

The exhausted error also carries the last bind failure, so a reserved range (`EACCES`) reads differently from an occupied port (`EADDRINUSE`) from the log alone.

## Changes

- `HttpProxy.ts`: log route request and upgrade failures with the route id and cause; give client-originated disconnects a distinct tagged error; settle before destroying sockets
- `Ports.ts`: walk the documented range with a co-prime stride from a derived start; attach the last bind failure to the exhausted error. No longer needs the platform crypto service, so `HostProcess` stops carrying that requirement
- `Service.ts`: record a failed wake preparation on the observation, guarded against a concurrent relaunch
- `docs/adr/0017`: record the derived scan start

`Service.ts` here is the wake path only; `restart` is unchanged because its caller already receives the error.

https://claude.ai/code/session_01LcUaThuQfDKPwi9mCE9XeD


packages/stack/src/Ports.ts
```diff
@@ -58,10 +58,12 @@ export const makePorts = (state: State.Interface) =>
...
 
             const owner = yield* Scope.Scope;
-            const first = yield* crypto.randomIntBetween(0, 29999);
             let failures = 0;
+            let lastFailure: PortError | undefined;
             for (let attempt = 0; attempt < 30000 && failures < 64; attempt++) {
-              const port = requested === "auto" ? 20000 + ((first + attempt) % 30000) : requested;
+              // Each probe is redrawn so a contiguous reserved range cannot exhaust the budget.
+              const port =
+                requested === "auto" ? yield* crypto.randomIntBetween(20000, 49999) : requested;
```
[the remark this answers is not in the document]
avallete: Confirmed and fixed in `e65b21b`, which also drops the random draw entirely.

The scan now walks the range with a fixed stride instead of sampling: `port = 20000 + ((start + attempt * 257) % 12768)`. Since 257 is co-prime with 12768 it visits every candidate exactly once, so there is no replacement and no omitted endpoint — the exclusive-upper-bound problem is gone with the sampling.

That also restores what this ADR section already specified (`20000..32767`, stride `257`, 64 bounded attempts); the previous code had drifted to a contiguous walk over `20000..49999`, which overlapped the Linux default ephemeral range. Only the start now differs from the ADR — it derives from the stack's project root, id, and listener key rather than a random draw, so allocation is reproducible — and the ADR text is updated to match.

Brute-forced the reserved-range behaviour over every start and block offset: a contiguous refusal block of any width within the span costs at most 39 consecutive failures, so the 64 allowance cannot be exhausted by a reserved range.

---

packages/stack/src/HttpProxy.ts
```diff
@@ -286,6 +289,11 @@ export const makeHttpProxy = (options: {
               response.end("Not Found");
             } else {
               yield* proxyRequest(request, response, route).pipe(
+                Effect.tapError((cause) =>
+                  cause._tag === "HttpProxyDisconnected"
+                    ? Effect.void
+                    : Effect.logError(`Route ${route.id} request failed`, cause),
+                ),
```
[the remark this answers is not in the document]
avallete: Correct, and fixed in `e65b21b`. The race only classified disconnects *during* target acquisition, so a client that went away mid-response still produced `HttpProxyError` and a new error log — exactly the noise this PR set out to avoid.

Client-originated paths now fail with `HttpProxyDisconnected`: `request` abort and the response-close check in `proxyRequest`, and the client socket error on an established upgrade. Upstream faults (`outgoing` error, `incoming` abort and error, `upstream` error) still fail with `HttpProxyError` and log.

Covered by `disconnects a pending upstream response quietly when its client closes`, which drives a real client that closes after the backend has the request and asserts no error log, synchronising on the route scope's finalizer so the assertion cannot race the log. It fails on the previous classification.

---

packages/stack/src/Service.ts
```diff
@@ -428,7 +428,9 @@ export const makeService = <Config>(
         if (existing.lifecycle === "running") return;
       }
       const nextConfig = candidate ?? (yield* Ref.get(config));
-      if (definition.prepare !== undefined) yield* definition.prepare(nextConfig);
+      // A wake has no caller to receive a preparation failure, so observers record it instead.
+      if (definition.prepare !== undefined)
+        yield* definition.prepare(nextConfig).pipe(Effect.tapError((error) => update({ error })));
```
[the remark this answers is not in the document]
avallete: Reproduced and fixed in `e65b21b`. The race is real: with a first preparation still pending, a second start can prepare, launch, and clear `error`, after which the first failure overwrote it — leaving a running, healthy service reporting a stale preparation error.

The record now goes through a guard that drops the failure unless the intent revision still matches the attempt and the observation is still registered and `stopped`, so a relaunched or invalidated service keeps its own state.

`keeps a stale preparation failure off a relaunched observation` drives that interleaving: the first preparation parks on a gate, a second start launches to running, then the gate opens and the first fails. It asserts the failure still reaches the caller while the observation stays `running` with `error` undefined, and it fails against the unguarded update.

---

packages/stack/src/HttpProxy.ts
```diff
@@ -131,33 +134,35 @@ const proxyRequest = Effect.fn("HttpProxy.proxyRequest")(
...
           cleanup();
           resume(result);
         };
-        const onError = (cause: Error) => {
+        const abandon = (result: Effect.Effect<void, HttpProxyError | HttpProxyDisconnected>) => {
           if (settled) return;
           outgoing?.destroy();
           incoming?.destroy();
-          finish(Effect.fail(errorFor(cause)));
+          finish(result);
```
jgoux: [P2] Settle the disconnect before destroying the upstream response

When a client closes after the backend has sent headers and a body chunk, `onClientGone` enters `abandon` with `HttpProxyDisconnected`. However, `incoming.destroy()` synchronously emits `aborted` while `settled` is still false and the `onError` listener is attached. That listener re-enters `abandon` without an error argument and settles the request as `HttpProxyError: undefined`, defeating the new quiet-disconnect classification. I reproduced this with a real streaming backend: wait for the client to receive the first chunk, close it, wait for target release, then assert no error logs. This fails on this head with `Route pending request failed HttpProxyError: undefined` and passes on the base. Please mark the operation settled and remove the abort listener before destruction, and cover this mid-response case; the existing test disconnects before any upstream response arrives.

packages/stack/src/Ports.ts
```diff
@@ -58,10 +66,14 @@ export const makePorts = (state: State.Interface) =>
...
-            const first = yield* crypto.randomIntBetween(0, 29999);
+            const start = scanStart(stack, request.key);
             let failures = 0;
-            for (let attempt = 0; attempt < 30000 && failures < 64; attempt++) {
-              const port = requested === "auto" ? 20000 + ((first + attempt) % 30000) : requested;
+            let lastFailure: PortError | undefined;
+            for (let attempt = 0; attempt < portSpan && failures < 64; attempt++) {
+              const port =
+                requested === "auto"
+                  ? portBase + ((start + attempt * portStride) % portSpan)
```
jgoux, line 75: Non-blocking: correct the port-scan guarantee in the PR description

The stride improves coverage, but coprimality guarantees a full traversal only if the scan runs through the entire span. The 64-failure limit does not guarantee escaping a contiguous block of any width. For example, with start offset 0, attempts 0–49 select offsets 0–12593; attempts 50–63 wrap to offsets 82–3423. A reserved block of ports 20000–32593 therefore covers all 64 attempts while leaving 174 ports free. Please remove the “any width … at most 39” guarantee. This does not require an allocator redesign: the stride/range change remains useful, while deterministic starting points are a separate policy choice.

jgoux (changes requested):
Requesting changes for the reproduced mid-response disconnect regression below. The existing 42 targeted integration tests pass, but a client that closes after receiving the first response chunk produces a spurious error log; the same reproduction passes against the base.

The port scan is a separate, reasonable fix; its description overstates the guarantee, as noted inline. A separate pre-existing limitation is that detached hosts discard stdout/stderr, so the new proxy logs are not retained in that topology.

packages/stack/src/HttpProxy.ts
avallete: Reproduced and fixed in `0756522`. Your diagnosis is exact — I wrote the test from your description before touching the code and got the failure verbatim:

```
AssertionError: expected [ Array(1) ] to deeply equal []
+   "Route streaming request failed HttpProxyError: undefined"
```

`abandon` destroyed before settling, so `incoming.destroy()` emitted `aborted` synchronously while `settled` was still false and `onError` was still attached; the nested call resettled with `errorFor(undefined)` and the outer `finish` became a no-op.

Both paths now settle and drop their listeners before destroying (`finish(result)` first, then the destroys) — the upgrade path had the same ordering, where the nested handler would have won with `Effect.void` instead. Late `error` events after settling still land on the retained listeners and no-op, as the existing comment there intends.

`stays quiet when a client closes after receiving part of the response` covers it: a streaming backend writes headers plus a chunk, the client closes once it has that chunk, and the assertion waits on the route scope's release before checking for logs. It fails on the previous head and passes now; ran it 15× consecutively with no flakes.

---

packages/stack/src/Ports.ts
avallete: You're right, and the description is corrected. My "any width" was an unsound generalisation: I measured a 10000-wide block (39 failures is correct for that width) and then stated it as a universal bound. Coprimality only gives a full traversal across a complete scan, which the 64-attempt limit truncates — and your counterexample checks out, all 64 attempts inside offsets 0–12593 with 174 ports still free.

I computed the actual threshold, exhaustively over every start (the scan is translation-invariant, so fixing the block at offset 0 and varying the start covers every arrangement):

| contiguous block | worst-case failures before a bindable port |
| --- | --- |
| 1 000 | 4 |
| 4 000 | 16 |
| 8 000 | 32 |
| 10 000 | 39 |
| 12 511 | 49 |
| 12 512 | ≥ 64 (can exhaust) |

So the honest statement, now in the description: a contiguous block up to 12511 of the 12768 ports is always escaped within the bound, and wider blocks can still exhaust it — that limit is the ADR's, not something the stride removes. No allocator redesign here, as you say; the range and stride change stands on restoring the documented behaviour.

I also noted your detached-host point in the description rather than leaving it implied: with `stdout`/`stderr` ignored for a detached owner (`HostProcess.ts:171-174`), the proxy log lines only land where the runtime shares the caller's process, and it's the recorded observation that names the cause via `stack status` and the e2e diagnostics in that topology. Happy to pick up retaining a detached owner's output separately if you want it tracked.

---

jgoux (approved):

